PR #107, opened 2023-09-26
Added load_today and pv_today for 2nd inverter to apps.yaml

gcoan:
If you have two inverters then each inverter operates independently and will separately record its own view of load & PV power.
Changed apps.yaml to include sensors from 2nd inverter


apps/predbat/config/apps.yaml
```diff
@@ -22,16 +22,20 @@ pred_bat:
   
   #
   # Sensors, currently more than one can be specified and they will be summed up automatically
-  # however if you have two inverters only set one of them as they will both read the same.
+  # If you have two inverters, each inverter will have its own load and PV generation data 
+  # but only set the import and export for one of them as these will both read the same.
```
springfall2008: I'm not sure this is true, the load energy sensor should be a CT clamp and hence cover the whole house, unless you are on 3 phase and you have different inverters for each phase.

For PV it depends, if you have two PV inverters it makes sense but many people don't have PV on both.

apps/predbat/config/apps.yaml
```diff
@@ -226,6 +226,10 @@ pred_bat:
   # Standing charge in pounds, can be set to a sensor or manually entered (e.g. 0.50 is 50p)
   # The default below will pick up the standing charge from the Octopus Plugin
   metric_standing_charge: 're:(sensor.(octopus_energy_|)electricity_[0-9a-z]+_[0-9a-z]+_current_standing_charge)'
+  #
+  # If you don't want to include the standing charge in batpred's cost prediction, set metric_standing_charge to zero.
```
springfall2008: Or just delete the line

gcoan: True, either way, I thought it worthwhile adding an explanation of this as I personally prefer to just focus on the consumption. I can't do anything about the standing charge and it was confusing me to include it

apps/predbat/config/apps.yaml
springfall2008: I've added some text about this on main now

apps/predbat/config/apps.yaml
springfall2008: I've added some text about this on main now

springfall2008 (approved):

PR #124, opened 2023-09-29
Remove refs. to Octopus Intelligent plugin

iainfogg:
I thought I'd get this done while it was fresh in my head.

I've also tried to make it consistent so it says Intelligent Octopus everywhere, rather than just saying Intelligent, or saying Octopus Intelligent, as I'm aware there's already quite a bit of confusion over the Intelligent Flux tariff, so hopefully a consistent text will help (or at least, not add to the confusion).

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,6 @@ If you want to buy me a beer then please use Paypal - tdlj@tdlj.net
   * [Energy rates](#energy-rates)
     + [Octopus Energy Plugin](#octopus-energy-plugin)
     + [Rate bands](#rate-bands)
-    + [Octopus Intelligent Plugin](#octopus-intelligent-plugin)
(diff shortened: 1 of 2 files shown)


springfall2008:
Rather confusingly it doesn't quite read correctly when you say 'The Octopus Energy Plugin' in some cases it needs to say 'You are using Intelligent Octopus with the Octopus Energy Plugin' as you could have the plugin but be on a different tariff that doesn't have smart charging.

Not sure if you want to update that or you want me to merge this and then do the final update?

iainfogg:
Do you know which specific bits you mean @springfall2008 ? Or do you want to edit it yourself to fix the bits you mean? Either way is fine :)

I think you should have edit access to it.

PR #194, opened 2023-10-19
Create dynamic-entities-card.yaml

DJBenson:
A dynamically created list of entities for [predbat](https://github.com/springfall2008/batpred/tree/main/apps/predbat) which groups the entities by type and is collapsed by default to prevent screen clutter. Requires [lovelace-collapsable-cards](https://github.com/RossMcMillan92/lovelace-collapsable-cards) and [lovelace-auto-entities](https://github.com/thomasloven/lovelace-auto-entities) custom cards as well as the stock vertical stack card.

image

### DIFF
--- a/dynamic-entities-card.yaml
+++ b/dynamic-entities-card.yaml
@@ -1,0 +1,116 @@
+type: vertical-stack
+title: Predbat 🦇
+cards:
+  - type: entity
+    entity: predbat.status
+  - type: custom:collapsable-cards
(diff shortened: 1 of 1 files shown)


springfall2008 (approved):

rjdiver:
How do you exclude the [Disabled] entities from showing, I see an exclude part in the code but not sure on the syntax to enter, if that is indeed the way to exclude those entries?

PR #224, opened 2023-10-30
Solis

fboundy:
Initial edits to support Solis inverters. The only changes to the main code should be where "inverter_type" is used to test whether to apply. I do populate that early in the initialize() function of the PredBat class.

I think the Solis will need to write to the inverter directly using the ModBus service so I am prepping the code to do this. All the calculations look reasonable so far and very similar to my own code.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,3 @@
+apps/predbat/config/apps.yaml
+solaredge.yaml
+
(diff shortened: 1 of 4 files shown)


springfall2008:
Hi,

I'd like to merge this, but predbat.py itself has been reformatted and so will create a massive conflict with other changes. Is it possible to undo the format changes and just include your functional changes?

fboundy:
That’s Black autoformatting it.

I’ll turn it off and re-submit the PR. Half of this evening’s edits undo yesterday’s anyway!

PR #230, opened 2023-10-31
Solis

fboundy:
Revised version supporting Solis inverters. Should be much cleaner than the previous PR with Black autoformat turned off.


springfall2008:
I'm not sure why but predbat.py is showing in the diff like a new file, maybe it's in the wrong directory?

fboundy:
I noticed that too. I’m make a new branch and a new PR.

PR #235, opened 2023-10-31
Solis Support

fboundy:
This looks better.

### DIFF
--- a/apps/predbat/config/ginlong_solis.yaml
+++ b/apps/predbat/config/ginlong_solis.yaml
@@ -1,0 +1,331 @@
+pred_bat:
+  module: predbat
+  class: PredBat
+  
+  # Sets the prefix for all created entities in HA - only change if you want to run more than once instance
+  prefix: predbat
+
+  # XXX: This is a configuration template, delete this line once you edit your configuration
+  # template: True
+
+  # Timezone to work in
+  timezone: Europe/London
+  
+  #
+  # Sensors, currently more than one can be specified and they will be summed up automatically
+  # however if you have two inverters only set one of them as they will both read the same.
+  #
+  load_today: 
+    - sensor.solis_daily_consumption
+  import_today: 
+    - sensor.solis_daily_energy_imported
+  export_today: 
+    - sensor.solis_daily_energy_exported
+  pv_today:
+    - sensor.solis_daily_generation   # This is a custom sensor - documentation TBD
+  #
+  # Controls/status - must by 1 per inverter
+  #
+  num_inverters: 1
+ 
+
+  battery_rate_max:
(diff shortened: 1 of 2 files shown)


springfall2008 (approved):

springfall2008:
Thanks for your change, I've merged it now but will submit a tidy up shortly. Let me know what you think about it once its in. The idea is that you shouldn't do inverter specific stuff unless really required, most of it should be configuration related.

fboundy:
OK - I think I get it. This might require a bit of a re-think on my part:

My current approach would have predbat writing to the modbus registers on the inverter but I could pull this out into a separate appDaemon app and just write to the appropriate HA entities instead. This might actually mean much less customisation within predbat.

What do you think?
________________________________

fboundy:
Just thinking a bit more about this I think it’s a much cleaner solution and probably wouldn’t require most of the changes in my PR of last night.

Can I suggest you revert it for now?

springfall2008:
I think like you say the Modbus stuff should be in a separate app, there maybe already one available for it I guess as you have some sort of HA integration? Predbat needs to control the inverter via the HA interface.

fboundy:
Yes there is one already. I’ve not used it previously but I’ve got it working fine now.

A couple of queries:

Battery Reserve - is this a minimum SOC that is held above the maximum Depth of Discharge? Solis calls this Backup SOC. Is is specified in % or kWh?

The Solis inverters have no way of setting a target SOC while charging so we may need a workaround by setting a reduced charge current which will meet the desired target SOC at the end of the charging slot.

Cheers

Francis

springfall2008:
Reserve % is indeed the level where the battery stops discharging, it’s saved for a power cut or suchlike. Predbat uses it to hold the current battery level when in a charge slot if no more charging is required.

I’m surprised you can’t set the % to charge up to, but I don’t think it’s an issue. Predbat will by default (if set reserve is enabled) cancel the charge once the % is reached and uses the reserve to hold the level until the end of the slot. I think if you hardwire set reserve enable to On it probably will be fine. I’d advise against changing charging current as it will confused predictions.

fboundy:
Separate issue I have is that there’s no direct equivalent of scheduled_charge_enable  or scheduled_discharge_enable on the Solis. It is done through a combination of the Inverter Mode Control switch (which is set bitwise). Bit 1 enables Timed Charging and then any Timed slots are active if the Start and End times are different. To enable/disable the Reserve, you also need to switch Bit 4.

To complicate this further the Solax integration that I am using in place of GivTCP wraps all this up in a Select entity using a string descriptor. I can work around this but I am struggling to get write_and_poll_options to write to the relevant entity. Anything I should know?

PR #236, opened 2023-11-01
Refactor additional inverter support

springfall2008:


springfall2008:
@fboundy this is a small refactor of your change

PR #262, opened 2023-11-05
Additional Inverter Documentation 

fboundy:
Mainly comments also added a new dict INVERTER_DEFS to help to identify the differences between inverter control paramters which I think will help in refactoring the Inverter class.


apps/predbat/predbat.py
```diff
@@ -301,8 +329,38 @@ def __init__(self, base, id=0, quiet=False):
...
-        Update inverter status
-        """
+        Update the following with inverter status.
+
+        Inverter Class Parameters
+        =========================
+
+            Parameter                          Type    Units    
+            ---------                          ----    -----    
+            self.rest_data                     bool
```
springfall2008, line 340: rest_data is a dictionary containing the REST data read from the API, but you can do an IF on it as when it's empty it's false

apps/predbat/predbat.py
```diff
@@ -498,7 +556,20 @@ def update_status(self, minutes_now, quiet=False):
 
     def adjust_reserve(self, reserve):
         """
-        Adjust the reserve target % in GivTCP
+        Adjust the output reserve target %
+
+        Inverter Class Parameters
+        =========================
+
+            None
```
springfall2008, line 565: Well, it's using self.rest_data, self.reserve_percent, self.reserve_max and references to base

fboundy: I did think about list what was being read but I'm mainly concerned that I can change the outputs of each function as required by a different inverter. I'm assuming that the inputs will be handled by the Predbat class.

apps/predbat/predbat.py
```diff
@@ -128,6 +128,34 @@
     {'name' : 'holiday_days_left',             'friendly_name' : 'Holiday days left',              'type' : 'input_number', 'min' : 0,   'max' : 28,    'step' : 1,    'unit' : 'days', 'icon' : 'mdi:clock-end'},
 ]
 
+INVERTER_DEF={ 
+    "GE": {
+        'has_rest_api': True,
+        'output_charge_control': "current",
```
springfall2008: These don't like right for GE, in that it has a change and discharge time enable and a target soc?

springfall2008 (approved):

springfall2008 (approved):

PR #282, opened 2023-11-07
Inverter support

fboundy:
This PR implements my revised model for supporting additional inverters. ALthough it possibly looks more complex I think it is simpler because it clearly identifies ways in which the new brand (in this case Solis) departs from GE. The approach is then do create "dummy" entities which mimic those which GE uses. This simplifies the re-use of existing code.

I'd suggest testing this branch on a GE setup before pulling it into main. I've tried to make sure that the existing funtionality isn't compromised but can't guarantee I haven't messed something up!


apps/predbat/predbat.py
```diff
@@ -257,20 +290,29 @@ def __init__(self, base, id=0, quiet=False):
...
+
             ivtime = self.base.get_arg('inverter_time', index=self.id, default=None)
 
         # Battery can not be zero size
         if self.soc_max <= 0:
             self.base.log("ERROR: Reported battery size from REST is {}, but it must be >0".format(self.soc_max))
             raise ValueError
 
-        # Battery rate max charge, discharge
+        # Battery rate max charge, discharge (all converted to kW/h)
```
springfall2008: I think it's actually kWh/minute confusingly enough

ginlong_solis.yaml
```diff
@@ -216,7 +218,7 @@ pred_bat:
   # Or manually set it to the correct sensor names e.g:
   # sensor.octopus_energy_electricity_xxxxxxxxxx_xxxxxxxxxxxxx_current_rate
   # sensor.octopus_energy_electricity_xxxxxxxxxx_xxxxxxxxxxxxx_export_current_rate
-  metric_octopus_import: 're:(sensor.(octopus_energy_|)electricity_[0-9a-z]+_[0-9a-z]+_current_rate)'
+  metric_octopus_import: sensor.octopus_energy_electricity_19m1337498_1610032016836_current_rate
```
springfall2008: This looks like a typo to hard-write your meter number?

fboundy: Correct! I was having issues with Octopus last night. Will edit this out.

apps/predbat/predbat.py
fboundy: Yes, of course.

fboundy:
I think this now works with my Solis inverter. Going to leave it running overnight and see if it does what it is supposed to.

springfall2008 (approved):

PR #294, opened 2023-11-09
Remove whitespace

iainfogg:
Just having a bit of a tidy-up :)

This PR removes a load of whitespace at the end of lines, and also ensures there's a single blank line at the end of each file. It was an automated job to do this.

Should be easily verifiable by clicking on the cog in the diff and doing 'Hide whitespace'.

Thinking of building something like this (plus other code quality checks) into the automated process at some point if that'd be helpful @springfall2008 . If so, I'll create a separate issue for that some time.

EDIT: The two files that do show, if you hide whitespace, are ones that had multiple blank lines at the end, and there's now only one.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -21,11 +21,11 @@
 			]
 		}
 	},
-	
+
 	// Use 'forwardPorts' to make a list of ports inside the container available locally.
 	// "forwardPorts": [9000],
 
-	// Use 'portsAttributes' to set default properties for specific forwarded ports. 
+	// Use 'portsAttributes' to set default properties for specific forwarded ports.
 	// More info: https://containers.dev/implementors/json_reference/#port-attributes
 	// "portsAttributes": {
 	// 	"9000": {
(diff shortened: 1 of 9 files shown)


springfall2008 (approved):

fboundy:
I think that would be good. Best done like this as a one-off. Some of the very long lines would benefit from cleaning up.

I use Black for a lot of my other projects and it really went to town on PredBat!

springfall2008:
It would be nice to have an auto-format in place, but I think the max line length should probably be a bit longer than default, not sure but you both but I use a 4k monitor so finding around 250 chars is okay, some still wrap that and probably should be split up

fboundy:
My eyes aren’t up to processing lines that long! I think PEP8 says 80 which is quite narrow but I’d be happy to go with 120.

iainfogg:
I've got a branch on my own fork which uses pre-commit to do various tasks.

Things I've tried doing with it include:
- Trimming whitespace at the end of lines
- Ensuring there's a blank line at the end of each file
- Running black
- Running pylint
- Doing spell checks
- Sorting python imports alphabetically
- Sorting file contents alphabetically (eg the custom spelling dictionary, or the requirements.txt file)
- Some other stuff which I've forgotten

How about I do a bit more on that, then I'll create a separate issue where we can discuss what's best? I suspect it'll take a long conversation to see what we want, and maybe going in stages is the best bet.

Also I'm looking into using pre-commit to actually do fixes on PRs itself, rather than just failing the PR, got an idea on how that can work.

PR #306, opened 2023-11-11
Support for Config Groups

fboundy:
To do:

- additional groups
- the ability to save custom groups


iainfogg:
Great to see this coming together so quickly!

I don't know if I'm right in thinking this, but I'd envisaged that each separate config group would set the same set of config items, rather than having different items in different groups.

That way, if you change tariffs you're getting the full set of config when you move, including resetting config items that were previously set for your last tariff.

What do you think?

springfall2008:
Looks really good, I suspect I might need to rethink the groups and options that are set. And I agree with @iainfogg that probably all of them need to change the same settings so that you can switch between them without side effect.

apps/predbat/predbat.py
```diff
@@ -43,6 +43,88 @@
...
+            "set_reserve_enable": True,
+            "set_reserve_hold": True,
+            "calculate_best_charge": True,
+            "combine_charge_slots": True,
+            "metric_min_improvement": 0,
+            "set_charge_window": True,
+            "best_soc_keep": 2.0,
+            "best_soc_min": 0.0,
+            "rate_low_threshold": 0.8,
+            "forecast_plan_hours": 24,
```
springfall2008: forecast_plan_hours is in config.yml currently so can't be changed here. We could move it into HA to allow dynamic configuration.

apps/predbat/predbat.py
```diff
@@ -8557,6 +8656,9 @@ def initialize(self):
             self.record_status("ERROR: Exception raised {}".format(e))
             raise e
 
+        self.config_grouping = self.get_arg("config_grouping", default=list(CONFIG_GROUPS.keys())[0])
```
springfall2008: I think the default here should be 'user defined' or similar, this will set the value of config_grouping in HA. The user would then select an option which would update the settings but the dropdown wouldn't change. We don't want the settings to overwrite every time the user changes a value

fboundy:
Yep that makes sense. Though perhaps the easiest thing is to have a default set and then each  group has to record the differences from that. Easier to manage I think?

If you leave it with me I’ll add that and a Custom workflow.

fboundy:
Thinking about this a little more:

Currently all the configuration variables are in separate variables such as self.set_soc_on. Wouldn’t it be easier to save them, group them etc if they were in a single dict: self.config? You’d then have self.config[‘set_soc_on’] for example.

You would know that all config was in one place.

You could still have CONFIG_ITEMS declared as is because this keeps the constant declarations out of the main code but on init you would simply populate self.config with self.config = {i[name]: i for i in CONFIG_ITEMS}.

The initial search and replace could probably be done with a regex to ease the pain.

With each config item and it’s associated HA entity indexed by its name the event_handler logic would also be much simpler.

iainfogg:
Is there any way to show what's changed in the UI do you think? Or do people just need to read the logs (which get overwritten in the standard setup fairly quickly?

apps/predbat/predbat.py
```diff
@@ -8284,16 +8371,28 @@ def select_event(self, event, data, kwargs):
...
                 self.log("select_event: {} = {}".format(entity, value))
                 self.expose_config(item["name"], value)
+                self.log(f"Entity: {entity} Group: {config_group_entity}")
+                if entity == config_group_entity:
+                    group = self.get_arg(entity, default=list(CONFIG_GROUPS.keys())[0])
+                    self.log(f"Loading default config for {group}")
+                    items = CONFIG_GROUPS[group]['items']
+                    for name in items:
+                        value=items[name]
+                        self.log(f"Setting {name} to {value}")
```
iainfogg: Is it worth also logging the current value?

fboundy:
Agreed - logging the change would be better.

apps/predbat/predbat.py
springfall2008: I changed forecast_plan_hours to an HA config element now

PR #308, opened 2023-11-11
Initial documentation for Solis inverters

fboundy:


docs/config-yml-settings.md
```diff
@@ -186,13 +184,14 @@ Connect to your cars sensors for accurate data:
 Control how your battery behaves during car charging:
 
 - **car_charging_from_battery** - When True the car can drain the home battery, Predbat will manage the correct level of battery accordingly. When False home battery discharge will be prevented when your car charges, all load from the car and home will be from the grid. This is achieved by setting the discharge rate to 0 during car charging and to the maximum otherwise, hence if you turn this switch Off you won't be able to change your discharge rate outside Predbat. The home battery can still charge from the grid/solar in either case. Only use this if Predbat knows your car charging plan, e.g. you are using Intelligent Octopus or you use the car slots in Predbat to control your car charging.
-    - CAUTION: If you turn this switch back on during a car charging session you will need to set your battery discharge rate back to maximum manually.
+
+  - CAUTION: If you turn this switch back on during a car charging session you will need to set your battery discharge rate back to maximum manually.
```
iainfogg: The proper Markdown spec is to have 4 spaces (or a tab) for an indented list, I think you've changed this plus some others to 2.

I went through them all when I redid the docs and set them all to 4 (GitHub was fairly forgiving, and would work with 2).

And if it did with with mkdocs, that'll be because fairly late on I added the sane lists extension, as one of the original lists wasn't working with the strict implementation of Markdown spec.

mkdocs.yml
```diff
@@ -19,6 +19,7 @@ nav:
     - creating-charts.md
     - predbat-plan-card.md
   - 'Predbat development':
+    - other-inverters.md
```
iainfogg, line 22: Personally I'd add this in another section, rather than in the section on developing on Predbat, as it's not really developer content.

iainfogg (commented):
Added a couple of comments

docs/config-yml-settings.md
```diff
@@ -230,21 +229,21 @@ The triggers count export energy until the next active charge slot only.
 
 For each trigger give a name, the minutes of export needed and the energy required in that time
 Multiple triggers can be set at once so in total you could use too much energy if all run
-Each trigger create an entity called 'binary_sensor.predbat_export_trigger_<name>' which will be turned On when the condition is valid
+Each trigger create an entity called 'binary*sensor.predbat_export_trigger*<name>' which will be turned On when the condition is valid
```
springfall2008: Typo, the star should be before binary?

docs/install.md
```diff
@@ -45,7 +51,7 @@
 
 Predbat needs a solar forecast in order to predict battery levels.
 
-If you don't have solar then comment out the Solar forecast part of the apps.yml: **pv_forecast_* **
+If you don't have solar then comment out the Solar forecast part of the apps.yml: **pv*forecast*\* **
```
springfall2008: Why star's in place of underscore?

fboundy:
I don’t remember editing this at all. May have been som auto-formatting nonsense!

fboundy:
Hmm - something odd going on here. Again not a change I remember making…

iainfogg:
Could it be that you'd already written the code in the original README then you copied it in here after I reworked the documentation?

The original README used 2 characters for indents, so that would explain it.

fboundy:
I don’t think so. I think it may be VScode doing some autoformatting. I’ve noticed that it changes some markdown - sorting out tables and the like but maybe it’s not following the same standards?

Will have a look tomorrow once the rain has set in

iainfogg:
Ok, that's strange as I used VS Code to do the work in the first place. It griped about some indents but never auto changed them for me.

I think it leans towards the GitHub more relaxed standards, rather than the strict standards.

springfall2008:
@fboundy were you going to fix the issues with the * before merging?

fboundy:
Yes. Leave this for now

springfall2008 (approved):

PR #317, opened 2023-11-12
Merge Main

fboundy:


iainfogg:
@fboundy I'd suggest that if you're able to, it's better to just merge `main` in as a merge commit on your branch, rather than doing a pull request to do it.

Here's an example where I did it, pulling in the latest from main.

https://github.com/iainfogg/batpred/commit/3c4ac76747b2a0586f2bcd731a4dbf28d01a43a7

I just run `git fetch` to get the latest commits from `main` pulled locally (assuming I'm on the Predbat repo already - if I'm on my fork I use the UI to sync my main to the Predbat main first), then while on my feature branch I run `git merge origin/main` (there's a flag for doing a commit message with that which I've forgotten right now).

That way, it keeps the pull request history just to things which are being merged into `main`, things which actually need the review aspect of pull requests.

fboundy:
Yep - figured I was doing something wrong there..

________________________________

PR #324, opened 2023-11-14
323 solis inverter hourmin not matching timed chargedischarge entity

fboundy:
Adds an extra check to makes sure that the Solis hour and minute entities are consistent with the "GE style" time


fboundy:
Ah - that's not the issue. It's that the Solis time can't span midnight! Closing this because it needs a different fix

PR #331, opened 2023-11-15
298 feature request group configs

fboundy:
Adds support for Config Groups:

Default values are defined in CONFIG_ITEMS
CONFIG_GROUPS is then a dict of differences.

If any configs are changed the Group becomes Custom. This is written to the attributes of the select.predbat_config entity so it can be read back if the user decides to go back to what they previously had.

I'll create a separate Issue for @springfall2008 to review the defaults and the groupings


apps/predbat/predbat.py
```diff
@@ -43,6 +43,56 @@
...
+        "calculate_best_discharge": True,
+        "calculate_discharge_first": True,
+        "combine_discharge_slots": False,
+        "set_discharge_window": True,
+        "metric_min_improvement": 0,
+        "metric_min_improvement_discharge": 0.1,
+        "rate_low_threshold": 0,
+        "rate_high_threshold": 0,
+        "metric_battery_cycle": 2,
+        "calculate_max_windows": 32,
```
springfall2008: Probably should be max_windows 40 now and metric_battery_cycle 0 (users are confused otherwise)

apps/predbat/predbat.py
```diff
@@ -43,6 +43,56 @@
...
+        "combine_charge_slots": True,
+        "combine_discharge_slots": False,
+        "set_discharge_window": True,
+        "metric_min_improvement": 0,
+        "metric_min_improvement_discharge": 0,
+        "rate_high_threshold": 0,
+        "set_discharge_freeze": True,
+        "set_charge_freeze": False,
+        "calculate_max_windows": 96,
+        "predbat_metric_battery_cycle": 3,
```
springfall2008: Typo on metric_battery cycle?

apps/predbat/predbat.py
```diff
@@ -177,112 +202,199 @@
...
-        "min": 0.00,
-        "max": 2.00,
+        "min": 0.0,
+        "max": 2.0,
         "step": 0.05,
         "unit": "multiple",
         "icon": "mdi:multiplication",
+        "default": 0.0,
     },
-    {"name": "car_charging_hold", "friendly_name": "Car charging hold", "type": "switch"},
```
springfall2008, line 233: This diff seems odd?

apps/predbat/predbat.py
```diff
@@ -8316,113 +8505,40 @@ def fetch_config_options(self):
...
+            "Inverter type {} max_windows {} num_cars {} debug enable is {} calculate_plan_every {}".format(
+                self.inverter_type, self.calculate_max_windows, self.num_cars, self.debug_enable, self.calculate_plan_every
+            )
+        )
+
         if self.inverter_type != "GE":
             default_enable_mode = False
             self.log("WARN: Using experimental inverter type {} - not all features are available".format(self.inverter_type))
 
-        self.calculate_best = self.get_arg("calculate_best", True)
```
springfall2008, line 8431: I don't understand this big delete?

apps/predbat/predbat.py
fboundy: Fixed

apps/predbat/predbat.py
fboundy: Fixed

apps/predbat/predbat.py
fboundy: I changed the order of the definitions in CONFIG_ITEMS so I could QC them against the various attributes that are initialised by them in fetch_config (see next comment). So the diffs look odd but it's just because of the change in order

apps/predbat/predbat.py
```diff
@@ -8276,21 +8454,32 @@ def fetch_config_options(self):
         Fetch all the configuration options
         """
 
-        self.debug_enable = self.get_arg("debug_enable", False)
-        self.previous_status = self.get_state(self.prefix + ".status")
-        self.calculate_max_windows = self.get_arg("calculate_max_windows", 48)
-        self.calculate_plan_every = max(self.get_arg("calculate_plan_every", 10), 5)
+        # Most attributes can be loaded directly
+        for item in CONFIG_ITEMS:
```
fboundy, line 8511: This is the for loop that populates all the attributes from `CONFIG_ITEMS'

apps/predbat/predbat.py
fboundy: All the attributes were manually set with a default. The for loop at the start of `fetch_update` now does most of this automatically:

`self.attribute` can be set as `self.__dict__['attribute'`]` so the code just loops through CONFIG_ITEMS and sets each attribute based on `CONFIG_ITEMS['name']`. This doesn't work for all of them where there is, for example, a units change or a conversion from 'loss' (3%) to efficiency (97%) so these are still handled explicitly in the code.

fboundy:
@springfall2008 what is outstanding on this?

springfall2008:
I'm going to test this PR now before merging

springfall2008:
So I just fired this up locally and found the config group selection menu to be empty:

image

springfall2008:
I must admit reading the code I'm a bit stuck on how this is supposed to work.

I was guessing that config group would have a drop-down where I can select one of the pre-defined configuration groups and then if I select it then it will load the other config values automatically. If I change the values then I assume it would go back to saying Custom.

fboundy:
I think you’re absolutely right to test this before accepting the PR - the fact I have different hardware means it’s entirely possible that I have screwed something up for GE while GS works fine

That said, let me just make absolutely sure that the remote branch is synced with my local one and that it works on my system.

springfall2008:
I suggest we review this in context of my cleanup change, I think the config groups is still a good idea but maybe simpler after my changes

iainfogg:
@fboundy just checking, do you think this is still useful now the cleanup change is done? And is this codebase up to date with the latest version?

Didn't want to spend time if it was no longer needed.

fboundy:
No - this is now so far behind that there’s little point. I will close the PR

PR #393, opened 2023-11-29
Update predbat.py, added rounding to self.log messages

gcoan:
Added rounding to logging of status messages where figures were being displayed with lots of unnecessary decimal places:
- cache retrieval aging
- PV forecast figures
- soc charging and discharging limits
- plan last updated time


apps/predbat/predbat.py
```diff
@@ -2146,7 +2146,7 @@ def download_predbat_releases_url(self, url):
             pdata = self.github_url_cache[url]["data"]
             age = now - stamp
             if age.seconds < (120 * 60):
-                self.log("Using cached GITHub data for {} age {} minutes".format(url, age.seconds / 60))
+                self.log("Using cached GITHub data for {} age {} minutes".format(url, round(age.seconds / 60,1)))
```
springfall2008: I think this is the same as self.dp1(x) - we should probably pick one of them :)

apps/predbat/predbat.py
```diff
@@ -7712,6 +7712,10 @@ def calculate_plan(self, recompute=True):
                 save="best",
                 end_record=self.end_record,
             )
+            # round charge_limit_best and discharge_limits_best to nearest whole number as soc's are integer values
+            self.charge_limit_best = [ int(round(elem,0)) for elem in self.charge_limit_best ]
```
springfall2008: I'm not sure this is correct, for legacy reasons the charge limit is in kWh while the discharge limit is in percent

gcoan:
Ok I’ll change both of them in line with your comments

apps/predbat/predbat.py
```diff
@@ -7712,6 +7718,10 @@ def calculate_plan(self, recompute=True):
                 save="best",
                 end_record=self.end_record,
             )
+            # round charge_limit_best (kWh) to 1 decimal place and discharge_limits_best (percentage) to nearest whole number  gc
+            self.charge_limit_best = [ self.dp1(elem) for elem in self.charge_limit_best ]
```
springfall2008: I'm not 100% sure 1 decimal place is enough here as it needs to be converted back into the right percentage later. I wonder if you are trying to fix the formatting of the text rather than really wanting to modified the save values that will be used to program the inverter?

springfall2008 (approved):

PR #396, opened 2023-11-30
Add word to custom dictionary and fix long line

iainfogg:
Add a word to the dictionary so the spell check passes.

Fix a long line in the markdown for energy-rates.md.

Also contains a couple of auto-added fixes by pre-commit.

Both these issues were causing a failure in the linting pipeline.


iainfogg:
@springfall2008 this is good to go, it fixes a couple of things that had come in after I finished the coding standards PR that were meaning that the code on main had faults, so they would have caused any other builds to fail.

Be good to get this in ASAP so it doesn't hold up any other PRs.

springfall2008 (approved):

PR #418, opened 2023-12-04
Auto opt into saving sessions, remove old assumed rate sensor

springfall2008:


apps/predbat/predbat.py
```diff
@@ -8549,20 +8534,34 @@ def fetch_sensor_data(self):
         # Octopus saving session
         octopus_saving_slots = []
         if "octopus_saving_session" in self.args:
+            saving_rate = 200  # Default rate if not reported
```
iainfogg, line 8537: Octopus told me that they're passing on £2.25, although clearly plenty / all have been higher than that so far, so I think it'd be safe to take £2.25 as a minimum.

PR #420, opened 2023-12-04
Predbat mode

springfall2008:


apps/predbat/predbat.py
```diff
@@ -8900,9 +8909,30 @@ def fetch_config_options(self):
...
+            self.set_charge_window = False
+        elif self.predbat_mode == PREDBAT_MODE_OPTIONS[PREDBAT_MODE_CONTROL_CHARGE]:
+            self.calculate_best_charge = True
+            self.calculate_best_discharge = False
+            self.set_charge_window = True
+        elif self.predbat_mode == PREDBAT_MODE_OPTIONS[PREDBAT_MODE_CONTROL_CHARGEDISCHARGE]:
+            self.calculate_best_charge = True
+            self.calculate_best_discharge = True
+            self.set_charge_window = True
+        else:  # PREDBAT_MODE_OPTIONS[PREDBAT_MODE_MONITOR]
```
iainfogg, line 8927: Personally I'd make this an explicit `elif`, followed by an `else` that throws an error due to an unexpected mode, so it's safer in the future if you add other values to the mode.

Or is this `else` primarily the mechanism to set the mode to MONITOR if no mode is currently set?

iainfogg (commented):
I know this is merged, but thought I'd drop a comment in now I've been able to look it over.

And I think this is a good step forward in further simplifying things, well done!

iainfogg:
Also, the mode dropdown doesn't appear in the auto entities card, it'd be good to coordinate with the person who build that to add another section for it (or maybe add it manually at the top of that card alongside the status).

iainfogg:
Now I've got it running and checked it out more, I also thought that the docs for the SOC only section don't make it clear (to me at least) what the benefit is of this mode, I couldn't figure why I'd want it to adjust the target percentage - but I expect there is a reason that you understand :)

https://springfall2008.github.io/batpred/customisation/#predbat-control-soc-only-mode

iainfogg:
And I think I would have had the mode options in a different order, with Monitor, Charge and Discharge, Charge, SOC - the first one for the default starting value, then the others in what I guess is reducing order of likeliness of selection. Not a big deal, just may be a tad more usable.

PR #429, opened 2023-12-05
Fixes

springfall2008:


iainfogg:
@springfall2008 Was the charge limit fix in this PR related to the float issue I had, or something else?

Just clarifying if it's relevant, the issue I had was work the car SOC entity, where mine stopped connecting to the manufacturer and presumably reported unknown, and that seemed to stop that home battery charging.

PR #445, opened 2023-12-10
Move template files into new directory

springfall2008:


iainfogg (changes requested):
Hi, I'm afraid those relative links to the template files won't work, for a couple of reasons:

- The whole repo is not present in the documentation, check out the branch gh-pages at https://github.com/springfall2008/batpred/tree/gh-pages and you'll see it's an html site that mkdocs has built - that overwrites the rest of the repo (at least the way we have it set up now,, though this may be changeable,I can't remember right now)

- The documentation site is present at https://springfall2008.github.io/batpred/ so if you wanted to use relative links, you'd need to use everything after the .io - so it would be /batpred/templates

I can check the first issue tomorrow if you like, or just change it to point to the main code, or put the content in a Markdown page in the actual documentation and have people copy from there.

If you need more help just let me know.

springfall2008:
For now I've just given the absolute links back to github, maybe we can find a more relative way later

docs/output-data.md
```diff
@@ -6,7 +6,7 @@ Each Predbat config item has an input_number or switch associated with it, you c
 configuration in your AppDaemon configuration area under the filename **predbat_dashboard.yaml**.
 You will need to open that file in a Home Assistant file editor and copy the contents into a new dashboard page.
 
-You can also create a dashboard page using [dynamic-entities-card.yaml](https://github.com/springfall2008/batpred/blob/main/dynamic-entities-card.yaml) for a
+You can also create a dashboard page using [dynamic-entities-card.yaml](https://raw.githubusercontent.com/springfall2008/batpred/tidy/templates/dynamic-entities-card.yaml) for a
```
iainfogg: This one has the tidy branch name in it, instead of main

iainfogg (commented):
I think there's just that one link to fix

docs/output-data.md
springfall2008: Well spotted

PR #455, opened 2023-12-12
Add mobile notification when joining octopus saving session. Corrected kWh and kW capitalisations

gcoan:
When Predbat auto-joins a new Octopus saving session, a mobile notification message is now sent

Corrected capitalisation of kW and kWh in Predbat log messages, alert messages and entities (unit of measurement and friendly names)

### DIFF
--- a/apps/predbat/predbat.py
+++ b/apps/predbat/predbat.py
@@ -24,7 +24,7 @@ TIME_FORMAT_SOLIS = "%Y-%m-%d %H:%M:%S"
 PREDICT_STEP = 5
 RUN_EVERY = 5
 
-# 240v x 100 amps x 3 phases / 1000 to kw / 60 minutes in an hour is the maximum kWh in a 1 minute period
+# 240v x 100 amps x 3 phases / 1000 to kW / 60 minutes in an hour is the maximum kWh in a 1 minute period
 MAX_INCREMENT = 240 * 100 * 3 / 1000 / 60
 
 SIMULATE = False  # Debug option, when set don't write to entities but simulate each 30 min period
@@ -160,7 +160,7 @@ CONFIG_ITEMS = [
         "min": 4,
         "max": 8.5,
         "step": 0.10,
-        "unit": "kw",
+        "unit": "kW",
         "icon": "mdi:ev-station",
         "default": 6.0,
     },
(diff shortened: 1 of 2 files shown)


apps/predbat/predbat.py
```diff
@@ -8650,9 +8650,10 @@ def fetch_sensor_data(self):
                         start = event.get("start", None)
                         end = event.get("end", None)
                         saving_rate = event.get("octopoints_per_kwh", saving_rate * octopoints_per_penny) / octopoints_per_penny  # Octopoints per pence
-                        if code:
+                        if code:  # Join the new Octopus saving event and send an alert
                             self.log("Joining Octopus saving event code {} start {} end {} price per kWh {}".format(code, start, end, saving_rate))
                             self.call_service("octopus_energy/join_octoplus_saving_session_event", event_code=code, entity_id=entity_id)
+                            self.base.call_notify("Predbat: Joined Octopus saving event: start {}, end {}, price per kWh {}".format(start, end, saving_rate))
```
springfall2008, line 8679: Do you think this should be enabled with 'self.set_status_notify' as per the status changes?

springfall2008 (approved):

apps/predbat/predbat.py
gcoan: I did wonder about that.  I concluded not, so the alert would always be sent, on the basis that:
1. Once comfortable with predbat users may be running it in the background and not receiving any notifications as a result
2. This was a significant "activity" that changes the way Predbat would behave, and users may not otherwise have been aware of the saving session and would want to be notified so they can change any other consumption behaviour (and potentially turn predbat off if they don't want the force discharge) - in effect it was a 'high priority' alert

springfall2008: > I did wonder about that. I concluded not, so the alert would always be sent, on the basis that:
> 
> 1. Once comfortable with predbat users may be running it in the background and not receiving any notifications as a result
> 2. This was a significant "activity" that changes the way Predbat would behave, and users may not otherwise have been aware of the saving session and would want to be notified so they can change any other consumption behaviour (and potentially turn predbat off if they don't want the force discharge) - in effect it was a 'high priority' alert

Okay for me

Issue #556, opened 2023-12-30
Changing control mode with read only enabled turns off charge schedule


mpartington:
**Describe the bug**
Read only mode should prevent predbat writing to the inverter, however if you change modes it appears to do the new reset, which results in the charging schedule being turned off.  I have a standalone automation to switch back on, hence why it quickly reverts in the screenshots
![Screenshot_2023-12-30-22-52-29-466_io homeassistant companion android~2](https://github.com/springfall2008/batpred/assets/106986397/7eb6e115-25b4-440a-b5e3-71e7a30bd2cf)
![Screenshot_2023-12-30-22-52-18-984_io homeassistant companion android](https://github.com/springfall2008/batpred/assets/106986397/04ce7663-eb93-411c-8acb-eb5078927c27)

**Expected behavior**
Read only should mean just that, no writes should be performed to the inverter. The status of Charging schedule should not be changed.

**Predbat version**
V7.14.28

**Environment details**

- Standard HAOS 

**Screenshots**
As attached

**Log file**
N/A

mpartington:
![Screenshot_2023-12-30-22-55-32-958_io homeassistant companion android](https://github.com/springfall2008/batpred/assets/106986397/dd56f5ea-a53b-413e-9529-4f0f3cda0cba)

mpartington:
Covered by #577

PR #557, opened 2023-12-31
Do not reset inverter in read-only mode even for mode

springfall2008:


springfall2008:
https://github.com/springfall2008/batpred/issues/556

PR #656, opened 2024-01-27
Code cleanup

marksie1988:
I have cleaned up some of the code following modifications: 

* Removed nested if statements
* Removed variables that were not in use
* Fixed issues with some format strings having the incorrect values
* Fixed an issue with dashboard_item attributes being incorrect
* Added sourcery / flake8 / black / isort for tidier code
* Updated requirements.txt with missing `adbase` (needed for dev to pick up auto completes etc) 

NOTE: I haven't fully tested everything as some of the changes were to parts that i don't use personally but i noticed were incorrect. Although this was only a refactoring so shouldn't negatively impact anything


springfall2008:
Thanks for all your code cleanups. You have a few automatic lint issues to correct before I can merge, let me know if you need any help with them.

springfall2008 (approved):

Issue #650, opened 2024-01-26
Low Power Charge Mode doesn't always self adjust to fill the available time period


mpartington:
I understand the intent is that if "Set Charge Low Power Mode" is true, then Predbat will manage the charge rate , aiming to achieve the target SoC 30 mins before the end of the charge window.

I have seen a number of instances on my system where it finished much earlier (up to 90 minutes).  Predbat doesn't seem to realise that it is ahead of schedule and to further dial down the charge rate.  I have since tweaked the charge profile in apps.yaml (to suit 3015) and reduced total charging losses to 4%.

It was actually quite accurate this morning, finishing 42 minutes before end of charge window (so 12 ahead of target), so not sure if it may relate to size of charge necessary.  I have checked, and the amount or energy added each night is consistent with an approximation based on starting SoC, so no major errors there (or soc jumps)

**Expected behavior**
Whilst inaccurate efficiency assumptions in the settings (or SoC readings) could initially lead to a faster of slower rate of charge than predicted, I would expect Predbat to compensate for this (up until the point the final 10% charge profile is limited by the BMS) by progressively reducing the charge rate per calculation cycle.  On the graphed example (25-Jan), this doesn't seem to have happened as expected (unfortunately I only have the log for the very final part of that charge)

Predbat Version:
v.7.15.1

**Environment details**

- AC3 with 2x 5.2kWh batteries running D0.203-A0.203 (new firmware) and 3015
- Standard HAOS installer
- Anything else?

**Screenshots**

<img width="848" alt="bp2" src="https://github.com/springfall2008/batpred/assets/106986397/536dcf08-3bd8-48be-b53c-161a38676ff0">

**Log file**
Logs for 25th (bad)  attached. I will continue to log and add bad examples to better show the issue
[25_Jan.log](https://github.com/springfall2008/batpred/files/14062700/25_Jan.log)

mpartington:
Added battery discharge Rate graph for same period. It doesn't fluctuate, but it is unusual that it increases the charge rate quite near to the end of charge, despite being ahead of schedule for the finish time
![Screenshot_2024-01-25-07-10-46-388_io homeassistant companion android](https://github.com/springfall2008/batpred/assets/106986397/4461fc3a-daa2-4ed8-877f-970bc862cd09)

mpartington:
Similar profile this morning. Hit 100% at 3am, charge rate is initially reduced, but then goes back up. Plan was aiming 97% at 3.30. wil
[...]

PR #657, opened 2024-01-28
Fix battery charge curve modelling for lower than max battery charge …

springfall2008:
…rate settings


springfall2008:
Fix for: https://github.com/springfall2008/batpred/issues/650

Issue #666, opened 2024-01-29
battery charge curve not being created by predbat


gcoan:
**Describe the bug**
Upgraded to 7.15.4 (after the crashes of 7.15.3 - thanks for the quick bugfix)

But the power curve is not created by predbat

getting the error
pred_bat: Note: Can not find battery charge curve, one of the required settings for soc_kw, battery_power and charge_rate are missing from apps.yaml

Looking at apps.yaml, none of these are set because I am using givtcp_rest and so all the inverter controls are commented out, vis:

```
  # When set use the REST API rather than HA entity for control, should be more reliable/faster to control
  # Set one per inverter
  # If using Docker then change homeassistant.local to the Docker IP address
  givtcp_rest:
  # - 'http://homeassistant.local:6345'
   - 'http://homeassistant.local:6346'

  # If not using REST then instead set the Control here (one for each inverter)
  # - you can delete this section if using REST
  #charge_rate:
  #  - number.givtcp_{geserial}_battery_charge_rate
  #  - number.givtcp2_{geserial2}_battery_charge_rate
  #discharge_rate:
  #  - number.givtcp_{geserial}_battery_discharge_rate
  #  - number.givtcp2_{geserial2}_battery_discharge_rate
  #battery_power:
  #  - number.givtcp_{geserial}_battery_power
  #  - number.givtcp2_{geserial2}_battery_power
  #pv_power:
  #  - number.givtcp_{geserial}_pv_power
  #  - number.givtcp2_{geserial2}_pv_power
  #load_power:
  #  - number.givtcp_{geserial}_load_power
  #  - number.givtcp2_{geserial2}_load_power
  #soc_kw:
  #  - sensor.givtcp_{geserial}_soc_kwh
  #  - sensor.givtcp2_{geserial2}_soc_kwh
 ```

**Expected behavior**
Battery curve created per new feature in 7.15.3/4

**Predbat version**
7.15.4

**Environment details**
HAOS
Gen 1 hybrid

**Comment by springfall2008** [comment]
Unfortunately REST doesn’t provide any history so the fix is just to comment these back in.  I probably need to change the documentation to say that

**Comment by gcoan** [comment]
OK, I didn't appreciate that, yes to update the documentation to make it clearer

**Comment by gcoan** [comment]
Its not working for me either.

I uncommented out the three non-rest lines the charge curve needs:
```
charge_rate:
    - number.h_{geserial}_battery_charge_rate
  battery_power:
[...]

PR #682, opened 2024-02-01
Automatic unit detected for Watts vs Kwh when reading historical data

springfall2008:


springfall2008:
https://github.com/springfall2008/batpred/issues/666

PR #780, opened 2024-02-21
Documentation Change - EV Charging how to example without/very little 3rd party integration

dandwhelan:
I have created a guide on setting up an MG4 EV and Hypervolt Charger. I feel a real-world example with all the steps laid out will be helpful for other users. This can be used by anybody, just solely MG4 and Hypervolt users. All they need is a means to turn the charger off and on.


springfall2008:
What happened with this one, why did you close it?

dandwhelan:
Trevor, I've no idea what I'm doing, it's my first ever push /pull

Issue #801, opened 2024-02-29
Ability to specify multiple battery_scaling values


gcoan:
**Is your feature request related to a problem? Please describe.**
I have been running with just a single 5.2kW battery on one inverter for some time, so had set battery_scaling to 0.8 so that the 80% DoD SoC was correctly computed.

Had a new 9.5 commissioned (separate inverter) and wanted to continue with the 5.2 being scaled correctly, but Predbat gives a warning when I try to set multiple battery scaling values as the code must only be expecting a single value for all batteries across all inverters:

apps.yaml:

```yaml
  battery_scaling: 
    - 1.0
    - 0.8
```

logfile:

2024-02-29 08:30:00.103140 INFO pred_bat: WARN: Return bad float value [1.0, 0.8] from battery_scaling using default 1.0

**Describe the solution you'd like**
Ability to set battery scaling on a per inverter basis, allowing multiple entries as PV load, givtcp_rest, etc already do

**Describe alternatives you've considered**
Set a single hybrid battery scaling across all inverters, but this will give the wrong capacity result for both battery.

springfall2008:
Now on main, please test :)

gcoan:
Thanks Trefor, you been away on holiday, its been very quiet !

Loaded the version on main, set the battery scaling list in apps.yaml and from the logfile it appears to be working:

```
2024-03-04 01:15:12.583733 INFO pred_bat: Inverter 0 with soc_max 9.52 kWh nominal_capacity 9.52 kWh battery rate raw 2600 w charge rate 2.6 kW discharge rate 2.6 kW battery_rate_min 0.0 w ac limit 5.0 kW export limit 5.0 kW reserve 4.0 % current_reserve 6.0 %
2024-03-04 01:15:14.879243 INFO pred_bat: Inverter 0 SOC: 0.48 kW 5 % Current charge rate 2600.0 w Current discharge rate 2600.0 w Current power 0.0 w Current voltage 52.0
...
2024-03-04 01:11:07.200954 INFO pred_bat: Inverter 1 with soc_max 4.18 kWh nominal_capacity 5.22 kWh battery rate raw 2600 w charge rate 2.6 kW discharge rate 2.6 kW battery_rate_min 0.0 w ac limit 5.0 kW export limit 5.0 kW reserve 4.0 % current_reserve 6.0 %
[...]

PR #813, opened 2024-03-03
Make battery scaling per inverter

springfall2008:


springfall2008:
Fix is for: https://github.com/springfall2008/batpred/issues/801

Issue #867, opened 2024-03-18
TypeError: argument of type 'NoneType' is not iterable


onetoomany:
From latest version:
 v7.16.8

```
2024-03-18 10:55:02.317144 WARNING pred_bat: ------------------------------------------------------------
2024-03-18 11:00:01.966605 WARNING pred_bat: ------------------------------------------------------------
2024-03-18 11:00:01.966914 WARNING pred_bat: Unexpected error in worker for App pred_bat:
2024-03-18 11:00:01.967169 WARNING pred_bat: Worker Ags: {'id': '2fc90665927d42ff87d845ca949acd9c', 'name': 'pred_bat', 'objectid': '9166e474f15c41a48e746bc20b80f3ca', 'type': 'scheduler', 'function': <bound method PredBat.run_time_loop of <predbat.PredBat object at 0x7f3d6ebd9110>>, 'pin_app': True, 'pin_thread': 0, 'kwargs': {'interval': 300, 'random_start': 0, 'random_end': 0, '__thread_id': 'thread-0'}}
2024-03-18 11:00:01.967434 WARNING pred_bat: ------------------------------------------------------------
2024-03-18 11:00:01.969599 WARNING pred_bat: Traceback (most recent call last):
  File "/usr/lib/python3.11/site-packages/appdaemon/threading.py", line 1022, in worker
    funcref(self.AD.sched.sanitize_timer_kwargs(app, args["kwargs"]))
  File "/usr/lib/python3.11/site-packages/appdaemon/adbase.py", line 35, in f_app_lock
    return f(*args, **kw)
           ^^^^^^^^^^^^^^
  File "/homeassistant/appdaemon/apps/batpred/predbat.py", line 12671, in run_time_loop
    raise e
  File "/homeassistant/appdaemon/apps/batpred/predbat.py", line 12667, in run_time_loop
    self.update_pred(scheduled=True)
  File "/usr/lib/python3.11/site-packages/appdaemon/adbase.py", line 35, in f_app_lock
    return f(*args, **kw)
           ^^^^^^^^^^^^^^
  File "/homeassistant/appdaemon/apps/batpred/predbat.py", line 11664, in update_pred
    status, status_extra = self.execute_plan()
                           ^^^^^^^^^^^^^^^^^^^
  File "/homeassistant/appdaemon/apps/batpred/predbat.py", line 10622, in execute_plan
    inverter.adjust_battery_target(self.charge_limit_percent_best[0], isCharging)
[...]

PR #875, opened 2024-03-19
Add if gate to set_current_from_power function call

stevebuk1:
To solve issue #867 on SolarEdge Inverters, which was caused by an error introduced in #843 made for Solis inverters.


duckfullstop (approved):
LGTM as a fix for the broken behaviour; but I'm wondering whether gating on `set_current_from_power()` itself makes more sense instead of requiring this explicit check every time? (especially because using it improperly causes an exception)

stevebuk1:
> LGTM as a fix for the broken behaviour; but I'm wondering whether gating on `set_current_from_power()` itself makes more sense instead of requiring this explicit check every time? (especially because using it improperly causes an exception)

It does make more sense, but its currently called from only 5 places and the other 4 correct ones use identical code, so I just copied that. I'm not that familiar with Python (I've only made PRs to Predbat in the last week), I wasn't sure of the scope of "inv_output_charge_control" and the last edit in a different area required a "get_args" call to access a variable, so went with the existing gating method.

Issue #877, opened 2024-03-20
SOLIS mode name changed


marcoaddario:
Starting from version 2024.03.2 the energy storage mode switch exposed by the Solax Modbus component changed resulting in batpred not been able to select the right mode for charge/discharge.

These are the new mode names displayed:
(changelog [here](https://github.com/wills106/homeassistant-solax-modbus/commit/0731c6d8e68ff69109bf79998e2f737cc22ae5ef))

```
1: "Self-Use - No Grid Charging",
3: "Timed Charge/Discharge - No Grid Charging",
17: "Backup/Reserve - No Grid Charging",
33: "Self-Use - No Timed Charge/Discharge",
35: "Self-Use",
37: "Off-Grid Mode",
41: "Battery Awaken",
43: "Battery Awaken + Timed Charge/Discharge",
49: "Backup/Reserve - No Timed Charge/Discharge",
51: "Backup/Reserve",
```
[...]


**Expected behavior**

SOLAX_SOLIS_MODES should be updated to reflect the new mode names for mode 1, 33 and 35
(this is the actual one)

```
SOLAX_SOLIS_MODES = {
    "Selfuse - No Grid Charging": 1,
    "Timed Charge/Discharge - No Grid Charging": 3,
    "Selfuse": 33,
    "Timed Charge/Discharge": 35,
    "Off-Grid Mode": 37,
    "Battery Awaken": 41,
    "Battery Awaken + Timed Charge/Discharge": 43,
    "Backup/Reserve": 51,
}

```...shall backward compatibility be granted?


**Predbat version**
7.16.8

**Environment details**

- Solis with modbus integration

**Log file**
2024-03-19 22:30:06.562730 INFO pred_bat: Setting ECO mode as no discharge window planned                                                                                                                                                                                                                       
2024-03-19 22:30:06.570776 INFO pred_bat: Inverter 0 Adjust force discharge to False, change times from 00:00:00 - 00:00:00 to 00:00:00 - 00:00:00                                                                                                                                                              │
2024-03-19 22:30:06.575503 INFO pred_bat: Car charging from battery is off, next slot for car 0 is 03-19 22:30:00 - 03-19 23:00:00                                                                                                                                                                              │
2024-03-19 22:30:06.582246 INFO pred_bat: Inverter 0 Current Target SOC is 100 already at target                                                                                                                                                                                                                
[...]

PR #896, opened 2024-03-30
Fix to manage Solis storage modes name changed in Solax modbus integration version 2024.03.2

marcoaddario:
This fix solves issue #877 changing the Solis operational names to reflect them as changed in Solax modbus integration from version 2024.03.2

### DIFF
--- a/apps/predbat/predbat.py
+++ b/apps/predbat/predbat.py
@@ -1012,11 +1012,11 @@ INVERTER_DEF = {
 
 
 SOLAX_SOLIS_MODES = {
-    "Selfuse - No Grid Charging": 1,
+    "Self-Use - No Grid Charging": 1,
     "Timed Charge/Discharge - No Grid Charging": 3,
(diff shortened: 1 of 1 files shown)


springfall2008:
I'm going to re-do this as configurable to not break existing users

springfall2008:
Can you check the changes I made on main to see if they work for you?

marcoaddario:
I think the change you made is much better from a backward compatibility point of view.
I have tried the version you pushed on main but I got an exception:

`2024-04-01 17:32:10.811781 INFO pred_bat: ERROR: Exception raised 'PredBat' object has no attribute 'arg_arg'`

marcoaddario:
I think line 3420 should be 
`            modes_new = self.base.get_arg("solax_modbus_new", True)
`
instead of
`            modes_new = self.base.arg_arg("solax_modbus_new", True)

if this is the case ;-) , it works perfectly to me

Thanks


`

marcoaddario:
perfect to me.
Thanks

PR #987, opened 2024-04-21
Move constants out of predbat.py

Meatballs1:
There's no decent examples of large multifile projects that I can find. 

I've prefixed all files with `predbat_` as it looks like many files often get bundled into the same folder with appdaemon?


Meatballs1:
Releases are zipped up by GitHub so you have a single file to track/download?

The sanity check also needs tweaking I think

Meatballs1:
Example code for updating from releases. I guess it may not work for `main`. Not sure how it would work with the add-on version?

```
    def download_predbat_release_from_github(self, tag, filename):
        # Filename unused?
        self.log(f"BEN Update tag {tag}")
        url = f"https://github.com/springfall2008/batpred/archive/refs/tags/{tag}.zip"
        url = "https://github.com/Meatballs1/batpred/files/15067489/batpred-7.16.17.zip"
        with requests.get(url, headers={}, stream=True) as r:
            if r.ok:
                self.log(f"Write new version {filename}")
                with open(filename, "wb") as f:
                    shutil.copyfileobj(r.raw, f)
            else:
                self.log("WARN: Downloading Predbat file failed, URL {}".format(url))

springfall2008:
Yes, the issue is it needs to work from 'main' also so you can't use the zip. 

Instead in Predbat.py there is a list:

PREDBAT_FILES = ["predbat.py"]

What happens is the downloader will get predbat.py from github and then find all the files in the new list and download those, so it should work provided this list remains in predbat.py.

The appdaemon-predbat add-on has a fixed predbat.py file in it also, but this could be changed to use the .zip file as its just an initial download script.

I assume your apps.yaml change only relates to appdaemon restarting when a file changes? In which case its really not important for normal users and only needed for developers changing code?

Meatballs1:
I think that main can probably be catered for by grabbing the zip file at archive/refs/heads/main.zip

I believe the module definitions are just required for the dynamic reloading, but I would have thought that is required for the update too, unless you can signal to appdaemon to reload? I havent considered the overwriting the actual config until you've just brought it up, it could probably be put into another file if it needed to.

I think it would be a better approach to have a update.json file on the repo rather than the array in predbat.py as the update mechanism is too closely coupled. This would make sense as a first step as it would really help simplify moving the update code out. I think moving the update code out first would make more sense than constants (which I picked as an "easy" first step 🙃)

PR #996, opened 2024-04-23
Decouple updates file array from predbat.py

Meatballs1:
Very initial stage to start decoupling the update code. 

I've not tested this as its a bit of a pain to do so in branch setting tags, but to get your thoughts.

### DIFF
--- a/apps/predbat/predbat.py
+++ b/apps/predbat/predbat.py
@@ -12752,9 +12752,10 @@ class PredBat(hass.Hass):
             updates_json = download_predbat_file_from_github(tag, "updates.json", os.path.join(this_path, f"updates.json.{tag}"))
             if updates_json:
                 updates = json.loads(updates_json)
-
+                file_list = ", ".join(updates)
+                self.log(f"Predbat update files: {file_list}"
+                         
                 for file in updates:
-                    self.log("Predbat update files{}".format(f["file"]))
                     self.download_predbat_file_from_github(tag, file, os.path.join(this_path, file + "." + tag))
 
                 # Kill the current threads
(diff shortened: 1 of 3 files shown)


apps/predbat/predbat.py
```diff
@@ -12750,28 +12749,14 @@ def download_predbat_version(self, version):
...
-                        if file != "predbat.py":
-                            self.download_predbat_file_from_github(tag, file, os.path.join(this_path, file + "." + tag))
+            updates_json = download_predbat_file_from_github(tag, "updates.json", os.path.join(this_path, f"updates.json.{tag}"))
+            if updates_json:
+                updates = json.loads(updates_json)
+                file_list = ", ".join(updates)
+                self.log(f"Predbat update files: {file_list}"
+                         
+                for file in updates:
+                    self.download_predbat_file_from_github(tag, file, os.path.join(this_path, file + "." + tag))
```
Meatballs1: Think we need to go up to parent folder here so everything is relative

springfall2008:
Seems like a sensible approach, I can give it a try when I get time

Meatballs1:
I will fix up the paths and see if I can find a sensible way to test. Might be worthwhile adding an optional configuration base URL in apps.yaml that can be used for testing

Meatballs1:
I've introduced the `update_override_url` in the configuration which lets us point to a different fork/branch which helps a lot with testing. 

I've tested with it pointing at `https://raw.githubusercontent.com/Meatballs1/batpred/update_decouple/` and works for my test HAOS with Appdaemon add on and predbat manually installed. Also tested in my live dockerised appdaemon install.

I suspect there may be issues if users haven't maintained the directory structure as per the repo e.g. `apps/predbat/predbat.py`? You can really just dump `predbat.py` anywhere `apps`  directory and it doesn't really care, but we are attempting to overwrite files based on the repo structure.

springfall2008:
I've not forgotten this its just not made it to the top of the list yet:)

springfall2008:
This is now done on main, in a slightly different way but the files are split up :)

PR #1088, opened 2024-05-13
Fix: parameter formatting error for get_history

ncains:
Added sensor_id to parameter config to prevent crash when calling get_history


springfall2008:
Thanks for the fix, interesting you manage to hit this condition as I didn't think it would ever get there

springfall2008 (approved):

PR #1157, opened 2024-06-01
Documentation updates for Solcast and Sunsynk

springfall2008:


springfall2008:
> The 'pull frequency', hope you have some logic to randomise the pull time as solcast is rate limiting hobbyist accounts and I've had to set my automation to eg 01:19:35 to reduce load on the solcast server. Merging solcast in to predbat reduces the dependencies and install for new users although personally I like being able to also use the solcast data in dashboards and the energy dashboard and I'd probably prefer to remain as is

Its kind of random as the update time is related to when its starts rather than fixed times

gcoan:
Great, thanks. 

Are you populating any sensors with the solar forecast (eg for dashboards) or only pulling it in for predbat use?

springfall2008:
> Great, thanks.
> 
> Are you populating any sensors with the solar forecast (eg for dashboards) or only pulling it in for predbat use?

Yes its now creating sensor.predbat_pv_today and sensor.predbat_pv_tomorrow with the data.

dandwhelan:
Would it be possible to use the solar forecast data in ha energy module?

Issue #1294, opened 2024-07-09
Solis failed to update charge/discharge times


marcoaddario:
**Describe the bug**
My setup has 2 modbus clients for the inverter. one is the original solis datalogger and another one is homeassistant. In this setup the bus is available only to a client a time.

In some situation (e.g. when the inverter modbus is busy uploading data to the solis website) the button.push action times out due to a too much aggressive timing.
It happened few times in pas weeks (solis site more busier than usual???) that predbad fails to save the charge or discharge time and on the successive run do not retries because the HA params are already set by the previous run.
As workaround I have increased the `write_and_poll_sleep` params in config.py to 4 seconds and it seems ok, but I think that the charge /discharge times in HA should be restored to the previous state if the button.press service do not complete successfully. In this way predbat should try again on the next run. 

**Expected behavior**
in case of unsuccessful inverter charge/discharge time set all the params should be restored to the previous state and the write_and_poll_sleep can be increased a little bit more to face specific bus vulnerabilities.

**Predbat version**

8.2.4 (but is not s behavior specific to this version)

**Environment details**

- solis integration via modbus (as in https://github.com/alienatedsec/solis-ha-modbus-cloud)
- Standard HAOS installer


**Log file**
```
2024-07-08 23:30:20.182826 INFO pred_bat: Inverter 0 set new start time to 00:00:00                                                                                                                                                                                                                                                   
2024-07-08 23:30:24.234441 INFO pred_bat: Inverter 0 Wrote 00:00:00 to discharge_start_time, successfully now 00:00:00                                                                                                                                                                                                                
2024-07-08 23:30:28.274043 INFO pred_bat: Inverter 0 Wrote 0 to discharge_start_hour, successfully now 0                                                                                                                                                                                                                              
2024-07-08 23:30:32.306195 INFO pred_bat: Inverter 0 Wrote 0 to discharge_start_minute, successfully now 0                                                                                                                                                                                                                            
2024-07-08 23:30:32.307780 INFO pred_bat: Inverter 0 Set new end time to 00:00:00 was 23:31:00                                                                                                                                                                                                                                        
2024-07-08 23:30:36.346442 INFO pred_bat: Inverter 0 Wrote 00:00:00 to discharge_end_time, successfully now 00:00:00                                                                                                                                                                                                                  
2024-07-08 23:30:40.393290 INFO pred_bat: Inverter 0 Wrote 0 to discharge_end_hour, successfully now 0                                                                                                                                                                                                                                
[...]

PR #1324, opened 2024-07-23
fix time press

heyiamluke:
fixes datetime issue between button press and current time


marcoaddario:
Hi. thank you very much for the PR.
this is going to fix https://github.com/springfall2008/batpred/issues/1294 but I have noticed that in my case, even if I have patpred in the Europe/London timezone (from the apps.yaml), and everything in HA is shown in the Europe/London TZ, the datetime for the button to be pressed is returned in UTC.

this is a debug I have added to print the `time_pressed` value
`2024-07-25 21:00:48.327981 INFO pred_bat: Debug: attempt 0 - time pressed: 2024-07-25 20:00:44.144563+00:00`

I had to either convert `time_pressed` in local_timezone or get localtime in UTC.
I don't know what is the best way to make it more compatible. probably converting time_pressed to HA localtimezone...

PR #1369, opened 2024-08-16
Fix ha_key crash if not using addon

springfall2008:


gcoan:
Sorry about that bug, I did try to get the code tested in different configurations of running predbat and did fear that it might not work correctly in a docker container.

TBH though I don’t quite understand the change @springfall2008 , you’ve just moved the code that finds the slug name from predbat into ha.py, but otherwise it looks the same

PR #1415, opened 2024-09-01
Make all local paths relative. This will mean it supports being behind an ingress.

kingjpc:
This change means you can setup this https://github.com/lovelylain/hass_ingress with something like the following to get Predbat to appear in the home assistant sidebar.

```
#/homeassistant/configuration.yaml
ingress:
  predbat:
    url: http://homeassistant.local:5052
    ui_mode: normal
    title: Predbat
    icon: mdi:solar-power
```

### DIFF
--- a/apps/predbat/web.py
+++ b/apps/predbat/web.py
@@ -208,9 +208,9 @@ class WebInterface:
         else:
             text += "<h2>Logfile (All)</h2>\n"
 
-        text += '- <a href="/log">All</a> '
-        text += '<a href="/log?warnings">Warnings</a> '
-        text += '<a href="/log?errors">Errors</a><br>\n'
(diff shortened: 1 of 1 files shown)


springfall2008:
Hi @kingjpc the change looks fine to me, I can pull it no problem

I'd like to go the final step and have the 'ingress' enabled in the add-on itself, but when I tried this it doesn't open. Any ideas how I can make the build in HA add-on ingress work?

kingjpc:
I will take a look see if I can enable it locally. I think it’s a different process as the ingress docs imply you don’t listen on a port anymore.

springfall2008:
I posted here also: https://community.home-assistant.io/t/cant-get-ha-add-on-ingress-to-work-what-am-i-doing-wrong/766070

PR #1476, opened 2024-09-23
Templated UI

iainfogg:
CURRENTLY DRAFT, DO NOT MERGE

This PR moves all the HTML templates, plus the simpler HTML page content generation, out into HTML template files. These files are then filled out by Jinja2 (a templating library).

`web.py` will ultimately just be responsible for gathering the data that's needed for each web page, and passing that data to the templating engine so it can build out the HTML that's required for each page.

However, for now, some pages still have logic in `web.py` which generates HTML, but now this is just for filling the `content` block in the middle of each template. Over time, these will transition out.

### DIFF
--- a/apps/predbat/predbat.py
+++ b/apps/predbat/predbat.py
@@ -17,6 +17,7 @@ import sys
 from datetime import datetime, timedelta
 import hashlib
 import traceback
+from environment import is_appdaemon_environment
 
 # Import AppDaemon or our standalone wrapper
 try:
@@ -32,7 +33,7 @@ from multiprocessing import Pool, cpu_count, set_start_method
 import asyncio
 import json
 
-THIS_VERSION = "v8.4.9"
+THIS_VERSION = "v8.4.8"
 PREDBAT_FILES = ["predbat.py", "config.py", "prediction.py", "utils.py", "inverter.py", "ha.py", "download.py", "unit_test.py", "web.py"]
 from download import predbat_update_move, predbat_update_download, check_install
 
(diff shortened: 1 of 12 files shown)


apps/predbat/web.py
```diff
@@ -430,28 +431,36 @@ async def html_log(self, request):
             start_line = line[0:27]
             rest_line = line[27:]
 
-            if "error" in line_lower:
-                text += "<tr><td>{}</td><td nowrap><font color=#ff3333>{}</font> {}</td></tr>\n".format(lineno, start_line, rest_line)
+            if (
```
iainfogg: @springfall2008 this bit is where lines are filtered in/out - lines that shouldn't be seen are simply not added to the list that's passed through to the template.

iainfogg:
@springfall2008 I created a PR to make it easier to see the differences, and also to make discussing it easier - it was really hard to see the comments you'd made on the commit, there was no way to simply show all comments.

Also, you flagged a couple of lines as having merge issues, I couldn't see any changes to them in the PR when I created it. Let me know in this PR if you think there's still an issue.

Can you tag me in any comments please so I get notified, otherwise they're a bit hard to spot in the midst of everything else?

apps/predbat/predbat.py
```diff
@@ -17,6 +17,7 @@
 from datetime import datetime, timedelta
 import hashlib
 import traceback
+from environment import is_appdaemon_environment
```
springfall2008, line 20: You will need to add environment.py to the file list for Predbat to download

iainfogg: Thanks, sorry this has been stuck for a bit - been busy with other stuff @springfall2008

springfall2008:
I don't spot any obvious errors apart from the one comment I made about the file list

springfall2008:
I may return to this at some point but the implementation moved on a lot and it was hard to merge

Issue #1652, opened 2024-11-24
Octopus free sessions incorrectly set negative agile rate to zero 


DrJohnM61:
**Describe the bug**
Agile rates that are negative are incorrectly set to zero when free session is active. 
**Expected behaviour**
According to the Octopus FAQ, the negative rate is still relevant and therefore, the negative plan rates should be retained as published by Octopus and only positive rates should be set to zero.

From the Octopus FAQ:

> Can I take part as an Agile customer?
> 
> Whether it's worth taking part largely depends on what the energy prices are doing at that time.
> 
> If it's a negative price period, your power is already better than free, so you'll get billed your power at your negative prices and won't be credited for any Free Electricity.
> 
> However, if your price is above zero, you're welcome to take part and benefit from our super-green sessions.

**Predbat version**

8.7.2

**Environment details**

- Givenergy AIO + Gateway
- Standard HAOS installer
- Anything else?

**Screenshots**
<img width="467" alt="image" src="https://github.com/user-attachments/assets/0d9fdb7b-6542-4787-9457-53fd2985cb69">

As can be seen above, the 8:00 slot is -2.9p however, the free electricity detection has set the rate to zero. However, the 8:30 agile slot is positive and has been correctly set to zero.

<img width="399" alt="image" src="https://github.com/user-attachments/assets/eacde268-2a58-403a-b1e6-9c6bbc3d68de">

**Log file**

springfall2008:
Fix is now on 'main'

PR #1653, opened 2024-11-24
Set minimum of free or negative agile rate when applying Octopus Free sessions

Mikebert4:
#1652 

I think this solves it? First time creating a PR so forgive me if I've missed a trick or three!


springfall2008:
I've merged this into a larger merge I made to 'main' so we can discharge this PR now

PR #1869, opened 2025-01-11
Max charge rate temperature curve adjustments

springfall2008:


WyndStryke:
If it helps, I found this information on the temperature vs charge rates on the Fox batteries:

https://foxesscommunity.com/viewtopic.php?p=6538#p6538

>  It appears the charge current (MaxChargeCurrent) switches at particular (CellTempLow) temperatures: 2A at 5.9°C and below; 5.5A for 6.0°C to 10.9°C; and 10A from 11°C. 

Not clear exactly **which** batteries the poster is referring to, though.

PR #1880, opened 2025-01-19
Add repeat to service templates

springfall2008:


jefft4:
What does the repeat on service templates change in terms of how things behave, @springfall2008 ?  Wondering if I need to remove repeat modbus sends on the other side of the service calls now.

PR #1942, opened 2025-02-01
New option octopus_intelligent_consider_full

springfall2008:


cyberkryten:
I've just enabled this to test and currently my car is charging (87% with limit at 90%) but Predbat shows no charge in the Car kWh column.  Without the switch is shows charge throughout the entire IOG low rate period - IOG uses my Zappi charger so doesn't know the Car SOC which is I think what this flag is for.

As a result, Predbat status is Exporting so it thinks it can export to the Grid, which it can't as the car+house is sinking the 8kW output of the inverter.

I can't see any Docs but I've read the code comments - have I misunderstood what this is for?

Screenshot 2025-02-03 at 13 31 30

Issue #1994, opened 2025-02-10
Add a Home Assistant Entity to Manually Force a Tariff Comparison


stefanwest1:
****Is your feature request related to a problem? Please describe.**
Currently, there is a "Force Comparison" button available in the Web UI. However, for users running PredBat in an AppDaemon container, the Web UI is not accessible. This means there is no easy way to trigger a tariff comparison from within Home Assistant.

**Describe the solution you'd like**
I would like a dedicated Home Assistant entity (e.g., input_boolean.predbat_force_comparison or button.predbat_force_comparison) that allows manually triggering a tariff comparison.

This entity should be available in Home Assistant so that automations, scripts, and UI interactions can trigger a tariff comparison.
When activated, it should set self.compare_tariffs = True inside PredBat, ensuring the comparison runs in the next cycle.

**Describe alternatives you've considered**

- Manually restarting the AppDaemon container (which forces a comparison on startup after manually setting self.compare_tariffs = True on restart, but is inefficient).

- Using Node-RED to call a custom API (but this requires modifying the AppDaemon script).

- Using AppDaemon’s Web UI (but it is not accessible for users running PredBat inside a Home Assistant appdaemon container).

**Additional context**
This feature would be extremely useful for those of us running AppDaemon as a separate container, as it would allow direct control from Home Assistant without relying on external modifications or workarounds.

stefanwest1:
Just to add, as a workaround for now. Under def initialize(self) in predbat.py I added

`        self.listen_state(self.force_comparison_from_ha, "input_button.predbat_force_comparison")`

And then at the bottom of predbat.py

```
    def force_comparison_from_ha(self, entity, attribute, old, new, kwargs):
        """
        When the Home Assistant button is pressed, set self.compare_tariffs = True.
        """
        self.log("Home Assistant button pressed: Forcing tariff comparison")
        self.compare_tariffs = True
```

And finally I just manually created input_button.predbat_force_comparison in HA and this seems to work well for now

stefanwest1:
Just another little update. In compare.py I also changed 

`end_record = int((my_predbat.minutes_now + 24 * 60 + 29) / 30) * 30 - my_predbat.minutes_now`

to:

`        minutes_until_midnight = (24 * 60) - my_predbat.minutes_now
        end_record = int((minutes_until_midnight + (24 * 60) + 29) / 30) * 30 - my_predbat.minutes_now`


[...]

PR #2019, opened 2025-02-15
Compare active entity

springfall2008:
https://github.com/springfall2008/batpred/issues/1994


pantherale0:
I think this should have been a button entity and a separate sensor storing the current state. A switch usually implies that a user can turn something off and on whereas this only really does anything if you turn it on.

PR #2078, opened 2025-03-06
Documentation updates

gcoan:
Documentation updates, tidy-up's and clarifications in response to recent issues


springfall2008:
Looks good will merge.

BTW: You can override house load scaling using the energy rates override with an increment of 0 pence and a load scaling factor.

springfall2008 (approved):

PR #2092, opened 2025-03-11
Feature/ems support

mgazza:
Complete punt to add ems auto-detect support.
needs testing

### DIFF
--- a/apps/predbat/gecloud.py
+++ b/apps/predbat/gecloud.py
@@ -908,14 +908,16 @@ class GECloudDirect:
         """
         device_list = await self.async_get_inverter_data_retry(GE_API_DEVICES)
         serials = []
-        if device_list is not None:
-            for device in device_list:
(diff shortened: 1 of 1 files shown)


apps/predbat/gecloud.py
```diff
@@ -904,20 +932,29 @@ async def async_get_device_info(self, serial):
...
+            inverter = device.get("inverter", {})
+            serial = inverter.get("serial", None)
+            model = inverter.get("info", {}).get("model", "").lower()
+            batteries = inverter.get("connections", {}).get("batteries", [])
+            if serial:
+                serial = serial.lower()
+                if "plant ems" in model:
+                    result["ems"] = serial
+                elif batteries:
+                    result["battery"].append(serial)
```
springfall2008, line 956: The list here isn't batteries but a list of inverters that have batteries, I'll take the PR and then rename it

PR #2117, opened 2025-03-17
When db_enable use the db for settings

mgazza:
Change the UI elements to use the db for settings rather than storing them in a seporate json file

### DIFF
--- a/apps/predbat/ui.py
+++ b/apps/predbat/ui.py
@@ -332,7 +332,7 @@ class UI:
         Load the current configuration from a json file
         """
         if self.ha_interface.db_primary:
-          return
+            return
 
         filepath = self.config_root + "/predbat_config.json"
         if os.path.exists(filepath):
@@ -362,7 +362,7 @@ class UI:
         """
 
         if self.ha_interface.db_primary:
-          return
+            return
 
         filepath = self.config_root + "/predbat_config.json"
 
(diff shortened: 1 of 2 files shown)


apps/predbat/ui.py
```diff
@@ -331,6 +331,9 @@ def load_current_config(self):
         """
         Load the current configuration from a json file
         """
+        if self.ha_interface.db_enable:
```
springfall2008: I think this should be db_primary as you can have db_enable without db_primary meaning it keeps the database but still has home assistant also

apps/predbat/userinterface.py
```diff
@@ -852,9 +855,12 @@ def load_user_config(self, quiet=True, register=False):
                 continue
 
             # Get from current state, if not from HA directly
-            ha_value = item.get("value", None)
-            if ha_value is None:
+            if self.ha_interface.db_enable:
                 ha_value = self.load_previous_value_from_ha(entity)
```
springfall2008, line 859: Why ignore current state?

mgazza: Maybe I'm not understanding correctly. `Item` is  from `self.CONFIG_ITEMS`.  `load_previous_value_from_ha` uses the current state underneath. Are you saying Item will have the current state?

springfall2008: If it has current state then you can use it, this state will be updated if its modified inside Predbat and then the change to the database is scheduled for the next DB update.

mgazza: I'll wind this block back and give it a test in the morning :D

Issue #1964, opened 2025-02-04
Predbat not predicting any battery usage


TonyHoyle:
**Describe the bug**
Predbat consistently suggests I'll use at most 0.005w of battery which means it never predicts any battery usage.  Other values seem reasonable and historical battery usage is plotted correctly.

Used the fox.yaml from https://raw.githubusercontent.com/springfall2008/batpred/main/templates/fox.yaml with some minor adjustments for Octopus and setting the soc_max to 10.36 to match my battery.

**Expected behaviour**
Some prediction of usage, even it it's not accurate yet due to shortage of data.

**Predbat version**

8.13.3

**Environment details**

- Fox ESS H1 G2, 2xEP5 batteries
- HAOS installer

**Screenshots**

![Image](https://github.com/user-attachments/assets/907b428a-06dc-4d73-859d-38bb93e72cf9)

![Image](https://github.com/user-attachments/assets/1ab71953-24e6-4000-81fb-e9dfa8f916e8)

**Log file**
Can you capture a log file from the time of the issue, debug mode is not normally required.

**Predbat debug yaml file**
[predbat_debug.yaml.txt](https://github.com/user-attachments/files/18658359/predbat_debug.yaml.txt)

Enable debug mode and let Predbat run for one cycle. Look in the 'debug' directory under the Predbat add-on and take the .yaml file for the time period in question.
Rename the predbat_debug_XX_XX_XX.yaml to predbat_debug_XX_XX_XX.yaml.txt and upload it to Github. This will allow your plan to be reproduced and also stores all your settings for review.

gcoan:
have you got some of the fox sensors set to kW rather than Watts or vice versa?

If you look in the apps.yaml documentation it explains what each sensor needs to be set to

TonyHoyle:
I don't see how that could be changed.. the sensors are what they are.. it's up to predbat to be compatible with them.  This is the example config from the documentation, which should work.

WyndStryke:
When I look at the power chart, the battery line looks normal to me.  So I'm not sure what the difference is between your config and mine.

![Image](https://github.com/user-attachments/assets/32b0b7f4-abf6-4293-aff5-d5503dbfede1)

> I don't see how that could be changed..

[...]

PR #2174, opened 2025-03-31
Improve fox template

simoncozens:
As detailed in #1964, Fox inverter users need:

1) Tony's W/kW conversion routines
2) Correct entity configuration to go through this conversion
3) Support for freeze charge

Confirmed working on my H1-G2.

### DIFF
--- a/templates/fox.yaml
+++ b/templates/fox.yaml
@@ -1,3 +1,8 @@
+# Note that you will need to install the W/kW conversion automations
+# from https://github.com/TonyHoyle/foxess_predbat - place the
+# predbatt.yaml file from that repository into your HA configuration
+# directory, and edit the HA configuration.yaml as show in the README.
+
 pred_bat:
   module: predbat
   class: PredBat
@@ -85,7 +90,7 @@ pred_bat:
     clock_time_format: "%Y-%m-%d %H:%M:%S"
     write_and_poll_sleep: 2
     has_time_window: False
-    support_charge_freeze: False
+    support_charge_freeze: True
     support_discharge_freeze: False
 
   # Services to control charging/discharging
@@ -101,6 +106,14 @@ pred_bat:
     service: select.select_option
     entity_id: select.work_mode
     option: "Force Discharge"
+  discharge_stop_service:
+    service: select.select_option
+    entity_id: select.work_mode
+    option: "Self Use"
+  discharge_freeze_service:
+    service: select.select_option
+    entity_id: select.work_mode
+    option: "Feed-in First"
 
   #
   # Run balance inverters every N seconds (0=disabled) - only for multi-inverter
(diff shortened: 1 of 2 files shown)


WyndStryke:
This is mine:


[Fox_template_work_in_progress.yaml.txt](https://github.com/user-attachments/files/19535295/Fox_template_work_in_progress.yaml.txt)

Among the bigger changes from the original are the following:

- Configuring the reserve functionality (has_reserve_soc: True and reserve: - number.min_soc_on_grid)
- Charge freeze and discharge freeze
- Configuring soc_max via a sensor not a literal
- Configuring battery_min_soc via a sensor
- Configuring temperature charge curves & SoC charge curves (but these will vary for everyone)
- Battery temperature & temperature history (battery_temperature: - sensor.bms_cell_temp_low, battery_temperature_history: - sensor.bms_cell_temp_low)

springfall2008:
Looks good, although it would probably be better to put the template stuff into the Predbat repo too?

templates/fox.yaml
```diff
@@ -85,7 +90,7 @@ pred_bat:
     clock_time_format: "%Y-%m-%d %H:%M:%S"
     write_and_poll_sleep: 2
     has_time_window: False
-    support_charge_freeze: False
+    support_charge_freeze: True
```
s0129, line 93: Is this correct? As is enabling charge_freeze but later adding the discharge_freeze_service and not the charge_freeze_service?

WyndStryke: > Is this correct? As is enabling charge_freeze but later adding the discharge_freeze_service and not the charge_freeze_service?

I think it would be good to merge in some or all of the following:

```

...
    has_reserve_soc: True
    support_charge_freeze: true
    support_discharge_freeze: true
...
  discharge_freeze_service:
    service: select.select_option
    entity_id: select.work_mode
    option: "Feed-in First"
  charge_freeze_service:
    service: select.select_option
    entity_id: select.work_mode
    option: "Back-up"
...

  # Reserve SoC is used for Freeze Charge and Hold Charge modes.  This must never be set to less than 10%
  # (to match 90% DoD) and preferably a little higher to be safe.  
  # Also see the predbat setting input_number.predbat_set_reserve_min which should be set to 10 minimum, 
  # and input_number.predbat_best_soc_min which needs to be set >= kWh equivalent of that.
  reserve:
    - number.min_soc_on_grid
...
  battery_min_soc:
    - sensor.min_soc
...
  # sensor.bms_kwh_remaining relies on recent firmware, replace with literal for the kWh nominal capacity if unavailable
  soc_max:
    - sensor.bms_kwh_remaining
...
  # sensor.bms_cell_temp_low relies on recent firmware, comment out if unavailable
  battery_temperature:
    - sensor.bms_cell_temp_low
  battery_temperature_history:
    - sensor.bms_cell_temp_low
...
  battery_scaling:
    - 1.0
...
  # Carbon Intensity data from National grid
  carbon_intensity: 're:(sensor.carbon_intensity_uk)'


```

Issue #2188, opened 2025-04-02
House battery being used when car is charging despite "Car charging hold" being set


stevedundee2:
**Describe the bug**
My house battery is being used when car is charging despite "Car charging hold" being set.  

The scenario is a little complicated. I plugged the car in to take advantage of solar but Intelligent Octopus Go decided it would like to charge the car immediately. So the car is getting power from 3 sources: solar, battery (never more than 2kW for some reason) and grid.

For added confusion, the battery is sometimes being charged (for occasional 30 minute slots) but mostly discharging.

I tried reducing the self-sufficiency metric from 5.0 (a recent change) to 3.0 (which I've used for a long time) but this didn't make a discernable difference.

**Expected behaviour**
The battery should not discharge when the car is being charged from the grid

**Predbat version**

8.17.3

**Environment details**

- GE 9.5kW with hybrid inverter and some solar panels
- Standard HAOS installer 
- The car is an ID.3. The electricity supplier is Octopus Intelligent Go

**Screenshots**

![Image](https://github.com/user-attachments/assets/8bff8774-551d-40aa-b39e-b3e1a7562b3b)

**Log file**

[predbat.log](https://github.com/user-attachments/files/19568371/predbat.log)

**Predbat debug yaml file**

[predbat_debug_14_10_00.yaml.txt](https://github.com/user-attachments/files/19568584/predbat_debug_14_10_00.yaml.txt)

Note that the car reached its target SOC in this cycle so it had stopped charging by 2.10pm

stevedundee2:
Predbat is on "Control charge" rather than "Control charge and discharge" to stop the car using a forced export. This is toggled in automations.

springfall2008:
Predbat gets the car slots from Octopus, are you sure they just haven't updated yet?

stevedundee2:
I suppose the behaviour could be explained by Octopus repeatedly changing its mind about the charging slots and predbat needing to play catch up. It has been tending to do that recently but I don't think it was in this case. The plan knew about the charging slots when it was in Demand status and those slots didn't seem to change.

Perhaps one to monitor. It doesn't often happen that the car is getting charged by solar diversion and even rarer that Octopus decides to charge during the day.

Steve

Jorthax:
I literally had this event yesterday, but I think it's to do with the speed at which information is available from Octopus via the API and PredBats consideration of it.

[...]

Issue #2184, opened 2025-04-01
Documentation - LuxPower


brickatius:
I've tried to create 'input_text.xxxxxxx' helpers for charge and discharge start and end times - for example as in the 'luxpower.yaml' template as follows:

charge_start_time:
    - input_text.lux_batpred_charge_start

but unless I've got something completely wrong these 'input_text'  don't seem to be writable by Predbat and an error occurs. Wouldn't it be better to use the actual predbat sensor names like this?

charge_start_time:
    - sensor.predbat_LuxPower_0_charge_start_time

You can then write a Home Assistant automation to target the LuxPower inverter start and end time sensors, as follows, whenever the predbat sensor changes. 

alias: Charge Time Change - Start
description: ""
triggers:
  - entity_id:
      - sensor.predbat_LuxPower_0_charge_start_time
    trigger: state
conditions: []
actions:
  - metadata: {}
    data:
      time: "{{ states('sensor.predbat_LuxPower_0_charge_start_time')}}"
    target:
      entity_id: time.lux_ac_charge_start1
    enabled: true
    action: time.set_value
mode: single

It works a treat and you can see the new charge and discharge times in the LuxPower app and online monitoring.

gcoan:
There's certainly something incomplete with the documentation because the template defines using input_text.lux_batpred_discharge_start etc as you say, but then there's no guidance as to what then needs to happen when these input_text's are written to.

Predbat should be able to write to input_text's

But looking at your automation, did you try setting the time.lux_ac_charge_start1 entity id's directly in apps.yaml so predbat can write directly to that entity

Per the documentation, predbat can write to time entities for charge/discharge start/end time https://springfall2008.github.io/batpred/apps-yaml/#schedule although this is a fairly new feature and maybe wasn't there when the luxpower template was developed.

Can you give it a go and if it works, let me know what changes are required to the documentation

Also, if there are other automations required or changes to the template, please flag it up

brickatius:
Hi Geoffrey, 

I’ll certainly give it a go. 
I did have Predbat set up very briefly on another machine which for reasons unrelated to Predbat died without backup. I bought another one and started from scratch. As it was only a couple of weeks ago I do remember having this:

charge_start_time:
   - time.lux_ac_charge_start1

[...]

Issue #2179, opened 2025-03-31
Zero load predicted Tuesday Mornings


bfayers:
**Describe the bug**
For some reason I'm getting zero load predicted between 03:30 and 08:00. It did the same thing last week, but it was between 02:30 and 07:00 (time difference due to clock change I guess).

In my predbat config I have:
```yaml
days_previous:
  - 7
  - 14
```

If I comment out the line with 14 then it goes back to what I would expect, predicting the load properly for that time. The strange thing is if I go and look at the history data for sensor.load_power in HASS it's as expected for this time window over the past 2 tuesdays.

**Expected behaviour**
Load to be predicted normally, as it is every other day and half hour period.

**Predbat version**

8.17.2

**Environment details**

- Inverter and battery setup
  - Fox H1 G2 & 2x Fox EP11
- Standard HAOS installer or Docker
  - HAOS

**Screenshots**

![Image](https://github.com/user-attachments/assets/0a81e13c-190c-4271-9fe3-2d20cf5de350)

**Log file**

[predbat.log](https://github.com/user-attachments/files/19540024/predbat.log)

**Predbat debug yaml file**

[predbat_debug_19_50_00.yaml.txt](https://github.com/user-attachments/files/19540048/predbat_debug_19_50_00.yaml.txt)

gcoan:
> if I go and look at the history data for sensor.load_power in HASS it's as expected for this time window over the past 2 tuesdays.

what do you have your purge_keep_days: 9 set to in configuration.yaml?

If you have not set this then by default HA will only store 10 days of history and so the previous Tuesday will be expunged.

When you look at the sensor history in HASS it will show you the sensor pulled from both the sensor history AND long term statistics.  Predbat only uses the sensor history.

bfayers:
> > if I go and look at the history data for sensor.load_power in HASS it's as expected for this time window over the past 2 tuesdays.
> 
> what do you have your purge_keep_days: 9 set to in configuration.yaml?


purge_keep_days is set to 14. This is only happening between those hours on Tuesday mornings.

gcoan:
try increasing purge_keep_days to 15 to ensure that there is always a full 14 days of history

bfayers:
I'll increase it but I guess I'll be waiting until next monday night / tuesday to see if it's made any difference

gcoan:
[...]

Issue #2095, opened 2025-03-12
Support for Kostal plenticore inverter


mbuhansen:
**Is your feature request related to a problem? Please describe.**
I m try to use Predbat to control my Kostal Plenticore inverter, i have made a APPS. yaml and testing now.

I can get kostal to charge and discharge by home assistat service call. i have also made a freeze charge button that work, inverter can charge battery but not use from battery.
When Predbat say hold in state, does it stop charge and discharge from battery completely..?? Can i call this by service api..??

The inverter is control by modbustcp, from home assistant, so i have made button i HA that start inverter to charge, dicharge or hold/lock battry use.
when inverter dont get any modbus message the battery control went back to internal use, and can both charge and discharge battery.

load_power sensor does is this power from grid or hous load both from griad and inverter..?? I have now set is to the last it measure the home power inverter calculate.

so does this apps yaml look correct..??

https://pastebin.com/unMeY171

mbuhansen:
Here the plan change from freeze charge to demand, but predbat dont call charge stop service..?? how can i be shure it turn off my no discharge button i use as freze charge function..??



2025-03-12 05:35:06.813198: Import threshold used for optimisation was 230.9p
[...]

PR #2220, opened 2025-04-06
Documentation updates, LuxPower and Solcast templates, ha.py

gcoan:
#2049 added predbat_active check to error detector
Update new Ohme statii for plugged in
#1522 SolarEdge multi batteries & template corrections
Added links to each inverter setup
Improve solcast entity explanation in apps.yaml
Comment out geserial2 lines if 1 inverter
#1872 correct yaml indentation, expand on parent/child yaml explanation
#2184 luxpower template sensors and luxpower.yaml tidy up
#2179 Warning about setting car_charging_energy
#2095 rates_gas for iboost (non Octopus customers)
#2184 luxpower template sensors in kWh, charge/discharge time sensors
#2188 add watch_list explanation to apps.yaml
#2201 #1206 Add clip kWh to output data description
ha.py suppress supervisor error msg for docker users, ensure slugid is retrieved when ha_url is set
Don't display clipped kWh value if clip amount is zero


apps/predbat/ha.py
```diff
@@ -507,22 +514,27 @@ def call_service(self, service, **kwargs):
...
+        :param core: True is this is a call to HA Core, False if it is a Supervisor call
+        :param silent: True if warning message from the API call is to be suppressed
         :return: The response from the API.
         """
         if core:
             url = self.ha_url + endpoint
+            key = self.ha_key
         else:
-            url = self.ha_url.replace("/core", "") + endpoint
+            # make an API call to the Supervisor using the supervisor token
```
springfall2008, line 527: I don't think this is right if you look at the use all calls to api_call are set with core True apart from the /addons/self/info one

gcoan: It is correct, but probably needs a bit of explaining ....

By default, if ha_url is not set in apps.yaml then ha_url defaults to http://supervisor/core

Most API calls are with core=true so are to http://supervisor/core/[key] 

However, on startup ha.py tries to determine the add-on slugid that predbat is running under. That has to be a call to the supervisor, so api_call is made with core=false.  At present the call is made to ha_url with the '/core' removed and the key at the end, so with ha_url not set in apps.yaml it will call http://supervisor/[key] e.g. http://supervisor/addons/self/info

The first problem comes if you are running HAOS and have set ha_url in apps.yaml.  I have this setup, have set ha_url so that I avoid the websocket timeouts that occasionally occur.
My ha_url is set to 'http://192.168.1.90:8123' - which is the HA IP address & port

The calls to api_call with core=true are thus made to http://192.168.1.90:8123/[key] and all is good

But in startup (when core=false) the code tries to make an API call to http://192.168.1.90:8123/addons/self/info which always fails with a 500 error because /addons/self/info is a supervisor API call not a HA API call.

So first part of this change is to not try to work out the supervisor call by substringing ha_url but instead to explicitly call http://supervisor/[key] if core=False

This ensures that the call to determine slug id correctly is made to the supervisor if ha_url is set in apps.yaml

Second thing changed in this is that if you try to make a supervisor API call with a docker install, it always fails as there is no supervisor running.  So I have added a silent=False option to api_call, and ONLY in this call to determine the slug id is silent=True.  This supresses the warning message you'd otherwise always get with docker installs when they try to find the slug id.

And third thing is I have changed the sequence of api calls in ha.py.  Currently it is:

a. check if able to talk to HA, api_call("/api/services")
b. determine slug id from supervisor, api_call("/addons/self/info", core=False)
c. open websocket to HA

Some people have reported errors on c, I believe because you call HA then the supervisor, then back to HA again.

I have resequenced the code to be:

1. determine slug id from supervisor, api_call("/addons/self/info", core=False)
2. check if able to talk to HA, api_call("/api/services")
3. open websocket to HA

And this then doesn't fail at c any more

Long explanation, sorry !

Issue #2247, opened 2025-04-15
8.18.2 - Completely Broken


ProximusAl:
Predbat crashed. Restarting in 5 seconds...
Starting Predbat standalone...
starting startup.py...........
Predbat files are installed correctly for version v8.18.2
**** Starting Standalone Predbat ****
2025-04-15 23:11:09.538331: Loading apps.yaml
2025-04-15 23:11:09.547298: Error: Exception raised can only concatenate str (not "NoneType") to str
2025-04-15 23:11:09.547825: Error: Traceback (most recent call last):
  File "/config/predbat.py", line 1189, in initialize
    self.ha_interface = HAInterface(self)
                        ^^^^^^^^^^^^^^^^^
  File "/config/ha.py", line 86, in __init__
    res = self.api_call("/addons/self/info", core=False, silent=True)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/config/ha.py", line 532, in api_call
    "Authorization": "Bearer " + key,
                     ~~~~~~~~~~^~~~~
TypeError: can only concatenate str (not "NoneType") to str
2025-04-15 23:11:09.548003: Error: set_state_wrapper - No HA interface available
2025-04-15 23:11:09.548015: Info: record_status Error: Exception raised can only concatenate str (not "NoneType") to str
Error: Failed to initialize predbat can only concatenate str (not "NoneType") to str
Traceback (most recent call last):
  File "/config/hass.py", line 49, in main
    p_han.initialize()
  File "/config/predbat.py", line 1238, in initialize
    raise e
  File "/config/predbat.py", line 1189, in initialize
    self.ha_interface = HAInterface(self)
                        ^^^^^^^^^^^^^^^^^
  File "/config/ha.py", line 86, in __init__
    res = self.api_call("/addons/self/info", core=False, silent=True)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/config/ha.py", line 532, in api_call
    "Authorization": "Bearer " + key,
                     ~~~~~~~~~~^~~~~
TypeError: can only concatenate str (not "NoneType") to str
2025-04-15 23:11:09.548362: Stopping Predbat
Bootstrap Predbat
Startup
Shutdown, sleeping 30 seconds before restarting
Predbat crashed. Restarting in 5 seconds...

[...]

PR #2250, opened 2025-04-16
#2247 suggested change to allow docker based predbat and hass calls

asnaedae:
fix for https://github.com/springfall2008/batpred/issues/2247

### DIFF
--- a/apps/predbat/ha.py
+++ b/apps/predbat/ha.py
@@ -528,6 +528,13 @@ class HAInterface:
             url = "http://supervisor" + endpoint
             key = os.environ.get("SUPERVISOR_TOKEN", None)
 
+            if not key:
+                if not silent:
+                    self.log("Warn: api_call failure, key not set")
+                    self.api_errors += 1
+                data = None
+                return data
+
         headers = {
             "Authorization": "Bearer " + key,
             "Content-Type": "application/json",


ProximusAl:
I've tested this and it works great.  Thanks...

apps/predbat/ha.py
```diff
@@ -528,6 +528,12 @@ def api_call(self, endpoint, data_in=None, post=False, core=True, silent=False):
             url = "http://supervisor" + endpoint
             key = os.environ.get("SUPERVISOR_TOKEN", None)
 
+            if not key:
+                if not silent:
+                    self.log("Warn: api_call failure, key not set")
+                    self.api_errors += 1
+            data = None
+            return data
```
gcoan: The indentation is not quite right on these two lines, it needs to be indented one further tab stop so that it's only returning if key is None

As it is at the moment it will always return without ever getting the supervisor info (which should be the happy path on HAOS)

gcoan (changes requested):

apps/predbat/ha.py
asnaedae: I'm a little confused here then, currently indenting the return further still causes the exception to be thrown on the supervisor call when using core+docker 

```
urllib3.exceptions.NameResolutionError: <urllib3.connection.HTTPConnection object at 0x798cebc1b770>: Failed to resolve 'supervisor' ([Errno -3] Try again)
```

I dont see any case where api_call with core=True is being called, so it'll always default to attempting the http://supervisor/ call and the exception isn't being suppressed in the docker/core path

ProximusAl:
This is what ChatGPT fixed mine to:

            # make an API call to the Supervisor using the supervisor token
            url = "http://supervisor" + endpoint
            key = os.environ.get("SUPERVISOR_TOKEN", None)

            if not key:
                if not silent:
                    self.log("Warn: api_call failure, key not set")
                self.api_errors += 1
                data = None
                return data

        headers = {
            "Authorization": "Bearer " + key,
            "Content-Type": "application/json",
            "Accept": "application/json",

gcoan:
> This is what ChatGPT fixed mine to:
> 
> 
> 
>             # make an API call to the Supervisor using the supervisor token
> 
>             url = "http://supervisor" + endpoint
> 
>             key = os.environ.get("SUPERVISOR_TOKEN", None)
> 
> 
> 
>             if not key:
> 
>                 if not silent:
> 
>                     self.log("Warn: api_call failure, key not set")
> 
>                 self.api_errors += 1
> 
>                 data = None
> 
>                 return data
> 
> 
> 
>         headers = {
> 
>             "Authorization": "Bearer " + key,
> 
>             "Content-Type": "application/json",
> 
>             "Accept": "application/json",
> 
> 

That looks correct

apps/predbat/ha.py
asnaedae: Not sure what was going on whilst making the changes, but seems to be working with the correct indents, change pushed

gcoan (approved):
Thanks

springfall2008:
I've already included this fix on 'main' for the .3 release

PR #2327, opened 2025-05-04
Support for Solis invertiers with split charge/discharge update buttons

rszemeti:
Some solis inverters have two buttons for charge/discharge time updates, the PR adds support for that while maintaining the combined "update_charge_discharge" feature.

### DIFF
--- a/apps/predbat/inverter.py
+++ b/apps/predbat/inverter.py
@@ -2307,7 +2307,7 @@ class Inverter:
 
             if old_charge_schedule_enable == "off" or old_charge_schedule_enable == "disable":
                 self.base.log("Inverter {} Turning on scheduled charge".format(self.id))
-    
+
     def press_and_poll_button(self):
         """
         Press charge/discharge update button(s) for the inverter.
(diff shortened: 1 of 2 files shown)


apps/predbat/inverter.py
```diff
@@ -2307,28 +2307,67 @@ def adjust_charge_window(self, charge_start_time, charge_end_time, minutes_now):
...
+            if charge_button:
+                if not self._press_single_button_and_poll(charge_button):
+                    success = False
+    
+            if discharge_button:
+                if not self._press_single_button_and_poll(discharge_button):
+                    success = False
+    
+            if not charge_button and not discharge_button:
+                self.base.log(f"Error: No update button found for inverter {self.id}")
```
springfall2008: Not sure this should be an error, before if the button wasn't set it was ignored?

rszemeti: That seems reasonable,  not all inverters need this, so if no buttons are set, just carry on, shall I fix on my side and re-issue a PR?

PR #2392, opened 2025-05-25
Reduce the number of tables needed,

mgazza:
Introduce migrations and reduce the number of tables needed.

### DIFF
--- a/apps/predbat/db_engine.py
+++ b/apps/predbat/db_engine.py
@@ -33,8 +33,8 @@ class DatabaseEngine:
         """
         # Use SHA256 hash and take first 8 bytes as integer for 64-bit signed integer
         # This gives us a deterministic ID that's the same across all instances
-        hash_bytes = hashlib.sha256(entity_id.encode('utf-8')).digest()[:8]
-        return int.from_bytes(hash_bytes, byteorder='big', signed=True)
+        hash_bytes = hashlib.sha256(entity_id.encode("utf-8")).digest()[:8]
+        return int.from_bytes(hash_bytes, byteorder="big", signed=True)
 
     def _close(self):
         """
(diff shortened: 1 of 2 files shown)


springfall2008:
But I probably wouldn't include migrations, I doubt anyone is using the DB apart from us right now, just start a new DB. If it was included the files would have to be added to the download list etc.

apps/predbat/db_engine.py
```diff
@@ -35,175 +36,159 @@ def _close(self):
...
+
+        attributes_json = json.dumps(attributes)
+
+        # Upsert into state
+        self.db_cursor.execute(
+            """
+                               INSERT INTO state (entity_id, state, attributes, updated_at)
+                               VALUES (?, ?, ?, ?, ?)
+                               ON CONFLICT(entity_id) DO UPDATE SET
+                                                                    state=excluded.state,
```
springfall2008: I don't see any reference to excluded

mgazza:
> But I probably wouldn't include migrations, I doubt anyone is using the DB apart from us right now, just start a new DB. If it was included the files would have to be added to the download list etc.

happy to drop the migration logic for now and just require a clean DB. I’ll remove the migration scripts and simplify accordingly unless we want to support older data later on.

apps/predbat/db_engine.py
mgazza: `excluded` is a special table alias in SQLite that refers to the row that was proposed for insertion but caused a conflict. So in:

```sql
ON CONFLICT(entity_id) DO UPDATE SET
    state = excluded.state
```

…it means we update the existing row's `state` to the `state` value from the conflicting row. You can see more in the SQLite docs under [UPSERT](https://www.sqlite.org/lang_UPSERT.html) — the `excluded.` keyword is described in the "ON CONFLICT DO UPDATE" section.

springfall2008:
I remember now why you want a states table also, as the larger data items you need to keep in the latest state but you don't want to keep the history for them. How do you handle this with one table?

mgazza:
> I remember now why you want a states table also, as the larger data items you need to keep in the latest state but you don't want to keep the history for them. How do you handle this with one table?

Thats a good point, having the state history and the entity(with the latest data) is fine.

mgazza:
re-worked to address your comments. Added a test that passed before and after my change. I can PR the test on its own if you like (I probably should)

mgazza:
Question: for history, do we need to keep history for columns other than value?

springfall2008:
> Question: for history, do we need to keep history for columns other than value?

Home assistant keeps the history of all attributes, this mirrors the same thing but it removes 'big' attributes first

PR #2424, opened 2025-06-05
Add the battery size as an output

mgazza:
Punting out the total battery size of the system


springfall2008:
Any reason why this can't just be an attribute in soc_kw?

mgazza:
Nope none.

springfall2008:
Will re-do this with an attribute

mgazza:
@springfall2008 you may want to look at [the last commit](https://github.com/springfall2008/batpred/pull/2424/commits/2be8b9ac464192e5112d5da9088d33184143b73d).

PR #2426, opened 2025-06-06
Adding current and final soc to soc_kw/soc_kw_best sensors

springfall2008:


apps/predbat/plan.py
```diff
@@ -3193,6 +3193,8 @@ def run_prediction(self, charge_limit, charge_window, export_window, export_limi
                         "unit_of_measurement": "kWh",
                         "first_charge_kwh": first_charge_soc,
                         "icon": "mdi:battery",
+                        "soc_now": dp3(self.soc_kw),
+                        "soc_max": dp3(self.soc_max),
```
mgazza, line 3197: could you add soc_pc also?

mgazza: if not its simple enough for me to so in the UI, I don't mind :D

springfall2008: What is soc_pc?

Issue #2366, opened 2025-05-14
New Gen3 Single phase high voltage inverter and batteries


TheSargey:
Hi folks, hope you're well.

I mentioned a while back I was upgrading to the latest GE inverter and batteries, it's 10kW inverter (Single phase) paired with 20kW of HV batteries.

Seems like the kit is still in it's infancy, had big issues getting the battery firmware upgraded (ended up having to go via USB direct into the battery stack) but we're up and running today.  Pretty frightening to see it suddenly pulling 10kW from the grid to charge the battery!

Anyway, I digress.  One issue I've got (which I've raised with GE directly) is the local modbus connection seems to vanish after a couple of hours, however the cloud information continues to work.  A quick restart of the inverter allows GivTCP to get it's connection back and things work as expected.

Whilst GE looks into this, I'm just looking at the restart options I have available in apps.yaml - namely:

  # When enabled automatic restart will restart the add-on if communication fails
  # Example below is auto-restart for GivTCP add-on itself
  auto_restart:
    - shell: 'rm -rf /homeassistant/GivTCP/*.pkl'
    - service: hassio/addon_restart
      addon: 533ea71a_givtcp

  #  Example on how to restart the inverter via GivTCP
  #  - service: switch.turn_on
  #    entity_id: switch.givtcp_{geserial}_reboot_invertor

If I simply enable the inverter restart by removing the hashes on the last two lines, does this send a restart via the cloud or locally?  I'm guessing locally, however the problem I have is there is no local control when the inverter "disappears" however I can restart the inverter via the app/cloud.  Just trying to get it so that predbat can restart the inverter when the clock skews and REST becomes unavailable (which is basically the sign that the inverter has for some reason gone to sleep locally).

Just wanted to get your thoughts on this.

Thanks as always for your help.

Sarge

gcoan:
All such commands are made via GivTCP locally through modbus, so wouldn't work for you.

Might be worth looking at the GivEnergy API's to see if there is an API to poke the inverter to restart itself.

Alternatively why not use the GE cloud integration already built into predbat and get the data from the cloud rather than locally via GivTCP.  https://springfall2008.github.io/batpred/inverter-setup/#givenergy-with-ge_cloud

I have used the older method of the ge_cloud integration that Trefor wrote when I had an issue with accidentally zapping all the history of my givtcp import sensor, so predbat broke,  and using that cloud integration worked perfectly.

[...]

PR #2441, opened 2025-06-08
Fix for GECloud download at midnight

springfall2008:
https://github.com/springfall2008/batpred/issues/2366


springfall2008:
https://github.com/springfall2008/batpred/issues/2366

PR #2502, opened 2025-06-22
Plan can be manually overiden using Web UI

springfall2008:


raldred:
This is a really nice addition thanks for all your hard work @springfall2008

Issue #1336, opened 2024-07-30
Two Car Charging Plans


dandwhelan:
**Describe the solution you'd like**
I've two electric cars with two different chargers. One car charges at 7.2Kw and the other charges at 3.6kw.. I would like predbat to have the functionality to independantly control the charging of two different cars, at two different two charging speeds. 

I'd like to see two charging slots available to each car considering the speed they will charge at. Also on the html plan you'd need to add a column for car 1 and car2.

**Describe alternatives you've considered**
Using the normal charging plan and make both switches turn on at the same time. But it wont work well.

Thanks

dandwhelan:
I'd still love this feature allowing me to specify the max charging speed of two cars on predbat :) @springfall2008 
PS, love the new web interface. great for troubleshooting.

springfall2008:
Now supported AFAIK

PR #2519, opened 2025-06-30
Feature/multi car charging and MW export in debug logs fix

dandwhelan:
Hi, I made a request a year ago and I've managed to implement it myself. It works on my system (docker) Please review. Thanks, Dan

https://github.com/springfall2008/batpred/issues/1336


apps/predbat/predbat.py
```diff
@@ -1481,3 +1481,15 @@ def run_time_loop_balance(self, cb_args):
                 self.log("Error: " + traceback.format_exc())
                 self.record_status("Error: Exception raised {}".format(e), debug=traceback.format_exc())
                 raise e
+
+    def balance_inverters(self):
```
springfall2008, line 1485: I think this is left over from a bad merge, balance inverters is in execute.py now

apps/predbat/prediction.py
```diff
@@ -414,6 +414,15 @@ def run_prediction(self, charge_limit, charge_window, export_window, export_limi
         inverter_loss = self.inverter_loss
         inverter_hybrid = self.inverter_hybrid
 
+        # Can we do better than the base plan?
```
springfall2008, line 417: This also seems like a bad merge

springfall2008:
I like the change and will certainly include it, although it looks like the merge has gone wrong.

I'm wondering what the point of the apps.yaml setting is as well as having it as a config item, should we just have four settings for up to four cars?

From a backward compatible point of view I'd probably keep car_charging_rate as is (not call it _1), also its really car zero.

dandwhelan:
I agree, good idea. This is my first pull/push to somebody else's repo, and
I've got to admit, it looks like a mess. I'll make some changes tonight and
get back to you tomorrow. I've been running predbat on my current version
for a few days. It's working well, I was chuffed when the plan was updated
for cars with different charging speeds.

apps/predbat/predbat.py
dandwhelan: cool, I'll remove it

apps/predbat/prediction.py
dandwhelan: Yes, I agree

PR #2527, opened 2025-07-01
Ev multi car power fixes

dandwhelan:
I think pull 2519 was very messy and this new pull will be clearer on whats happened... at least for me... https://github.com/springfall2008/batpred/pull/2519


springfall2008:
I've sort of re-done this now in the latest release, you can configure the car charging rate for each car, it will create a new rate setting. Can you review?

dandwhelan:
Alot fancier than I could have done. Annoyingly my nissan leaf integration has broken so I can't check if it works

dandwhelan:
I'm sure it does

springfall2008:
This is now implemented in the last release

PR #2550, opened 2025-07-09
Update energy-rates.md | Added example spot integration template

ziat007:
Czech Republic spot integration and how it can be integrated into predbat.


ziat007:
I would suggest adding a separate wiki page where templates are set up for each country/integration. I could go more into detail there on the formatting predbat requires.

PR #2636, opened 2025-08-30
Fox inverter grid power

JustinWingChungHui:
grid_power is missing from the fox.yaml.  I had to create up a template sensor to supply the values.  

Is the best way to add comments into the fox template?

### DIFF
--- a/templates/fox.yaml
+++ b/templates/fox.yaml
@@ -141,7 +141,7 @@ pred_bat:
   battery_min_soc:
     - number.min_soc
   
-  # You will need to create a template sensor should be negative for import and positive for export
+  # You will need to create a template sensor that should be negative for import and positive for export
   # Example template code:
   #    {% set a = states('sensor.grid_consumption') | float(0) %}
   #    {% set b = states('sensor.feed_in') | float(0) %}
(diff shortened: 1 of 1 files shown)


WyndStryke:
Do you have sensor.grid_ct from the FoxESS modbus integration available?  It's in kW, positive export, negative import.

Some sensors are inverter-specific, and some are disabled by default (so would need to be enabled).  I'm not sure whether this one is in that group.


```
  #
  # Sensors, more than one can be specified and they will be summed up automatically
  #
  # For two inverters the load today would normally be the master load sensor only (to cover the entire house)
  # If you have three phase and one inverter per phase then you would need three load sensors
  #
  # For pv_today if you have multiple solar inverter inputs then you should include one entry for each inverter
  #
  load_today:
    - sensor.shelly_3em_c8_ch_b_daily_energy
  import_today:
    - sensor.grid_consumption_energy_today
  export_today:
    - sensor.feed_in_energy_today
  pv_today:
    - sensor.solar_energy_today
  grid_power:
    - sensor.grid_ct

```

Adding in grid_power: - sensor.grid_ct   on my system now displays activity from the grid on the power flow chart, whereas previously it was displaying no grid activity.

JustinWingChungHui:
You are right!  Let me amend that PR

PR #2709, opened 2025-10-01
Refactor Solcast from mixin to component architecture

mgazza:
- Convert Solcast from mixin pattern to true component accepting base parameter
- Add autonomous lifecycle with async start/stop methods
- Publish standard entities (sensor.pv_forecast_today/tomorrow) without prefixes
- Update Fetch to read forecast entities via configuration rather than direct calls

This enables Solcast to follow the same pattern as GivEnergy and other components, with proper dependency injection and entity-based communication.


springfall2008:
Doesn't this have to go into components.py and have changes to predbat.py?

mgazza:
yeah its a draft. i've done the work in my local branch, attempted to pull it out on its own, this PR is a draft so need more work.

Issue #2504, opened 2025-06-23
Support for bottlecapdave's Carbon Intensity Integration


markbw999:
Hi Trefor and team, the carbon intensity UK integration has been broken for me (and I think others) for a few weeks now. I don't think it's the integration itself, rather the source website which is failing to resolve postcodes when determining the user's region.
Bottlecapdave has created an alternative integration which works for me as when configuring it you get to select your region from a dropdown rather than entering your postcode.
It creates a sensor for the current intensity level and two events for the current and next day forecasts. The data looks good but when I try the event in apps.yaml predbat doesn't handle it.
I asked Dave whether the event could be replaced with a sensor that would work with predbat and this was his response...
"The events cannot be a value of a sensor as the data is too large. My other integration uses this setup and is compatible with predbat, but I'm not sure if there is something special predbat is doing." I assume the other integration he refers to is Octopus.
Could you add support for Dave's integration as an alternative to carbon intensity UK?
Thanks, Mark

springfall2008:
I've commented on: https://github.com/BottlecapDave/HomeAssistant-CarbonIntensity/issues/25

PR #2803, opened 2025-10-22
Add Support for HomeAssistant-CarbonIntensity Integration

TCWORLD:
Allows predbat to extract carbon forecast data provided by the BottlecapDave/HomeAssistant-CarbonIntensity integration.

Essentially if the carbon intensity sensor is named `xx_current_rating`, then it will locate the `event.xx_current_day_rates` and `event.xx_next_day_rates` entities and extract the forecast data from them. This will provide between 24 and 48 hours of forecast data depending on the time of day. 

This relates to #2504 and https://github.com/BottlecapDave/HomeAssistant-CarbonIntensity/issues/25.

image

### DIFF
--- a/apps/predbat/fetch.py
+++ b/apps/predbat/fetch.py
@@ -2107,21 +2107,24 @@ class Fetch:
                 # Extract data from event sensors
                 intensity_key = "intensity_forecast"
                 #  - For current day
-                current_rate_id = entity_id.replace("_current_rating", "_current_day_rates").replace("sensor.", "event.")
-                data_current = self.get_state_wrapper(entity_id=current_rate_id, attribute="rates")
-                if data_current:
-                    data_all += data_current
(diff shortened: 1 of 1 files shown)


TCWORLD:
At the moment this is not guaranteed to provide 48 hours worth of forecast data. It will always give at least 24 hours. 
The issue is that the integration only gives us today and tomorrow, much like the octopus integration. When the plan extends into the second day into the future, predbat defaults the carbon footprint to 0's and so seems to want to try and wait.

I've started a [PR](https://github.com/BottlecapDave/HomeAssistant-CarbonIntensity/pull/26) on the other integration to provide the "second day" into the future data which would allow predbat to get the full 48hrs at all times.

springfall2008:
This is why I didn't include this feature, with only 24 hours data its not very useful. 

I was planning to just add carbon directly into Predbat

TCWORLD:
Cool, given you've added it in directly, I'll close this as its not really necessary anymore.

PR #2897, opened 2025-11-12
fix: Optimize GECloud memory usage by clearing cached data after fetch

mgazza:
## Summary

Fixes OOM (Out of Memory) crashes in GECloud data fetching by eliminating duplicate data storage in RAM cache.

## Problem

GECloud was storing fetched data in three separate locations:
1. `self.mdata` - needed for processing by `minute_data()`
2. `self.ge_url_cache[url]["data"]` - RAM cache (unnecessary duplication)
3. YAML file on disk - for persistence across runs

For typical usage (8 days of historical data at ~5-minute intervals = ~2,304 data points), this resulted in significant memory overhead from the duplicate RAM cache storage.

## Solution

This PR clears the cached data from RAM after all data has been accumulated into `self.mdata`:
- Keeps only metadata (`stamp`, `next`) in the RAM cache
- Data is still saved to disk via `save_ge_cache()` for future runs
- Processing continues normally using `self.mdata`

## Changes

- `apps/predbat/gecloud.py:1538` - Added comment noting cached data is temporary
- `apps/predbat/gecloud.py:1541` - Added comment on temporary data storage
- `apps/predbat/gecloud.py:1606-1610` - Added cleanup loop to clear cached data after fetch

## Testing

- [ ] Verify OOM crashes no longer occur during GECloud data fetching
- [ ] Confirm data is still correctly processed and available via `get_data()`
- [ ] Check that disk cache is properly saved and loaded

## Memory Savings

Eliminates ~2,304 × 5 fields × 8 bytes per field = ~92 KB per fetch cycle from RAM cache duplication (actual savings may be higher depending on pagination and data density).

### DIFF
--- a/apps/predbat/gecloud.py
+++ b/apps/predbat/gecloud.py
@@ -1603,14 +1603,15 @@ class GECloudData:
         self.oldest_data_time = last_updated_time
         self.mdata = mdata
 
-        # Memory optimization: Clear cached data from RAM now that we've accumulated everything
+        # Save GE URL cache to disk for next time (MUST happen before clearing RAM cache)
+        self.save_ge_cache()
+
+        # Memory optimization: Clear cached data from RAM now that we've saved to disk
         # This prevents duplicate storage - data is only in self.mdata and disk cache
         for url_key in list(self.ge_url_cache.keys()):
             if "data" in self.ge_url_cache[url_key]:
                 del self.ge_url_cache[url_key]["data"]
 
-        # Save GE URL cache to disk for next time
-        self.save_ge_cache()
         return True
 
     def get_data(self):


apps/predbat/gecloud.py
```diff
@@ -1602,6 +1603,12 @@ async def download_ge_data(self, now_utc):
         self.oldest_data_time = last_updated_time
         self.mdata = mdata
 
+        # Memory optimization: Clear cached data from RAM now that we've accumulated everything
+        # This prevents duplicate storage - data is only in self.mdata and disk cache
+        for url_key in list(self.ge_url_cache.keys()):
+            if "data" in self.ge_url_cache[url_key]:
+                del self.ge_url_cache[url_key]["data"]
```
springfall2008, line 1613: This isn't right, it will not save the data to disk either as you deleted it before the save

mgazza:
Good catch @springfall2008! You're absolutely right - the data needs to be saved to disk BEFORE clearing it from RAM.

The correct order should be:
1. Accumulate all data into `self.mdata`
2. **Save `self.ge_url_cache` to disk** (includes the `data` field)
3. Clear the `data` field from RAM cache to save memory

The current PR has steps 2 and 3 swapped, which would result in an empty disk cache.

Here's the corrected code:
```python
self.oldest_data_time = last_updated_time
self.mdata = mdata

# Save GE URL cache to disk for next time (MUST happen before clearing RAM cache)
self.save_ge_cache()

# Memory optimization: Clear cached data from RAM now that we've saved to disk
# This prevents duplicate storage - data is only in self.mdata and disk cache
for url_key in list(self.ge_url_cache.keys()):
    if "data" in self.ge_url_cache[url_key]:
        del self.ge_url_cache[url_key]["data"]

return True
```

This way:
- ✅ Disk cache gets the full data (including `data` field)
- ✅ RAM cache only keeps metadata (`stamp`, `next`)
- ✅ Processing uses `self.mdata` as intended
- ✅ Memory is saved by eliminating duplicate storage in RAM

I'll update the PR with this fix.

springfall2008:
Still not correct as you would have to clear the cache so it reloads next time?

springfall2008:
In https://github.com/springfall2008/batpred/pull/2924

PR #2932, opened 2025-11-17
fix: Add missing 'previous' parameter to async_get_evc_device

### LINKED ISSUE #2931: GECloud: async_get_evc_device missing 'previous' parameter causes TypeError
## Bug Description

In v8.27.23, the `async_get_evc_device` method is called with a `previous` parameter but the method definition doesn't accept it, causing a TypeError that breaks the main loop for users with EVC (Electric Vehicle Charger) devices.

## Error Message

```
Error: GECloud: Exception in main loop GECloudDirect.async_get_evc_device() takes 2 positional arguments but 3 were given
```

## Root Cause

In `apps/predbat/gecloud.py`:

**Line 854 (call site):**
```python
self.evc_device[uuid] = await self.async_get_evc_device(uuid, self.evc_device.get(uuid, {}))
```

**Line 1152 (method definition):**
```python
async def async_get_evc_device(self, uuid):  # Missing 'previous={}' parameter
```

## Impact

- Every customer with an EVC device hits this error every 60 seconds
- The main polling loop crashes and restarts
- EVC device data is not being properly fetched/cached

## Expected Behavior

The method signature should match the call pattern, similar to `async_get_evc_device_data` which correctly has `previous={}`:

```python
async def async_get_evc_device(self, uuid, previous={}):
```

## Fix

Add the `previous={}` parameter to the method definition and return `previous` on failure instead of an empty dict.

## Version

v8.27.23

mgazza:
## Summary

Fixes #2931

- Adds missing `previous={}` parameter to `async_get_evc_device` method definition
- Returns `previous` value on API failure to preserve cached data (consistent with other methods)

## Problem

The `async_get_evc_device` method was being called with a `previous` parameter for caching:
```python
self.evc_device[uuid] = await self.async_get_evc_device(uuid, self.evc_device.get(uuid, {}))
```

But the method definition didn't accept it:
```python
async def async_get_evc_device(self, uuid):  # Missing 'previous' parameter
```

This caused a TypeError that broke the main loop for all users with EVC (Electric Vehicle Charger) devices:
```
Error: GECloud: Exception in main loop GECloudDirect.async_get_evc_device() takes 2 positional arguments but 3 were given
```

## Test plan

- [x] Verify method signature matches the call pattern at line 854
- [x] Ensure consistency with `async_get_evc_device_data` which already has `previous={}` parameter
- [ ] Test with an EVC device to confirm the main loop no longer crashes

### DIFF
--- a/apps/predbat/gecloud.py
+++ b/apps/predbat/gecloud.py
@@ -1149,7 +1149,7 @@ class GECloudDirect:
 
         return command_info
 
-    async def async_get_evc_device(self, uuid):
+    async def async_get_evc_device(self, uuid, previous={}):
         """
         Get EVC device
         """
@@ -1164,7 +1164,7 @@ class GECloudDirect:
             status = device.get("status", None)
             type = device.get("type", None)
             return {"uuid": uuid, "alias": alias, "serial_number": serial_number, "status": status, "online": online, "type": type, "went_offline_at": went_offline_at}
-        return {}
+        return previous
 
     async def async_get_smart_devices(self):
         """


springfall2008:
In: https://github.com/springfall2008/batpred/pull/2933

PR #2951, opened 2025-11-19
Add start of day offset for PV/Import/Export/Load

### LINKED ISSUE #2925: "Current PV"/"Import"/"Export" sensors don't correctly track source entities.
**Describe the bug**

The "Current PV' sensor is supposed to mirror the sensor in the pv_today config entity. However at some point in the last few days it seems to no longer be mirroring properly.

By the looks of it, there seems to be sometimes a residual amount from the previous day that's being subtracted from it.

**Expected behaviour**

Current PV and pv_today should be the same.

**Predbat version**
V8.27.21

**Environment details**

- Inverter and battery setup
- Anything else?

**Screenshots**
If applicable, add screenshots to help explain your problem. The most useful ones can be your battery chart, the Predbat HTML plan and your current settings in HA.

Image

**Log file**
Can you capture a log file from the time of the issue, debug mode is not normally required.

You can download the logfile from the WebUI on the Dash tab select **predbat_debug.log**

**Predbat debug yaml file**
This is important for any plan related issues.

Once you have captured the issue go to the Web UI, in the Dash tab click on **predbat_debug.yaml** and wait 30 seconds for it to download. Select 'keep' if your Web Browser thinks the file is dangerous.

Rename the download file to predbat_debug.yaml.txt and upload it to Github. This will allow your plan to be reproduced and also stores all your settings for review.

TCWORLD:
The predbat code currently subtracts the sensor value from minute 0 of the day from the `load_today`, `export_today`, `import_today`, and `pv_today` sensors.

However if the sensors reset to zero slightly into the day due to e.g. communication latencies, this causes the previous days data to be subtracted from the current days data leading to incorrect data being used.

This adds a new config entity "todays_sensor_reset" (feel free to rename) which allows offsetting the "start of day" data value to the set number of minutes in the day. 

- For example it it takes 1 minute for sensor values to reset to zero, you can now set the offset to say 2, and predbat will subtract the data from minute 2 from the days sensor values instead of minute 0.
- The default remains 0 indicating the first minute data of the day as it was before.

Fixes #2925.

### DIFF
--- a/apps/predbat/fetch.py
+++ b/apps/predbat/fetch.py
@@ -326,19 +326,19 @@ class Fetch:
         else:
             return total / total_weight
 
-    def get_now_from_cumulative(self, data, minutes_now, offset = 0, backwards=True):
+    def get_now_from_cumulative(self, data, minutes_now, offset=0, backwards=True):
         """
         Get the now value from an cumulative incrementing series e.g. kWh total -> kWh this minute
         by subtracting the start of today from the current minutes value.
         Allows skipping `offset` minutes from the start of the data to allow for e.g. sensors
         that reset to zero with a slight latency.
         """
-        offset = max(offset, 0) # Disallow negative offsets as those would be yesterday.
+        offset = max(offset, 0)  # Disallow negative offsets as those would be yesterday.
         if backwards:
             return max(data.get(0, 0) - data.get(max(minutes_now - offset, 0), 0), 0)
         else:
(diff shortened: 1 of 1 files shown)


TCWORLD:
I'm not sure whether this should be a config entity or a control input. I tried the latter but it just crashed prebat (must have done something wrong), so I've switched it to be a config entity.

springfall2008:
Great change, but I think I want to make this automatic, let me do a PR now

springfall2008:
In this PR: https://github.com/springfall2008/batpred/pull/2960

PR #2965, opened 2025-11-22
Fix #2937: Add per-car manual SoC controls

### LINKED ISSUE #2937: 2nd cars not charging as expected - predbat_car_charging_manual_soc_kwh does not increase
### Discussed in https://github.com/springfall2008/batpred/discussions/2890

<div type='discussions-op-text'>

<sup>Originally posted by **ppmt** November 10, 2025</sup>
a few weeks ago I configure predbat for my new car and all worked well. Now we have 2 electric cars and I just can't get it to charge as I want. 

I suspect it is a bad config on my side but.... 
Here is my config for both cars
Screenshot From 2025-11-09 20-31-52

Le me know if this is not enough info. 

So when I plug a car I have a switch which I use to say which car is connected. When I select the first one (lulu) then it works perfectly. I can set the initial SOC (none of my car can report it!), I can also say how much I want to charge the car and predbat does its magic and all is good.

When I select the second one (ioniq) and do the same it behaves completely differently. whatever initial SOC I enter predbat seems to assume it needs to charge the car from 0%. That is annoying but we could work around that.
So If I work around and say to charge 10k it will select the correct amount of slot to deliver the 10k.

What is not good is that the sensor input_number.predbat_car_charging_manual_soc_kwh does not increase as the car is charging. So as far as predbat is concerned the car is not charging and it constantly reschedules new slow. The result is that the car get overcharged until it is full.

 It works perfectly when I charge the first car. I can see that input_number.predbat_car_charging_manual_soc_kwh is increasing as the car is charging.
 
 As I said I suspect I have a bad config but I can't work it out and would appreciate some guidance.
 
 Thank you</div>

springfall2008:
## Summary
Fixes #2937 - Second car not charging as expected when using manual SoC mode

## Problem
When using manual SoC mode with multiple cars, only car 0's manual SoC (predbat_car_charging_manual_soc_kwh) would increment during charging. Cars 1+ would continue charging beyond their target because their manual SoC value remained at the initial setting.

## Solution
- Added per-car manual SoC controls following the same pattern as other car settings (e.g., car_charging_rate_1/2/3)
- Added config entries for cars 0-3:
  - switch.predbat_car_charging_manual_soc (car 0) and _1/_2/_3 variants
  - input_number.predbat_car_charging_manual_soc_kwh (car 0) and _1/_2/_3 variants
- Changed car_charging_manual_soc from single boolean to array
- Updated SoC reading logic in fetch.py to check per-car manual_soc array and use postfix
- Updated SoC increment logic in predbat.py to update the correct per-car entity using postfix
- Updated documentation in car-charging.md and customisation.md

## Changes Made
**config.py:**
- Added 6 new CONFIG_ITEMS entries for cars 1-3 manual SoC controls
- Updated friendly names for car 0 to clarify "(Car 0)"

**predbat.py:**
- Changed car_charging_manual_soc initialization from False to []
- Updated SoC increment logic (lines 876-880) to check array and use postfix

**fetch.py:**
- Initialize car_charging_manual_soc as array (line 1878-1882)
- Read per-car settings using postfix pattern (line 699-703)

**Documentation:**
- Updated car-charging.md with per-car manual SoC references
- Updated customisation.md to document _1/_2/_3 suffixed controls

## Testing
- [x] Code compiles without errors
- [x] Follows existing pattern for per-car settings
- [x] Documentation updated
- [ ] Tested with multi-car setup (requires user testing)

## Backward Compatibility
✅ Existing users with car 0 manual SoC enabled will continue to work unchanged


gcoan (commented):
@springfall2008 the changes all look good, they were the way I was intending to do it, adding more switch.predbat_car_charging_manual_soc's and then more soc_kwh input_numbers.

Thanks

ppmt:
I should be able to test the fix in the next couple of days. Sorry for the delays

ppmt:
@gcoan , @springfall2008 , I am not sure if this is the right place to report.

I am on Predbat 8.28.4 and configured 2 EVs. Last night I charged the second car in my config and I am happy to report that it worked.

My car charged and stopped once the desired level was attained.  I will charge using the fist car in the config tonight but I don't anticipate any issue there.

I am now going to finalize the document writing what I did to have a nice charging station in HA and will publish it in the discussion section.

Thank you very much to both of you.

gcoan:
Excellent, great news that it works for you.

Yes please do add something to the discussion area and I can always update the main documentation if there are any key points missing - and please do highlight them here if you notice anything. 

Cheers

gcoan:
Thanks @ppmt for the writeup, I have added a link to this to the Predbat documentation in my fork.

ppmt:
I hope it will be useful for someone. I am really happy with how it works now!

Thanks a lot for your help @gcoan !

gcoan:
> Thanks @ppmt for the writeup, I have added a link to this to the Predbat documentation in my fork.

the link didn't come through from my commit.  here it is https://github.com/gcoan/batpred/blob/main/docs/apps-yaml.md#multiple-electric-cars

PR #3012, opened 2025-12-01
Remove Minimum Drop for "Incrementing" Data near Midnight

TCWORLD:
If it is incrementing data, then any drop is technically invalid and should really be ignored.

By requiring a minimum drop of 1.0, then for cases with a unit of say kWh, on days where the entity (e.g. PV production, or export energy) has less than 1kWh overall for the day, then the resulting data can be wrong by a substantial amount. 

See #2925 for further discussion.

### DIFF
--- a/apps/predbat/utils.py
+++ b/apps/predbat/utils.py
@@ -453,8 +453,8 @@ def minute_data(
                 if smoothing:
                     near_midnight = (last_updated_time.time() < time(0, 6)) or (last_updated_time.time() > time(23, 58))
                     if clean_increment and state < last_state and (near_midnight or (last_state - state >= 1)):
-                        # If there is a large drop in the data or we are near the end of the day where the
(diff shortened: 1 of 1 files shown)


springfall2008:
I'm not sure about this, it was added because the GivEnergy incrementing data goes up and down by 1 due to rounding errors (the sensor must be constructed from other data). If we remove it then it can create errors.

TCWORLD:
@springfall2008  Do you have an example of the data doing that? I believe you, it just amazes me there's 1kWh of rounding errors

Either way, need to find a best of both solution as the hard limit messes up any data where there is <1kWh total increment during a day - I'll have another think about it, maybe a configurable limit.

TCWORLD:
I've updated my changes so that the minimum drop is still in place during the day, however near midnight the restriction is removed.

I think the time check is correct, though it might need some verification regarding time-zones. Is the data imported in local time or UTC?

PR #3046, opened 2025-12-08
Update link to current SigEnergy integration in docs

DFEvans:
We've updated the template config for the new integration, but the summary still points at the old integration.


DFEvans:
Hmm, something went weird with that commit

Issue #3084, opened 2025-12-19
Database Primary Issue


dandwhelan:
**Describe the bug**
Unable to get_history when enabling db_primary: true 

**Expected behaviour**
I'm wanting to move away from Homeassistant storing all the data and I'd like Predbat DB Manager storing all the data. I've moved to GE Cloud Data but still using HA GivTCP to control the invertor. Am I missing something or not understanding the feature correctly? I've tried this a few times over the past month but go back to the 'normal' setup due to lack of time to troubleshoot. If an entity in the app.yaml is in HA, then feature wont work? 

**Predbat version**

v8.29.13
**Environment details**
Standalone Docker
HA Docker

[predbat_debug.yaml](https://github.com/user-attachments/files/24267337/predbat_debug.yaml)
Thanks in advance and I love seeing Predbat constantly evolve over the years!

dandwhelan:
[predbat.txt](https://github.com/user-attachments/files/24267367/predbat.txt)

dandwhelan:
I've managed to fix this. 
Deleted the predbat.db from the predbat folder. Turned off the mirror from HA, enabled gathering data from GeCloud and made the internal DB the primary, and it works. Then I added predbat to the exclusion list to record data in HA. I'm not interested in the HA graphs; the Web app is good enough. The only reason I want Predbat to store the data instead of HA, is that predbat seems to use a lot of space, and every few months, my HA would crash and wipe all the databases. Hopefully, this will resolve the issue. Predheat might not work, but it doesn't matter too much, as the predictions don't affect the upcoming schedule.

gcoan:
> The only reason I want Predbat to store the data instead of HA, is that predbat seems to use a lot of space

Yes Predbat can use a lot of HA database space if untamed.  I have done some work on this and plan to document configuration changes that will dramatically reduce the amount of HA database space that Predbat uses.

Glad you got it sorted

dandwhelan:
database indexing fix resolves this issue. It came back 4 times after closing this issue. #3114

PR #3114, opened 2025-12-23
 Internal Database Disk IO improvements and CPU reduction

dandwhelan:
# Predbat Performance Optimisation Guide

FYI - This feature was heavily created with AI, yet it works, and my disk activity and CPU have plummeted since using it. 
## Overview of Changes
We have implemented several optimisations to `hass.py` and the database engine to address high CPU usage and excessive Disk I/O. These changes are designed to make Predbat run more efficiently, particularly on resource-constrained devices like Raspberry Pis or older hardware.

### 1. Database Write Batching (Critical for Disk I/O)
**The Change:** 
- We successfully enabled `WAL` (Write-Ahead Logging) mode for SQLite.
- We removed a "force commit" command (`self.db.commit()`) that was accidentally running after every single entity update.
- We configured `db_commit_interval: 30` (seconds).

**The Rationale:**
Previously, every time Predbat updated a sensor or entity state, it forced a write to the physical disk immediately.
- **High Disk I/O:** This caused the disk to be constantly active (writing 100s of times per minute), which is the primary killer of SD cards and causes system slowdowns.
- **The Fix:** Now, Predbat keeps changes in memory and only flushes them to the disk file (`predbat.db-wal`) once every 30 seconds. This drastically reduces "wear and tear" on your storage and frees up system resources.

### 2. Throttling File Modification Checks (CPU Usage)
**The Change:**
- The `check_modified` function (which scans all `.py` files to see if code has changed) was running every **1 second**.
- It is now throttled to run every **30 seconds**.

**The Rationale:**
Scanning the file system every second consumes unnecessary CPU cycles (`sys` usage). By reducing this frequency, `hass.py` spends less time checking for file changes and more time doing actual work (or sleeping), lowering overall CPU usage.

### 3. Main Loop Throttling
**The Change:**
- Configured `hass_loop_interval` to **5 seconds** (up from 1s) when in `performance_mode`.

**The Rationale:**
If there is no work to do, there is no need to wake up every second. Sleeping for 5 seconds reduces the idle load on the CPU.

---

## Explanation of Log Warnings

You noticed logs like:
`Warn: Callback ... took 9.98 seconds`

**What this means:**
These warnings indicate that the main Predbat calculation logic (planning, fetching data, solving) took 9.98 seconds to complete. During this time, the "loop" is blocked—meaning it can't engage in other quick tasks until the calculation finishes.

**Is this something to worry about?**
- **Generally, No:** Predbat performs complex calculations (linear programming, matrix operations) to generate the battery plan. It is normal for this to take several seconds (5–20s) on typical hardware, especially every 5 minutes when new data is fetched.
- **Context:** The log entries you shared show this happening roughly every 5 minutes (15:15, 15:20, 15:25, 15:30). This coincides with the main plan update cycle.
- **Performance Impact:** Since we bumped the `hass_loop_interval` to 5s, a 10s calculation just means one or two "heartbeats" are skipped. As long as it finishes successfully (which it does), this is acceptable behavior for a heavy calculation task.

**Summary:** The recent optimizations won't necessarily make the *calculation* faster (that depends on CPU speed), but they stop the *overhead* (disk writes, file checks) from slowing it down further.

### DIFF
--- a/docs/customisation.md
+++ b/docs/customisation.md
@@ -83,7 +83,8 @@ Predbat includes a dedicated **Performance Mode** designed for resource-constrai
 When `performance_mode` is enabled (see [apps.yaml basics](apps-yaml.md#performance-tuning)), Predbat optimizes its operation by:
 
 - **Throttling the Main Loop:** Increasing the idle sleep time from 1s to 5s to reduce CPU usage.
-- **Batching Database Writes:** Using Write-Ahead Logging (WAL) and committing to disk based on the configurable `db_commit_interval` (default **0** for immediate commits). Increasing this interval (for example to 30 seconds) can drastically reduce Disk I/O and SD card wear.
+- **Batching Database Writes:** Using Write-Ahead Logging (WAL) and committing to disk based on the configurable `db_commit_interval` (default **0** for immediate commits).
+- Increasing this interval (for example to 30 seconds) can drastically reduce Disk I/O and SD card wear.
 - **Throttling File Checks:** Reducing the frequency of checking for file modifications.
 
 By default Predbat controls the inverter every 5 minutes and updates the plan every 10 minutes.


apps/predbat/web.py
```diff
@@ -3633,10 +3611,7 @@ async def html_components(self, request):
             time_ago_text = format_time_ago(last_updated_time)
 
             # Create component card
-            card_class = "active" if is_active else "inactive"
-            if is_active and not is_alive:
```
springfall2008, line 3637: Are these changes related or perhaps a bad merge?

dandwhelan:
I've made a few changes to the original merge. Firstly I've fixed and issue I had with the database not loading get history, it was all related to no indexing being enabled on the database, I closed the request but it came back time and time again. #3084   I've also added an option to hide all the info from populating in the log file. I've not made any changes to your original intended design of predbat web and the addons are optional extras which would need to be enabled by the user. I've fumbled by way through the github pull push stuff, hopefully this one works.

springfall2008:
I've sort of re-implemented this here: https://github.com/springfall2008/batpred/pull/3122 and https://github.com/springfall2008/batpred/commit/c3b8e200bfa450783fa25b57ed2592996926e397#diff-163c9ad3856902e95974c000176343b241d39f041802a6eb8b65aefb3255f673

I like the performance improvements but not very keen on lots of new options, let me know what you think

dandwhelan:
> I've sort of re-implemented this here: #3122 and [c3b8e20#diff-163c9ad3856902e95974c000176343b241d39f041802a6eb8b65aefb3255f673](https://github.com/springfall2008/batpred/commit/c3b8e200bfa450783fa25b57ed2592996926e397#diff-163c9ad3856902e95974c000176343b241d39f041802a6eb8b65aefb3255f673)
> 
> I like the performance improvements but not very keen on lots of new options, let me know what you think

I like it, it looks alot cleaner than the original implementation. The main reason for changing my copy of predbat was to reduce the constantly database interactions every second along with Haas using 10% of my cpu on the pi5. anything to improve performance is welcome.

dandwhelan:
At the moment in your latest build, the wal-cache is constantly updating/being checked every second. I don't think it should work that like. I had this issue but resolved it (can't remember how, it was logic related(isnt everything))

springfall2008:
Sorry what is the wal-cache ?

springfall2008:
Fix for this issue added into the .9 release

Issue #3130, opened 2025-12-27
Ohme Direct link is wrong in documentation


murkle:
The link here
https://springfall2008.github.io/batpred/devices/#ohme-direct
for Ohme Direct points to:
https://springfall2008.github.io/batpred/apps-yaml/#ohme-car-charger---direct-integration

I think it should be:
https://springfall2008.github.io/batpred/apps-yaml/#ohme-car-charger-direct-integration

<img width="219" height="214" alt="Image" src="https://github.com/user-attachments/assets/6b7519ee-b58c-424c-9dba-056776057d25" />

gcoan:
Thanks for pointing out @murkle, I've corrected it now.  I think Visual Studio got confused by the link name having a hyphen in it.  I've removed that and updated the devices link

Issue #3051, opened 2025-12-08
UK: Axle Grid Event Support


WhatIsThisAllAboutSilly:
A company is paying people to use their batteries to stabilise the local grid when it's needed, paying £1/kWh. https://vpp.axle.energy/landing/grid

However, they take control of the battery via in my case, the FoxESS website.

However they do provide a REST API (https://vpp.axle.energy/landing/home-assistant) which provides:

ATTRIBUTE	TYPE	DESCRIPTION
start_time	ISO 8601 datetime	When the next grid event starts
end_time	ISO 8601 datetime	When the grid event ends
import_export	string: "import" | "export"	Whether to import from or export to the grid
updated_at	ISO 8601 datetime	When the event data was last updated

Predbat obviously needs to be aware of this to ensure that it knows that the battery during these periods will be charging/discharging at max rate.

Presumably the £1/kWh may change so I'm going to ask Axle if they could also as part of this API set a £/kWh rate so Predbat can use that as part of its calculations.

So I think the basic idea would be that predbat checks the entity with the axle event info in, and adds £1p to those slot(s).

Not sure how the charge slot thing would work, but lets start with having export functionality there..!

gcoan:
I have joined the Axle scheme.  I did wonder about a FR for Predbat to take account of this, but wondered if the number of users is quite small.

At the moment I have just created an automation to alert me to the event so I can see what happens and what notice I get.

Possible options then are:
- put predbat into read only mode so Axle can control the inverter
- put in some force exports to match the Axle plan
- put in a rate override so Predbat takes account of it in its plan

all of these can be done via automations without predbat enhancements so that was the way I was thinking of going for now.

Since we have given Axle control of our inverter they will send the remote commands to our inverter, as well as Predbat instructing it, so there is a potential for clash unless we put predbat into read only,  but RO has to be done sufficiently far in advance for Axle to then take control.
So no perfect solution.

Nilogax:
Any chance you could share what you've built @gcoan?  First I've heard of Axle but it looks like it's worth further exploration.

Ah - ignore that!  Just going through the signup and they've supplied config for the HA sensors - nice!

gcoan:
No problem, if you want to sign up, here's my referral code https://vpp.axle.energy/landing/grid?ref=R-VNNIZUYO - we both get a bonus

What I have at the moment is:

1. A REST sensor in configuration.yaml to capture Axle events:

```
  # Axle VPP import/export event details
  - platform: rest
[...]

PR #3187, opened 2026-01-05
Alert notifications for Axle events, documentation enhancements

gcoan:
Code:
- #3051 Send notification when new Axle session is scheduled
- Axle: move updated_at to separate attribute to enable triggering on change of current_event

Documentation:
- Inverter setup, LuxPower and Kostal Plenticore, fix YAML formatting issue
- #3130 Correct Ohme Direct link
- Battery curve percentage steps can be given as integers or strings
- Add missing image files

### DIFF
--- a/apps/predbat/tests/test_axle.py
+++ b/apps/predbat/tests/test_axle.py
@@ -39,7 +39,7 @@ class MockAxleAPI(AxleAPI):
             "import_export": None,
             "pence_per_kwh": None,
         }
-        self.updated_at: None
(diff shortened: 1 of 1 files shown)


gcoan:
thanks for merging @springfall2008 I forgot to update the PR description, but this also fixes a bug that causes the Axle event rate increment when an event is ongoing to be doubled.  Caused by get_axle_sessions returning both current and historic events and a session is written to historic as soon as it starts

PR #3219, opened 2026-01-11
Add rate coverage validation to prevent battery control with missing rates

mgazza:
## Summary
- Add `validate_rate_coverage()` function to check rate data availability before making battery control decisions
- Skip plan recalculation when rate coverage is below 95% to prevent incorrect optimization
- Log warnings and set "safe mode" status when rate data is unavailable

## Problem
When rate data is unavailable (e.g., due to API failures, NATS outages, or network issues), the rate dictionary can have gaps. Previously, missing rates would default to `0`, which could lead to incorrect optimization decisions like unnecessary battery exports when the system thinks export rates are zero.

This has been observed in production when NATS connectivity issues caused rate data to be unavailable, resulting in batteries being exported when they shouldn't have been.

## Solution
This PR adds rate coverage validation that:
1. Checks if we have rate data for at least 95% of the forecast period (allows for minor gaps during rate transitions)
2. If coverage is insufficient, logs a warning and skips plan recalculation
3. Preserves any existing valid plan rather than making decisions with bad data
4. Sets a status message ("Rate data unavailable - safe mode") visible to users

## Test plan
- [ ] Verify normal operation with full rate data (validation should pass)
- [ ] Test with empty rate dictionaries (should trigger safe mode)
- [ ] Test with partial rate data below 95% coverage (should trigger safe mode)
- [ ] Confirm existing tests still pass

### DIFF
--- a/apps/predbat/fetch.py
+++ b/apps/predbat/fetch.py
@@ -1504,6 +1504,51 @@ class Fetch:
         if print:
             self.log("Gas rates: min {}{}, max {}{}, average {}{}".format(self.rate_gas_min, curr, self.rate_gas_max, curr, self.rate_gas_average, curr))
 
+    def validate_rate_coverage(self, rates, rate_name, required_minutes=None):
+        """
+        Check if we have rate data for the required forecast period.
+
+        This validation ensures we don't make battery control decisions based on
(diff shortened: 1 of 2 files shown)


springfall2008:
As discussed safer to do this is min/max rates as export rates can be missing

PR #3246, opened 2026-01-15
Improved LuxPower Freeze Charging support: documentation updates, template improvements, and warning suppress

### LINKED ISSUE #2916: Freeze charge - LuxPower inverters.
**Predbat version** - 8.27.19 (and previous)
**apps.yaml (redacted**) - [apps-4.yaml.txt](https://github.com/user-attachments/files/23549961/apps-4.yaml.txt)
**Debug yaml** [predbat_debug.yaml.txt](https://github.com/user-attachments/files/23550069/predbat_debug.yaml.txt)

**The Issue**
I'm trying to get a LuxPower inverter to behave as closely as possible to what a GE inverter does when 'FrzChrg' (charge_freeze_service) is called by the plan. The description from the docs is this:

**_Freeze charging_** - _The battery is charging but the current battery level (SoC) is frozen (held). Think of it as a charge to the current battery level. The grid or solar covers any house load. If there is a shortfall of Solar power to meet house load, the excess house load is met from grid import, but if there is excess Solar power above the house load, the excess solar will be used to charge the battery,_

This reads just like Demand mode but with any house load, which might have come from the battery, coming from the grid instead. 

**Proposed implementation**
LuxPower inverters have a Charge Priority / Charge First option which when enabled stops the battery from discharging. All of the house load comes from the grid. Any solar power is used to charge the battery. Whilst this is identical to what is described above when there is no solar, it needs to be manipulated when there is solar power. To this end I have a template sensor helper (Solar/Home difference) to calculate the difference in wattage between the PV output and the house load thus:
```
{{ states("sensor.lux_solar_output_live") | float / 1 - states("sensor.lux_home_consumption_live") | float / 1 }}
```
A positive value would indicate that no action needs to be taken and Charge Priority is not required because 
a) The battery is not discharging
b) There is sufficient solar power to meet the house load, therefore grid import is not needed. 
c) Excess solar is used to charge the battery
A negative value means that the battery would be discharging to meet the house load which is not what we want. To keep the battery SoC frozen, Charge Priority needs to be switched on. 
_Caveat - when Charge Priority is enabled, none of the solar power is used to meet the house load but goes to charging the battery instead. There is a trade-off between more power being pulled from the grid than the FrzChrg description above but on the other hand more power has gone into the battery. I don't think this is very significant because the next time the Plan is calculated the battery SoC will be slightly higher than the plan was expecting and will be adjusted accordingly._  

I have an automation, triggered by the Solar/Home helper, to switch Charge Priority ON/OFF as required.
```
alias: Lux Predbat Freeze Charge
description: >-
  Predbat charge_freeze_service switching. Turn on Charge Priority if Home
  Comsumption is greater than Solar Output.
triggers:
  - trigger: state
    entity_id:
      - sensor.solar_home_difference
conditions: []
actions:
  - choose:
      - conditions:
          - condition: numeric_state
            entity_id: sensor.solar_home_difference
            below: 0
        sequence:
          - action: switch.turn_on
            metadata: {}
            data: {}
            target:
              entity_id: switch.lux_charge_priority
      - conditions:
          - condition: numeric_state
            entity_id: sensor.solar_home_difference
            above: 0
        sequence:
          - action: switch.turn_off
            metadata: {}
            data: {}
            target:
              entity_id: switch.lux_charge_priority
mode: single
``` 
This automation is enabled when the charge_freeze_service is called like this:
```
charge_freeze_service:
    - service: automation.turn_on
      entity_id: "automation.lux_predbat_freeze_charge"
```
You can see from the log and the automation timeline that everything is working as expected. The automation is running and switches the Charge Priority ON of OFF triggered by the state of the helper. Positive value = OFF. Negative value = ON.
```
2025-11-10 12:21:30.431662: Inverter 0 Calling service charge_freeze_service
domain charge service_name automation/turn_on with data {'entity_id': 'automation.lux_predbat_freeze_charge'}
```

Image

When Charge Priority is OFF as in Option 2 in the image above you would expect to see something like this

Image

but we're getting this instead

Image

**The problem**
It would appear that Predbat is ignoring the state of the Charge Priority switch in the automation altogether and just 'doing its own thing' when FrzChrg is called. 
I would really like to get this working and would appreciate any help or advice.

_Apologies for this being long-winded but I wanted to give as much info as possible. Some of the dates and times I've used in the examples don't match up but the problem is persistent._

brickatius:
## PR Description

This PR improves support for LuxPower inverters when using Predbat Freeze Charging and Freeze Export, focusing on documentation clarity, configuration correctness, and stable runtime behaviour.

### Summary of changes
**Documentation**

- Revises the LuxPower section of Inverter Setup to fully document how Freeze Charging and Freeze Export are implemented for LuxPower inverters.

- Explains the required Home Assistant helpers and automations.

- Introduces and documents the new maintain_freeze_charge_status inverter option, including why it is required for LuxPower.

### Configuration

Updates the luxpower.yaml example template to:

- Enable Freeze Charging / Freeze Export where supported.

- Reference `battery_min_soc` from the LuxPython integration to avoid discrepancies between user configuration and inverter state.

- Reference `battery_rate_max` from an input_number helper rather than a hard-coded value.

- Adds `maintain_freeze_charge_status` to the inverter configuration options and documents its purpose.

### Runtime behaviour

- Updates inverter.py to support `maintain_freeze_charge_status`.

- When enabled, Predbat suppresses expected warnings caused by manipulating `scheduled_charge_enable` during Freeze Charging. This prevents `predbat.status` from transitioning into `Warn:` states during normal Freeze Charging operation, allowing existing automations (e.g. Freeze exit handling) to behave reliably.

- Suppression only applies during Freeze Charging and does not affect normal charging behaviour or genuine error conditions.

**Rationale**

LuxPower inverters do not provide native Freeze Charging support and rely on controlled manipulation of `scheduled_charge_enable`. During Freeze Charging, this can legitimately cause Predbat to observe a mismatch between expected and reported state, resulting in warnings that are expected but disruptive.

`maintain_freeze_charge_status` ensures Predbat treats these cases as expected behaviour only while Freeze Charging is active, keeping predbat.status stable and avoiding unintended automation triggers.

**Notes**

No functional changes to Predbat planning logic.

Changes are opt-in and limited to inverters that enable maintain_freeze_charge_status.

Fixes #2916

### DIFF
--- a/docs/inverter-setup.md
+++ b/docs/inverter-setup.md
@@ -1216,12 +1216,11 @@ recorder:
 While LuxPower inverters cannot exactly replicate Predbat’s native Freeze Charging behaviour, these automations achieve an equivalent outcome. Any small differences are corrected the next time Predbat recalculates its plan.
 
 ³
-You may see warnings like those below in the Predbat logs. They can safely be ignored. They are to be expected and confirm that the Predbat Override automation is working as intended.
+You will see a warning like this below in the Predbat logs. It can safely be ignored. It is to be expected and confirms that the Predbat Override automation is working as intended.
 
 ```yaml
 Predbat log:
-Completed run status Freeze charging with Errors reported (check log)
-Warn: record_status – Inverter 0 write to scheduled_charge_enable failed
+
 Warn: Inverter 0 Trying to write scheduled_charge_enable to True didn’t complete (got off)
 
 ```


brickatius:
Revised PR with file edits rather than file uploads. Thanks to @gcoan for the advice.

apps/predbat/inverter.py
```diff
@@ -1665,6 +1666,21 @@ def write_and_poll_switch(self, name, entity_id, new_value):
                 self.count_register_writes += 1
             return True
         else:
+            # Optional suppression for inverters (e.g. LuxPower) that don't reliably reflect
+            # scheduled_charge_enable writes during Freeze charging.
+            status = self.base.get_state_wrapper(entity_id="predbat.status", default="") or ""
+
+            suppress = (
```
springfall2008: This is a bit too hard-wired, what if we just make supress inverter warnings to record_status on an option?

gcoan: also shouldn't hard code 'predbat' prefix, should get the prefix name from the existing self prefix and construct the entity name dynamically (in case people have changed the prefix)

brickatius:
Fear not! I'm working on a much more generic solution which won't suppress any warnings but will retain the status depending on the flag in any apps.yaml as follows.  
```yaml
 suppress_record_status_on_write_fail:
  - scheduled_charge_enable
 ``` 
Change inverter.py brand definitions. 

```py

self.inv_suppress_record_status_on_write_fail = self.base.get_arg(
    "suppress_record_status_on_write_fail",
    default=[],
    index=self.id,
) or []
if isinstance(self.inv_suppress_record_status_on_write_fail, str):
    self.inv_suppress_record_status_on_write_fail = [self.inv_suppress_record_status_on_write_fail]
```
and the failure branch to continue to log the warnings - useful confirmation that manipulating `scheduled_charge_enable` or anything else in the yaml flag is actually working. 

```py

else:
    # Always log the warning so failures remain visible in logs
    self.base.log(
        "Warn: Inverter {} Trying to write {} to {} didn't complete got {}".format(
            self.id,
            name,
            new_value,
            self.base.get_state_wrapper(entity_id=entity_id),
        )
    )

    # Optional suppression: prevent polluting predbat.status via record_status
    # for specific "write" names configured in apps.yaml.
    suppress_record_status = name in getattr(
        self,
        "inv_suppress_record_status_on_write_fail",
        [],
    )

    if not suppress_record_status:
        self.base.record_status(
            "Warn: Inverter {} write to {} failed".format(self.id, name),
            had_errors=True,
        )

    return False
```

I can make it "Freeze charging" status specific if you want but that would again be too hard-wired. 

I'll tweak the docs to reflect these changes. 


Please confirm that you are happy for me to continue along these lines.

apps/predbat/inverter.py
```diff
@@ -1665,6 +1666,17 @@ def write_and_poll_switch(self, name, entity_id, new_value):
                 self.count_register_writes += 1
             return True
         else:
+            # Optional suppression for inverters (e.g. LuxPower) that don't reliably reflect
+            # scheduled_charge_enable writes during Freeze charging.
+            status = self.base.get_state_wrapper(entity_id="predbat.status", default="") or ""
```
springfall2008: We should use predbat.status here, why read state that predbat already has? Probably better just to have an global option to supress record status for warnings on inverter writes

brickatius:
Revised PR attn: @springfall2008 @gcoan 

- Replaced Freeze charge specific logic with a generic `suppress_status_on_write_fail mapping.
- Changes to inverter.py. Now logs write failures but doesn't change 'predbat.status` when suppression is enabled.
- Documentation and template updated accordingly.

gcoan:
Richard, thanks for doing all this, I've been reading your updates to the documentation.

Good to see you have added your new option to the 'I want to add an unsupported inverter' doc, but reading it, I am still unclear of the use case by which you would configure the new suppress setting:

> ### suppress_record_status_on_write_fail
> 
> This option lets you tell Predbat to ignore certain expected write failure warnings.
> 
> Example:
> 
> ```yaml
> suppress_record_status_on_write_fail:
>   scheduled_charge_enable: False
> ```
> When True, Predbat will still log the warning but it will not change `predbat_status` allowing the current Predbat status to be maintained. 
> This can be useful when Predbat expects a control to be enabled, but it has been intentionally disabled elsewhere, causing repeated write attempts to fail as part of normal operation.   

I think you are using it slightly differently for Luxpower, predbat writes to a control and luxpower rejects it, so you need to suppress this error?
Maybe I understood it wrongly what the change is you have made.

Second much more minor comment, the apps.yaml fragment, can you indent everything by two spaces please.  All additions to apps.yaml have to be embedded under the predbat module at the top, so all the apps.yaml fragments we include in the documentation have this intent so that when people copy/paste from the documentation they have half a chance of getting the indentation correct !

brickatius:
Geoffrey (@gcoan),
Something in your last comment got me thinking.......

It is a fact that Predbat is a bit sluggish so I had a picture of old Mr. Predbat shuffling down a corridor. As he went past he turned on the light which I immediately turned off. Good old Mr. Predbat wasn't very happy about this and immedeately started cursing under his breath. When he came round again, about 5 minutes later, he turned the light on again but this time I waited until he had gone round the corner before turning it off again. Because he hadn't seen me turn it off, he had nothing to complain about. 
The solution to the problem is so simple, I can't understand why I didn't think of it before. 
- In the override automation that turns off `scheduled_charge_enable`, add a short delay after it is triggered before the action takes place. 
- As a result, the write succeeds. 
- Because the write is successful there are no warnings and there is no need to do anything at all about suppressing a change in `predbat_status`. I'm sure that Trefor @springfall2008 will be happy that no changes are needed to `inverter.py` 
- Job done!

Because of recent updates, inverter.py is now out of sync, so the simplest solution is for me to close this PR, rather than trying to undo the changes. 

I'll do a new PR with the neccesary changes to apps.yaml and inverter-setup.md

PS I had indent the luxpower.yaml the way it was because the `inverter:` block isn't a yaml list and this bit

```yaml
 suppress_record_status_on_write_fail:
    scheduled_charge_enable: False
```
had to go outside of it. I could have made it a list but that would have taken the lid of a whole can of worms.

gcoan:
Glad that my comment gave you a brainwave about an easier way to do this,  suppressing warnings isn't really good, so sounds like you're on track to a much simpler solution.

Speed of predbat very much depends on what you are running it on and what config settings you have.  Trefor has made a number of performance improvements over time.  I run my HA on a 'tiny' windows 10 PC as a virtual server, and I get predbat plan updates completing in well under a minute, sometimes as quick as 5 seconds.  Last year Predbat could take up to 5 minutes for an execution run so its a massive improvement.

brickatius:
I presume there aren't any other issues about the rest of what I was doing? I'm in the process of editing files now so it would be nice to know if there are any alterations I should make before doing another commit. 

I've set the delay in the Luxpower Freeze Predbat Override automation to 10 seconds after having tried 2. Do you think that is long enough?

PR #3261, opened 2026-01-19
Updated Freeze Charging & Exporting for LuxPower inverters

brickatius:
**Description**

This update refines the LuxPower documentation and example template to showcase an improved Freeze Charging configuration for LuxPower inverters.

Introducing a brief delay before modifying `scheduled_charge_enable` results in a more stable and predictable behaviour. This adjustment:

- **Eliminates write‑warning noise during Freeze Charging**
- **Removes the need for `maintain_freeze_charge_status` or `suppress_record_status_on_write_fail` entries in `luxpower.yaml`**
- **Avoids any requirement to patch `inverter.py` to suppress warnings or `record_status` updates**

This simplifies configuration, reduces maintenance overhead, and aligns the integration with best‑practice automation patterns.

**Closing**

Feedback is welcome, and I’m happy to adjust anything that helps this land smoothly.

### DIFF
--- a/docs/inverter-setup.md
+++ b/docs/inverter-setup.md
@@ -669,7 +669,9 @@ device class: Battery
 state class: Measurement
 ```
 
-- Create the following number helper. The maximum value (in Watts) can be found in your inverter datasheet. A more accurate figure can be obtained by observing the flow chart in the Monitor section of the LuxPower app/portal or by inspecting `sensor.lux_battery_flow_live` when the battery is force charging or discharging.
+- Create the following number helper. The maximum value (in Watts) can be found in your inverter data sheet.
+ A more accurate figure can be obtained by observing the flow chart in the Monitor section of the LuxPower app/portal
+ or by inspecting `sensor.lux_battery_flow_live` when the battery is force charging or discharging.
 
 ```yaml
 name: Battery Rate Max
@@ -684,7 +686,11 @@ unit of measurement: W
 ---
 
 **Important:**
-The Freeze Charging and Freeze Exporting setup described below relies on a set of carefully designed helpers and automations that work together. Each component has a specific role in safely entering, maintaining, and exiting Freeze Charging mode. Removing or skipping any part can lead to missed triggers, stuck AC charging, or incomplete cleanup. For reliable operation, make sure all helpers and automations in this section are created exactly as described before using Freeze Charging or Freeze Exporting modes. All of the automations apart from LuxPower HA Startup Reset remain disabled when Predbat is not Freeze Charging.
+The Freeze Charging and Freeze Exporting setup described below relies on a set of carefully designed helpers and automations that work together.
+Each component has a specific role in safely entering, maintaining, and exiting Freeze Charging mode.
+Removing or skipping any part can lead to missed triggers, stuck AC charging, or incomplete cleanup.
+For reliable operation, make sure all helpers and automations in this section are created exactly as described before using Freeze Charging or Freeze Exporting modes.
+All of the automations apart from LuxPower HA Startup Reset remain disabled when Predbat is not Freeze Charging.
 
 ---
 
@@ -723,7 +729,8 @@ name: Freeze Charge Guard
 entity_id: input_boolean.freeze_charge_guard
 ```
 
-The `freeze_charge_guard` helper acts as a lifecycle gate. It is enabled only when Predbat explicitly requests Freeze Charging and is cleared on exit, watchdog abort, or Home Assistant restart. All Freeze Charging automations check this guard to prevent unintended operation.
+The `freeze_charge_guard` helper acts as a lifecycle gate. It is enabled only when Predbat explicitly requests Freeze Charging and is cleared on exit, watchdog abort, or Home Assistant restart.
+All Freeze Charging automations check this guard to prevent unintended operation.
 
 **Binary sensor template helper**¹
 
@@ -890,7 +897,7 @@ mode: single
 ---
 
 - Create the **Freeze Charge Exit** automation to cleanly restore inverter state when Freeze Charging ends.
-
+<!-- cspell:ignore startswith -->
 ```yaml
 alias: LuxPower Freeze Charge Exit
 description: |
@@ -962,7 +969,9 @@ mode: single
 
 ```
 
-Occasionally, when a Manual Freeze Charge is requested, Predbat may immediately decide that **Hold Charging** is the more appropriate state based on current conditions. In this case, Freeze Charging automations may remain enabled even though Predbat reports Hold Charging. The watchdog safely exits Freeze Charging after a short grace period.
+Occasionally, when a Manual Freeze Charge is requested, Predbat may immediately decide that **Hold Charging** is the more appropriate state based on current conditions.
+In this case, Freeze Charging automations may remain enabled even though Predbat reports Hold Charging.
+The watchdog safely exits Freeze Charging after a short grace period.
 
 - Create the **Freeze Charge Watchdog** automation to handle cases where Manual Freeze Charging immediately transitions to **Hold Charging**.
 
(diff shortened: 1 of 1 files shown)


brickatius:
I appreciate that this isn't the right place for this. However, please accept my sincere apologies for the messiness of this PR. It results primarily from my ignorance of how GitHub works (I only learnt about pre-commit checks yesterday) but also from the death of my desktop computer which was running the latest version of Ubuntu. I have had to resort to the resurrection of a Mac dating back to 2013!

brickatius:
Is there anything further I need to do to get this PR over the line ( @springfall2008 @gcoan )?
The unverified commit about the Troubleshooting section authored by Richard is from me. 
I've made all the corrections flagged up by the workflow checks, however if you want me to close this PR and start afresh, I'm happy to do so.

springfall2008:
Looks good, will merge it

ppmt:
How should I proceed if I want to merge my existing configuration file with the new version?

gcoan:
> How should I proceed if I want to merge my existing configuration file with the new version?

you'll have to copy the new predbat template and re-apply your configuration to it.  There's no easy way to merge apps.yaml versions other than in  file editor

brickatius:
> How should I proceed if I want to merge my existing configuration file with the new version?

As @gcoan says but you don't necessarily need to replace the whole template, only the last couple of lines of the 'inverter:' section down as far as 'battery_voltage'

ppmt:
Ok. I will try to put them side by side and work from there.

Is the new config the "recommended" way to go as far as settings?

brickatius:
> Ok. I will try to put them side by side and work from there.
> 
> 
> 
> Is the new config the "recommended" way to go as far as settings?

It's up to you. Using the system charge and grid cut-off entities avoids any confusion with different values being set in apps.yaml and the inverter. 
Freeze charge & export are integral parts of Predbat so it's worth a try. This time of year you'll see more Freeze charging than exporting. It's all about maximising cost saving.

ppmt:
It probably isn't the place to ask but what is the difference between Freeze Charging and Holding Charge?

gcoan:
> It probably isn't the place to ask but what is the difference between Freeze Charging and Holding Charge?

they are effectively the same https://springfall2008.github.io/batpred/what-does-predbat-do/#predbat-status

brickatius:
> It probably isn't the place to ask but what is the difference between Freeze Charging and Holding Charge?

In both cases the battery will never discharge. 
The description of Freeze charging is a bit confusing. Basically it's the same as ordinary Demand mode but any house load that would have come from the battery comes from the grid instead. 
In LuxPower terms - 
Hold charging:
Exactly the same as Charge Priority / Charge First i.e. all the house load comes from the grid. 
Freeze charging:
Solar > House - house load from solar (identical to a GivEnergy inverter)
Solar < House - house load from grid (slightly different from GE inverter. Solar doesn't contribute to house load but there's a trade-off, more power in the battery which Predbat will account for)
No solar - house load from grid (identical to a GE inverter).

ppmt:
Thanks for the explanation. When you say that none of the power will come from the battery, I assume it is based on cost comparisons (ie exporting is more efficient that charging the battery?)
Basically what I am asking is that when the electricity is expensive (like evening) then Predbat will use the battery ?

brickatius:
Correct. I've been trialing this for a couple of months (with Octopus Flux and Agile import / fixed export) and I haven't seen Predbat Freeze charge during the most expensive import periods.

ppmt:
Ok I have enabled it now by following the documentation and all seems ok. Except for one thing.

In the documentation is says:

```
Enable Freeze Charging

    Ensure switch.predbat_set_charge_freeze is turned On.
```
The switch doesn't exist for me.

gcoan:
> When you say that none of the power will come from the battery, I assume it is based on cost comparisons (ie exporting is more efficient that charging the battery?)
> Basically what I am asking is that when the electricity is expensive (like evening) then Predbat will use the battery ?

Its not really a case of what the export rate is, freeze charge is used when the import rate is such that it is more economical to directly import from the grid to run the house (rather than deplete the battery),  but its not cheap enough to make it worthwhile charging the battery.  Running off grid means you don't incur the battery discharge and inverter conversion losses.

> Enable Freeze Charging
>
>    Ensure switch.predbat_set_charge_freeze is turned On.
>
> The switch doesn't exist for me.

check the documentation https://springfall2008.github.io/batpred/customisation/ you will see this is an expert mode switch so you need to turn expert mode on first

brickatius:
> Ok I have enabled it now by following the documentation and all seems ok. Except for one thing.
> 
> 
> 
> In the documentation is says:
> 
> 
> 
> ```
> 
> Enable Freeze Charging
> 
> 
> 
>     Ensure switch.predbat_set_charge_freeze is turned On.
> 
> ```
> 
> The switch doesn't exist for me. 

Look in the Predbat User Interface (UI) under 'Config', scroll down and you will see 'Set Charge Freeze' and also 'Set Export Freeze' 
Thanks for the heads up. I need to clarify the documentation.

ppmt:
Yes it is all setup now but as @gcoan said I had to enable expert to get that one to appear.

I will monitor over the next few days to see how it behaves

Thanks all for your patience and help :+1: 

image


PS: on a completely different subject. I noticed that the luxpower_app template is not using the new format for the carbon_intensity

gcoan:
I've updated the documentation to highlight the need to turn expert mode on, and have changed ALL the templates for the new carbon intensity config:

brickatius:
@ppmt Don't be alarmed if you see a few seconds of battery charging from the grid when Freeze charging starts. This is expected. It's caused by a delay between Predbat enabling the LuxPower AC Charge and the Stop AC Charge SoC% being set to the present battery SoC%.

brickatius:
> I've updated the documentation to highlight the need to turn expert mode on, and have changed ALL the templates for the new carbon intensity config:
> 
> 

Geoffrey, it would be clearer still if we mention the need to go to the Web UI/Config to set Expert Mode to True and then Set Charge Freeze to True, rather than referring to the Predbat entity itself. Likewise at the end of the Freeze export section. Are you happy to do that?

gcoan:
> Geoffrey, it would be clearer still if we mention the need to go to the Web UI/Config to set Expert Mode to True and then Set Charge Freeze to True, rather than referring to the Predbat entity itself. Likewise at the end of the Freeze export section. Are you happy to do that?

No

The danger of being too prescriptive is it starts an endless journey into telling people where to click and what to do. Some people don't use the web interface, some don't even use Home Assistant.

And there are lots and lots of places in the documentation where settings are described and cross referenced together and if we had to add more detail then the docs would be even longer than they are now!  Inverter setup is already over 2500 lines long.

I'm happy to point out that set charge freeze is an expert mode switch and you need to turn expert mode on first, but TBH even that is pushing it, the switch is described in the customisation section as being expert mode so that should have been enough.
Also, we don't remind people to turn this switch on (and expert mode) for all the other inverter types that have charge freeze functionality, so it's already being over-generous !

brickatius:
> > Geoffrey, it would be clearer still if we mention the need to go to the Web UI/Config to set Expert Mode to True and then Set Charge Freeze to True, rather than referring to the Predbat entity itself. Likewise at the end of the Freeze export section. Are you happy to do that?
> 
> No
> 
> The danger of being too prescriptive is it starts an endless journey into telling people where to click and what to do. Some people don't use the web interface, some don't even use Home Assistant.
> 
> And there are lots and lots of places in the documentation where settings are described and cross referenced together and if we had to add more detail then the docs would be even longer than they are now!  Inverter setup is already over 2500 lines long.
> 
> I'm happy to point out that set charge freeze is an expert mode switch and you need to turn expert mode on first, but TBH even that is pushing it, the switch is described in the customisation section as being expert mode so that should have been enough.
> Also, we don't remind people to turn this switch on (and expert mode) for all the other inverter types that have charge freeze functionality, so it's already being over-generous !
> 

Ok. Understood.

Issue #2785, opened 2025-10-18
predbat.cost_today  ammount jumps after unplugging car from charge


retr083:
**Describe the bug**
After charging on IOG outside the off peak times with slots given through the ohme predbat component that predbat.cost_today jumps up in price. The predbat.cost_today_car also shows 28.22p/kWh after charging has finished

**Expected behaviour**
predbat.cost_today should not increase 

**Predbat version**

v8.25.3

**Environment details**

-givenergy 5kw invertor/ 9.5kwh battery
- Predbat & Givtcp addons
- Anything else?

**Screenshots**

<img width="560" height="644" alt="Image" src="https://github.com/user-attachments/assets/379814f0-0cb6-4187-b7c9-7ed2322ee4a7" />

<img width="667" height="141" alt="Image" src="https://github.com/user-attachments/assets/ccab84f5-3b1b-461f-9125-3742851b8507" />

gcoan:
I am pretty sure this has been raised before and the conclusion was that this is just the way it works

Predbat recalculates cost today each time it runs, and when you unplug the car,  Octopus removes the cheap rate IOG slots from your account and so the predbat recalculation of cost goes up.

BenCathro:
I'm also on IOG and use an Ohme charger with an AIO home battery. I've been struggling to pin down some inconsistencies in cost calculations among other issues and this might be part of it.

@gcoan, couple questions if you don't mind answering:

1 - Does Octopus discard all bonus cheap slots that are reported by the api for all types of car/charger when unplugged or is this just an Ohme issue?

2 - Does this mean the household cost tracking is also messed up as when Predbat charges up the home battery during a bonus slot this is then recorded as full price after unplugging?

Bonus question. Can Predbat be configured to pull cheap slot pricing from a helper that records when cheap slots happen?

Thank you.

gcoan:
> @gcoan, couple questions if you don't mind answering:

I'll have a go, but I don't have an EV so my answers are based on my understanding of what happens and what other people have said happens

> 1 - Does Octopus discard all bonus cheap slots that are reported by the api for all types of car/charger when unplugged or is this just an Ohme issue?

I think it happens for all EV chargers, as I said, I'm pretty sure I have seen similar reports in the past.  You can search the predbat github issues, filter on closed, to see prior conversations

> 2 - Does this mean the household cost tracking is also messed up as when Predbat charges up the home battery during a bonus slot this is then recorded as full price after unplugging?

[...]

PR #3340, opened 2026-02-09
Persist rates and overrides 

raldred:
This PR is work in progress.

I'm looking to propose a solution to the issue where rate are lost for services like IOG and Axle.
If a car is unplugged, the off-peak slots are lost, so cost history is calculated in correctly.
Similar issue occurs if predbat is restarted. Rate overrides are lost.

This aim here is to create a store to persist rates the slot they apply to has past. History of what rates were, should never be forgotten.

### DIFF
--- a/apps/predbat/fetch.py
+++ b/apps/predbat/fetch.py
@@ -911,7 +911,7 @@ class Fetch:
         if self.rate_store:
             today = datetime.now()
             stored_import, stored_export = self.rate_store.load_rates(today)
-            
+
             if stored_import or stored_export:
                 # Merge frozen past rates into current rate tables
                 for minute in range(0, self.minutes_now):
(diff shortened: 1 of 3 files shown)


raldred:
@springfall2008 Just putting this up. No rush. It's not complete yet, hence draft status.
Are you happy for me to pursue? Do you have any thoughts on the approach, originally discussed in https://github.com/springfall2008/batpred/issues/2785

springfall2008:
I understand what you are trying to do here, allow an accurate record of the historical rates after the various overrides.

However there are two much simpler approaches:

1. Just record the current rate predbat is using in a HA sensor, now we have the history.
or
2. Save the rates from the actual rate table built in fetch, thus removing the need to change lots of components.

The downside to the HA history one is its kind of timing dependent, in that if predbat doesn't update on each period then it might have gaps.

I could implement 2 with very few lines of code, unless you see a downside to that?

raldred:
Thanks for the feedback! You're absolutely right I think with point 2.

I intentionally created this as separate surface area to avoid polluting the main modules and to avoid stepping on your toes while we figure out the best pattern.
However, in hindsight, it's just added extra complication.

I'm just looking at refactoring this to use the rates table in fetch directly, which would eliminate the need for components to call out to the rate store. So that should be much simpler.

The main issue I haven't been able to solve here, there's a race condition with the current architecture. By default, we only update sensor data every N minutes. If something changes relating to rates from Octopus, Car, Axle, or other components and the change is short lived, we could miss it.

The only real solution would be more frequent sensor updates. Users can achieve this by using the watch list feature, but I expect most people aren't doing that by default.

I don't think it's right to address here, my main goal was, as you say to have a much more accurate record for history and cost calculation, I think that should be a nice improvement on it's own.

I'll update this PR later once I've refactored.

PR #3381, opened 2026-02-16
Disable car charging subtraction in ML if no cars defined

DFEvans:
The ML Load prediction will default to subtracting car charging load, assuming that `car_charging_hold` default to true ([the docs state it is off by default](https://springfall2008.github.io/batpred/apps-yaml/#car-charging-filtering)), and not verifying whether any cars are defined.

Additionally, the `car_delta` value is left in a stale state from previous iterations if car load was previously detected and but is then disabled.

Disable `car_charging_hold` by default, check that `num_cars>=1` before removing car load, and ensure that `car_delta` is reset to 0 if there is no charging load detected.

### DIFF
--- a/apps/predbat/config.py
+++ b/apps/predbat/config.py
@@ -596,7 +596,7 @@ CONFIG_ITEMS = [
         "name": "car_charging_hold",
         "friendly_name": "Car charging hold",
         "type": "switch",
-        "default": True,
+        "default": False,
     },
     {
         "name": "car_charging_manual_soc",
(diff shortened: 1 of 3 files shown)


DFEvans:
Do shout if the changes to defaults aren't sensible - it's the "don't subtract car load if there's no car" and "set car_delta to 0.0 if there is no detected charge" that makes life better.

springfall2008:
The way predbat is implemented in the rest of the code car_charging_hold still works even with num_cars=0, as the user may not be controlling the car inside predbat but just filtering out the load.

DFEvans:
Fair point. It looks like you've done other changes around making sure the values are reset to zero correctly, so will close this

PR #3385, opened 2026-02-17
HA cache fixes

DFEvans:
The HA cache stores data as mutable lists, and returns these lists when a cache hit occurs. However, this means that if a caller modifies the result directly, then these changes are passed back to the cache.

Additionally, the cache only ever added new entries. If it was queried for a longer history (e.g. 7 days, then 8 days), it reported itself as having the longer length, but did not actually contain those values.

Ensure that the cache always returns an independent copy of its own data, and populate entries older than the earliest entry in cache when updating.

### DIFF
--- a/apps/predbat/ha.py
+++ b/apps/predbat/ha.py
@@ -88,7 +88,7 @@ class HAHistory(ComponentBase):
                 if tracked:
                     self.update_entity(entity_id, history_data)
                 result = [history_data]
-        
+
         return deepcopy(result)
 
     def prune_history(self, now):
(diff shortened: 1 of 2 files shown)


springfall2008:
Did you actually find some code that modifies the returned cache? Making a copy of the data will increase memory footprint.

springfall2008:
I merged this PR and then backed out the copy and instead I'm adding copy to any location that modified the history, I think this is better for performance :)

springfall2008:
https://github.com/springfall2008/batpred/pull/3386

DFEvans:
> Did you actually find some code that modifies the returned cache? Making a copy of the data will increase memory footprint.

> I merged this PR and then backed out the copy and instead I'm adding copy to any location that modified the history, I think this is better for performance :)

Yep, that's fair - I did it entirely as a precaution, mainly thinking that it's easier to catch in one place than in each use. You can tell I'm from the part of programming where memory is cheap and plentiful!

PR #3428, opened 2026-02-24
Fix MCP startup transport close by adding streamable HTTP behaviour

adamdoran:
Serve an SSE channel on GET /mcp for streamable HTTP clients, stop returning a JSON-RPC body for notifications/initialised (HTTP 202 instead), and keep initialise capabilities schema compliant for protocol 2024-11-05.

These changes enable the use of Predbat MCP with the OpenAI Codex CLI.


springfall2008:
Why did you close this one?

adamdoran:
That was unintentional. I'll open another one later.

PR #3484, opened 2026-03-04
Fix rate text descriptions to use tariff range

warksit:
## Summary
- Rate text descriptions ("cheap", "expensive", etc.) in the status message now compare rates to other rates on the tariff, instead of using the battery charging threshold
- Previously, the charging threshold made most rates appear "expensive" or "very expensive" on multi-tier tariffs like Cosy Octopus, even for the cheapest available slots
- Now the cheapest third of rates are "cheap", middle third "expensive", and top third "very expensive"

## Test plan
- [ ] Cosy Octopus: 14p = cheap, 28p = expensive, 43p = very expensive
- [ ] Agile: rates spread across the three bands proportionally
- [ ] Flat rate tariff: shows "fixed"
- [ ] Negative/free rates still handled correctly

### DIFF
--- a/apps/predbat/output.py
+++ b/apps/predbat/output.py
@@ -588,14 +588,16 @@ class Output:
                 text = "zero"
             elif rate < 0:
                 text = "negative"
-            elif rate <= (export_cost_threshold * 0.5):
-                text = "very low"
-            elif rate < export_cost_threshold:
-                text = "low"
-            elif rate <= (export_cost_threshold * 1.5):
-                text = "good"
             else:
-                text = "very good"
+                rate_frac = (rate - self.rate_export_min) / (self.rate_export_max - self.rate_export_min)
+                if rate_frac <= 0.25:
+                    text = "very low"
+                elif rate_frac <= 0.5:
+                    text = "low"
+                elif rate_frac <= 0.75:
+                    text = "good"
+                else:
+                    text = "very good"
         return text
 
     def get_rate_text(self, minute, export=False, with_value=False):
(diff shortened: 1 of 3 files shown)


apps/predbat/output.py
```diff
@@ -581,14 +573,14 @@ def band_rate_text(self, rate, export=False):
...
                 text = "negative"
-            elif rate <= import_cost_threshold * 0.5:
-                text = "very cheap"
-            elif rate <= import_cost_threshold:
-                text = "cheap"
-            elif rate <= (import_cost_threshold * 1.5):
-                text = "expensive"
             else:
-                text = "very expensive"
+                rate_frac = (rate - self.rate_min) / (self.rate_max - self.rate_min)
```
springfall2008, line 577: This could give a divide by zero error on a fixed rate tariff?

warksit: Yes, will add a check for that and resolve the test error (thought I had run them before submitting, sorry)

warksit:
Thanks for the review. The divide-by-zero was already guarded (the `rate_min == rate_max` check returns "fixed" before reaching the division), but there was a real bug: I removed the `export_cost_threshold` local variable definition when switching import to `rate_frac`, but left the export branch still referencing it — causing a `NameError` on any export rate lookup. Fixed in f6d4b6c.

Export now uses the same `rate_frac` approach with 4 bands (very low / low / good / very good, at 0.25/0.5/0.75 boundaries). Also added dedicated `band_rate_text` tests covering flat rate, Cosy Octopus, and Flux tariffs (import + export).

Issue #1761, opened 2024-12-19
Getting more from Solar Assistant 


xhemp:
Decided to open a feature request from Solar Assistant.
@sparkyhughes managed to enable has_target_soc and support_charge_freeze.
I have been testing on my side and now got just one error which is weird: 

2024-12-19 13:10:51.958903: Warn: Inverter 0 Trying to write 67 to charge_limit didn't complete got 66
2024-12-19 13:10:52.024017: Warn: record_status Warn: Inverter 0 write to charge_limit failed

Although it's weird as I am able to change it on Home Assistant and Solar Assistant accepts it.
And it looks to me it's indeed being accepted but PredBat says it is not, interesting.

xhemp:
Checking if @springfall2008 can have a look at this.

xhemp:
@sparkyhughes I can see my inverter trying to charge the battery with the grid after a few minutes even though battery first grid charge is set to disabled, can you please check if you see the same?

sparkyhughes:
Does Solar-Assistant Inverter settings show "Battery first grid charge"  as  "Disabled"? - in SA under Power menu -> Inverter Settings
Enabled and Disabled start with a capital letter, so went with that in the apps.yaml file

xhemp:
Yes, it's showing Disabled on HA and SA, it works at first but then it starts charging from the battery after a few minutes which is strange, I am retesting now and will record a video.

sparkyhughes:
I know without the stop charge value being changed from 100, if there is enough solar, and the inverter is on Battery first mode, it will charge the battery.  It shouldn't from the grid though

xhemp:
Can't replicate, will wait until a Charge Freeze again to test!

sparkyhughes:
One other SA / Growatt issue I've seen overnight relates to Charge Hold.
When this was activated last night, it stops the charge (changes the Growatt mode to Load First), sets the target SOC to 10% and sets the discharge rate to 0W.  
However, Growatt Inverters (mine is SPH model) doesn't have a discharge rate for Load First mode, so essentially the battery discharges into the car.

```
2024-12-20 03:00:28.523696: Inverter 0 current discharge rate is 3000.0W and new target is 0W
2024-12-20 03:00:33.586022: Inverter 0 Wrote 0 to discharge_rate, successfully now 0
2024-12-20 03:00:33.994493: Disabling battery discharge while the car 0 is charging
[...]

PR #3533, opened 2026-03-09
feat: Extend Solar Assistant (Growatt) functionality - Schedules and target SoC

ryt51V:
I'm a bit green on pull requests, so please bear with me and sorry if this is a bit too wordy.

### Changes
Solar Assistant (SA) inverter definition in config.py:
- Enable: **has_charge_enable_time**, **has_discharge_enable_time**, **has_target_soc**, **has_reserve_soc**, **charge_time_entity_is_option**
- Disable: **target_soc_used_for_discharge**
- Set **charge_time_format** (HH:MM:SS)

Solar Assistant Growatt SPH/SPA templates:
- Set **reserve** to _both_ the load first and grid first stop discharge entities.
- For the charge schedule ("battery first"), set **charge_start_time**, **charge_end_time**, **charge_limit** and **scheduled_charge_enable**
- For the discharge schedule ("grid first"), set **discharge_start_time**, **discharge_end_time**, and **scheduled_discharge_enable**
- Remove irrelevant comments / unused functionality.  (In particular the template used to set **energy_control_switch**, which I don't think is functional for SA?)

### Motivation
Predbat doesn't currently use all of the features available in Solar Assistant for Growatt inverters.  There is an open feature request #1761 for this, although I only just found that and it didn't feed into this pull request.

SA exposes the charging schedules for the **Battery first** (charge) and **Grid first** (discharge) modes.  I'm not sure what the project's thoughts are on this, but in my opinion it's preferable to using these to service calls to reduce the chance of a Predbat failure leaving the inverter in Battery First or Grid First states 24/7.  So this change allows using those, with the templates using both modes' time period 1.

For scheduled charge, SA exposes the target SoC ("Stop Charging SOC" in ShinePhone), so this change allows Predbat to set that.

Similarly, SA exposes the reserve SoC ("Stop Discharging SOC" in ShinePhone).  This is under separate settings/entities for **Load first** (demand) and **Grid first** modes, so I've added both to the template.  (For some reason the template already had the Grid first entity added, despite **has_reserve_soc** not being enabled in config.py, meaning it wouldn't work?)

### Testing
I first made these config.py changes a few months ago and have repeated them after upgrading Predbat.

Overall it's worked really well, but note main issues/limitations of my testing:
1. I have a Growatt SPH system.  I'm assuming that SPA systems are functionally equivalent in SA.

2. I only have my Predbat in **Control charge** mode, as for my TOU tariff it doesn't make much sense to ever force export.  Therefore I haven't tested the Grid First / discharge side of this.

3. For some reason, after restarting Home Assistant, the scheduled time entities (particularly **growatt_sph_battery_first_slot_1_end**) may be set to **unknown** or blank.  This is presumably a bug with the SA integration in HA, as the SA web interface shows the scheduled time no problem.  This does not stop manually setting the unknown entity in HA.\
\
In previous versions of Predbat this caused an error, and Predbat wouldn't attempt to set the unknown entity.  I'm waiting to retest on Predbat v8.34.2 and see if this is still an issue.

I also don't know if this would be a breaking change for existing installs that are missing the new settings in apps.yaml.  Would Predbat handle this gracefully?

In any case, ideally someone else on SA/Growatt can test out this PR.


ryt51V:
I think I've made headway with one of the issues with this PR that I mentioned:

> For some reason, after restarting Home Assistant, the scheduled time entities (particularly growatt_sph_battery_first_slot_1_end) may be set to unknown or blank. This is presumably a bug with the SA integration in HA, as the SA web interface shows the scheduled time no problem. This does not stop manually setting the unknown entity in HA.
> 
> In previous versions of Predbat this caused an error, and Predbat wouldn't attempt to set the unknown entity. I'm waiting to retest on Predbat v8.34.2 and see if this is still an issue.

First, it looks like this in v8.34.2 this appears as a warning in Predbat rather than an error.

Second, it seems the issue is due to how Solar Assistant handles different changes to the work mode.  If you manually enable **Battery First** in Solar Assistant, what it does is configure the first schedule to 00:00-23:59 and then enables it.  I'm guessing this is because Growatt doesn't expose manual control of work mode, and so Solar Assistant "fakes" manual control by modifying the schedules as required.

On the Home Assistant side it can't seem to handle that 23:59 and instead shows it as "unknown".  This is possibly because **select.growatt_sph_battery_first_slot_1_end** is a select entity that only has times available in 15 minute increments.

The behaviour appears to be the same if you set the Work Mode using the HA entity **select.growatt_sph_work_mode_priority** - which is what we're currently configuring in apps.yaml (charge_start_service / charge_stop_service / discharge_start_service).

I'm not yet sure why this only seems to cause a problem for Predbat after restarting HA.  However I've noticed in the predbat logs that it seems to be both setting schedules _and_  calling the defined service (charge_start_service / charge_stop_service / discharge_start_service), which I'm assuming is by design when you have both schedules and service calls configured.

The most obvious workaround is to simply not define the service calls in apps.yaml, and instead purely rely on the schedules.  (Though this problem could still trigger if the work mode is manually changed in HA/SA.)

In my opinion schedules is still surely the preferred way to manage SA/Growatt.  (Particularly because I'm not sure how Predbat can set the charge limit when not part of a schedule?)

I'll leave my setup to run with only schedules enabled for a few days and verify it works as expected.

PR #3557, opened 2026-03-13
DRAFT: feat: add GatewayMQTT component for ESP32 MQTT integration

mgazza:
## Summary

- Add GWMQTT inverter type to `INVERTER_DEF` in config.py for ESP32 gateway MQTT integration
- Add `gateway.py` — full MQTT component (aiomqtt) with telemetry decoding, command publishing, token refresh, exponential backoff reconnection
- Register `GatewayMQTT` in `COMPONENT_LIST` (components.py) with conditional import
- Add per-inverter HA-style entity injection (sensor., number., switch., select. prefixes with serial suffix)
- Add `automatic_config()` for PredBat `set_arg()` registration on first telemetry
- Add plan change detection to skip redundant ExecutionPlan publishes
- Add energy counter, schedule time, battery SoH/rate_max, inverter_time entities
- Add Python protobuf bindings and proto schema (PlanEntry, ExecutionPlan)
- Add MQTT integration test for plan publish format

## Test plan

- [ ] Verify gateway.py imports cleanly with aiomqtt + protobuf installed
- [ ] Verify conditional import degrades gracefully when aiomqtt not installed
- [ ] Verify GWMQTT inverter type is recognized by plan engine
- [ ] Verify entity names match between _inject_inverter_entities and automatic_config set_arg mappings
- [ ] Run existing gateway MQTT integration test

### DIFF
--- a/apps/predbat/tests/test_gateway.py
+++ b/apps/predbat/tests/test_gateway.py
@@ -9,7 +9,7 @@ import math
 
 sys.path.insert(0, os.path.join(os.path.dirname(__file__), ".."))
 
-from proto import gateway_status_pb2 as pb
+import gateway_status_pb2 as pb
 
 import importlib.util
 


mgazza:
## Code Review: GatewayMQTT Component

**Verdict: Approve**

### Correctness
- `decode_telemetry()` static method and `ENTITY_MAP` approach is clean and testable
- Token refresh via `oauth-refresh` with `provider: "predbat_gateway"` is consistent with FoxESS OAuth pattern
- `_check_token_refresh` correctly uses `_refresh_in_progress` guard to prevent concurrent refreshes
- `is_alive()` handles the "gateway offline but broker connected" case well

### Minor Issues (follow-up)

1. **`instance_id` extraction** (`gateway.py`): `self.args.get("user_id", "")` — verify `user_id` is injected by `ComponentBase` framework since it's not in the explicit args defined in `components.py`

2. **Reconnect backoff has no cap**: `_RECONNECT_BASE_DELAY * (2 ** attempt)` with 10 max attempts means max delay is ~85 min. Consider capping at 60s

3. **`build_command` selective fields**: Only `mode`, `power_w`, `target_soc` kwargs are included — other kwargs silently dropped. Fine for v1, note for future commands

### Test Coverage
12 tests covering protobuf round-trip, entity mapping, plan serialization, command format, EMS entities, and JWT token handling. Good coverage for v1.

springfall2008:
From an infra perspective the 'proto' directory won't be downloaded by default by the installed
And there's a new library here from google, do we know how big it is ? 
It might make sense to guard this in some way so if the library is missing its safe?

mgazza:
OSS community cant really use this stand alone so im not sure what we should do about this.

mgazza:
Good points, let me address each:

**Proto directory**: The `proto/` dir contains pre-compiled Python bindings (just `.py` files) — no `.proto` compilation needed at runtime. We can move them into the `apps/predbat/` tree so they ship with the standard install, or alternatively bundle them directly inside `gateway.py` since there are only two message types.

**Protobuf library size**: `protobuf` (google-protobuf) is ~1.8MB installed. It's a pure-Python package with no native compilation needed. That said, if size is a concern we could use lightweight alternatives like `proto-plus` or even hand-roll the decode since our messages are simple (2 message types, ~15 fields total).

**Import guard**: Agreed — I'll wrap the protobuf import in a try/except so if the library isn't installed the component simply doesn't register. PredBat continues to work normally without it. This is SaaS-only functionality anyway so OSS users won't be affected.

I'll push the import guard fix shortly.

apps/predbat/gateway.py
```diff
@@ -0,0 +1,924 @@
...
+        pfx = f"{self.prefix}_gateway_{suffix}"
+
+        bat = inv.battery
+        self.set_state_wrapper(f"sensor.{pfx}_soc", bat.soc_percent)
+        self.set_state_wrapper(f"sensor.{pfx}_battery_power", bat.power_w)
+        self.set_state_wrapper(f"sensor.{pfx}_battery_voltage", bat.voltage_v)
+        self.set_state_wrapper(f"sensor.{pfx}_battery_current", bat.current_a)
+        self.set_state_wrapper(f"sensor.{pfx}_battery_temperature", bat.temperature_c)
+        if bat.capacity_wh:
+            self.set_state_wrapper(f"sensor.{pfx}_battery_capacity", bat.capacity_wh)
```
mgazza: `soc_max` unit mismatch: this injects raw Wh from the proto (e.g. `15970`) but PredBat's `soc_max` expects kWh (e.g. `15.97`) when `soc_units: "%"`. The energy counter entities already divide by 1000 — this needs the same.

```suggestion
            self.set_state_wrapper(f"sensor.{pfx}_battery_capacity", round(bat.capacity_wh / 1000.0, 2))
```

mgazza:
### Code review

Found 1 issue:

1. `battery.rate_max_w` is referenced in `ENTITY_MAP`, `_inject_inverter_entities`, and `automatic_config`, but the `BatteryStatus` proto message only defines 8 fields (`soc_percent` through `capacity_wh`) — there is no `rate_max_w` field. In proto3, accessing a non-existent field silently returns 0, so `if bat.rate_max_w > 0` is always false and the battery rate max entity/arg is never populated from live gateway data.

https://github.com/springfall2008/batpred/blob/2052d61e0107d4f9a785d8221ccc4a0ca86270a0/apps/predbat/gateway.py#L57-L59

Proto schema (no `rate_max_w` field):

https://github.com/springfall2008/batpred/blob/2052d61e0107d4f9a785d8221ccc4a0ca86270a0/apps/predbat/proto/gateway_status.proto#L8-L18

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

mgazza:
### Code review

Found 3 issues:

1. **`soc_max` injected as Wh instead of kWh** -- `_inject_inverter_entities` stores `bat.capacity_wh` raw (e.g. 9500 for a 9.5 kWh battery) into the `battery_capacity` entity. `automatic_config` then registers this entity as `soc_max`, but PredBat's core engine treats `soc_max` as kWh. This makes the battery appear 1000x larger than it is, breaking all charge/discharge calculations. The energy counter entities in the same file correctly convert `Wh / 1000.0` -- battery capacity should do the same.

https://github.com/springfall2008/batpred/blob/4cecb2e596db58a72b0d016ef434f2d2fd6b1714/apps/predbat/gateway.py#L371-L373

2. **Protobuf version mismatch raises `RuntimeError`, not `ImportError`** -- The generated `gateway_status_pb2.py` calls `ValidateProtobufRuntimeVersion(..., 6, 33, 4, ...)` which raises a `RuntimeError` (or `VersionError`) if the installed protobuf version doesn't match. Both `gateway.py` (line 24) and `components.py` (line 40) only catch `ImportError`. If a user has protobuf installed at the wrong version, the uncaught exception crashes PredBat startup for all users, not just gateway users. Fix: broaden the except to `except (ImportError, Exception)` or specifically catch `RuntimeError`.

https://github.com/springfall2008/batpred/blob/4cecb2e596db58a72b0d016ef434f2d2fd6b1714/apps/predbat/gateway.py#L20-L26

https://github.com/springfall2008/batpred/blob/4cecb2e596db58a72b0d016ef434f2d2fd6b1714/apps/predbat/components.py#L36-L42

3. **Token refresh fires every 60 seconds when JWT parse fails** -- When `extract_jwt_expiry()` returns 0 (malformed token), `mqtt_token_expires_at` stays 0 (falsy). The early-return condition `if self.mqtt_token_expires_at and not self.token_needs_refresh(...)` short-circuits, skipping the return. The method falls through to the HTTP refresh call on every 60-second housekeeping cycle indefinitely. Fix: track whether extraction was already attempted, or set a sentinel value on parse failure.

https://github.com/springfall2008/batpred/blob/4cecb2e596db58a72b0d016ef434f2d2fd6b1714/apps/predbat/gateway.py#L718-L723

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

mgazza:
### Bug: test asserts raw Wh instead of converted kWh for battery_capacity

`_inject_inverter_entities` correctly converts `capacity_wh` to kWh:
```python
self.set_state_wrapper(f"sensor.{pfx}_battery_capacity", round(bat.capacity_wh / 1000.0, 2))
```

But the test asserts the unconverted value:
```python
assert entities["predbat_gateway_battery_capacity"] == 9500  # should be 9.5
```

This test will fail. Fix: `assert entities["predbat_gateway_battery_capacity"] == 9.5`

mgazza:
### Code review

Found 6 issues:

1. **Empty plan early-return prevents gateway schedule clearing.** `_on_plan_executed` returns early when both `charge_windows` and `export_windows` are empty, but never publishes an empty `ExecutionPlan` to clear the retained MQTT schedule. The gateway keeps executing the last published plan even when PredBat decides no action is needed.

https://github.com/springfall2008/batpred/blob/1948a9ff1bfeee9e5e41ea10c37ce181c54a2d6e/apps/predbat/gateway.py#L286-L289

2. **Proto3 `HasField()` on singular message fields will raise `ValueError` at runtime.** `inv.HasField("battery")`, `inv.HasField("schedule")`, and `inv.HasField("energy")` are called on regular embedded message fields in a `proto3` schema. In proto3, `HasField()` is only valid for `optional` fields or `oneof` members. These calls will crash `_inject_inverter_entities()` on every telemetry receipt.

https://github.com/springfall2008/batpred/blob/1948a9ff1bfeee9e5e41ea10c37ce181c54a2d6e/apps/predbat/gateway.py#L614-L645

3. **`depth_of_discharge_pct` does not exist in `BatteryStatus` proto.** `inv.battery.depth_of_discharge_pct` is referenced but the field is not defined in `gateway_status.proto`. The proto defines: `soc_percent`, `voltage_v`, `current_a`, `power_w`, `temperature_c`, `soh_percent`, `cycle_count`, `capacity_wh`. Accessing a non-existent attribute on a proto3 generated message raises `AttributeError`.

https://github.com/springfall2008/batpred/blob/1948a9ff1bfeee9e5e41ea10c37ce181c54a2d6e/apps/predbat/gateway.py#L637-L640

4. **Charge/discharge rate conversion uses `* 1000` instead of `* MINUTE_WATT`.** `battery_rate_max_charge` is stored as kWh/min (divided by `MINUTE_WATT = 60000`). Converting back to watts requires `* 60000`, not `* 1000`. The current code produces rates 60x too small (e.g. 3kW battery reports 50W to the gateway).

https://github.com/springfall2008/batpred/blob/1948a9ff1bfeee9e5e41ea10c37ce181c54a2d6e/apps/predbat/predbat.py#L842-L845

5. **`test_gateway.py` not registered in CI test runner.** Tests are pytest-style but `unit_test.py` TEST_REGISTRY is not updated. The repo runs tests via `coverage/run_all` -> `unit_test.py`, not pytest. These tests will not execute in CI. Same issue was flagged and required as a fix for PR #3635.

https://github.com/springfall2008/batpred/blob/1948a9ff1bfeee9e5e41ea10c37ce181c54a2d6e/apps/predbat/tests/test_gateway.py#L1-L10

6. **`update_success_timestamp()` only called on telemetry receipt.** If the ESP32 device goes offline for 60+ minutes, `Components.is_alive()` will mark the component dead even though the MQTT connection is healthy and `run()` returns True. Should also update on successful connection/housekeeping, not just telemetry. Same pattern was required for PR #3635.

https://github.com/springfall2008/batpred/blob/1948a9ff1bfeee9e5e41ea10c37ce181c54a2d6e/apps/predbat/gateway.py#L460-L465

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #3562, opened 2026-03-14
Goodwe Added support for manual charging/discharging settings 🔌

icysun:
此PR质量不达标，已关闭。需要重新修复。


springfall2008:
I assume this is just an example, its not integrated into Predbat

I'm not sure how I could integrate/test it without access to one of these systems

PR #3580, opened 2026-03-16
Feat  add battery rate max charge dc for hybrid inverter dc solar charging

F17HEH:
### Problem

On hybrid inverters (e.g. Sigenergy SigenStor) the solar panels connect
directly to the DC bus. During ECO mode the battery charges from solar
without the power passing through the AC stage, meaning it is not subject
to `inverter_limit` (the AC throughput cap).

Predbat was applying `inverter_limit` to all ECO mode battery charging
regardless of inverter type. This caused two compounding inaccuracies:

1. **Charge rate** — the model underestimated how fast the battery fills
   from solar on bright days, because it capped DC charging at the AC
   inverter ceiling rather than the battery's true DC charge rate.

2. **Round-trip loss** — the model applied `battery_loss` (AC charging
   efficiency, typically 4-5%) to DC solar charging, which bypasses AC
   conversion and is more efficient (~2%).

The knock-on effect of both is that the optimiser sees more predicted
solar clipping than will actually occur and compensates by planning a
higher overnight grid charge target — costing money unnecessarily on
Agile-style tariffs.

Additionally, `has_reserve_soc` was incorrectly set to `False` for
hybrid inverters, preventing Predbat from using the reserve SoC control.

---

#### `apps/predbat/fetch.py`

- Read `battery_rate_max_charge_dc` (watts) from `apps.yaml` and convert
  to kWh/min. Added `else` clause so the attribute is always set (`None`
  when unconfigured), preventing `AttributeError` at runtime.
- Read `battery_loss_dc` as a loss fraction, defaulting to `battery_loss`
  if not configured so existing installs are completely unaffected.

#### `apps/predbat/prediction.py`

- In ECO mode when `inverter_hybrid` is `True` and `pv_now > 0`, compute
  a fresh charge curve using `battery_rate_max_charge_dc` as the ceiling
  instead of `charge_rate_now_curve` capped by `inverter_limit`. This
  correctly models the DC bus charging path.
- For the SoC update step, detect DC solar charging (`battery_state == "e+"`
  on a hybrid with solar) and apply `battery_loss_dc` instead of
  `battery_loss`. This gives an accurate model of the higher DC charging
  efficiency.
- The `__init__` copy block already copies both new attributes from `base`
  into the prediction instance so threaded runs have access to them.

#### `apps/predbat/config.py`

- Set `has_reserve_soc` to `True` for hybrid inverters. This was
  incorrectly `False`, preventing Predbat from using the reserve SoC
  control on systems that support it.

#### `docs/apps-yaml.md`

- Added `### battery_rate_max_charge_dc` section under Inverter control
  configurations, documenting the setting, when to use it, how to
  calculate the value, and the companion `battery_loss_dc` setting.

---

### No behaviour change when

- `inverter_hybrid` is `False` (AC-coupled systems)
- `battery_rate_max_charge_dc` is not set in `apps.yaml` (falls back to
  `battery_rate_max_charge`)
- `battery_loss_dc` is not set in `apps.yaml` (falls back to
  `battery_loss`)

---

### Example configuration (Sigenergy, 2 × SigenStor BAT 10.0)
```yaml
# Each BAT 10.0 is rated at 4600W DC — two units = 9200W
battery_rate_max_charge_dc:
  - 9200

# DC solar charging bypasses AC conversion (~98% efficient vs ~95% for AC)
battery_loss_dc: 0.02
```

---

Tested on a Sigenergy system (6kW hybrid inverter, 18kWh battery, 7.2kW
panels) on Octopus Agile. Previously the plan showed the battery charging
at the AC-limited rate during solar peak and planned higher overnight grid
charges than were needed. After this change the predicted charge rate
correctly reflects DC bus capacity and overnight charge targets are reduced
on days with good solar forecast.

No regressions observed on overcast days or overnight grid charging windows.

### DIFF
--- a/templates/sigenergy_sigenstor.yaml
+++ b/templates/sigenergy_sigenstor.yaml
@@ -76,7 +76,7 @@ pred_bat:
 
   # If you have a hybrid inverter with DC-coupled batteries, solar can charge the battery
   # faster than the AC inverter limit allows. Set this to the max DC charge rate in watts.
-  # Defaults to battery_rate_max if not set.
+  # If not set, Predbat uses the normal AC charge limit (battery_rate_max_charge) and falls back to the existing ECO behavior capped by inverter_limit.
   battery_rate_max_charge_dc:
     - 9200 # 2 x BAT 10
   battery_loss_dc: 0.02


ward0:
Could this also working for Huawei inverters? Can i help testing?

F17HEH:
> Could this also working for Huawei inverters? Can i help testing?

Yes should work on any DC coupled invertor.

F17HEH:
@springfall2008 this is all ready for review again.

F17HEH:
@springfall2008 - lets hope this is the last one :)

springfall2008:
Sorry just started to look at this.

I can see why it makes sense to add battery_rate_max_dc if its higher than the ac limit.
But why add a new loss value, we have battery charge loss setting already?

apps/predbat/prediction.py
```diff
@@ -899,7 +903,15 @@ def run_prediction(self, charge_limit, charge_window, export_window, export_limi
                     battery_draw = min(diff, discharge_rate_now_curve_step, inverter_limit, battery_to_min)
                     battery_state = "e-"
                 else:
-                    battery_draw = max(diff, -charge_rate_now_curve_step, -inverter_limit, -battery_to_max)
+                    if inverter_hybrid and pv_now > 0 and battery_rate_max_charge_dc is not None:
```
springfall2008, line 906: Why if pv_now > 0, it could be 0.00001 and trigger this condition?

springfall2008:
Looking at this further there are all sorts of issues with this change, for example what happens if you are force battery charging and you have Solar and the solar is therefore doing DC charging while some part is AC from the grid?

Currently there is no battery loss dc as we use battery loss for that and inverter loss is ignored for the DC portion.

But also you change ignores the current charge rate setting in the inverter if there is any solar in eco mode, which seems odd.

I think this needs a rethink.

Maybe we can just not have the DC part of the charge count to the inverter limit, would that work?

F17HEH:
Ill have a review next week and see what i can come up with.

springfall2008:
I've written my own proposed, simpler, version of this - please review: Fix bug where inverter limit was wrongly applied to DC charging in EC

springfall2008:
This is now released in .8, so we can close this PR. Feel free to raise a new ticket or PR if you have issues with the new code.

PR #3605, opened 2026-03-19
Feat: Info tab on Web Console Log view

gcoan:
New feature:

- Info tab on Web Console Log view to only show Info messages
- Battery charge/discharge curve log messages changed to Info messages


gcoan:
Sorry Trefor I had committed this but it got missed out of the PR merge you just did

PR #3637, opened 2026-03-23
Add sample recorder filter configuration to FAQ

### LINKED ISSUE #3615: docs: FAQ states predbat entity attributes aren't stored in the database, but they are
**Describe the bug**
https://springfall2008.github.io/batpred/faq/#predbat-is-causing-exceed-maximum-size-warning-messages-in-the-home-assistant-core-log says "the Predbat entity attributes aren't stored in the database anyway", but they are, and they take up a lot of space (750MB locally).

**Predbat version**
8.34.7

**Environment details**
HAOS

**Screenshots**
Image

**Comments**
Should the documentation be updated or should the attributes actually not be stored? The warning says that any attributes over 16k aren't stored, so maybe they should be restricted before being sent to HA rather than lost.

daviessm:
Add a sample recorder filter configuration which stops HA saving the history of Predbat's large attributes in the recorder database.

With this configuration, only a couple of entities still have attributes that are too large - leave the sample logger filter there for those.

Fixes #3615 

Draft - I want to test this for a bit longer before merge.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -202,60 +202,61 @@ or to encourage discharging just before import rates fall overnight.
 
 ## Predbat is causing 'exceed maximum size' warning messages in the Home Assistant Core log
 
-If you have a large **input_number.predbat_forecast_plan_hours** then you may see warning messages in the Home Assistant Core log about the size of a number of Predbat entities,
-the message will be "State attributes for predbat.*XXXX* exceed maximum size of 16384 bytes".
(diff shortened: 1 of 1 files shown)


springfall2008:
I'm not 100% sure on this, do all the charts in the web UI work with these exclusions?

daviessm:
I'm going to leave it running for a week and check the charts after, but it looks all right to my untrained eye so far.

daviessm:
See charts below, it seems to work for me:
image
image
image
image
image
image
image
image
image
image

PR #3687, opened 2026-03-29
Update energydataservice.py

Scholdan:
fix(predbat): resolve DST-related minute offset bug in Energi Data Service rates


springfall2008:
Fixes put onto a new PR

PR #3689, opened 2026-03-29
fix(utils): compute midnight_utc with correct DST offset

bpinto:
**Disclaimer:** I'm no python programmer so I used AI, I asked it to create a function for generating a `local_midnight` function that would return a "midnight" with the current time TZ.

I haven't read it carefully enough to think about alternative solutions, but maybe there is a better solution if we remove `midnight_utc`.

P.S.: I haven't been able to test this code yet since I'm using the docker add-on and it mingles its own files with Predbat's, so I'm finding an alternative solution to test this. I've opened this PR in case it's useful, but please close it, if it's not the correct fix.

---

On DST transition days, datetime.replace(hour=0) preserves the current UTC offset instead of the offset in effect at midnight. For example on UK spring-forward day, now at 14:00+01:00 produces midnight as 00:00+01:00 (actually 23:00 UTC previous day) instead of the correct 00:00+00:00. This shifts every rate minute index by 60 minutes, breaking charge/discharge schedules.

Add a local_midnight() helper in utils.py that strips tzinfo and re-localises via pytz.localize(), which resolves the correct offset for the given date/time. For fixed-offset timezones (no DST) the simple replace() path is preserved.

Replace all broken .replace(hour=0) patterns across predbat.py, gecloud.py, fox.py, octopus.py, solis.py, predheat.py, and the test_execute.py mock with calls to local_midnight().

Add a DST-specific test in test_fetch_octopus_rates.py verifying that rates with mixed +00:00/+01:00 offsets land at the correct minute indices.

### DIFF
--- a/apps/predbat/fox.py
+++ b/apps/predbat/fox.py
@@ -25,6 +25,7 @@ import argparse
 import random
 from component_base import ComponentBase
 from oauth_mixin import OAuthMixin
+from utils import local_midnight
 
 # Define TIME_FORMAT_HA locally to avoid dependency issues
 TIME_FORMAT_HA = "%Y-%m-%dT%H:%M:%S%z"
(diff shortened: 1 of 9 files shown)


bpinto:
I'll be back home in a few hours to read the comments and review this more carefully. I should then be able to test these changes locally first.

bpinto:
@springfall2008 I still couldn't test this, hopefully if you like my proposal https://github.com/springfall2008/batpred/pull/3703 I should be able to test these changes. Unfortunately I didn't have the time yesterday to debug this as it would have been better to test this during the DST changing day.

P.S.: I haven't changed the variable naming but the `_utc` suffix isn't accurate since those are using the local timezone which may or may not be UTC timezone. I could change on this PR if you are okay, I didn't change it because that would increase the diff.

PR #3723, opened 2026-04-02
Fix DC charge rate calculation for export freeze

### LINKED ISSUE #3657: Manual export freeze slots not displayed correctly at 100% SoC
**Describe the bug**
It appears to only be at 100% SoC, but manually overriding a slot to 'force export freeze' does not display correctly in the plan. Something kind of changes in the inbuilt predbat UI, but nothing in the 'original' card.

I did try at a slot where soc was not 100% and it works correctly 

**Expected behaviour**
Confirmation of the override to be shown in the plan when soc is at 100%

**Predbat version**

V8.34.11

**Environment details**

- Giv AC3 and 2x 5.2 batteries
- Standard HAOS installer

**Screenshots**

Image
Image
Image

**Log file**

[predbat.log](https://github.com/user-attachments/files/26249023/predbat.log)

You can download the logfile from the WebUI on the Dash tab select **predbat.log**

**Predbat debug yaml file**

[predbat_debug-1.bin.txt](https://github.com/user-attachments/files/26249036/predbat_debug-1.bin.txt)

daviessm:
DC charge rate calculation for hybrid inverters was always using the max battery rate even when the slot was set to freeze export. Calculate the charge rate using the minimum battery rate if the slot should be export freezing, the same as AC charge rate.

Fixes #3657

### DIFF
--- a/apps/predbat/prediction.py
+++ b/apps/predbat/prediction.py
@@ -928,8 +928,13 @@ class Prediction:
                 else:
                     # Battery draw is only subject to inverter limit for the AC part
                     if inverter_hybrid:
+                        charge_rate_now_dc = battery_rate_max_charge_dc
+                        # Freeze mode
+                        if set_export_freeze and export_window_active and export_limit_now < 100.0 and (export_limit_now == 99.0 or set_export_freeze_only):
+                            charge_rate_now_dc = battery_rate_min  # 0
+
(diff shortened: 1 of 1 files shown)


WyndStryke:
Tested the change locally and it seems to be good (Freeze export works as expected during the planning phase when the hybrid flag = true).

PR #3739, opened 2026-04-04
Gateway fixes for soc chart

springfall2008:


apps/predbat/gateway.py
```diff
@@ -815,25 +813,18 @@ async def _publish_predbat_data(self):
...
-                            # Single-slot block — extend into next block
-                            current_block_state = state
-                            block_state_name = state
-                            block_soc.append(int(float(row.get("soc_percent", 0) or 0)))
-                        else:
-                            block_done = True
+                    if slot_idx == 0:
+                        block_state_name = state
+
+                    if slot_idx < 24:
```
bpinto, line 819: Is the usage of `24` here because a day typically has 24 hours? What about DST, wouldn't that cause issues on days that have more than 24 hours?

springfall2008: It's 24 x 30 minute periods which is only 12 hours?

bpinto: Ah my bad, so it isn't related to hours but intervals.

PR #3788, opened 2026-04-14
[Draft] Feature: Circuit Breaker for SolisCloud API Throttling

CraigCallender:
# Feature: Circuit Breaker for SolisCloud API Throttling

## Note

I'm currently testing this locally, which is why it's a draft PR. Just posting here to get any thoughts on it before I submit, but feel free to ignore it as well until I finalise it.

## Problem

The SolisCloud API implements rate limiting in a non-standard way: instead of returning HTTP 429 responses, it returns HTTP 200 with an error code in the JSON payload (e.g. `B0600: "Datalogger returns data abnormally. Please try again later"`). Predbat detects this error and logs it, but the existing `_with_retry` logic treated it like any transient failure and immediately retried — and then continued executing the next API call in the same cycle.

The problem compounds because operations like `write_time_windows_if_changed()` in V2 mode can require up to ~40+ API calls per cycle (read → write → verify for each changed register across 6 slots). When the first call hits the rate limit, the retry loop exhausts its 30-second window retrying that single call, then the cascade continues — each subsequent call also fails and retries, rapidly exhausting the API token budget. By the time the next 60-second cycle begins, the quota is still depleted, so those calls fail too, leading to a self-reinforcing failure loop that cannot recover without manually stopping Predbat and waiting ~5 minutes.

### `solis.py` — Circuit breaker pattern

A three-state circuit breaker (CLOSED → OPEN → HALF_OPEN) has been added to the `SolisAPI` class, gating all outgoing API calls at `_execute_request`.

**New constants:**

- `SOLIS_CB_FAILURE_THRESHOLD = 3` — consecutive non-throttle failures before tripping
- `SOLIS_CB_INITIAL_RECOVERY_TIME = 60` — seconds before first retry after trip
- `SOLIS_CB_MAX_RECOVERY_TIME = 900` — 15-minute cap on recovery wait
- `SOLIS_CB_BACKOFF_MULTIPLIER = 2` — recovery time doubles on each successive trip
- `SOLIS_CB_THROTTLE_CODES = {"B0600", "B0173", "B0115"}` — codes that trip the breaker immediately

**State machine:**

- **CLOSED** (normal): all API calls pass through; non-throttle failures increment a counter; the breaker trips when the counter reaches the threshold.
- **OPEN** (tripped): all API calls are blocked immediately with `SolisAPIError("Circuit breaker OPEN")` — no HTTP request is made. Recovery time uses exponential backoff: 60s → 120s → 240s → 480s → 900s (capped), resetting to 60s on full recovery.
- **HALF_OPEN** (probing): after the recovery window elapses, one probe request is allowed through. Success → CLOSED (full reset). Failure → OPEN again with doubled recovery time.

**Known throttle codes trip the breaker immediately** (on the first failure, not after the threshold), preventing any cascade.

**`_with_retry` bails immediately** when the circuit is not CLOSED, so the 30-second retry window is not wasted after a trip — critical for preventing the multi-call cascade in `write_time_windows_if_changed`.

**`run()` short-circuits** when the circuit is OPEN: logs one clean "Circuit breaker OPEN, skipping API calls this cycle (Xs remaining)" message per cycle instead of per-call failures flooding the log, and still calls `publish_entities()` so HA entities stay alive with stale-but-valid cached data.

**`record_api_call`** uses the `"rate_limit"` failure reason (already a valid metric label) for throttle-code failures, enabling Prometheus dashboards to distinguish throttling from other errors.

### `tests/test_solis.py` — 10 new circuit breaker tests

- `test_cb_closed_to_open_on_threshold` — CLOSED → OPEN after `SOLIS_CB_FAILURE_THRESHOLD` consecutive non-throttle failures
- `test_cb_immediate_trip_on_throttle_code` — CLOSED → OPEN immediately on B0600/B0173/B0115, no threshold required
- `test_cb_open_blocks_calls` — OPEN state raises `SolisAPIError` without making any HTTP request
- `test_cb_open_to_half_open_after_recovery` — OPEN → HALF_OPEN after recovery time elapses
- `test_cb_half_open_to_closed_on_success` — HALF_OPEN → CLOSED on successful probe; trip count and recovery time fully reset
- `test_cb_half_open_retrips_on_failure` — HALF_OPEN → OPEN on failed probe; recovery time doubles
- `test_cb_backoff_sequence_and_cap` — validates 60s → 120s → 240s → 480s → 900s sequence and cap
- `test_cb_full_recovery_resets_trip_count` — confirms trip count and recovery time reset to initial values after full recovery
- `test_cb_with_retry_bails_on_open` — `_with_retry` makes exactly 1 call (no retries) when circuit is OPEN
- `test_cb_run_skips_when_open` — `run()` returns `False` and calls `publish_entities()` but makes no API calls when circuit is OPEN

### DIFF
--- a/.cspell/custom-dictionary-workspace.txt
+++ b/.cspell/custom-dictionary-workspace.txt
@@ -299,6 +299,7 @@ remotecontrol
 resetmidnight
 resultid
 resultmid
+retrips
 rname
 Roboto
 rowspan
--- a/apps/predbat/solis.py
+++ b/apps/predbat/solis.py
@@ -32,6 +32,13 @@ SOLIS_MAX_RETRY_TIME = 30  # seconds
 SOLIS_INITIAL_RETRY_DELAY = 1  # seconds
 SOLIS_REQUEST_TIMEOUT = 30  # seconds
 
+# Circuit breaker configuration
+SOLIS_CB_FAILURE_THRESHOLD = 3          # consecutive non-throttle failures before tripping
+SOLIS_CB_INITIAL_RECOVERY_TIME = 60     # seconds before first retry after trip
+SOLIS_CB_MAX_RECOVERY_TIME = 900        # 15-minute cap on recovery wait
+SOLIS_CB_BACKOFF_MULTIPLIER = 2         # double recovery time on each successive trip
+SOLIS_CB_THROTTLE_CODES = {"B0600", "B0173", "B0115"}  # API response codes that trip the breaker immediately
+
 # CID Constants (Control IDs for inverter registers)
 SOLIS_CID_STORAGE_MODE = 636
 SOLIS_CID_BATTERY_RESERVE_SOC = 157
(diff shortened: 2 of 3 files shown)


springfall2008:
I suppose the question is why fail the API requests rather than wait, is it because you have to wait too long for their rate throttling to clear?

PR #3885, opened 2026-05-07
Add dynamic additional load forecasts

### LINKED ISSUE #2917: improved method to send external loads as input to predbat via automations
Predbat works best when it has good forecast of loads, solar, tariffs. I see a gap in the current methods available to inform it of loads. This feature request is to allow an enhancement to the current API and / or manual adjustment methods available. See [this discussion](https://github.com/springfall2008/batpred/discussions/2894) of one example.

Ideally it would allow:

- multiple inputs that are labelled and managed by the label - e.g. heating load, DHW load, cooking dinner load etc
- start and end-times (or duration) of the load
- load delta
- optionally load profile e.g. at start it is higher, at steady state lower

The linked discussion shows the options I have considered - i.e. use of the API or the manual select. They both work with limitations - e.g. the select method is "all or nothing" - you can set the load delta once or turn it off, nothing in between and no ability to manage multiple inputs.

**Additional context**
I don't see this as replacement of the existing methods - i.e. the load forecast built into predbat or the use of predAI. Both these mechanisms are effective and simple to use for forecasting loads based on history. I suspect they will cover most of the future load with sufficient accuracy for predbat to do it's thing. This would allow users to complement the load forecast with external data. In my example heating demand varies daily based on external temp etc, so a historical-based forecast is less meaningful.

Scholdan:
## Summary

This PR adds named additional house load forecasts so Predbat can include known future appliance/load events in the forward plan. The main use case is loads that are not well represented by historical usage, for example a dishwasher, cooking, hot water, or heating demand.

Fixes #2917.

It supports both static `apps.yaml` configuration and dynamic one-shot requests from Home Assistant automations.

### Static YAML Load Forecasts

Adds `house_load_additional_forecast` entries in `apps.yaml`, keyed by `name`.

Supported fields include:

- `name`
- `enabled`
- `mode`: `fixed` or `flexible`
- `start_time`
- `end_time`
- `duration`
- `energy`: total kWh across the whole load
- `slot_energy`: advanced kWh per Predbat plan slot
- `weighting`: optional profile shaping across the duration

Static YAML entries are intended for recurring or known configuration-driven load injection. They do not get delete buttons and are edited or removed through `apps.yaml`.

### Dynamic Home Assistant API Forecasts

Adds/uses `select.predbat_load_forecast_delta_api` for runtime one-shot forecasts from Home Assistant automations.

Example:

```yaml
action: select.select_option
target:
  entity_id: select.predbat_load_forecast_delta_api
data:
  option: "dishwasher?enabled=true&mode=flexible&end_time=07:00&duration=5&energy=0.7"
```

These dynamic forecasts are one-shot requests:

- Published as per-load binary sensors such as `binary_sensor.predbat_load_forecast_delta_dishwasher`
- Given delete buttons such as `button.predbat_load_forecast_delta_dishwasher_delete`
- Automatically expired after their scheduled finish time
- Can be cancelled early with the delete button
- Can be recreated by sending the select command again

A service path is also supported via `predbat.update_load_forecast_delta` for integrations that prefer a service-style update.

### Published HA Attributes

Each named load forecast publishes useful attributes including:

- `enabled`
- `mode`
- `energy`
- `slot_energy`
- `duration`
- `load_mode`
- `plan_interval_minutes`
- `slots`
- `total_energy`
- `requested_start`
- `requested_end`
- `suggested_start`
- `suggested_end`
- `selection_reason`
- `selection_locked`
- `candidate_count`
- `source`
- `auto_expire`
- `expires_at`
- `target_times`

These attributes make the selected plan inspectable in Home Assistant and usable from automations.

## Flexible Load Scheduling

Adds `mode=flexible` for appliance loads that can run anytime within a deadline.

Flexible semantics:

- `start_time` means earliest allowed start
- `end_time` means the load must be done by this time
- Omitted `end_time` means the remaining forecast horizon
- Omitted `start_time` differs by source:
  - YAML/static entries roll with the current plan slot on each refresh
  - API/service one-shot requests freeze the current plan slot when the request is received

Predbat chooses the flexible start time using the full Predbat optimiser metric, not just raw import rate or raw prediction cost. This means selection considers the current optimised plan, battery value, PV10 weighting, import/export prices, losses, cycle/keep penalties, self-sufficiency settings, and other predicted load.

## One-Shot Lifecycle

Dynamic flexible forecasts follow the practical appliance lifecycle:

```text
pending request
→ suggested_start/suggested_end published as a movable suggestion
→ reselected on later replans until suggested_start is reached
→ locked once now reaches suggested_start
→ only remaining future target_times are published while running
→ archived to additional-load history at completion/expiry
→ expired/deleted once now reaches suggested_end
```

Before `suggested_start`, the selected window is suggestion-only. Predbat publishes `suggested_start` and `suggested_end`, but does not publish committed `target_times`, `slots`, or `total_energy` for the load sensor yet. This keeps the request movable and allows later planning runs to choose a better window if the wider prediction metric changes.

Once `suggested_start` has passed, the load is treated as committed/running because a real appliance such as a dishwasher cannot normally be moved after it has started. At that point `selection_locked` becomes true, future replans no longer move it, and the sensor publishes only the remaining future `target_times` until auto-expiry at `suggested_end`.

Completed/expired dynamic additional loads are archived only at the end of the scheduled slot, not when merely suggested or started. This avoids subtracting loads from historical learning before Predbat knows the scheduled slot has completed.

## Additional Load History

Adds Predbat-owned persistence for completed additional load slots:

- Sensor: `sensor.predbat_load_forecast_delta_history`
- State: number of stored records
- Attribute: `records`
- Restored on startup/restart
- Pruned to the configured previous-days horizon and capped to avoid unbounded growth

Archived completed loads are subtracted from future historical load filtering so one-shot appliance forecasts do not permanently inflate learned household demand.

## Optimisation Behaviour

This PR also fixes flexible-load planning details discovered during testing:

- Flexible candidates now conserve configured energy exactly once across Predbat prediction buckets.
- Flexible selection is performed after the first real optimisation pass so it evaluates against the real optimised charge/export plan, not provisional windows.
- Predbat runs at most one extra optimisation pass after flexible-load selection, and only when the unlocked selected start changed.
- Locked/running flexible loads are excluded from reselection and cannot be moved once started.

## Requested Start Drift Fix

This PR fixes an issue found during testing where dynamic API loads with omitted `start_time` had `requested_start` drift forward on each refresh.

The fix stamps `_requested_start_minutes` when a one-shot API/service request is first parsed or created. This keeps `requested_start` stable across replans while still allowing a fresh start time when the same command is sent again.

A follow-up edge case is also handled: if stale selected flexible metadata exists before the frozen requested start, the selected start is clamped to `requested_start`, so `suggested_start` cannot be earlier than the published requested window.

## Textual Plan

The textual plan now includes confirmed additional loads once they have target slots, for example:

```text
- Additional load dishwasher from 10:00 to 12:00 using 1.20 kWh is planned.
```

Pending flexible loads are not included until Predbat has selected an actual window.

## Documentation

Docs were added/updated in:

- `docs/apps-yaml.md`
- `docs/manual-api.md`

The docs cover:

- Static `house_load_additional_forecast` configuration
- `energy` vs `slot_energy`
- Fixed vs flexible load behaviour
- API one-shot behaviour
- Delete buttons and expiry
- Frozen API `start_time` behaviour vs rolling YAML behaviour
- A generic dishwasher scheduling example with helper, request automation, start-at-`suggested_start` automation, and manual-start cleanup automation

## Validation

Latest local validation run successfully:

```text
./run_all --test additional_load_forecast
./run_all --quick
./venv/bin/python -m py_compile ../apps/predbat/plan.py ../apps/predbat/tests/test_additional_load_forecast.py
git diff --check
```

The targeted test coverage includes:

- Disabled and enabled load forecasts
- Fixed loads using `start_time`/`duration`
- Fixed loads using `start_time`/`end_time`
- Total `energy` distribution
- Advanced `slot_energy`
- Weighting behaviour
- Select API updates
- Service updates
- Delete button cleanup
- Auto-expiry
- Stale entity cleanup
- Additional-load history archive, restore, pruning, dedupe, and historical-load filtering
- Flexible pending state before plan selection
- Flexible full-metric selection rather than raw cost selection
- Flexible candidate energy conservation
- Flexible API suggestion-only state before `suggested_start`
- Flexible API reselection before `suggested_start`
- Flexible API lock after `suggested_start`
- Selector-level protection that locked/running flexible loads are not reselected
- Expiry after locked flexible load reaches `suggested_end`
- API omitted `start_time` freeze
- YAML omitted `start_time` rolling behaviour
- Stale selected metadata not starting before frozen `requested_start`
- Textual plan output for confirmed additional loads


Scholdan:
I found two edge cases while reviewing this and pushed fixes in `c851bf93`.

1. Partial-duration loads were not fully represented in the prediction data when the duration did not line up exactly with the plan interval. The attributes could show the configured total kWh, but the planner could see less energy. The fix keeps the published slot energy as-is and scales the per-minute adjustment for partial slots.

2. Forecast names that get sanitized for Home Assistant entity IDs, for example `Dishwasher Eco` becoming `dishwasher_eco`, could fail to map back to the original forecast name for service updates or delete buttons. The fix resolves entity suffixes against the active forecasts, overrides, and API commands before updating or deleting.

I added regression coverage for both cases and ran:

- `coverage/./run_all --test additional_load_forecast`
- `coverage/./run_pre_commit`

Scholdan (commented):
I re-checked the PR after `c851bf93`.

The two issues I found earlier are now fixed and covered by tests:

- Partial-duration loads now keep the configured total energy in the planner.
- Sanitized Home Assistant entity names now map back to the original forecast names for service updates and delete buttons.

I did not find any new blocking issues in the updated diff.

Checks looked good:

- `coverage/./run_all --test additional_load_forecast`
- `coverage/./run_pre_commit`

Scholdan:
I simplified the dynamic update side of this PR.

Changes:
- Kept the dynamic API select-backed only via `select.predbat_load_forecast_delta_api`, which matches Predbat's existing manual API pattern and preserves restart persistence.
- Removed the separate `predbat.update_load_forecast_delta` service path because it duplicated the API and did not use the same persistence mechanism.
- Centralized additional-load forecast record construction, timestamp formatting, and API command parsing in `fetch.py` to reduce duplicated logic without changing behavior.

I kept the remaining behavior intact: static YAML forecasts, dynamic one-shot forecasts, flexible scheduling lifecycle, restart metadata, `slot_energy`, `weighting`, delete buttons, and auto-expiry.

Validation:
- `coverage/./run_all --test additional_load_forecast`
- `coverage/./run_pre_commit`

Scholdan:
@springfall2008 I see the planner's moving to C++. Does that change how this fits? Happy to rebase the flexible-load logic onto the new architecture if you point me at it — or if the Python planning path is frozen/being replaced, tell me and I'll hold off reworking it.

apps/predbat/additional_load.py
```diff
@@ -0,0 +1,1273 @@
...
+
+    def parse_additional_load_api_command(self, api_command):
+        """
+        Parse one load_forecast_delta_api command into a forecast override.
+        """
+        if "?" not in api_command:
+            self.log("Warn: Bad load_forecast_delta_api command {}, expected name?start_time=...&duration=...".format(api_command))
+            return None
+
+        name, command_args = self.additional_load_command_args(api_command)
```
springfall2008, line 516: **Overnight/wrap-around flexible windows can collapse to "start now", defeating the configured window.**

Hand-traced this with `start_time=22:00`, `end_time=06:00`, `duration=2h`, queried at `minutes_now_slot=08:30` (510): the sorted `windows` list's first candidate is `(-120, 360)` (yesterday's occurrence). `usable_start = max(-120, 510) = 510`; since `510 + 120 > 360`, `window_end` is bumped by `+24*60` to `1800` **in place**, and the very next check `usable_start + duration_minutes <= window_end` now passes, so the function returns `(510, 1800)` — i.e. "may start right now at 08:30", even though the configured window is 22:00–06:00.

The bug is that extending `window_end` by a day doesn't re-validate that `usable_start` actually falls within a real occurrence of the window — it only checks the duration fits before the (now arbitrarily extended) end. A load with an overnight-only deadline can end up scheduled in broad daylight.

Suggest: when `usable_start + duration_minutes > window_end`, advance to the *next* window's start rather than mutating `window_end` and re-testing the same `usable_start`.

apps/predbat/plan.py
```diff
@@ -1663,6 +1609,32 @@ def calculate_plan(self, recompute=True, debug_mode=False, publish=True):
             self.plan_last_updated = self.now_utc
             self.plan_last_updated_minutes = self.minutes_now
 
+        additional_load_adjust_before = self.house_load_additional_forecast_adjust.copy()
+        flexible_selected, load_minutes_step, load_minutes_step10 = self.select_flexible_additional_loads(load_minutes_step, load_minutes_step10, pv_forecast_minute_step, pv_forecast_minute10_step)
```
springfall2008: **`select_flexible_additional_loads()` now runs unconditionally on every `calculate_plan()` call, including `recompute=False`.**

This call sits outside the `if self.calculate_best and recompute:` block above (8-space indent vs. the block's 12-space indent). `calculate_plan(recompute=False)` is the fast path the main 5-minute loop takes on most cycles once the plan is fresh, specifically to skip expensive replanning. With even one `mode: flexible` additional load configured (default search window = full 48h forecast horizon), this now runs the full candidate search — up to ~96 candidates × 2 `Prediction()` objects × 2 `run_prediction()` calls each — on every cycle regardless of whether anything changed, silently reintroducing the cost `recompute=False` exists to avoid.

apps/predbat/plan.py
```diff
@@ -1663,6 +1609,32 @@ def calculate_plan(self, recompute=True, debug_mode=False, publish=True):
...
+                for minute in self.house_load_additional_forecast_adjust.keys() | additional_load_adjust_before.keys()
+                if self.house_load_additional_forecast_adjust.get(minute, 0.0) != additional_load_adjust_before.get(minute, 0.0)
+            }
+            load_minutes_step90 = self.add_additional_load_to_step_data(load_minutes_step90, selected_load_adjust)
+            self.load_minutes_step = load_minutes_step
+            self.load_minutes_step10 = load_minutes_step10
+            self.load_minutes_step90 = load_minutes_step90
+            self.prediction = Prediction(self, pv_forecast_minute_step, pv_forecast_minute10_step, load_minutes_step, load_minutes_step10, pv_forecast_minute90_step, load_minutes_step90)
+            self.prediction.batch_threads = resolve_batch_threads(self.get_arg("threads", "auto"), cpu_count())
+            if recompute and self.calculate_best and self.house_load_additional_flexible_selection_changed:
```
springfall2008: **Flexible-load re-optimisation bypasses the plan retention/improvement gate.**

Every other path that replaces `charge_limit_best`/`charge_window_best` goes through the "Plan comparison" block a few lines above (`plan_scoring_pair` / `should_replace_plan`, gated by `metric_min_improvement_plan`) specifically to prevent the plan flapping between near-equal options each cycle. This new branch instead calls `optimise_best_windows_once(...)` again and unconditionally overwrites `self.plan_preclip` (and, via that call, `self.charge_limit_best`/`self.charge_window_best`/`self.export_window_best`/`self.export_limits_best` in place) with no comparison against the plan just accepted moments earlier.

`house_load_additional_flexible_selection_changed` is set on a bare `existing_start != best_start` in `select_flexible_additional_loads`, with no minimum-improvement threshold of its own — so a flexible load whose best candidate start ties or nearly ties between adjacent slots cycle-to-cycle (forecast noise, floating-point) can force an unconditional full plan replacement every such cycle, defeating the anti-flapping protection right where it's most needed.

springfall2008 (commented):
Automated review (high effort, 10 finder agents across correctness/reuse/simplification/efficiency angles, cross-verified against the diff before posting). Summary of the highest-confidence findings below as inline comments. A few lower-severity items not inlined: `get_additional_load_bool` duplicates the string-to-bool coercion already in `get_arg` (userinterface.py) with a slightly different truthy set; the new `delete_state`/DELETE code path has no coverage in `TestHAInterface` (`tests/test_infra.py`), so it's untested by this PR's own suite; the `.cspell` addition of `delayedstart` doesn't appear to be used anywhere in the diff.

apps/predbat/additional_load.py
```diff
@@ -0,0 +1,1273 @@
...
+        """
+        return self.additional_load_entity_name(name).replace("binary_sensor.", "button.", 1) + "_delete"
+
+    def additional_load_name_from_entity(self, entity_id):
+        """
+        Return additional load forecast name from a binary sensor or button entity id.
+        """
+        marker = "_load_forecast_delta_"
+        if entity_id and marker in entity_id:
+            safe_name = entity_id.split(marker, 1)[1].replace("_delete", "")
```
springfall2008: **Unanchored `.replace("_delete", "")` strips the substring anywhere in the name, not just the trailing button suffix.**

`entity_id.split(marker, 1)[1].replace("_delete", "")` is a global replace. For a load whose HA-safe name itself contains `_delete` — e.g. a load named `auto_delete_timer` → entity `button.predbat_load_forecast_delta_auto_delete_timer_delete` — this strips *both* occurrences, producing `"auto_timer"` instead of `"auto_delete_timer"`. No candidate matches that mangled name, so `resolve_additional_load_name` falls through and returns the wrong value; the delete button then silently fails ("Ignoring delete for inactive additional load forecast auto_timer") for any load whose name contains "delete" as a substring.

Suggest anchoring the strip to the end: `safe_name[:-len("_delete")] if safe_name.endswith("_delete") else safe_name`.

apps/predbat/additional_load.py
```diff
@@ -0,0 +1,1273 @@
...
+
+    def additional_load_stamp_to_minutes(self, stamp):
+        """
+        Convert a durable timestamp string into forecast minutes from the current midnight.
+        """
+        try:
+            stamp_datetime = datetime.fromtimestamp(int(stamp), tz=self.midnight_utc.tzinfo) if str(stamp).isdigit() else datetime.fromisoformat(str(stamp))
+        except (ValueError, TypeError):
+            return None
+        return int((stamp_datetime - self.midnight_utc).total_seconds() / 60)
```
springfall2008: **Datetime subtraction is outside the `try/except`, so a naive ISO timestamp raises an uncaught `TypeError`.**

Only the `datetime.fromisoformat(...)` parse on the previous lines is guarded by `except (ValueError, TypeError): return None`. The following `stamp_datetime - self.midnight_utc` is not — if `stamp` is a non-digit ISO string without a timezone offset, `fromisoformat` returns a naive datetime, and subtracting it from the tz-aware `self.midnight_utc` raises `TypeError: can't subtract offset-naive and offset-aware datetimes`, uncaught. Every other malformed-input path in this file logs a `Warn:` and degrades gracefully instead. This function (and its near-duplicate `additional_load_history_record_minutes`) is reachable from `load_additional_load_history()`, called on essentially every minute-sample, so a single malformed/foreign record in persisted storage would crash repeatedly rather than being skipped.

apps/predbat/ha.py
```diff
@@ -1098,13 +1098,26 @@ def call_service(self, service, **kwargs):
             data_frame = {"domain": domain, "service": service, "service_data": data}
             return run_async(self.base.trigger_callback(data_frame))
 
-    def api_call(self, endpoint, data_in=None, post=False, core=True, silent=False):
+    def delete_state(self, entity_id):
+        """
```
springfall2008, line 1111: **`delete_state()` never cleans the SQLite-backed DB mirror, only the in-memory dicts.**

Compare `set_state`, which persists via `self.db_manager.set_state_db(...)` when `db_enable and (db_mirror_ha or db_primary)`. `delete_state` only pops `db_mirror_list`/`state_data` and, if `ha_key` is set, issues the real HA DELETE — `db_manager.py`'s IPC dispatcher has no delete/remove command at all. In `db_primary` mode (the no-HA / Predbat.com / Docker path this codebase supports), the SQLite row for a "deleted" entity survives, and the next `update_states_db()` call (fired on init and on every config change) repopulates `self.state_data` from the stale DB row — the entity silently resurrects even though the caller believes cleanup succeeded. This path also isn't exercised by any test, since `TestHAInterface` doesn't define `delete_state` at all.

apps/predbat/additional_load.py
```diff
@@ -0,0 +1,1273 @@
...
+        if end_minutes is None:
+            end_minutes = start_minutes + int(duration * 60)
+        end_minutes = int(end_minutes)
+        if end_minutes <= start_minutes:
+            end_minutes += 24 * 60
+        if duration <= 0:
+            duration = (end_minutes - start_minutes) / 60.0
+        if duration <= 0:
+            return False
+        periods = int((int(duration * 60) + plan_interval - 1) / plan_interval)
```
springfall2008: **Float round-trip in `int(duration * 60)` before ceiling division can silently drop the load's final scheduling slot.**

`periods = int((int(duration * 60) + plan_interval - 1) / plan_interval)` truncates the minutes before ceiling-dividing. E.g. `duration=4.1` (hours): `4.1*60 == 245.99999999999997` in Python, so `int(...)` gives `245` instead of `246`, and with `plan_interval_minutes=5` this yields `periods=49` instead of the correct `50` — the final 5-minute slot (and its energy) is silently dropped, with no warning logged. Masked at the default `plan_interval_minutes=30` (truncation doesn't cross a period boundary there) but reproducible at 5/10-minute intervals, which are plain user-settable values. Same pattern repeats at the other three `int(duration * 60)` call sites in this file (`get_additional_load_window`, `fetch_additional_load_forecast`, `get_additional_load_text`).

apps/predbat/additional_load.py
```diff
@@ -0,0 +1,1273 @@
...
+            if start_minutes is None or end_minutes is None or duration_minutes <= 0:
+                continue
+
+            candidate = max(start_minutes, self.minutes_now)
+            candidate = int((candidate + plan_interval - 1) / plan_interval) * plan_interval
+            latest_start = min(end_minutes - duration_minutes, self.minutes_now + flexible_max_minutes - duration_minutes)
+            if latest_start < candidate:
+                continue
+
+            baseline_prediction = Prediction(self, pv_forecast_minute_step, pv_forecast_minute10_step, working_load_step, working_load_step10)
```
springfall2008: **Missing `kernel_static_cache` reuse across the candidate search — redoes load-independent kernel setup for every candidate.**

Both `baseline_prediction = Prediction(self, ...)` here and `candidate_prediction = Prediction(self, ...)` a few lines below leave `kernel_static_cache` at its default `None`, so each call rebuilds the whole load-independent static context (rates, PV, temperature-cap curves, carbon, gas, iboost, car slots) from scratch via `create_kernel_context`. This exact scenario — many `Prediction()`s differing only in load forecast — is already solved elsewhere in this codebase: `marginal.py` builds one `kernel_static_cache = {}` and reuses it across 28 `Prediction()` calls (comment there: "computes the load-independent arrays once instead of 28 times, which is nearly all of the work here"). With the default full-horizon search this loop can build ~96 candidates per flexible load per cycle, each redoing that same setup work unnecessarily.

apps/predbat/userinterface.py
```diff
@@ -1337,6 +1367,12 @@ def api_select(self, config_item, value):
         if value.startswith("+"):
             # Ignore selections which are just the current value
             return
+        if config_item == "load_forecast_delta_api" and value != "off" and "[" not in value:
```
springfall2008: **Deleting a `load_forecast_delta_api` entry via the select's own bracketed `"[name]"` remove option leaks stale metadata.**

This block is gated by `and "[" not in value`, so when the standard bracketed removal syntax is used (the same convention `manual_api`/`manual_times` use, and what `api_select_update` itself generates for the dropdown's remove entries), the whole block — including `self.house_load_additional_forecast_overrides.pop(name, None)` — is skipped. Only the dedicated delete button (`delete_additional_load_forecast`) actually clears `house_load_additional_forecast_overrides`. If a user removes an entry via the generic `[name]` select option instead, stale `_selected_start_minutes` / `_expires_minutes` / `_selection_locked` fields are left behind; `get_additional_load_forecast_config`'s merge applies `house_load_additional_forecast_overrides` *after* a fresh API override for the same name, so a later legitimate command for that name can be silently clobbered by the leftover metadata.

apps/predbat/userinterface.py
```diff
@@ -1255,6 +1279,12 @@ def manual_select(self, config_item, value):
         if value.startswith("+"):
             # Ignore selections which are just the current value
             return
+        if config_item == "load_forecast_delta_api" and value != "off":
```
springfall2008: **This `load_forecast_delta_api` special-case block in `manual_select` is unreachable.**

`config.py`'s CONFIG_ITEMS entry for `load_forecast_delta_api` sets only `"api": True` (no `"manual"`/`"manual_rate"` key), and `select_event`'s dispatcher routes strictly on `if item.get("manual"): manual_select(...) elif item.get("api"): api_select(...)` — so this item is always routed to `api_select`, never here. The real (reachable) copy in `api_select` has a bug this one doesn't (see the `api_select` comment on the bracket-removal gap) — worth removing this dead copy so there's one metadata-preservation implementation to maintain instead of two, only one of which is ever exercised or covered by tests.

PR #3904, opened 2026-05-10
Implement thread-safe log file management

daviessm:
When debugging I was getting a lot of crashes because files couldn't be renamed due to multiple threads trying to rename files at the same time. Use a mutex to prevent this and also ignore errors caused by missing files - hacky, but better than crashing.


springfall2008:
Didn't quite work as the lock broke the execution pool PICKLE, I'll re-do it another way

PR #3915, opened 2026-05-12
cleanup: Fix battery discharge diagram to go the right direction

Kedstar99:
When my battery is charging up on a solar day, the diagram is going the wrong way? Would attach a diagram but i think github is preventing me from adding it?

Any case from the comments itself should be clear. Line 346-347 states battery is charging, battery is discharging. If it is charging why is it going from the battery to the house?

### DIFF
--- a/apps/predbat/web.py
+++ b/apps/predbat/web.py
@@ -421,7 +421,7 @@ class WebInterface(ComponentBase):
             """.format(
                 dp0(pv_power)
             )
-        if battery_charging:
+        if battery_discharging:
             # Calculate animation speed based on power flow - faster for higher power
             battery_speed = max(0.5, min(3.0, 2.0 - (abs(battery_power) / 3000)))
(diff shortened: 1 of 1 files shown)


springfall2008:
I suspect the issue is your battery sensor works the other way around, you likely need to use 'battery_power_invert: True' in apps.yaml

Kedstar99:
The code comments don't make sense for this function.... If my PR is wrong here then the comments of the function don't make sense. @springfall2008 
```
 battery_charging = battery_power >= 10  # Battery is charging (positive value)
 battery_discharging = battery_power <= -10  # Battery is discharging (negative value)
if battery_charging:
   <!-- Battery to House Arrow -->
   <!-- Moving dots for Battery to House -->
else
   <!-- House to Battery Arrow -->
   <!-- Moving dots for House to Battery -->

```

PR #3958, opened 2026-05-25
Add historical rate-mean fallback for short forward rate windows

Scanlia:
This adds an optional fallback for `rate_replicate` that uses historical rate data instead of a flat last-known rate when the rate sensor only publishes a short forward window (e.g. Amber Australia ~16h). When enabled, Predbat queries HA recorder history and builds a per-half-hour-of-day mean profile, restoring the diurnal price shape for the plan horizon.

## Changes

**`fetch.py`** — New `build_rate_history_buckets(entity_id, days, scaling)` method. Walks HA history via step-function lookup (one sample per half-hour slot per day), resistant to transient spikes. Empty buckets fall back to overall mean. Modified `rate_replicate` to accept optional `history_buckets` — history only fills slots with no 24h-back match; existing fallback chain is unchanged.

**`config.py`** — New `rate_history_days_average` input_number (0–30, default 0 to disable). Four new APPS_SCHEMA entries for source entities and scaling multipliers.

**`test_rate_history_average.py`** — 9 sub-tests: disabled, no history, unavailable states, scaling, diurnal shape, spike resistance, replication integration, 24h-back preference, backward compatibility.

**`docs/apps-yaml.md`** + **`docs/energy-rates.md`** — Documented all 5 new config options.

## Config

```yaml
  rate_history_days_average: 7        # 0 to disable (default)
  rate_history_source_import: sensor.amber_general_price
  rate_history_scaling_import: 100.0  # $/kWh → c/kWh
```

## Compatibility

Feature is off by default. No behaviour change when `rate_history_days_average` is 0.


springfall2008:
Seems like a good idea, but predbat already keeps the history of rates so we can probably just wire it directly to that predbat current rate sensor for the history?

PR #3961, opened 2026-05-25
Fix TypeError in dp1/dp2/dp3/dp4 when value is None

Scanlia:
The `dp1`/`dp2`/`dp3`/`dp4` rounding helpers in `utils.py` crash with `TypeError` when passed `None` (e.g. an inverter entity not yet initialised). Add an early `None` guard to each.

```python
def dp1(value):
    if value is None:
        return None
    return round(value, 1)
```

Same pattern for `dp2`, `dp3`, `dp4`. 1 file changed, 8 lines inserted.

### DIFF
--- a/apps/predbat/utils.py
+++ b/apps/predbat/utils.py
@@ -851,6 +851,8 @@ def dp1(value):
     """
     Round to 1 decimal place
     """
+    if value is None:
+        return None
     return round(value, 1)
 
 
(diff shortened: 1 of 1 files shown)


springfall2008:
Hi @Scanlia,

Thanks for your contribution.

While I agree your code change is correct, it is likely to have a performance impact due to the number of times these functions are called. 

I wonder if this working around a bug elsewhere?

Scanlia:
Thanks springbat! You were right I can see you have guarded it at the source now in a035037 , many thanks!

Issue #3828, opened 2026-04-24
Clipping issues due to export limit


davesbell:
I expect I'm doing something wrong or not understanding how something works. I don't think this is actually a bug, just user error.

**Describe the bug**
Predbat doesn't detect that clipping is going to happen, and keep that information

**Expected behaviour**
I've got an 8kW export limit, but a 12kW array. I would expect when I'm getting towards 8kW of generation, Predbat would notice that and limit what is in the battery to take the overflow. 

It does seem to do this for the following day, but when that day rolls around it all seems to be lost.

**Predbat version**

v8.37.3

**Environment details**

- Sigen Sigenstore 12kW invertor with 20(ish) kWh battery storage
- Standard HAOS install
- 2 solar arrays. 1 north, 1 south. Both in Solcast.

**Screenshots**

<img width="518" height="511" alt="Image" src="https://github.com/user-attachments/assets/22619e43-013f-4a14-9e5b-6f0115f0b6a6" /> This shows that generation will be just below my export limit today. I'm not sure this will actually be the case however.

**Log file**

[predbat.log](https://github.com/user-attachments/files/27043824/predbat.log)

**Predbat debug yaml file**

[predbat_debug.yaml.txt](https://github.com/user-attachments/files/27043834/predbat_debug.yaml.txt)

[predbat_plan.html](https://github.com/user-attachments/files/27043842/predbat_plan.html)

[apps.yaml.txt](https://github.com/user-attachments/files/27043851/apps.yaml.txt)

daviessm:
See also #1206

gcoan:
> Describe the bug
> Predbat doesn't detect that clipping is going to happen, and keep that information
> 
> Expected behaviour
> I've got an 8kW export limit, but a 12kW array. I would expect when I'm getting towards 8kW of generation, Predbat would notice that and limit what is in the battery to take the overflow.
> 
> It does seem to do this for the following day, but when that day rolls around it all seems to be lost.

Could you explain a bit more about what is actually happening, I don't really follow it based on the graph.

In terms of what Predbat does (or should do) assuming you have set the export limit and inverter/battery rates correctly in apps.yaml:

First predbat doesn't control export limit. It relies on the inverter to do that, but it should be aware of it and take account of it ...

If you turn the HTML plan debug on then you will see an extra clipping kWh column which is where predbat predicts there will be clipping.

[...]

Issue #3481, opened 2026-03-04
Feature request - Keep export as priority and store excess in battery if clipping occurs


maverik-77:
**Is your feature request related to a problem? Please describe.**
During Summer when people have PV production that goes higher than DNO specified limits, there is clipping.  I would like predbat to implement a new logic to store extra PV energy left after export in the battery. 

**Describe the solution you'd like**
Would Predbat be able to predict the generation from Solcast for a given day and if clipping is expected, export from battery based on SOC before PV generation starts and then keep export as priority until PV production reaches 5kW and if it exceeds, store the excess beyond 5kW in the battery.

**Describe alternatives you've considered**
In the past i have tried exporting in the morning manually and play around with washer Dryer etc but it is not a workable situation long term.

**Additional context**
I have 5kW Givenergy Gen2 Hybrid inverter and a 4kW Solis String inverter and total of 12kW PV, though DNO set PV limit through inverters is 8kW. 
Here my DNO export limit is set at 5kW and the inverter export limit is 5kW too. On the other hand battery to grid/home export limit of 3.9kW as limited by inverter.  The grid to battery is limited by inverter to 3.9kW too when charging. On the other hand, when PV is on, the inverter can handle 8.3kW total (3.3kW to battery and 5kW to the house/grid at the same time). When 9kW is being produced by the PV in total, 5kW gets exported and maybe .6kW gets used by home and the remaining gets clipped.

Currently Givenergy limits itself to 5kW export and if Solis sees the export at 5kW it derates itself accordingly to keep limit at 5kW.

gcoan:
> Would Predbat be able to predict the generation from Solcast for a given day and if clipping is expected, export from battery based on SOC before PV generation starts and then keep export as priority until PV production reaches 5kW and if it exceeds, store the excess beyond 5kW in the battery.

Predbat should do this automatically for you, if it detects clipping will occur, it should automatically force export to reduce the SoC level so excess solar doesn't get clipped and instead gets charged in the battery

You might need to artificially reduce your export limit that predbat thinks you have to force it to occur as solar could be higher than predicted and thus unpredicted clipping occurs

maverik-77:
Thanks for that, may be i have my configs wrong. May need assistance to fix it.

I have the inverter and export limit set as below. Moreover i only have the givenergy inverter added in for PV power. Do i need to add the Solis here too.

```
  pv_power:
  - sensor.givenergy_{geserial}_pv_power
[...]

PR #4023, opened 2026-06-07
Model curtailment on a negative export price (#3986)

### LINKED ISSUE #3986: Feature request: model a rate-conditional export limit (curtailment under a price threshold)
# Feature request: model a rate-conditional export limit (curtailment under a price threshold)

## Is your feature request related to a problem?

When the export price goes negative (dynamic tariff), exporting PV is a direct cost. Many of us prevent this by curtailing the inverter to net-zero export during negative-price slots. In my case an automation sets the SolarEdge site export limit to 0 when `export_price < 0`.

The problem is that Predbat has no way to know this curtailment happens. With a static `export_limit` (mine is 17000 W), Predbat models the export as proceeding normally during those slots. That has two consequences:

1. **Inaccurate cost/return projection.** Predbat books an export cost for negative-price slots that I actually avoid through curtailment, and it models the PV as exported rather than curtailed. The day's projected economics don't match reality.
2. **PV calibration pollution (separate but related).** Because the inverter throttles back PV production during curtailment, measured generation drops. With `metric_pv_calibration_enable` on (the default), Predbat treats this as panel underperformance and reduces the forecast. The docs already warn about this and the workaround is to turn calibration off — but that's a workaround, and it means Predbat is flying blind about *why* the generation is lower.

This is closely related to three existing reports about export-limit / curtailment modelling that are still open:

- #3442 — Sigenergy: does Predbat assign a cost to curtailed energy and value lowering SOC ahead of high-generation days? (DNO export limit, FIT)
- #3481 — feature request: keep export as priority and store excess in battery if clipping occurs (DNO export limit)
- #3828 — clipping due to export limit not detected/retained for the current day (DNO export limit)

Those three are all about a **static** export limit (DNO caps) that should be modelled and optimised around. My case is the same mechanism but **rate-conditional** (the limit only applies when the price is below a threshold). A single feature would cover all four situations.

## Describe the solution you'd like

A rate-conditional export limit in the prediction model. Concretely:

- A new optional config (e.g. `export_limit_rate_threshold` and an associated limit value, or an extension of the existing `export_limit`) that says: *when the slot's export rate is below threshold X, model the export limit as Y (e.g. 0) instead of the normal value.*
- Predbat does not drive the curtailment, it only models it. The actual curtailment is left to the user's inverter setting or automation. This keeps the feature inverter-agnostic and avoids any new control failure modes.

The building blocks already exist in `run_prediction` (apps/predbat/prediction.py):

- The per-slot export rate is already available, `export_rate = rate_export.get(minute_absolute, 0)`.
- The export-limit clipping already happens per-slot in the "Export limit, clip PV output" block, using `export_limit` and accumulating `clipped_today`.
- The cost function already penalises export at a negative rate (`metric -= export_rate * energy`).

So the change is small and localised: derive an effective per-slot export limit from the rate before the clipping block, e.g.

```python
# near the top of the slot loop, after export_rate is read
effective_export_limit = export_limit
if export_limit_rate_threshold is not None and export_rate < export_limit_rate_threshold:
    effective_export_limit = export_limit_negative   # e.g. 0, * step

# in the "Export limit, clip PV output" block, use effective_export_limit
if diff < 0 and abs(diff) > effective_export_limit:
    over_limit = abs(diff) - effective_export_limit
    ...
```

With that in place, during sub-threshold slots Predbat models net export as capped (PV above house load + battery headroom is curtailed, not exported), the return projection matches reality, and the optimiser sees the stored-PV value of keeping battery headroom ahead of a sunny negative-price window (avoided future import) rather than booking a phantom export cost.

For the static DNO cases (#3442 / #3481 / #3828) the same plumbing applies with the threshold disabled — the limit is just always active. The one extra step those cases additionally want is for the curtailed energy (`clipped_today`) to carry a cost/lost-value signal into the optimiser so it actively lowers SOC ahead of high-generation days. That's a deeper change to the cost function and could be a follow-up; the rate-conditional limit modelling stands on its own.

## The same curtailment signal also fixes PV calibration pollution

There's a second, separate problem with the same root cause, and knowing when curtailment happens fixes both.

`pv_calibration` in `solcast.py` calibrates the forecast by comparing ~7 days of measured `pv_today` against the historical forecast (`sensor.<prefix>_pv_forecast_h0`), per slot and per day, and derives a scaling factor `actual / forecast`. It makes no distinction between "panels underperformed" and "the inverter was curtailed", it just sees actual < forecast and scales the forecast down. The lower clamp is `PV_CALIBRATION_LOWEST = 0.20`, so a curtailed history can drag the forecast down to 20% of its real value (this matches the 42 kWh → ~7 kWh collapse reported in #3373).

So every time export is curtailed during negative-price slots, measured generation for those slots drops, and the calibration reads it as structural panel underperformance and pulls the future forecast down, even though the panels were fine and only the export was withheld.

The current documented workaround is to set `metric_pv_calibration_enable` off. That's fine for Solcast users, because the docs point them to Solcast's auto-dampening which already accounts for shutoff periods. But Forecast.Solar users have no equivalent. Forecast.Solar is a pure weather forecast with no feedback from actual production, so turning calibration off removes their only feedback layer, leaving only a manual, static `pv_scaling` knob. 

The fix is the same signal: exclude curtailed slots from the calibration input. When building `pv_power_hist_by_slot` and `past_day_actual` (around the history-aggregation loop in `pv_calibration`), skip slots whose historical export rate was below the threshold. Calibration then keeps working on normal days (real shading / forecast-bias correction) but is no longer polluted by deliberate curtailment.

That means one piece of knowledge "when is export curtailed" resolves two independent issues:

1. Modelling export correctly instead of booking an avoided cost (the export-limit problem above).
2. Keeping PV calibration clean so the forecast isn't dragged down by curtailment, which specifically rescues Forecast.Solar users from the current off-switch trade-off.

That's the argument for treating these together rather than as separate fixes.

## Describe alternatives you've considered

- **Predbat driving the curtailment** (a switch/service that sets the inverter's export limit). Rejected as the primary approach: it would need a per-inverter control implementation, introduces a second actor that can conflict with existing charge/discharge control and user automations, and adds failure modes (limit left hanging if a service call fails). Modelling-only is simpler, safer, and works for everyone. A control hook could be added later as an optional layer on top, via the existing service-API pattern.
- **Dynamically updating `export_limit` from an entity.** Doesn't work: `export_limit` is a system-wide scalar applied to all forecast slots equally, not a per-slot series. A momentary value would either block all future export or allow it everywhere — it can't represent "export OK now, not in 3 hours."
- **Inflating the export rate / `pv_metric10_weight`** (suggested in #3442). Indirect and imprecise; doesn't model the physical curtailment.

## Additional context

- Setup: 2× SolarEdge SE10K-RWB48BFN4 (leader/follower), 2× SE Home Battery (28.35 kWh total), ~36 kWp across three planes, dynamic tariff. Forecast via Forecast.Solar.
- Predbat v8.39.7
- Happy to work on a PR for the rate-conditional modelling

tieskuh:
## What this does

Implements the request in #3986: lets Predbat **model** the curtailment of export that many of us apply when the export price goes negative (e.g. an automation that sets the inverter''s export limit to zero while `export_price < 0`). Predbat does not drive the curtailment — it only models it, so the feature is inverter-agnostic and adds no new control failure modes.

## Why

With a static `export_limit`, Predbat models export as proceeding normally during negative-price slots, causing two problems:

1. **Inaccurate projection** — it books a phantom export cost/credit and models PV as exported for slots that are actually curtailed.
2. **PV calibration pollution** — when the inverter throttles generation during curtailment, measured generation drops and `pv_calibration` reads it as panel underperformance, dragging the forecast down (clamped as low as 20%). This hits Forecast.Solar users hardest, as they have no auto-dampening.

## What''s included

- **New control `select.predbat_curtail_on_negative_export_price`** (expert mode, default `off`):
  - `off` — no curtailment modelled (existing behaviour, unchanged).
  - `curtail_excess` — grid export modelled as blocked whenever the export rate is negative, while PV still charges the battery and supplies the house.
  - `solar_production_off` — PV modelled as fully off during negative-price slots, for inverters that can only disable generation rather than limit export.
- **Prediction** (`prediction.py`) — per slot, when enabled and `export_rate < 0`, the effective export limit is set to zero (and PV to zero in `solar_production_off`).
- **PV calibration fix** (`solcast.py`) — negative-price slots are excluded from the calibration input (both the actual and forecast aggregates), using the recorded history of the `predbat.rates_export` entity, so calibration can stay enabled without curtailment being mistaken for underperformance.
- **Plan view** (`output.py`) — curtailed PV is shown in a distinct colour; no new columns/fields, so the Predbat table card is unaffected.
- **Docs** — `customisation.md`, `energy-rates.md`, `apps-yaml.md`.
- **Tests** — unit tests for the prediction modes and the calibration exclusion.

## Design note

The original issue proposed a configurable price *threshold*. In testing, the only sensible threshold turned out to be 0 (curtail when the price is negative); a configurable value mainly added confusion, and export fees can be folded into the rate sensor instead. So the trigger is fixed at `< 0`, and the control is a single `select` that also captures the two real curtailment behaviours (export-only vs full PV-off).

## Off by default

With the default `off`, nothing changes for existing users. The feature is modelling-only; the actual curtailment stays with the user''s inverter or automation.

## Testing

- Unit tests added (model scenarios for both modes + a positive-price guard; calibration-exclusion test including a guard that the correct rate entity is queried).
- Validated on a live SolarEdge (dual-inverter) + Forecast.Solar system: the plan correctly curtails (colour + zero export) during negative-price slots and leaves positive-price export untouched. An A/B toggle confirmed the calibration exclusion fires only on negative-price days and leaves other days byte-identical.

Closes #3986.

### DIFF
--- a/apps/predbat/config.py
+++ b/apps/predbat/config.py
@@ -603,6 +603,20 @@ CONFIG_ITEMS = [
         "default": 0.0,
     },
     {
+        # Model curtailment of export when the export price is negative (issue #3986). Predbat does not drive
+        # the curtailment, it only models it - the actual curtailment stays with the user's inverter or automation.
+        # "off" (the default) disables the feature. "curtail_excess" models grid export as blocked while PV still
+        # charges the battery and supplies the house. "solar_production_off" models the PV as fully off, for
+        # inverters that can only disable generation rather than limit export.
+        "name": "curtail_on_negative_export_price",
+        "friendly_name": "Curtail on negative export price",
+        "type": "select",
+        "options": ["off", "curtail_excess", "solar_production_off"],
+        "icon": "mdi:transmission-tower-export",
+        "enable": "expert_mode",
+        "default": "off",
+    },
+    {
         "name": "car_charging_hold",
         "friendly_name": "Car charging hold (remove car charging energy from load data)",
         "type": "switch",
--- a/apps/predbat/fetch.py
+++ b/apps/predbat/fetch.py
@@ -2319,6 +2319,7 @@ class Fetch:
         self.iboost_solar_excess = self.get_arg("iboost_solar_excess")
         self.iboost_rate_threshold = self.get_arg("iboost_rate_threshold")
         self.iboost_rate_threshold_export = self.get_arg("iboost_rate_threshold_export")
+        self.curtail_on_negative_export_price = self.get_arg("curtail_on_negative_export_price", "off")
         self.iboost_charging = self.get_arg("iboost_charging")
         self.iboost_gas_scale = self.get_arg("iboost_gas_scale")
         self.iboost_max_energy = self.get_arg("iboost_max_energy")
--- a/apps/predbat/output.py
+++ b/apps/predbat/output.py
@@ -1178,6 +1178,12 @@ class Output:
             elif pv_forecast == 0.0:
                 pv_forecast = "&#9866;"
 
+            # Mark PV modelled as curtailed on a negative export price with a distinct colour, so curtailed solar
+            # is visually separated from normally-exported solar (issue #3986). With the feature off this never
+            # triggers, leaving the plan unchanged for existing users.
+            if raw_pv_forecast > 0.0 and self.curtail_on_negative_export_price != "off" and rate_value_export < 0:
+                pv_color = "#C8A2C8"
+
             pv_forecast = str(pv_forecast)
             if plan_debug and pv_forecast10 > 0.0:
                 pv_forecast += " (%s)" % (str(pv_forecast10))
--- a/apps/predbat/predbat.py
+++ b/apps/predbat/predbat.py
@@ -686,6 +686,7 @@ class PredBat(hass.Hass, Octopus, Energidataservice, Stromligning, Fetch, Plan, 
         self.iboost_smart_threshold = 0
         self.iboost_rate_threshold = 9999
         self.iboost_rate_threshold_export = 9999
+        self.curtail_on_negative_export_price = "off"
         self.iboost_plan = []
         self.iboost_energy_subtract = True
         self.iboost_running = False
--- a/apps/predbat/prediction.py
+++ b/apps/predbat/prediction.py
@@ -157,6 +157,7 @@ class Prediction:
             self.inverter_hybrid = base.inverter_hybrid
             self.inverter_limit = base.inverter_limit
             self.export_limit = base.export_limit
+            self.curtail_on_negative_export_price = base.curtail_on_negative_export_price
             self.pv_ac_limit = base.pv_ac_limit
             self.battery_rate_min = base.battery_rate_min
             self.battery_rate_max_charge = base.battery_rate_max_charge
@@ -513,6 +514,7 @@ class Prediction:
         car_energy_reported_load = self.car_energy_reported_load
         inverter_limit = self.inverter_limit * step
         export_limit = self.export_limit * step
+        curtail_on_negative_export_price = self.curtail_on_negative_export_price
         pv_ac_limit = self.pv_ac_limit * step
         set_charge_low_power = self.set_charge_window and self.set_charge_low_power and (save in ["best", "best10", "test"])
         carbon_enable = self.carbon_enable
@@ -578,6 +580,13 @@ class Prediction:
             if io_adjusted.get(minute_absolute, 0) and pv10 and minute > 30:
                 import_rate = self.rate_max  # Assume in worst case that slot goes away and max rate applies
             export_rate = rate_export.get(minute_absolute, 0)
+
+            # Model curtailment of export when the export price is negative (issue #3986). In curtail_excess
+            # mode the export is modelled as curtailed to net-zero while PV still charges the battery and serves
+            # the house; in solar_production_off mode the PV is modelled as fully off (see where pv_now is read).
+            # Predbat does not drive the curtailment, it only models it.
+            export_curtailed = curtail_on_negative_export_price != "off" and export_rate < 0
+            effective_export_limit = 0 if export_curtailed else export_limit
 
             # Alert?
             alert_keep = all_active_keep.get(minute_absolute, 0)
@@ -656,6 +665,9 @@ class Prediction:
 
             # Get load and pv forecast, total up for all values in the step
             pv_now = pv_forecast_minute_step_flat[minute]
+            # Model PV as fully off during curtailed slots for inverters that can only disable generation (issue #3986)
+            if export_curtailed and curtail_on_negative_export_price == "solar_production_off":
+                pv_now = 0
             load_yesterday = load_minutes_step_flat[minute]
 
             # Count PV kWh
@@ -812,8 +824,8 @@ class Prediction:
 
                 # Exceed export limit?
                 diff = get_diff(battery_draw, pv_dc, pv_ac, load_yesterday, inverter_loss, inverter_loss_recp)
-                if diff < 0 and abs(diff) > export_limit:
-                    over_limit = abs(diff) - export_limit
+                if diff < 0 and abs(diff) > effective_export_limit:
+                    over_limit = abs(diff) - effective_export_limit
                     reduce_by = over_limit
 
                     if reduce_by > battery_draw:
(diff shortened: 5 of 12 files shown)


springfall2008:
Looks good, but the use of 'off' is incorrect it should be False/True for apps.yaml settings. Let me know if you are okay to update?

tieskuh:
image
<br>
Happy to update if you'd like. I chose 'off ' as option because it is used in the existing convention for selects: manual_charge, manual_export, manual_demand, the manual_*_rates, manual_soc, manual_api, the freeze selects, etc. all use off as their default/option, so keeping it would be consistent with those (and the YAML-boolean caveat would apply to them equally). It's also a 3-way select (curtail_excess / solar_production_off / off), so it can't become a plain True/False.

I'm happy either way: keep off for consistency, or rename the disabled option to e.g. disabled. What would you prefer?


Below a screenshot of the plan tomorrow with the negative prices and curtailment:
image

I ran several A/B tests for a week to see if calibration works correctly.

tieskuh:
Cross-referencing #4036 (Native Solar Clipping Buffer) for coordination — the two are complementary, but they touch overlapping files, so whichever merges second will need a rebase.

Both deal with PV that can''t be exported, but via a different trigger and mechanism:

| | This PR (#4023) | #4036 |
|---|---|---|
| Trigger | export price is negative (rate-conditional) | PV exceeds a power limit (inverter AC / DNO export) |
| Goal | avoid a phantom export cost + keep PV calibration clean | avoid solar lost to clipping |
| Approach | modelling-only (models the curtailment) | active control (reserves battery SoC headroom) |
| Related issue | #3986 | #1206 (+ the static-limit family #3442 / #3481 / #3828) |

The original request in #3986 noted that the static export/DNO-limit cases would need a deeper change that actively lowers SoC ahead of high-generation days — which is exactly what #4036 tackles. So they complement rather than duplicate each other.

Overlap heads-up: both touch `prediction.py` (the export/clip block), `solcast.py`, `output.py`, `config.py`, `fetch.py` and several test files, so a rebase + conflict resolution will be needed for whichever lands second. Happy to rebase this one on top of #4036 if that goes first.

tieskuh:
@springfall2008 Hi Trefor, please let me know if you have any questions, need test results. Also if you want me to change or add anything, please let me know!

tieskuh:
**Live A/B test of the PV-calibration exclusion (real data)**

To check the calibration fix on real data, I ran the PV calibration on my live system
with the feature on (`curtail_excess`) and off, everything else identical, and compared
the per-day forecast/actual that calibration feeds on. Only days that actually had
negative-price slots should change; every other day should be byte-identical (the forecast
input is the same — only the exclusion differs).

Per-day calibration input — forecast kWh / actual kWh (implied scaling):

| Past day | Feature ON (`curtail_excess`) | Feature OFF | Changed? |
|---|---|---|---|
| 1 | 79.97 / 80.44 | 79.97 / 80.44 | — |
| 2 | 98.77 / 130.41 | 98.77 / 130.41 | — |
| 3 | 96.02 / 121.75 | 96.02 / 121.75 | — |
| 4 | 74.85 / 72.57 (×0.97) | 114.95 / 92.64 (×0.81) | negative-price day |
| 5 | 78.56 / 107.12 | 78.56 / 107.12 | — |
| 6 | 76.58 / 82.53 | 76.58 / 82.53 | — |
| 7 | 53.48 / 63.05 (×1.18) | 90.14 / 69.94 (×0.78) | negative-price day |

What this shows:
- Days without negative prices (1, 2, 3, 5, 6) are byte-identical between on/off — the
  exclusion leaves normal days untouched.
- The two days that had negative-price slots (4 and 7) change exactly as intended. With the
  feature off the curtailed slots are included and those days look like panel underperformance
  (×0.81 and ×0.78); with the feature on they are excluded and the remaining (non-curtailed)
  slots show the panels actually performed fine (×0.97 and ×1.18).
- On those two days the excluded slots had measured generation at roughly 50% and 19% of
  forecast — i.e. genuine generation suppression during curtailment — which is exactly the
  pollution that would otherwise drag the forecast down.

So calibration excludes only the curtailed (negative-price) slots, using the recorded
`predbat.rates_export` history, and the forecast is no longer suppressed by deliberate curtailment.


**Below are some screenshot of the plan view with the feature on and off**

This is what it looks like with the option 'curtail_excess'
image

This is what it looks like with the option 'solar_power_off'
image

This is what it looks like with the option 'off'
image

As you can see it models both scenario's correctly. I have had it running for about two weeks now without any issue.

PR #4036, opened 2026-06-09
Feature: Native Solar Clipping Buffer (Resolves #1206)

### LINKED ISSUE #1206: Clipping Buffer (hybrid inverter)
I currently experience clipping when solar generation exceeds inverter AC capacity, and there is no or limited battery capacity to make use of inverter DC capacity.

It would be great to see additional features designed around preventing clipping. Predbat does already attempt to accommodate clipping within the plan, however is closely optimised according to solar forecast and low / negative import rates (as intended!), which can come at the cost of reduced flexibility when e.g. battery is charged, and solar generation exceeds forecast rates. This is particularly evident in summer since the solar forecast looks at average generation across each hour, but in reality there could be peak generation for 10 minutes while there is a break in clouds, then 10 minutes of cloud cover etc.

The feature could work by allocating some overhead / buffer in the battery to accommodate clipping, especially during summer / peak generation months. The buffer should be active during period where generation is set to exceed inverter capacity. Once risk of clipping is over, the battery should revert to regular target SOC (e.g. 100%). The start / end time and amount of spare capacity could be determined based on e.g.:
1)	Manual user setting
2)	Ratio of import to export rate (e.g. allocate a higher buffer when export rate is significantly greater than import)
3)	Sunset / sunrise times, plus an offset

Additionally, battery charge power should be set to maximum throughout this period, and priority given to exporting generation (battery is only charged at the minimum rate which avoids clipping).

I currently run an automation which triggers manual discharge earlier in the day to create this buffer capacity within the battery.

**Additional context**

rholligan:
Resolves #1206. This PR introduces a native clipping buffer to Predbat. It prevents solar energy loss on high-generation days by reserving a dynamic buffer in the battery SoC, specifically targeting periods where PV generation is forecast to exceed inverter AC capacity or DNO export limits.

image

**Key Features:**
   * Dynamic Buffer Calculation: Computes the required buffer (sensor.predbat_clipping_buffer_kwh) based on the solar forecast area exceeding the "Effective
     Clipping Limit" (choosing the most restrictive hardware or DNO constraint).
   * Grid-Charge Capping: Restricts overnight grid charging to ensure headroom for the calculated buffer.
   * Proactive Risk Management: Includes a two-stage safety system. It uses a risk_threshold (trigger) and a configurable clipping_buffer_safety_margin
     (default 5% of window solar) to protect against un-forecasted cloud-edge spikes.
   * Dynamic Buffer Decay: Reserved space shrinks as the solar peak passes, releasing capacity for evening use.
   * Web UI Integration: Adds a "Clipping Analysis" chart to the Predbat Web UI, displaying PV curves, clipping ceilings, and the decaying buffer.
   * Improved Peak Reporting: Refined the clipping status text to report the actual high-risk generation period (minutes exceeding the limit) rather than the broad protection window, providing clearer information for the user.

  **Configuration Parameters:**
   * clipping_buffer_enable: Master toggle.
   * clipping_buffer_forecast: Selection of forecast model (recommends pv_clearsky).
   * clipping_buffer_safety_margin: Percentage of window solar to reserve as a safety factor (default 0.05).
   * clipping_buffer_risk_threshold: Sensitivity factor to trigger the safety margin logic (default 0.80).
   * clipping_buffer_can_discharge: Discharge behavior (None, Cost Optimal, or Always).
   * clipping_buffer_min_kwh / max_kwh: Manual size constraints.
   * clipping_buffer_limit_override: Manual power threshold override (Watts).

  **Testing & Documentation:**
   * Verified on live 5kW hardware; effectively mitigated 6.28kW solar spikes by holding battery headroom.
   * Includes unit tests in apps/predbat/tests/test_clipping.py.
   * Full feature documentation available in docs/clipping.md.

Developed with Gemini ✦


tieskuh:
Cross-referencing #4023 (Model curtailment on a negative export price) — our PRs look complementary and touch some of the same files, so flagging for coordination.

| | #4023 | This PR (#4036) |
|---|---|---|
| Trigger | export price is negative (rate-conditional) | PV exceeds a power limit (inverter AC / DNO export) |
| Goal | avoid a phantom export cost + keep PV calibration clean | avoid solar lost to clipping |
| Approach | modelling-only (models the curtailment) | active control (reserves battery SoC headroom) |
| Related issue | #3986 | #1206 (+ #3442 / #3481 / #3828) |

Yours handles clipping against a power limit with active SoC buffering; mine models price-driven curtailment and keeps PV calibration clean (modelling-only). We overlap in `prediction.py` / `solcast.py` / `output.py` / `config.py` and the tests, so whichever merges second will need a rebase. Happy to rebase mine on top if yours lands first.

springfall2008:
This looks very smart, I guess the question I have at the higher level is why does PredBat not handle the clipping correctly already, could it do so natively without having to work around it?

rholligan:
Thanks @springfall2008!

IMO the need for a solution like the clipping buffer stems from the combination of hardware limits and forecast resolution:

1. On free tier, Solcast provides smoothed average generation over 30-minute blocks. A block forecasting a 3kW average might mask local variability / maximums (e.g. 10 minutes of 6kW peak sun, followed by 20 minutes of cloud).
2. On systems with high DC-to-AC ratios (e.g. a 5kW hybrid inverter with 7kWp panels), that 6kW spike can only be captured if the battery has empty capacity to absorb the DC excess.
3. Because Predbat plans against the 30-minute average (3kW), it assumes generation will stay below the 5kW inverter limit. Consequently, the plan might optimise for e.g. cheap import rates and charges the battery to 100%.
4. When the 6kW spike actually occurs, the inverter is maxed out at 5kW, the battery is at 100%, and the 1kW DC excess is clipped and lost. The optimiser can't natively prevent this because the spike is hidden in the forecast average.

This feature introduces a proactive safety margin to ensure there is always a buffer of empty battery space ready to absorb those local above average spikes.

rholligan:
Thanks @tieskuh for flagging - more than happy to handle the rebase depending what first! Great work on #4023.

springfall2008:
>On free tier, Solcast provides smoothed average generation over 30-minute blocks. A block forecasting a 3kW average might mask local variability / maximums (e.g. 10 minutes of 6kW peak sun, followed by 20 minutes of cloud).

Inside PredBat solar there is a cloud model which creates a wave formation on the solar in 5 minute intervals throughout the 30 minute period basic on the variation between 10% and 90% forecasts. 

I'm wondering why it isn't working for you and if we can debug that?

daviessm:
> I'm wondering why it isn't working for you and if we can debug that?

I think there are probably a few things affecting it. This happens for me as well, and it seems to be at least partially because Predbat assumes that in freeze export mode the battery won't be charged whereas for FoxESS at least, it is, when the generation is above the export limit + house load. This causes Predbat to think that battery will fill slower than it actually does. See #3555 for a bit of a discussion on that.

rholligan:
Hey @springfall2008, taking your suggestion, I've had a stab at integrating the manual clipping overrides directly into the cloud model.

I'm still testing on my setup to evaluate real life results, but if you have a sec I'd appreciate you your eyes on the approach to see if it better aligns with your expectations.

You can check out the branch here: https://github.com/rholligan/batpred/tree/clipping-cloud-model

daviessm:
@springfall2008  Another example unrelated to my last: with an Axle event tomorrow, Predbat doesn't want to force export at all and looks like it's predicting to clip about 13kWh (export limit of 5kW), plus a random charge at 10am?

image

[predbat(5).log](https://github.com/user-attachments/files/28933943/predbat.5.log)
[apps.yaml.txt](https://github.com/user-attachments/files/28933944/apps.yaml.txt)

InfernoJaffa:
This is exactly what I need for my setup. I have oversized PV and regularly hit clipping/export limits on good solar days, especially when Predbat has filled the battery too early and there’s no headroom left to absorb the excess.

I’m currently having to use a fairly involved HA gatekeeper automation, with a lot of tuning, to try to prevent this: [https://github.com/springfall2008/batpred/issues/1206#issuecomment-4224843054](url)

A native configurable clipping buffer in Predbat feels like the right solution, since Predbat already has the forecast, battery state and optimisation context. I’d be very interested in this.

rholligan:
@springfall2008 raised #4086 with clipping protection integrated into the existing cloud model - let me know what you think!

giddie:
> I'm wondering why it isn't working for you and if we can debug that?

I think the issue may be that even at 90%, Solcast doesn't capture the intensity of direct sunlight spikes. Are the waves calculated from forecast energy for the slot, or max power?

springfall2008:
No new inline review threads or CI failures to address on this PR. Context from `main` since it was opened, which affects the discussion above:

- `prediction.py` on `main` now accounts for PV clipping natively in the simulation (`predict_clipped_best` / `clipped_today` against `pv_ac_limit`), so clipped energy is tracked in predictions rather than only post-hoc.
- The plan search weighs a p90 (upside) PV scenario via `pv_metric90_weight` / `calculate_pv90_plan` (default weight 0.15), so plans that only hold up under the central forecast are penalised when PV could come in above it — which is the masked-peak situation described in the rationale comment from June.

That doesn't fully supersede the SoC-headroom control proposed here (the p90 weight prices the risk in plan selection but doesn't reserve headroom), but it changes the baseline this diff was built against. Worth reconciling the three open, overlapping branches (#4023, this PR, #4086) against current `main` before any of them rebases — the p90 machinery changed the same `prediction.py` / `plan.py` surfaces this PR touches.

tieskuh:
I am still running https://github.com/springfall2008/batpred/pull/4023 on my personal fork which works great. I am happy to rebase if you are considering merging it.

springfall2008:
Tried merging `origin/main` into this branch today as part of routine cleanup: the merge stops with **49 conflict hunks across 12 files**, and the conflicts are architectural rather than textual, so I've aborted rather than guess at a resolution. The branch needs a manual rebase. What changed upstream that makes this non-trivial:

1. **Clipping is now accounted for natively in the prediction engine.** `run_prediction` in `prediction.py` clamps PV at `pv_ac_limit` and tracks `clipped_today` / `predict_clipped_best` internally. This PR builds the same accounting by hand via a `mitigated_today` element threaded through the prediction return tuple in ~26 places, so every one of those hunks is a semantic collision, not a text merge.

2. **The prediction call architecture was replaced.** The `wrapped_run_prediction_*` / `multiprocessing.Pool` structure this branch extends was removed; scenario predictions now go through the batched `prediction_kernel` (`enqueue_prediction` / handle-based `queue_run_prediction_*`), and the `run_prediction` return tuple changed shape. Porting the `mitigated_today` plumbing means re-implementing it through the new kernel path.

3. **The plan metric changed.** `compute_metric` in `plan.py` now takes a pv90 scenario (`soc90`/`cost90`/`final_iboost90`), and the optimiser's candidate-evaluation code this branch modifies was restructured around it.

Because of (1) in particular, a rebase should probably start by deciding which parts of the buffer logic are still needed at all: with native clipped-energy accounting and the pv90 upside already priced into the metric on main (see #1206 discussion), the remaining new value here is the SoC-headroom reservation and its web-UI chart. I can't resolve that reconciliation safely by hand-merging 49 hunks, so leaving it to a human-driven rebase.

chalfontchubby:
Claude here, on chalfontchubby's behalf. He asked me to look at why the auto-rebase bounced and whether this is worth reviving. Short version: the design is sound and I have data supporting it from #4865, but the conflicts aren't the real cost and I'd suggest not rebasing at all.

## The conflicts are mostly easy

12 files, 56 hunks, base is 13 June — **1177 commits** behind main. The two biggest look alarming and aren't:

- **plan.py, 173 vs 286 lines** — main added `plan_fragmentation()`, this PR added `calculate_clipping_buffer()`, at the same insertion point. Disjoint; keep both.
- **plan.py, 40 vs 8** — main added the whole pv90 scenario block; the PR side is only a tuple-unpack reformat. Take main.

About 45 of the 56 are one-liners, dictionary entries or adjacent additions. Only one is genuine rework: **solcast.py**, where main replaced the hand-rolled irradiance integration with a `gti_hourly_to_period_kwh()` helper. Since `pv_clearsky` is this feature's default forecast source, the clear-sky tracking has to be re-expressed inside that helper.

## The real cost is the C++ kernel

`run_prediction`'s hot loop is now mirrored in `prediction_kernel.cpp` under an explicit PARITY RULE — changes need the kernel updated in lockstep, `KERNEL_PARITY_REVISION` and `PK_PARITY_REVISION` bumped, and `./run_all --test kernel_parity` green.

Four of this PR's prediction.py hunks land inside that loop, and one changes `run_prediction`'s return arity (`mitigated_today`), which the kernel also marshals. None of that existed in June. That's the bulk of the remaining work, and it's C++, not conflict resolution.

## Four things I'd want changed regardless of rebase

1. **The hybrid clipping guard is dropped.** The PR replaces
   `if not inverter_hybrid and pv_ac_limit > 0 and pv_now > pv_ac_limit` with a generic `self.clipping_limit` test. On main that block is specifically for AC-coupled/microinverter systems; hybrids clip later against `inverter_limit` and `export_limit`. With both set to the same value a hybrid would be clipped twice. Worth checking — that's exactly the configuration in #4865.

2. **The metric penalty looks double-counted.** `revenue_lost` is added to both `metric` and `metric_keep`, but `compute_metric()` then does `metric += battery_cycle * metric_battery_cycle + metric_keep` on a `cost` that already contains it.

3. **`clipped_today` gets redefined.** `clipped_today -= soak_up / battery_loss` changes an existing published series from "PV lost" to "PV lost after mitigation" — the plan's Clip kWh column and `predict_clipped` silently change meaning. Better as a separate `mitigated` series alongside it.

4. **"Always" mode sets `export_limit_now` inside the prediction.** The simulation invents an action the planner didn't choose, so the executed plan won't match the simulated one.

## A sketch of the redo

The key realisation is that most of this doesn't belong in the hot loop at all. Splitting it three ways gets the feature in far sooner:

**Phase 1 — planning only, no kernel work.** Compute the buffer in `plan.py` and express it as a cap on `charge_limit_best` per charge window, plus an export floor. Both are things the planner and the prediction already understand, so the hot loop needs *no* changes and the kernel is untouched. "Don't charge above X" and "don't export below X" cover the grid-charge capping and the dynamic decay — the bulk of the described behaviour. This also fixes (4) for free, because the constraint lives in the plan and so gets executed.

**Phase 2 — forecast and visibility.** Clear-sky output inside `gti_hourly_to_period_kwh()`, `sensor.predbat_clipping_buffer_kwh`, the plan column and the web chart. Independent of phase 1 and separately reviewable.

**Phase 3 — only if measured to matter.** The in-loop "soak" behaviour, where clipped energy is allowed past the effective ceiling. That genuinely can't be expressed as a plan constraint, and it's the one piece that has to pay the kernel-parity cost. Its own PR, with `kernel_parity` green.

Two design questions that main has raised since June:

- **pv90 didn't exist when this was written.** There's now a third scenario blended into the metric at `pv_metric90_weight`, and it's the one where clipping actually shows up. The buffer needs defined behaviour there — arguably it should be *sized* from the upside scenario while being *costed* against the nominal one.
- **The config surface is seven items.** Starting with enable plus a max, and deriving the rest, would be an easier sell; the safety-margin and risk-threshold knobs could follow once there's field data.

## Supporting data

From #4865 — 8 days of 1-minute history from a 9.4 kW array behind a 5.5 kW limit, simulating exactly this buffer:

| day | generated | clipped, battery full | buffer for full recovery | recycled |
|---|---|---|---|---|
| broken cloud | 36.6 kWh | 3.50 kWh | **2.5 kWh** | 1.3x |
| clear sky | 57.2 kWh | 7.96 kWh | 7.5 kWh | 1.0x |
| other 6 days | 8-29 kWh | <=0.03 kWh | - | - |

That supports the feature and the sizing. It also supports the `pv_clearsky` default over a Solcast percentile: on the day examined, Solcast's pv90 peaked about 18% *below* that array's demonstrated clear-sky ceiling, so percentile-derived sizing would have undershot.

Worth reviving. Just probably not by rebasing this branch.

---

*Analysis by Claude, run and reviewed by chalfontchubby. Conflict figures from `git merge-tree` against main at 136249d7; buffer simulation from the debug files attached to #4865.*

chalfontchubby:
@rholligan — Claude again, on chalfontchubby's behalf. My previous comment jumped straight to "rebuild rather than rebase" without asking you, which wasn't really my call to make. It's your PR and your design, so: which would you prefer?

**If you'd rather it were rebuilt on current main**, you get credited as the author of the work — the design, the buffer sizing, the clear-sky approach and the config model are all yours. That means the new PR opening with your name and a link back here, a `Co-Authored-By:` trailer where your code survives in substance, and an explicit list of where the redo diverges from what you wrote and why, so nothing looks like a silent rewrite. This PR would be closed pointing at the replacement, not left to rot.

**If you'd rather push ahead with this branch**, that's entirely reasonable and I'd help with it instead. Being straight about the cost either way:

- The kernel-parity work is the same in both paths. Four of the prediction.py changes sit in the loop that's now mirrored in `prediction_kernel.cpp`, and the return-arity change touches its marshalling. Nobody avoids that by choosing one route over the other.
- What the rebase adds on top is 56 conflict hunks across 1177 commits — mostly mechanical, but `solcast.py` needs the clear-sky tracking moved inside the new `gti_hourly_to_period_kwh()` helper, and the pv90 scenario that landed since June needs a decision either way.
- What the rebase costs in review terms is that it stays one 2300-line change. My phase suggestion wasn't about the conflicts — it was that phases 1 and 2 (buffer as a plan-side constraint on `charge_limit`, then the forecast and UI) need no kernel changes at all, so they could land while phase 3 is still being written. That's the only real argument for the redo, and it applies just as well as a way of *splitting* this branch.

Reasons you might well prefer to keep going here that I can't weigh from outside: you've tested it on live 5 kW hardware and I haven't, the web UI and docs are already written, and you know the code. Any of those could outweigh the above.

The four things I flagged in my last comment (the dropped `not inverter_hybrid` guard, the apparent double-count of `revenue_lost`, the redefinition of `clipped_today`, and "Always" mode setting `export_limit_now` inside the prediction) are worth a look whichever route you take — I read those off the diff rather than running them, so I may have misread.

No rush, and no pressure either way. Just say which you'd prefer and we'll go with it.

---

*Written by Claude, on chalfontchubby's behalf.*

rholligan:
Thanks @chalfontchubby, @springfall2008 - you're right, main has moved on massively since June and this branch hasn't been kept up to date. 

I've been progressing clipping protection directly into the existing cloud-model structure over on #4086 - let's shift the discussion on merging into main there. 

That said, I'll close this PR (or let me know if you have a better idea).

PR #4078, opened 2026-06-17
feat(solis): OAuth 2.0 bearer auth mode alongside HMAC api-key

mgazza:
Adds a dual-mode auth path to the Solis driver via the shared `OAuthMixin`, so SolisCloud can be driven with an OAuth 2.0 Bearer token (refreshed through the SaaS `oauth-refresh` edge function) as an alternative to the existing HMAC api-key. **Existing api-key users are unaffected** — `auth_method` defaults to `api_key` and the HMAC path is unchanged.

### Changes
- `initialize()` accepts `auth_method`/`access_token`/`token_expires_at`/`token_hash`; OAuth mode uses the OAuth host (`api-oauth2.soliscloud.com`) with a Bearer header.
- `_build_headers` branches Bearer (OAuth) vs HMAC-SHA1 (api-key); `_execute_request` proactively refreshes and handles 401/403; cid/value control payloads unchanged.
- `components.py`: `api_key`/`api_secret` now optional, with a startup guard that requires them in api-key mode; adds `solis_auth_method`/`access_token`/`token_expires_at`/`token_hash` args.
- `_with_retry` aborts fast when OAuth refresh has permanently failed (no retry-window hang).
- Test: OAuth/HMAC header selection.

> Note: OAuth read/control endpoint paths reuse the HMAC paths at the OAuth host (design "approach 1") — to be confirmed by a manual smoke test before production use. Gated off by default in the SaaS until then.


apps/predbat/components.py
```diff
@@ -390,8 +390,13 @@
         "name": "Solis Cloud API",
         "event_filter": "predbat_solis_",
         "args": {
-            "api_key": {"required": True, "config": "solis_api_key"},
-            "api_secret": {"required": True, "config": "solis_api_secret"},
+            # api_key/api_secret (HMAC) OR auth_method=oauth+access_token must be supplied.
```
springfall2008: This can't work, all options are not required so Solis will be enabled in all configs

apps/predbat/solis.py
```diff
@@ -325,7 +339,12 @@ def _format_date(self, dt=None):
         return dt.strftime("%a, %d %b %Y %H:%M:%S GMT")
 
     def _build_headers(self, endpoint, payload):
-        """Build HTTP headers with HMAC-SHA1 authorization"""
+        """Build HTTP headers — Bearer (OAuth) or HMAC-SHA1 (api-key)"""
+        if getattr(self, "auth_method", "api_key") == "oauth":
```
springfall2008: Why getattr when auth_method is defined in initialize?

apps/predbat/solis.py
```diff
@@ -405,6 +433,10 @@ async def _with_retry(self, operation, max_retry_time=SOLIS_MAX_RETRY_TIME):
             try:
                 return await operation()
             except SolisAPIError as err:
+                # OAuth permanently failed (needs reauth) — abort immediately rather than
+                # burning the full retry window hitting the API with a known-bad token.
+                if getattr(self, "oauth_failed", False):
```
springfall2008: Where is auth_failed set?

apps/predbat/components.py
```diff
@@ -390,8 +390,13 @@
...
+            "api_key": {"required": False, "config": "solis_api_key"},
```
springfall2008, line 394: At least one required key is mandatory or all predbat's will start solis

apps/predbat/components.py
springfall2008, line 393: This can't work, all options are not required so Solis will be enabled in all configs

springfall2008 (dismissed):
This can't work, all options are not required so Solis will be enabled in all configs

apps/predbat/solis.py
mgazza: Good catch — `_init_oauth()` (called unconditionally from `initialize()`) always sets `self.auth_method`, so the `getattr` default was dead defensiveness. Switched to `self.auth_method` here and at the other two call sites (`_execute_request` 401 handler) in edaf90b7.

apps/predbat/solis.py
mgazza: It's `oauth_failed` (not `auth_failed`): set in `OAuthMixin._init_oauth()` — always initialised to `False` — and flipped to `True` on a `needs_reauth` result in `_do_refresh()`. The `getattr(self, "oauth_failed", False)` made it look like it might not exist; removed the `getattr` so the source is just `self.oauth_failed` in edaf90b7.

apps/predbat/components.py
mgazza: Fixed in edaf90b7. Added `"required_or": ["api_key", "access_token"]` to the solis component — the same gating mechanism Solcast and Octopus use — so Solis only activates when an HMAC `api_key` or an OAuth `access_token` is present. With neither configured the component is no longer constructed.

apps/predbat/components.py
mgazza: Agreed — that was the bug. Making all args optional (needed so OAuth-only configs validate) meant `have_all_args` stayed true for everyone. Fixed with `"required_or": ["api_key", "access_token"]` in edaf90b7, and added `test_solis_not_activated_without_credentials` (plus positive cases for each auth mode) so this can't regress.

apps/predbat/solis.py
```diff
@@ -355,6 +374,12 @@ def _build_headers(self, endpoint, payload):
     async def _execute_request(self, endpoint, payload):
         """Execute HTTP POST request to Solis API"""
         url = f"{self.base_url}{endpoint}"
+        # OAuth: proactively refresh the token before the call (no-op for api-key mode).
+        # If the token can't be obtained (refresh failed / needs reauth), fail fast — the
+        # oauth_failed flag below makes _with_retry abort instead of looping with a bad token.
+        if getattr(self, "auth_method", "api_key") == "oauth":
+            if not await self.check_and_refresh_oauth_token():
+                raise SolisAPIError("Solis OAuth token unavailable — reconnect required", status_code=401)
```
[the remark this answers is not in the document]
mgazza: Fixed in edaf90b7. After `check_and_refresh_oauth_token()` returns true we now also verify `self.access_token` is set and raise a 401 `SolisAPIError` otherwise, so we never send `Authorization: Bearer None` or burn the retry window. Covered by `test_oauth_execute_request_aborts_when_token_missing`.

apps/predbat/components.py
```diff
@@ -390,8 +390,13 @@
...
+            "api_secret": {"required": False, "config": "solis_api_secret"},
+            "auth_method": {"required": False, "config": "solis_auth_method", "default": "api_key"},
+            "access_token": {"required": False, "config": "solis_access_token"},
+            "token_expires_at": {"required": False, "config": "solis_token_expires_at"},
+            "token_hash": {"required": False, "config": "solis_token_hash"},
```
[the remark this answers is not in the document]
mgazza: Added `solis_auth_method`, `solis_access_token`, `solis_token_expires_at` and `solis_token_hash` to `APPS_SCHEMA` in config.py (edaf90b7), matching the existing `fox_*` OAuth key typing.

apps/predbat/tests/test_solis.py
```diff
@@ -664,6 +664,38 @@ async def test_run_read_only_skips_write():
...
+    api.access_token = "tok123"
+    headers = api._build_headers("/v2/api/atRead", {"inverterSn": "X", "cid": 636})
+    if headers.get("Authorization") != "Bearer tok123":
+        print("ERROR: OAuth _build_headers should set Bearer Authorization, got {}".format(headers.get("Authorization")))
+        failed = True
+    if "Content-MD5" in headers:
+        print("ERROR: OAuth _build_headers should not include Content-MD5")
+        failed = True
+
+    # API-key mode (default mock has no auth_method) — HMAC 'API <key>:<sig>'
```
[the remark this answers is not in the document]
mgazza: Updated in edaf90b7. `MockSolisAPI.__init__` now sets `auth_method="api_key"` (plus `access_token=None` and `oauth_failed=False`) to mirror what `initialize()`/`_init_oauth()` do, and the comment now reads "mock defaults auth_method to api_key, as initialize() does". This was also required because the production code now reads `self.auth_method` directly.

apps/predbat/solis.py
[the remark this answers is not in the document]
mgazza: Added focused coverage in edaf90b7: `test_oauth_execute_request_refreshes_before_call` (pre-call refresh + new Bearer header), `test_oauth_execute_request_aborts_when_token_missing` (fail-fast on missing token) and `test_with_retry_aborts_on_oauth_failed` (retry loop aborts once `oauth_failed` is set).

PR #4081, opened 2026-06-18
Lattice 1/3: device-mapping framework (read-only)

mgazza:
**Reworked to read-only** after maintainer feedback (control is premature without a common API; the near-term value is *mapping devices + sensors*). Control is now a separate, deferred design — see [lattice-spec#4](https://github.com/Predictive-Cloud-Ltd/lattice-spec/pull/4).

First of a 3-PR train that maps the devices on a site and inventories their sensors. **Read-only, flag-gated (`lattice_projection_enable`, default off), touches no control path.**

**Commits:**
1. read-only core — `lattice.py`: Fragment/Node/Sensor/AccessPath, `merge_fragments` (one node per device by serial, ranked access paths), `resolve_sensor`, `device_fragment`.
2. projection + component + flag — `LatticeProjection` (discover producers, merge, resolve sensor entity) + `LatticeComponent` (logs the merged device map) + the enable switch.

**Tests:** 12 standalone unit tests (`python -m pytest apps/predbat/test_lattice.py`).

Train: **PR1 (this)** → 2 gateway/GE-Cloud producers → 3 Fox/Solax/Solis producers.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4083, opened 2026-06-18
Lattice 3/4: hot-path control wiring (flag-gated)

mgazza:
**PR 3/4** of the Lattice merge train. Stacked on #2. **This is the one that touches the live control path** — review closely.

**Commits (story):**
1. site control plane — `execute.lattice_express_control` (resolve + execute via held task, log failures, return False so the normal write is never lost)
2. inverter expresses intent upward — `adjust_charge_rate/discharge_rate/battery_target/reserve` hand intent to the site plane via `_lattice_expresses`; bespoke write skipped only when Lattice owns it, else runs unchanged

Gated by `lattice_projection_enable` (default off) → identical behaviour today. Harness-verified (`run_all --test inverter --test execute`).

### DIFF
--- a/apps/predbat/components.py
+++ b/apps/predbat/components.py
@@ -44,6 +44,7 @@ except (ImportError, Exception):
     HAS_GATEWAY = False
     GatewayMQTT = None
 from load_ml_component import LoadMLComponent
+from lattice_component import LatticeComponent
 from datetime import datetime, timezone, timedelta
 import asyncio
 import os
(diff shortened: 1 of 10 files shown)


mgazza:
Closing — this was the live-control hot-path. Per maintainer feedback, generic control is premature without a common (schedule/shape-based) control API, which is now being designed separately in lattice-spec (#4). The train has been reworked to read-only device mapping + sensors (#4081/#4082/#4084); control will return once the common API is agreed.

PR #4099, opened 2026-06-20
Gateway: Fix plan controls for freeze charge/export

springfall2008:


kabadisha:
Just trying to understand this PR properly...

>Implement “freeze charge” detection (charge limit equal to reserve)

Can you expand on this? How does the charge limit being equal to the reserve qualify as a charge freeze? Or am I misunderstanding?

I also like the logic of setting the power rate to zero to execute a freeze. This would finally allow me to support charge freeze on my Sunsynk inverter. However, why set the target SoC to 100% or reserve?
Wouldn't it be more accurate to set the target SoC to the current SoC?

That way, it would work even for inverters that don't have a control for charge/discharge power settings.

Cheers :-)

Issue #2984, opened 2025-11-26
Discharging early @15p at the expense of importing later at 30p.


AndySmithers:
Lately I have noticed a few strange choices.
For tomorrow, for example:

Predbat is choosing to force export a couple of times during the day (albeit at low rate) even though the plan shows the battery running out before the IOG low price slot.
This makes no sense to me.

![Image](https://github.com/user-attachments/assets/6ac0efb1-4876-44c3-a8f3-c009ab3c8565)

[predbat.log](https://github.com/user-attachments/files/23774601/predbat.log)

gcoan:
please provide your debug file as per the issue template so your configuration can be understood.  Without this, very hard to understand what the problem might be

thanks

AndySmithers:
Done..

[predbat_debug.yaml](https://github.com/user-attachments/files/23776534/predbat_debug.yaml)

springfall2008:
Really odd as I can see your "bad" plan but when I re-run it myself it comes out as expected

<img width="1326" height="1079" alt="Image" src="https://github.com/user-attachments/assets/f627fa7f-93c6-4c48-966b-2e0c54e2d5d8" />

<img width="1326" height="1079" alt="Image" src="https://github.com/user-attachments/assets/27dc62da-b576-4479-ae22-d44a04dfd586" />

springfall2008:
Looking at your original plan from the debug yaml the PV amounts are not the same either

<img width="1083" height="1079" alt="Image" src="https://github.com/user-attachments/assets/88ac948d-50df-4dbf-8043-9477f584b2f5" />

AndySmithers:
Well, this is all 3 files just generated again. It looks even worse now!

[predbat.log](https://github.com/user-attachments/files/23778811/predbat.log)
[predbat_debug.yaml](https://github.com/user-attachments/files/23778812/predbat_debug.yaml)

[predbat_plan.html](https://github.com/user-attachments/files/23778865/predbat_plan.html)

AndySmithers:
And, when I turn off Export Low Power Mode, all those spurious exports disappear.

springfall2008:
I've figured it out now.

You have metric_battery_cycle set to 1.0 meaning you are adding 1p/kWh to charging and exporting of the battery.

In the slow exports are preventing the battery from charging so much (due to PV) but having no real impact on cost (as you will run out of battery either way before the end of the day).

If you turn off metric battery cycle (set to 0) the issue should go away and your costs will go down.

[...]

PR #4118, opened 2026-06-24
Sustain in-progress forced export across plan cycles (anti-flapping on flat peaks)

tieskuh:
## Problem

On an evening with a high export price that stays nearly flat across several half-hour slots while PV is still producing, forced export does not run continuously through the peak. Predbat re-plans every ~5 minutes and the per-slot forced-export decision oscillates: the battery exports for a few minutes, the slot then flips to "hold", PV charges the battery instead of exporting, then it flips back. Net effect is that energy is shifted out of the most expensive slot into later, cheaper ones.

The offline plan is clean and continuous — the oscillation only shows up across live recomputes.

## Root cause

`optimise_export` already has a hysteresis to keep an existing export window:

```python
if window_n < 2 and this_export_limit < 99.0 and self.export_window and self.isExporting:
    ...
    metric -= max(0.5, self.metric_min_improvement_export)
```

But the binding selection gate is on the raw `cost`, not the metric:

```python
if (metric <= off_metric) and (metric <= best_metric) and ((cost + min_improvement_scaled) <= off_cost):
```

On a near-flat multi-slot peak, exporting the current slot versus holding it and exporting the adjacent equal-priced slot is a cost coin-toss (tipped by the intraday load/PV adjustment each cycle). Because the hysteresis only lowers `metric` and never touches the cost gate, it cannot hold the export and the decision flip-flops between recomputes.

## Fix

When that same hysteresis condition fires (we are already exporting within this window) also relax the cost gate by the commitment amount, so an in-progress forced export is sustained across recomputes:

```python
if keep_export:
    min_improvement_scaled = min(min_improvement_scaled, -max(0.5, self.metric_min_improvement_export))
```

This only sustains an export that is **already running** (`self.isExporting`, `window_n < 2`, current time inside the window). It never opens a new export window, so it cannot reintroduce the `metric_keep` gaming guarded by #2984. The relaxation is bounded (`max(0.5, metric_min_improvement_export)`), so if holding becomes clearly better the export is still released.

## Tests

Adds `apps/predbat/tests/test_export_commitment.py` (registered as `export_commitment`):

- fresh plan, not exporting, large min-improvement → export correctly gated out (100%)
- already exporting within the window → export retained under commitment (<100%)
- stale export flag outside the prior window → no spurious export (guard)

The test fails without the patch (the in-progress export is dropped to 100%) and passes with it. `optimise_levels`, `optimise_windows`, `compute_metric`, `execute` and `model` groups still pass; black / ruff (F401) / interrogate / cspell are clean.

## Live validation

Validated on a real SolarEdge hybrid install across two consecutive evenings with similar near-flat two-hour export peaks — the two peak half-hours within ~2c/kWh of each other, which is the condition that triggers the oscillation (the absolute height is not what drives it):

- **Before (unpatched, ~0.69/0.67 peak):** the storage-mode select flipped mid-peak — `Discharge to Maximize Export` → `Maximize Self Consumption` → `Discharge to Maximize Export` within the first 15 minutes — and during the "hold" the remaining PV charged the battery instead of exporting, shifting energy out of the dearest slot into cheaper later ones.
- **After (this patch, ~0.94/0.96 peak):** a single transition into `Discharge to Maximize Export` at the start of the peak, held continuously for the full two hours, returning to demand only when the price dropped. **Zero mode-flips**; the battery held a steady ~10 kW discharge with no charging dips; SoC fell monotonically (100% → ~23%) with no saw-tooth.

A single evening is not conclusive on its own, but the oscillation condition was present and the before/after contrast on the same install is clear.

### DIFF
--- a/apps/predbat/plan.py
+++ b/apps/predbat/plan.py
@@ -1721,7 +1721,11 @@ class Plan:
                 dwindow = self.export_window[0]
                 if self.minutes_now >= pwindow["start"] and self.minutes_now < pwindow["end"] and ((self.minutes_now >= dwindow["start"] and self.minutes_now < dwindow["end"]) or (dwindow["end"] == pwindow["start"])):
                     metric -= max(0.5, self.metric_min_improvement_export)
-                    keep_export = True
+                    # Only relax the cost gate (further below) for an option that actually covers the current
+                    # minute. The option start is varied during optimisation, so a future-starting option is
+                    # not the in-progress export and must not receive the cost-gate commitment.
+                    if start <= self.minutes_now:
+                        keep_export = True
 
             # Round metric to 4 DP
             metric = dp4(metric)
--- a/apps/predbat/plan.py
+++ b/apps/predbat/plan.py
@@ -1715,11 +1715,13 @@ class Plan:
                 metric -= 0.001
 
             # Adjust to try to keep existing windows
+            keep_export = False
             if window_n < 2 and this_export_limit < 99.0 and self.export_window and self.isExporting:
                 pwindow = export_window[window_n]
                 dwindow = self.export_window[0]
                 if self.minutes_now >= pwindow["start"] and self.minutes_now < pwindow["end"] and ((self.minutes_now >= dwindow["start"] and self.minutes_now < dwindow["end"]) or (dwindow["end"] == pwindow["start"])):
                     metric -= max(0.5, self.metric_min_improvement_export)
+                    keep_export = True
 
             # Round metric to 4 DP
             metric = dp4(metric)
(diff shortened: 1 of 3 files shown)


tieskuh:
Thanks for the review — both points addressed in the follow-up commit.

**1. Commitment firing for a future-starting option.** Good catch. The cost-gate relaxation is now gated on `start <= self.minutes_now`, so it only applies to an option that actually covers the current minute (a genuinely in-progress export). The pre-existing metric nudge just above it is left unchanged; for a future-starting option that nudge is harmless in practice — the option starting at `window["start"]` exports the most and dominates selection — but the new cost-gate relaxation is the stronger lever, so it's now scoped tightly to in-progress options.

**2. Test leaving mutated state.** The test now restores the fields it changes (`isExporting`, `export_window`, the export-freeze flags and the `metric_min_improvement_export*` thresholds) before returning, so it can't make later tests in the shared-instance suite order-dependent.

The regression test still fails without the patch (the in-progress export drops to 100%) and passes with it; `optimise_levels` / `optimise_windows` / `compute_metric` / `execute` / `model` groups remain green.

Issue #4120, opened 2026-06-24
Prevent opting into Octopus Saving Sessions


parklife200:
**Describe the bug**
I had opted out of saving sessions, but somehow been opted back in and am now taking part. I have been opted in today for a session which conflicts with an Axle VPP event. Savings session from 19:30 - 20:30 and Axle from 19:00 - 20:00.
I am using the predbat direct octopus integration

**Expected behaviour**
An option to prevent being signed up for Octopus saving sessions, or at the very least not to sign up for an event that conflicts with an Axle event

**Predbat version**
8.42.0

**Environment details**
GivEngergy 5kW hybrid and 9.5kW battery
Standard HAOS

**Screenshots**

<img width="1088" height="496" alt="Image" src="https://github.com/user-attachments/assets/7e439293-fded-456c-8e59-6747168806b8" />

**Log file**

[predbat.log](https://github.com/user-attachments/files/29292088/predbat.log)

**Predbat debug yaml file**
[predbat_debug.yaml.txt](https://github.com/user-attachments/files/29292087/predbat_debug.yaml.txt)

giddie:
Just curious: I saw the same event overlap, but there actually isn't any _conflict_, I believe? Both events were driven by the same grid conditions. One event requires reduced import, and the other requires increased export. You can fulfil both requirements at the same time.

In my case, PredBat first planned to import a little after the Axle event, but postponed the import after joining the Octopus event too, which I thought was pretty neat.

But yeah, naturally it shouldn't be re-joining you after you opted out :p

Jonno12345:
> Just curious: I saw the same event overlap, but there actually isn't any _conflict_, I believe? Both events were driven by the same grid conditions. One event requires reduced import, and the other requires increased export. You can fulfil both requirements at the same time.
> 
> In my case, PredBat first planned to import a little after the Axle event, but postponed the import after joining the Octopus event too, which I thought was pretty neat.
> 
> But yeah, naturally it shouldn't be re-joining you after you opted out :p

This is what I was thinking on reading the change too - in fact it's worse than that because Octopus actively reward you for exporting during a Saving Session in Octopoints. I was getting a fantastic export rate of agile + saving session + axle of nearly £2 per kWh. I'm not sure why you'd want to opt out...!

dulochpark:
Thanks for doing these enhancements @springfall2008

CossieRob:
[...]

PR #4128, opened 2026-06-25
Octopus saving sessions: auto-join toggle and Axle conflict avoidance (#4120)

springfall2008:
## Summary

Addresses [#4120](https://github.com/springfall2008/batpred/issues/4120) — a user who had opted out of Octopus saving sessions was auto-joined back in, and the joined session conflicted with an Axle VPP event.

Previously Predbat auto-joined **every** available Octopus saving session unconditionally. The only opt-out was to remove the `octopus_saving_session` entry from `apps.yaml` entirely — which also throws away battery planning for sessions you *do* want to join.

This PR adds two complementary controls:

1. **`switch.predbat_octopus_saving_auto_join`** (On by default) — gate auto-join. When Off, Predbat no longer joins available sessions on your behalf, but still plans battery activity for sessions you've already joined.
2. **Axle conflict avoidance** — when auto-join is On, Predbat skips any available saving session that overlaps (any time overlap) an Axle VPP session, so it never commits to two conflicting events for the same period. The Axle session takes priority.

## Changes

- `config.py` — new `octopus_saving_auto_join` switch config item (default `True`).
- `fetch.py` — fetch Axle sessions before Octopus sessions and pass the list into `fetch_octopus_sessions()` (safe reorder; Axle fetch has no dependency on the Octopus fetch).
- `octopus.py` — gate the auto-join loop on the new switch; add `_saving_event_conflicts_axle()` half-open interval overlap test and skip conflicting events before joining (without stamping the 2h throttle, so a later cycle can still join if the Axle event disappears).
- `tests/test_saving_session.py` — new `test_saving_session_axle_conflict` and `test_saving_session_auto_join_toggle`, registered in `unit_test.py`.
- `docs/energy-rates.md` — document the new switch and the Axle-priority behaviour.

## Testing

- `./run_all -k saving_session` — all saving session tests pass (including the two new ones).
- `./run_all -k octopus_` and `-k axle` — pass.
- `./run_pre_commit` — ruff, black, cspell, markdownlint, and the quick test suite all pass.

### DIFF
--- a/apps/predbat/config.py
+++ b/apps/predbat/config.py
@@ -706,6 +706,12 @@ CONFIG_ITEMS = [
         "default": True,
     },
     {
+        "name": "octopus_saving_auto_join",
+        "friendly_name": "Octopus Saving Session Auto Join",
+        "type": "switch",
+        "default": True,
+    },
+    {
         "name": "octopus_intelligent_ignore_unplugged",
         "friendly_name": "Ignore Intelligent slots when car is unplugged",
         "type": "switch",
--- a/apps/predbat/fetch.py
+++ b/apps/predbat/fetch.py
@@ -950,9 +950,10 @@ class Fetch:
             # Basic rates defined by user over time
             self.rate_export = self.basic_rates(self.get_arg("rates_export", [], indirect=False), "rates_export")
 
-        # Fetch octopus saving sessions and free sessions
-        self.octopus_free_slots, self.octopus_saving_slots = self.fetch_octopus_sessions()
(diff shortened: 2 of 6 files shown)


robelmes:
Axle conflict avoidance — when auto-join is On, Predbat skips any available saving session that overlaps (any time overlap) an Axle VPP session, so it never commits to two conflicting events for the same period. The Axle session takes priority.

Not sure why you can't have both?  Are they really conflicting?  Saving session is met by doing export.

Would it not make sense to have the Axle conflict avoidance an optional thing rather than blanket application?

PR #4162, opened 2026-07-01
feat(gateway): publish rate_min/rate_max/export_rate for appliance RAG

mgazza:
## What
Publishes the user's real electricity **rate anchors** so the gateway display can colour appliances by absolute cost.

## Why
The gateway range-screen appliance RAG scored each appliance against its own per-load marginal-cost curve (no shared scale), which inverted EV vs washer (a cheap 12p/kWh wash showing RED next to an 18p/kWh EV trending GREEN). The firmware fix anchors colours to real rates — it needs those rates in the `predbat_data` payload.

## How
- **`marginal.py`**: expose `rate_min` / `rate_max` on the marginal-energy-costs sensor attributes.
- **`gateway.py`**: pure `extract_rate_anchors()` helper (rounds to 0.1p, all-or-nothing → `None`); include `rate_min` / `rate_max` / `export_rate` (from `grid_export_now`) in the `predbat_data` MQTT payload. Emits explicit `null` when unavailable so the firmware (`is<float>()`) falls back cleanly.

## Tests
`unit_test.py --test marginal_costs` and `--test gateway` green: `TestRateAnchors` (pure helper: valid→rounded, missing/non-numeric→None) + `TestPublishPredbatData` (payload present-and-rounded / null-when-absent) + marginal sensor attribute assertions.

## Coordination
Companion firmware change: **predbat-gateway** branch `fix/range-rag-anchored`. Backward-safe either order; feature takes effect once both ship.

### DIFF
--- a/.cspell/custom-dictionary-workspace.txt
+++ b/.cspell/custom-dictionary-workspace.txt
@@ -323,6 +323,8 @@ remotecontrol
 resetmidnight
 resultid
 resultmid
+rexp
+rmax
 rname
 Roboto
 rowspan
--- a/.cspell/custom-dictionary-workspace.txt
+++ b/.cspell/custom-dictionary-workspace.txt
@@ -323,8 +323,6 @@ remotecontrol
 resetmidnight
 resultid
 resultmid
-rexp
-rmax
 rname
 Roboto
 rowspan
--- a/apps/predbat/fetch.py
+++ b/apps/predbat/fetch.py
@@ -968,6 +968,7 @@ class Fetch:
             self.rate_scan(self.rate_import, print=False)
             self.rate_max_base = self.rate_max  # True peak rate before saving sessions / overrides inflate it
             self.rate_min_base = self.rate_min  # True off-peak rate before free sessions / overrides deflate it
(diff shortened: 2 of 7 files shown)


mgazza:
### xhigh code review → 3 batpred-side fixes (commit `947cf143`; gateway+marginal tests green; opus re-review clean)

- **F2 — anchor to the TRUE tariff.** Now publishes `rate_min_base`/`rate_max_base` and a new `rate_export_base` snapshot (captured in `fetch.py` *before* the saving/free-session + override loaders distort `rate_import`/`rate_export`) instead of the session-distorted `rate_min`/`rate_max`/`grid_export_now`. Previously an Octopus saving/free session skewed the gateway's span / near-peak / cheap-floor thresholds (peak appliances → AMBER not RED, or export inflation → all-GREEN).
- **F5 — NaN/Inf guard.** `extract_rate_anchors` now rejects non-finite → `None`; otherwise a NaN/Inf anchor serialises as bare `NaN` and the gateway drops the entire `predbat_data` payload.
- **F6 — precision.** Anchors rounded to 2dp to match the marginal-cost matrix, avoiding a rounding-only GREEN↔AMBER flip at the break-even boundary.

Re-reviewed (opus): all three resolved; snapshot ordering confirmed pre-distortion; no regressions.

PR #4239, opened 2026-07-12
fix(solis): don't force TOU bit on CID 636 when no V2 slot is active

springfall2008:
## Summary
- On V2-mode Solis inverters (per-slot CIDs), `write_time_windows_if_changed` always forced the storage-mode TOU bit (bit 1 of CID 636) on when choosing Self-Use/Feed-in-priority, regardless of whether slot 1 actually has a charge or discharge window configured.
- The physical inverter only retains that bit when a schedule is actually active, and silently clears it on read-back otherwise — this produced a permanent write/verify-fail loop (`Failed to verify CID 636 storage mode ..., wrote 35 but read back 33`) every cycle once both charge and discharge were disabled for slot 1.
- The V1 branch already has the correct behaviour (`in_charge_slot or in_discharge_slot` gating, falling back to the `"... - No Timed Charge/Discharge"` mode variant). This mirrors that same gating into the V2 branch, based on whether slot 1 has any enabled window.
- Cross-checked storage-mode bit numbering, CID 636, and the V2 per-slot CIDs against the independent [mkuthan/solis-cloud-control](https://github.com/mkuthan/solis-cloud-control) integration — identical mapping. That project also deliberately withholds manual TOU-bit control specifically for TOU-V2 inverters, corroborating that this bit shouldn't be forced by client code on V2.

## Test plan
- [x] Added `test_write_time_windows_v2_no_active_slot` regression test (confirmed it fails against the old code, passes after the fix)
- [x] `./run_all --test solis` — full Solis suite passes
- [x] `pre-commit run` (ruff, black, cspell) passes on changed files

### DIFF
--- a/apps/predbat/solis.py
+++ b/apps/predbat/solis.py
@@ -658,6 +658,7 @@ class SolisAPI(ComponentBase, OAuthMixin):
                 max_discharge_current_amps = float(self.cached_values.get(inverter_sn, {}).get(SOLIS_CID_BATTERY_MAX_DISCHARGE_CURRENT, 60.0))
                 charge_current = max_charge_current_amps
                 discharge_current = max_discharge_current_amps
+                slot1_active = False  # Whether slot 1 has any charge/discharge window configured
 
                 # There appears to be charge current limit on the slots which confused as slot 1 has a higher limit than the others,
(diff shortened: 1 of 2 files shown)


giceton:
Thanks for sorting this annoyance!

PR #4247, opened 2026-07-14
fix(car-charging): hold battery discharge during car charging regardless of car_energy_reported_load (#4246)

### LINKED ISSUE #4246: Battery discharges during car charging when car_energy_reported_load: Off, despite predbat.status showing 'Charging' (shared-busbar topology)
> 🤖 **Note:** This report was drafted with the assistance of Claude Code. The reporter's original observations and setup details are their own; the "Code-verified root cause" and "Possible fix options" sections below were produced by having Claude read the relevant source directly and trace the logic paths against the reported symptoms.

## Setup

- Sigenergy SigenStor EC 5.0 SP, TypQxQ Sigenergy-Local-Modbus integration
- EV charger (Hypervolt) wired **upstream of the CT clamp/meter** — car, solar, battery and grid all share the same physical connection point before metering. This is not "car fed via export" in a billing sense — it is electrically true: car draw and battery discharge are two loads/sources on the same busbar, with the grid making up any difference.
- `switch.predbat_car_energy_reported_load` = Off (correct per docs for this topology)
- `car_charging_now` was configured (Hypervolt `charging_readiness` sensor) and confirmed correctly detecting charging in real time — since removed, per docs recommendation, as Octopus Intelligent slot detection is already working and using both together is not recommended
- `switch.predbat_calculate_export_during_charge` = Off, `switch.predbat_combine_charge` = On
- `octopus_intelligent_slot`, `octopus_ready_time`, `octopus_charge_limit` all correctly configured and resolving via the Octopus Energy integration

## Direct evidence — `predbat.status` itself contradicts reality

During a confirmed car-charging window (16:30–17:00), the Logbook shows `predbat.status = "Charging"` for the entire window. Real telemetry for the same period shows the battery actively exporting/discharging, not charging — these are opposite actions. This is not a secondary/mirror diagnostic entity; `predbat.status` is Predbat's own canonical, headline statement of what it currently believes it is doing, and it was flatly wrong about the battery's actual behaviour for the full 30-minute window.

## Ruling out a general control failure

Predbat's ability to command this inverter is independently proven fine — a manual forced export test worked cleanly, and scheduled overnight charge windows show the discharge-limit register (`number.sigen_plant_ess_max_discharging_limit`) toggling 5.5kW↔0kW correctly and precisely in step with Predbat's own plan. So "Predbat can't control this inverter" is ruled out. That narrows the explanation to two possibilities: (a) Predbat's status/plan correctly decided "Charging" but the corresponding command was never actually issued for this specific car-charging scenario, or (b) the correct command was issued but Sigenergy's own onboard AI (independently known to be misbehaving since an unrelated grid outage) silently overrode it. Either way, the fault is specific to this decision path, not a general communication problem. (Code review below confirms it is (a), not (b) or a Sigenergy issue — see "Code-verified root cause.")

## On lag — checked and largely ruled out as the cause

A separate `binary_sensor` tracking "car charging AND exporting simultaneously" showed only a ~2 minute delay between the car starting to charge and the bad condition being flagged (measured before `watch_list` was even added) — so Predbat detects the car charging quickly. The ~30 minute duration of the mismatch is not a detection-lag problem; Predbat appears to know quickly that the car is charging, but something between that recognition and actually enforcing it at the inverter doesn't happen for the full duration of the slot.

## Observed behaviour (repeated on 3+ separate occasions across 2 days)

While the car is charging during a live Octopus Intelligent Go dispatch slot (confirmed genuine — grid import correctly showing 8p for the slot), the battery discharges heavily (multiple kW, confirmed via live power-flow/Sankey diagram) rather than holding. It's not clear from the outside whether this discharge is intended to serve the car's load, is an attempted export that the car happens to be absorbing (given the shared upstream connection), or some combination — `predbat.status` says "Charging" throughout, which doesn't match either interpretation. `number.sigen_plant_ess_max_discharging_limit` (the register Predbat itself writes to enforce charge/hold — confirmed working correctly for Predbat's own scheduled overnight windows) remained fully open (5.5kW) throughout these events; it was never set to 0 to prevent this discharge, regardless of what the intended action actually was.

## Real export opportunities missed as a direct result

On at least one occurrence, genuine Outgoing export rates of ~19p were available later the same day, comfortably clearing the ~18p round-trip breakeven (8p import + ~10p battery cycle wear) and representing real profit. The battery capacity spent discharging into/through the car during the cheap-import slot was capacity that would otherwise have been available to sell at that 19p rate — so this isn't just a small per-kWh loss during the car-charging slot itself, it's a knock-on loss of genuine same-day export revenue.

## Why this is a real loss, not just a missed optimisation

Round-trip battery cost (charge + discharge cycle wear, ~5p+5p in this config) is higher than the 8p Octopus Intelligent Go import rate — i.e. during this specific slot, simply importing grid power directly to serve the car is cheaper than discharging already-stored battery energy to serve the same load. So this isn't a marginal missed export opportunity — the battery is actively the worse choice here, and Predbat is choosing it anyway.

## Related issues and how this differs

Related issue and how this differs from it: #2188 describes a similar-sounding symptom (battery discharging into an ID.3 on Octopus Intelligent Go despite car-charging-hold being set), and one comment there attributes it to Predbat's plan taking ~5 minutes to catch up after an ad-hoc dispatch slot appears, with the maintainer's only response being to ask whether the data had simply not updated yet. Two things distinguish this report from that one:

- The mismatch persists for roughly 30 minutes, not resolving after ~5 as in #2188 — and per the detection-lag check above, this isn't simply slow recognition (recognition is ~2 minutes); the gap is between Predbat knowing the car is charging and actually enforcing that at the inverter for the remainder of the slot. At ~4kW discharge for 30 minutes, that's ~2kWh of battery capacity spent — both the small per-kWh cost differential (~4p) and, more significantly, the loss of that 2kWh of capacity that would otherwise have been available for the day's genuine export/avoided-import window later. This is a repeatable, non-trivial cost, not noise at the edge of a refresh cycle.
- #2188 does not involve `car_energy_reported_load: Off` — that reporter's car load is on the house-load side of the CT clamp. This report's root cause (per the code review below) is specific to the `car_energy_reported_load: Off` topology, where the only discharge-prevention mechanism is documented as having no effect at all — a distinct, and possibly more fundamental, issue than polling lag.

Related existing reports that may share a root cause: #2188, #2380, #3325.

---

## Code-verified root cause (Claude-assisted analysis)

Two call sites both gate the discharge-hold mechanism on `switch.predbat_car_energy_reported_load`, and both silently no-op when it is Off:

- [`apps/predbat/execute.py#L437`](https://github.com/springfall2008/batpred/blob/main/apps/predbat/execute.py#L437) — the "car charging from battery disable" block. Its guard is:
  ```python
  if self.set_charge_window and not self.car_charging_from_battery and self.car_energy_reported_load:
  ```
  When `car_energy_reported_load` is Off, this whole block — which calls `inverter.adjust_discharge_rate(0)` and flips `status` to `"Hold for car"` — is skipped entirely.

- [`apps/predbat/prediction.py#L701-L723`](https://github.com/springfall2008/batpred/blob/main/apps/predbat/prediction.py#L701-L723) — the equivalent logic in the planning/cost model. Same pattern: `discharge_rate_now = battery_rate_min` only executes `if self.car_energy_reported_load:`. In the `else` branch, car load is shunted into `car_load_energy_bypass` instead and never reaches `load_yesterday`, the value that drives the battery-vs-grid decision.

Downstream consequences, both matching the reported symptoms exactly:

1. **Status stays "Charging"**: `status` is set to `"Charging"` earlier at [`execute.py#L234`](https://github.com/springfall2008/batpred/blob/main/apps/predbat/execute.py#L234) and is only ever overwritten to `"Hold for car"` inside the skipped block above. With the block skipped, the status string never changes — explaining why `predbat.status` said "Charging" for the full 30-minute window while telemetry showed discharge.
2. **Discharge register left open at max, not 0**: `resetDischarge` is initialised `True` at the top of the inverter loop ([`execute.py#L79`](https://github.com/springfall2008/batpred/blob/main/apps/predbat/execute.py#L79)) and is only cleared to `False` inside the skipped car-holding block (or the "Hold charging"/"Freeze charging" branches, which don't apply here). Since it's never cleared, the fallback at [`execute.py#L493-L494`](https://github.com/springfall2008/batpred/blob/main/apps/predbat/execute.py#L493-L494) fires:
   ```python
   if resetDischarge:
       inverter.adjust_discharge_rate(inverter.battery_rate_max_discharge * MINUTE_WATT)
   ```
   This actively sets the discharge limit to **max** (5.5kW in the reporter's case) rather than leaving it untouched or setting it to 0 — a stronger and more damaging outcome than mere inaction.
3. **No cost-comparison exists in this mode**: the only logic touching `car_load_energy_bypass` ([`prediction.py#L1152-L1156`](https://github.com/springfall2008/batpred/blob/main/apps/predbat/prediction.py#L1152-L1156)) is a one-sided discount applied to *assumed export revenue*, not a discharge-avoidance or battery-vs-grid cost comparison. So there is genuinely no mechanism, in this configuration, that could have made the "import beats battery cycle cost" decision the reporter describes.

### Documentation discrepancy (secondary, not the cause of the bug)

`docs/car-charging.md` (around line 48) states that when `car_energy_reported_load` is Off, Predbat will "Automatically set `switch.predbat_car_charging_from_battery` to On in the model." No such auto-assignment exists in code — checked all three write sites (`fetch.py:1932`, `prediction.py:137`, `prediction_kernel.py:413`); the switch is read verbatim from user config and never forced. The real mechanism is the `and self.car_energy_reported_load` guard shown above, not an auto-toggle of the other switch. The net documented behaviour ("has no effect") is directionally correct, but the doc misdescribes the mechanism — worth a docs fix regardless of how the functional gap is resolved.

### Design assumption behind the current behaviour

The guard exists because Predbat assumes `car_energy_reported_load: Off` implies the car is on a **physically separate circuit** the battery cannot reach (typical "car charger wired directly to the grid connection, outside the CT clamp" topology). Under that assumption, zeroing battery discharge during car charging is pointless since the battery can't serve the car anyway, so the code intentionally skips it and instead just declines to credit export revenue for that period.

That assumption does not hold for topologies where the CT clamp placement is the *only* reason `car_energy_reported_load` is Off, but the car, battery, solar and grid are all still electrically joined upstream of the meter (as in this report — car and battery share a busbar, `car_energy_reported_load: Off` is set purely because the car isn't inside the CT clamp's measurement point, not because it's electrically isolated). Predbat currently has no way to distinguish these two topologies — both present identically as `car_energy_reported_load: Off`.

### Possible fix options

1. **Decouple the discharge-hold mechanism from `car_energy_reported_load`.** Remove (or make independently configurable) the `and self.car_energy_reported_load` guard in both `execute.py:437` and `prediction.py:715`, so `car_charging_from_battery = Off` always zeroes discharge during car-charging windows regardless of CT clamp topology. Simplest change, smallest surface area. Risk: for users where the assumption genuinely holds (car truly isolated from the battery), this would needlessly restrict discharge/export during car charging with no benefit — a behaviour regression for that population, though a fairly conservative one (worst case: some missed export income, not incorrect operation).

2. **Add a new switch/config option to opt back into discharge-holding when `car_energy_reported_load` is Off** (e.g. `car_charging_shared_busbar` or similar), defaulting Off to preserve current behaviour for the "genuinely isolated circuit" case, and lets topologies like this reporter's opt in explicitly. Slightly more config surface but no behaviour change for existing users who rely on the current assumption. Would need a docs update explaining when to enable it.

3. **Add the missing cost-comparison instead of a hard discharge-zero.** Rather than an all-or-nothing hold, compute round-trip battery cost vs. current import rate (as the reporter suggests: ~10p battery cycle cost vs 8p Octopus IOG import) and only permit discharge during car-charging windows when it's actually cheaper than import. This is the most correct long-term fix and generalises beyond this one topology (also relevant to `switch.predbat_calculate_export_during_charge` and similar heuristics elsewhere), but is the largest change — needs new cost logic in `prediction.py`, plus equivalent handling in `execute.py`'s live-control path, and is more work to get right/test than options 1-2.

4. **Fix `predbat.status` independently of the discharge fix.** Regardless of which discharge fix is chosen, `status` should not silently remain `"Charging"` when the car-holding logic is bypassed — at minimum it's a diagnostic/trust problem (this report explicitly cites the wrong status as primary evidence something is broken). Could set a distinct status (e.g. `"Charging (car, export blind)"`) when in the `car_energy_reported_load: Off` + car-charging state, independent of whether discharge is actually being held.

### Suggested triage priority

Options 2+4 together look like the lowest-risk, most surgical fix: add an opt-in switch for shared-busbar topologies (preserves existing behaviour for the isolated-circuit case that the current design correctly serves) and fix the status string so it never again claims "Charging" while contradicting live discharge telemetry. Option 3 (proper cost comparison) is the "correct" long-term answer and worth a follow-up issue, but is a larger scope than this specific report needs to be actionable.

---

**Question for maintainers:** Is there a way to get the cost-comparison / discharge-avoidance behaviour of `car_charging_from_battery` to apply in the `car_energy_reported_load: Off` configuration, given that the "battery cannot supply the car" assumption doesn't hold for a topology where car, battery and grid share a connection point upstream of metering? Or is this a genuine gap where nothing currently prevents this specific loss?

Happy to provide the full `apps.yaml` car-charging config section, plan table screenshots, and discharge-limit history graph if useful.

chalfontchubby:
> 🤖 Drafted with Claude Code assistance.

Fixes #4246. Contains two commits: a regression test that reproduces the bug against `main`, followed by a fix.

## The fix

Removes the `and self.car_energy_reported_load` condition from the discharge-hold gate in `execute.py` and `prediction.py` (with the matching change mirrored in `prediction_kernel.cpp` to keep C++/Python parity), so `car_charging_from_battery = Off` always holds discharge during car-charging windows, regardless of CT-clamp visibility.

This is **"option 1"** from the fix options discussed in #4246 - the simplest of a few possible approaches, chosen because:
- `car_energy_reported_load` only ever gated CT-clamp *load-history/export-accounting* logic elsewhere in the codebase (confirmed by auditing every remaining use) - conditioning the *discharge-hold* decision on it as well looks like a category error rather than intentional design, since that switch says nothing about whether the battery can physically reach the car.
- Charging and discharge-holding are controlled by independent code paths (`adjust_charge_window`/charge-rate vs. `adjust_pause_mode`/discharge-rate), so this doesn't risk interfering with a legitimate "charge now, rate is cheap" decision - verified both by code inspection and the included regression test.

It is **not** the only way to close this gap - #4246 also discusses:
- **Option 2**: an opt-in switch, defaulting Off, for shared-busbar topologies specifically - preserves current behaviour for any (largely theoretical) genuinely-isolated-circuit population, at the cost of more config surface.
- **Option 3**: a proper cost comparison (round-trip battery cost vs. import rate) instead of an unconditional hold - the more "correct" long-term answer, but meaningfully larger in scope (also touches the compiled kernel's cost model, not just this one gate) and out of scope for this PR.

Went with option 1 here as the smallest, most surgical change that fixes the reported behaviour; happy to rework this PR toward option 2 or 3 instead if a maintainer prefers a different tradeoff.

## Verification

- New/updated regression test (`no_discharge_car_demand1b` in `test_execute.py`) passes against the fix, fails against `main`.
- Full local suite (`unit_test.py`, no `--quick`) passes, including `kernel_parity` and `model_kernel` (C++/Python engines stay in sync).
- `run_pre_commit` passes cleanly (ruff, black, cspell, etc.).

### DIFF
--- a/apps/predbat/execute.py
+++ b/apps/predbat/execute.py
@@ -432,9 +432,11 @@ class Execute:
 
                 status_freeze_export = " [Freeze exporting]"
 
-            # Car charging from battery disable?
+            # Car charging from battery disable? Applies regardless of car_energy_reported_load - that
+            # switch only controls CT-clamp load-history/export accounting, not whether the battery is
+            # physically able to serve the car (it shares the same busbar either way).
             carHolding = False
-            if self.set_charge_window and not self.car_charging_from_battery and self.car_energy_reported_load:
+            if self.set_charge_window and not self.car_charging_from_battery:
                 for car_n in range(self.num_cars):
                     if self.car_charging_slots[car_n]:
                         window = self.car_charging_slots[car_n][0]
--- a/apps/predbat/execute.py
+++ b/apps/predbat/execute.py
@@ -433,8 +433,8 @@ class Execute:
                 status_freeze_export = " [Freeze exporting]"
 
             # Car charging from battery disable? Applies regardless of car_energy_reported_load - that
-            # switch only controls CT-clamp load-history/export accounting, not whether the battery is
-            # physically able to serve the car (it shares the same busbar either way).
+            # switch only controls whether EV energy is included in the CT-clamp house-load model, not
+            # whether we enforce the discharge hold.
             carHolding = False
             if self.set_charge_window and not self.car_charging_from_battery:
                 for car_n in range(self.num_cars):
--- a/apps/predbat/prediction.py
+++ b/apps/predbat/prediction.py
@@ -720,7 +720,7 @@ class Prediction:
                             car_load_energy_bypass += car_load_scale / self.car_charging_loss
 
                         # Model not allowing the car to charge from the battery - applies regardless of
-                        # car_energy_reported_load, as the battery shares the same busbar either way
+                        # car_energy_reported_load, which only controls CT-clamp house-load inclusion
                         if (car_load_scale > 0) and (not self.car_charging_from_battery) and set_charge_window:
                             discharge_rate_now = battery_rate_min  # 0
 
--- a/apps/predbat/prediction.py
+++ b/apps/predbat/prediction.py
@@ -716,11 +716,13 @@ class Prediction:
                             # Only add load if the car is reporting it as load, otherwise its outside the CT Clamp
                             car_amount_premium += car_load_scale / self.car_charging_loss
                             load_yesterday += car_amount_premium
-                            # Model not allowing the car to charge from the battery
-                            if (car_load_scale > 0) and (not self.car_charging_from_battery) and set_charge_window:
-                                discharge_rate_now = battery_rate_min  # 0
                         else:
                             car_load_energy_bypass += car_load_scale / self.car_charging_loss
+
+                        # Model not allowing the car to charge from the battery - applies regardless of
+                        # car_energy_reported_load, as the battery shares the same busbar either way
+                        if (car_load_scale > 0) and (not self.car_charging_from_battery) and set_charge_window:
+                            discharge_rate_now = battery_rate_min  # 0
 
             # Iboost
             iboost_rate_okay = True
--- a/apps/predbat/prediction_kernel.cpp
+++ b/apps/predbat/prediction_kernel.cpp
@@ -503,12 +503,14 @@ int32_t pk_run(int64_t handle, const PkScenario *s, PkResult *out)
                         // Note: mirrors the Python engine exactly - the cumulative premium amount is added per car
                         car_amount_premium += car_load_scale / c->car_charging_loss;
                         load_yesterday += car_amount_premium;
-                        // Model not allowing the car to charge from the battery
-                        if ((car_load_scale > 0) && (!c->car_charging_from_battery) && c->set_charge_window) {
-                            discharge_rate_now = battery_rate_min; // 0
-                        }
                     } else {
                         car_load_energy_bypass += car_load_scale / c->car_charging_loss;
+                    }
+
+                    // Model not allowing the car to charge from the battery - applies regardless of
+                    // car_energy_reported_load, as the battery shares the same busbar either way
+                    if ((car_load_scale > 0) && (!c->car_charging_from_battery) && c->set_charge_window) {
+                        discharge_rate_now = battery_rate_min; // 0
                     }
                 }
             }
--- a/apps/predbat/prediction_kernel.cpp
+++ b/apps/predbat/prediction_kernel.cpp
@@ -508,7 +508,7 @@ int32_t pk_run(int64_t handle, const PkScenario *s, PkResult *out)
                     }
 
                     // Model not allowing the car to charge from the battery - applies regardless of
-                    // car_energy_reported_load, as the battery shares the same busbar either way
+                    // car_energy_reported_load, which only controls CT-clamp house-load inclusion
                     if ((car_load_scale > 0) && (!c->car_charging_from_battery) && c->set_charge_window) {
                         discharge_rate_now = battery_rate_min; // 0
                     }
--- a/apps/predbat/tests/test_execute.py
+++ b/apps/predbat/tests/test_execute.py
@@ -2170,6 +2170,25 @@ def run_execute_tests(my_predbat):
     if failed:
         return failed
 
+    # Battery and car share a busbar upstream of the CT clamp (car_energy_reported_load=False is a
+    # monitoring blind spot, not electrical isolation), so discharge must still be held here exactly
+    # as it is for the reported_load=True case above (no_discharge_car_demand1) - see issue #4246.
+    failed |= run_execute_test(
+        my_predbat,
+        "no_discharge_car_demand1c_shared_busbar",
+        set_charge_window=True,
+        set_export_window=True,
+        soc_kw=100,
+        assert_status="Hold for car",
+        assert_pause_discharge=True,
+        car_slot=charge_window_best_slot,
+        assert_immediate_soc_target=100,
+        car_charging_from_battery=False,
+        car_energy_reported_load=False,
+    )
+    if failed:
+        return failed
+
     failed |= run_execute_test(
         my_predbat,
         "no_discharge_car_demand2",
--- a/apps/predbat/tests/test_execute.py
+++ b/apps/predbat/tests/test_execute.py
@@ -2154,28 +2154,11 @@ def run_execute_tests(my_predbat):
     if failed:
         return failed
 
+    # Discharge-hold applies regardless of car_energy_reported_load - that switch only controls
+    # CT-clamp load-history/export accounting, not whether the battery can physically serve the car.
     failed |= run_execute_test(
         my_predbat,
         "no_discharge_car_demand1b",
-        set_charge_window=True,
-        set_export_window=True,
-        soc_kw=100,
-        assert_status="Demand",
-        assert_pause_discharge=False,
-        car_slot=charge_window_best_slot,
-        assert_immediate_soc_target=100,
-        car_charging_from_battery=False,
-        car_energy_reported_load=False,
-    )
-    if failed:
-        return failed
-
-    # Battery and car share a busbar upstream of the CT clamp (car_energy_reported_load=False is a
-    # monitoring blind spot, not electrical isolation), so discharge must still be held here exactly
-    # as it is for the reported_load=True case above (no_discharge_car_demand1) - see issue #4246.
-    failed |= run_execute_test(
-        my_predbat,
-        "no_discharge_car_demand1c_shared_busbar",
         set_charge_window=True,
         set_export_window=True,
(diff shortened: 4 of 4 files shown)


springfall2008:
Thanks for this, @chalfontchubby — the fix itself is good. The red CI here isn't the change; it's that the pre-built kernel binaries (`prediction_kernel_lib_*.so`) need regenerating after the `prediction_kernel.cpp` edit, and the `kernel-binaries` auto-commit job can't run on fork PRs (it needs `contents: write` on the head branch), so the checked-in binary stays stale and fails the parity verify step.

I've brought your commits into a branch in the main repo — `fix/car-charging-discharge-hold` — so the kernel binaries can be rebuilt and committed automatically, and I'll merge from there. Your original commits (and authorship) are preserved on that branch.

Closing this in favour of that branch — full credit to you for the fix. Superseded by the work carried over from this PR (#4247).

PR #4254, opened 2026-07-14
fix(axle): don't fail the run for a user-disabled component

mgazza:
## Problem

When a user turns **Axle automation off** in the UI but leaves a stored API key, the Axle component stays loaded and keeps polling (component enablement is `required_or: [api_key, managed_mode]`). If that key has been rotated/expired, every fetch returns **HTTP 401**, so the component reads *active-but-not-alive*.

`record_final_run_status()` treats any active-but-not-alive component as a **hard run failure** — even though the user opted out and the plan itself computed fine:

```
AxleAPI: Fetching latest VPP event data (BYOK)
Warn: AxleAPI: Request failed with status 401
Error: Complete run status Charging with component errors: Axle Energy
Warn: record_status Error: Complete run status ... with component errors: Axle Energy
```

A run stamped `Error` (`had_errors=True`) isn't committed cleanly, so battery control **thrashes** — we've seen instances export at the outgoing rate then re-import at the peak import rate on alternating runs, costing users money — and `components_healthy` flaps off on every fetch run.

This is the residual case behind a recent fleet incident: an Axle key-rotation left a batch of opted-out users still erroring every run, because turning Axle *off* stops the planner acting on sessions (`automatic=False`) but does **not** stop the component polling and 401-ing.

## Fix

Add a `health_exempt` concept so a component the user has disabled can be *degraded* without failing the whole run:

- `component_base.py` — `health_exempt()` defaults to `False`.
- `axle.py` — override returns `not self.automatic` (exempt only while automation is user-disabled).
- `components.py` — manager delegator `health_exempt(name)`.
- `predbat.py` `record_final_run_status()` — an active-but-not-alive component that is `health_exempt` is reported as `"degraded"`; it does **not** increment `error_count`, join `failed_components`, or flip `all_healthy`. Non-exempt components are unchanged.

Net effect: with Axle automation off, a broken/rotated key surfaces as *degraded* (still visible, `failures_total` and warn logs intact) but never fails the run or marks the instance unhealthy — the user opted out. With automation **on**, an unhealthy Axle still fails the run exactly as before.

## Tests

- `test_component_health_status.py`: two new cases — an exempt-unhealthy component records a **success** status; a mix of exempt + non-exempt records an error naming **only** the non-exempt component. (`FakeComponents` gains a `health_exempt` stand-in.)
- `test_axle.py`: new `health_exempt` subtest — `True` when `automatic=False`, `False` when `automatic=True`.

All `component_health_status` and `axle` (31/31) tests pass; `black --check` clean.

### DIFF
--- a/apps/predbat/axle.py
+++ b/apps/predbat/axle.py
@@ -579,13 +579,17 @@ class AxleAPI(ComponentBase):
         )
 
     def health_exempt(self):
-        """Axle is health-exempt while the user has disabled automation.
+        """Axle is health-exempt only when no axle_session entity is configured.
 
-        With automation off (automatic=False) the component still polls so the event
-        sensor stays honest, but a broken or rotated credential returning errors must not
-        fail the Predbat run or mark the instance unhealthy — the user has opted out.
+        Axle stays loaded whenever an api_key is present, but it only influences the plan
+        when an axle_session entity is set — either the automatic default (axle_automatic=True)
+        or a manual override. When neither is configured the user has effectively disabled Axle
+        (e.g. the SaaS "Axle off" toggle leaves the key but no session), so a broken or rotated
+        credential returning errors must not fail the run or mark the instance unhealthy. A user
+        who names axle_session manually is still actively using Axle and is NOT exempt. This
+        mirrors the participation check in fetch_axle_sessions().
         """
-        return not self.automatic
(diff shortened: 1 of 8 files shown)


springfall2008:
I think this is incorrect, the component is enabled with the key and not with automatic

PR #4256, opened 2026-07-16
fix(github): don't cache failed GitHub API responses for the full 2-h…

chalfontchubby:
Symptom: After clearing problems with GitHub api calls for my account, present continues to complain about being unable to download GitHub version information, leading to som confusion as to the continuing problem. download_predbat_releases_url fetches the GitHub version and caches it for a 2 hour window to reduce api calls to a minimum. 
However it never checked the HTTP status before treating the response body as valid data - GitHub's API returns a JSON body even on errors (e.g. rate limiting), so a 403 was parsed fine and cached as if it were real release data for 2 hours. That meant a transient rate limit kept showing "Unable to download Predbat version information" for up to 2 hours after the limit had actually cleared.

Now checks response.ok before parsing/caching, and on any failure (network error, non-2xx, bad JSON) falls back to the last successfully cached data instead of caching the failure - so the next 5-minute plan cycle retries rather than waiting out the full cache window. Also adds a request timeout, since this is an external call with no bound on how long it can hang.

### DIFF
--- a/apps/predbat/github.py
+++ b/apps/predbat/github.py
@@ -68,32 +68,43 @@ class GitHub:
     def download_predbat_releases_url(self, url):
         """
         Download release data from GitHub, but use the cache for 2 hours
+
+        On failure (network error, non-2xx status, or bad JSON) falls back to the last
+        successfully cached data if any, and never caches the failure itself - this way
+        a transient error (e.g. rate limiting) is retried on the next plan cycle rather
+        than getting stuck for the full 2-hour cache window after it's already cleared.
         """
         self._load_github_url_cache_from_storage()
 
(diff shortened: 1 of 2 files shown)


chalfontchubby:
(Hi Trefor - don't think we ever overlapped at IMG - I left in 2012 - Rik Allen. Used to run the video simulation team)

springfall2008:
> (Hi Trefor - don't think we ever overlapped at IMG - I left in 2012 - Rik Allen. Used to run the video simulation team)

I joined in Jan 2012 so likely we had a short overlap, I'm still in touch with Mark Rigby who you likely know

PR #4267, opened 2026-07-17
fix: fetch rates when currently in an active planned intelligent disp…

### LINKED ISSUE #4261: Kraken documentation updates
Thanks for the great work on this. I've just moved from IOG to EDF GoElectric and have managed to migrate most things across in Predbat. There are a couple of documentation changes that would make this simpler for others in future:

- I couldn't way to generate an API key for EDF so just used username and password. Make it clear that these are the same as the EDF username and password.
- Kraken also generates a dispatches entity within Predbat that allows Predbat to understand when charge slots are. I used octopus_intelligent_slot: 'binary_sensor.predbat_kraken_a_d081455d_intelligent_dispatch_000000001c3c', but octopus_intelligent_slot: 're:(binary_sensor.octopus_energy([0-9a-z_]+|)_intelligent_dispatching)' also works on testing
- Keep switch.predbat_octopus_intelligent_charging turned on

The only problem I've found at the moment is that Predbat does not update the import rate during the charge slots, remaining at the higher rate.

Easen:
…atch

attempts to fix #4261 (https://github.com/springfall2008/batpred/issues/4261)

### DIFF
--- a/apps/predbat/kraken.py
+++ b/apps/predbat/kraken.py
@@ -1234,7 +1234,7 @@ class KrakenAPI(ComponentBase, _AUTH_BASE):
             new_completed = self._normalize_dispatches(data.get("completedDispatches"), completed=True)
             device["completed_dispatches"] = self._merge_completed_dispatches(device.get("completed_dispatches", []), new_completed)
 
-    def _publish_dispatch_sensors(self):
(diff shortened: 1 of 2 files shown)


springfall2008:
This doesn't seem like it would fix the original issue, fetching the rates again during import won't change them

Issue #4248, opened 2026-07-14
Forthcoming changes to octopus intelligent go


chalfontchubby:
As a heads up...
According to https://octopus.energy/blog/intelligent-octopus-go-smarter-charging-for-a-greener-grid/ there are changes coming to octopus that will impact those on the tariff and include car charger load in their predate visible ct clamp.

It implies the electricity meter on its own is insufficient to determine what power is going to the car/battery since you may have overnight cheap usage for the house while car consumption is expensive. 

My own setup has the ct clamp positioned to exclude car power so should be unaffected. 

---------------------------------------

We’ve split out your car charging and your home use, creating a dynamic 4-rate system. It sounds complicated, but it boils down to four simple rules:
Your home off-peak: Your house always gets the cheap off-peak rate between 23:30 and 05:30. Reliable, consistent, cheap green electrons every day.

Your car allowance: Your car gets up to 6 hours of off-peak charging per day (measured midday to midday). We might schedule this during your home's off-peak window, or at other times of the day when the grid is greenest.

Your freebie (the bonus home rate): If we schedule your car to charge outside of the 23:30–05:30 window, your home also gets the off-peak rate for that half-hour when it’s within your car allowance.

Your standard rate: The default for your home, outside its off-peak window. If your car charges for more than 6 hours (or you decide to Boost charge), the car rate flips back to this rate (even if it happens at 2:00 AM while your house is still getting the off-peak rate).

jHamshaw:
Also, Octopus have confirmed that if a smart slot is issued during daytime hours (i.e. outside of the 23:30 - 05:30 house night fixed cheap rate range) and the car doesn't charge (for whatever reason - maybe full from yesterday and still plugged in or even just a charge fail) the those IOG slots are "rescinded" from the house cheap rate during the day. If the EV charges at all (even just for a few minutes) the FAQ's indicate whole 30-minute slot remains at cheap rate for the house (which makes sense as that's house the house billing at least, now (still) works (presumably with some side-calculation based on EV or Charger consumption as controlled and measured by IOG? This still seems to have a 24-48 hour lag in the Octopus app as happened pre charge-cap).

Predbat may have already jumped on that intelligent slot and started charging the house battery in a slot that then reverts to high rate after a few minutes if the EV doesn't charge. Should an "EV has actually started charging" check be applied before the inverter command is sent to charge the house battery?

chalfontchubby:
[...]

PR #4270, opened 2026-07-18
fix(octopus): don't commit the house battery to an unconfirmed Intell…

chalfontchubby:
> 🤖 Drafted with Claude Code assistance.

Related to #4248 and the discussion there (in particular @jHamshaw's report of the house battery discharging into a slot that later gets rescinded).

## The problem

Octopus Intelligent Go can issue a daytime "smart dispatch" slot (outside the fixed 23:30-05:30 off-peak window). That slot is still Octopus's own provisional/revisable plan - if the car never actually draws power during it, Octopus retroactively rescinds the cheap rate. `rate_add_io_slots` previously wrote the low rate into `rate_import` for every dispatch slot, planned or completed, with no distinction - so the house battery could commit to charging on a rate that later reverts to full price.

## The fix

Adds a new expert-mode switch, **`octopus_intelligent_confirm_slots`** (default **Off**): when enabled, an out-of-window slot is only treated as low-rate once it's confirmed - either as a completed (metered) dispatch, or via `car_charging_now` confirming real-time draw. The fixed overnight window is always trusted regardless, since it's guaranteed cheap by the tariff itself, not the dispatch mechanism.

This composes cleanly with the existing `octopus_slot_max` 6-hour cap (`octopus.py:2649`, `if confirmed and slots_per_day[day_offset] < octopus_slot_max:`) - both are independent, ANDed conditions, no conflict.

**Why default Off, with a warning rather than defaulting On:** investigation showed `car_charging_now` is really the *only* confirmation signal fast enough to matter for a live slot - it ships commented out by default, existing docs discourage setting it for most users, and Octopus's "completed dispatch" confirmation may not arrive until after the slot has ended. Given how sweeping the consequence is without `car_charging_now` (daytime slots become *permanently* untrusted, not just occasionally), this seemed too large a behaviour change to ship as a silent default-On. Instead: default Off, plus a startup warning if the switch is on, Octopus Intelligent Charging is enabled, and `car_charging_now` isn't set - so turning it on without the signal it needs doesn't fail silently.

**Why not infer it from load instead, and skip needing `car_charging_now` at all?** Predbat's `dynamic_load` mechanism has a car-slot cancellation check that looks like a candidate, but it isn't one on closer inspection: `load_car_slot` isn't derived from load telemetry, it's just "are we inside a `car_charging_slots` window" - and `car_charging_slots` is itself built directly from the same provisional Octopus dispatch data being confirmed here, so using it as confirmation would be circular. The only thing it does in response to observed load is one-directional and negative (it can cancel an assumed slot if load stays surprisingly low; there's no symmetric path that positively confirms one), and it's gated behind `metric_dynamic_load_adjust`, which defaults Off anyway.

Beyond the existing mechanism, a purpose-built load-spike heuristic wouldn't fare much better here. A single-phase EV charger's draw overlaps in both magnitude and duration with common household loads (oven, electric shower, immersion heater, tumble dryer) - disambiguating "the car is charging" from "someone's cooking" out of one aggregate whole-house signal is a real signal-disaggregation problem (NILM), not something a threshold check reliably solves. It's also worst exactly when this fix matters most - daytime, when other household loads are most active. And real charging current isn't flat: cold-battery preconditioning, BMS tapering near full SOC, and load-balancing EVSEs can all put genuine charging well under a fixed threshold, so no single cutoff avoids both false positives (from other appliances) and false negatives (from a slow-charging car) at once.

## Also included

- Docs (`docs/car-charging.md`) updated to state plainly that `car_charging_now` is effectively required for this protection to do anything on daytime slots, regardless of `car_energy_reported_load`.
- A worked `car_charging_now` example (including that it can point at a charger-level sensor rather than the car's own API - useful for chargers Octopus controls directly, e.g. Hypervolt) added to the main `apps.yaml` template and all 13 inverter templates carrying the same block.
- Full test coverage for the new confirmation logic (`test_rate_add_io_slots.py`) and the startup warning (`test_fetch_config_options.py`).

## Verification

- Full local suite passes (591/591 non-flaky tests; the one known-flaky `debug_cases` case is unrelated, see prior PR discussion).
- `run_pre_commit` passes cleanly.
- Manually verified `octopus_slot_max` and `octopus_intelligent_confirm_slots` compose as independent AND conditions with no interference.


springfall2008:
I guess the only issue with this is many people don't have car_charging_now set up, I wonder if we could do it based on load?

chalfontchubby:
If car load is included in predbats visible that might be viable. But it would involve some assumptions about the load on the house that might be prone to risk. Car_charging_now is kinda useful in this case - don't know if there is something (quickly) visible in the octopus integration that could confirm the slot has been used. 

For all of this, I don't want to get in the way, and only want to help if it's useful. Happy to be "managed" or told to back off as needs be.

jHamshaw:
> I guess the only issue with this is many people don't have car_charging_now set up, I wonder if we could do it based on load?

Mine is set up like this (Zappi integrated with IOG, 2 x EV's) : 

```
  car_charging_now:
  - sensor.myenergi_zappi_xxxxxxxx_status
  - sensor.myenergi_zappi_xxxxxxxx_status

  car_charging_now_response:
  - "Charging"
  - "Boosting"
  - "SmartBoost"
```

chalfontchubby:
Circling back on the load-based idea from a couple weeks ago - on reflection I don't think it's reliable enough to gate a decision like this on. The risk is a false positive: Predbat commits the house battery to a large import on the assumption the car will soak it up, and if the car doesn't actually pick up charging (session cancelled, already topped up, charger fault, whatever) that import still happens - potentially at a premium rate. `car_charging_now` at least confirms the car is genuinely pulling current right now; a load-based signal without that confirmation is a guess dressed up as a fact.

@jHamshaw thanks for the concrete config - that's a good example of `car_charging_now` being a normal, achievable setup (Zappi/IOG here), which for me strengthens the case for keeping it as the required signal rather than building a load-based workaround for people who haven't set it up.

Happy to support a load-based fallback as an explicit, clearly-warned opt-in for people who can't configure `car_charging_now`, but I'd want it defaulted off and documented as the riskier path - not something we quietly fall back to.

Fred224:
My Ohme has an ischarging sensor, and a power sensor on the wire is an easy addition which gives an clear signal between idle (or a few watts) and charging 100's+ Watts, with nothing in between.

PR #4275, opened 2026-07-20
Suppress LoadML component errors while forecasts are calculating

nickgee31:
Prevents Predbat’s main status from reporting ML Load Forecaster as an error while LoadML is actively calculating.

The underlying run status is preserved, so Predbat continues to show the appropriate state such as Demand, Export, or Idle. LoadML failures outside calculation—and failures from other components—still report errors normally.

Adds regression tests covering both calculating and genuine LoadML failure states.


apps/predbat/predbat.py
```diff
@@ -763,7 +763,10 @@ def record_final_run_status(self, status, status_extra):
                     component_status[component_name] = "error"
                     all_healthy = False
                     error_count += 1
-                    failed_components.append(COMPONENT_LIST.get(component_name, {}).get("name", component_name))
+                    component = self.components.get_component(component_name)
+                    is_load_ml_calculating = component_name == "load_ml" and component and component.is_calculating()
```
springfall2008: Could we do this in a less hard-wired way, e.g. add a is_calculating to the base class that returns False and then override it for load_ml to return True when calculating and then remove the =="load_ml" here?

nickgee31: Yeah good idea, have added that now

PR #4282, opened 2026-07-22
fix(octopus): don't re-mint tokens on a CDN 403, and cache the EV catalogue

mgazza:
## Problem

A live customer's Octopus Direct component locked itself out for 9+ hours and could not recover without a restart.

Octopus's CloudFront edge began returning `403 Request blocked` — an HTML error page, not a GraphQL error. `async_graphql_query` treats any 401/403 as a revoked JWT, so it discarded a still-valid Kraken token and immediately re-minted. But `obtainKrakenToken` sits behind the same block, so the mint failed too — and every subsequent query repeated the cycle, ~42 times an hour, indefinitely. The component could never self-heal.

Diagnosis from inside the affected pod:

- A **bogus** API key returns a normal `KT-CT-1139` GraphQL error at **HTTP 200** — so the key was valid and the egress IP was not banned.
- A repeated probe returned GraphQL `"Too many requests"` — the block is rate limiting, scoped to the account, not a credential problem.
- Both IPv4 and IPv6 egress returned 200, ruling out a blanket hosting-provider IP block.

Meanwhile `sensor.*_import_rates` kept updating from the public REST rate endpoint, which masked the failure: `components.is_alive()` only fails a component once its last success is >60 min old, and the REST path still refreshed that timestamp. The run flapped between healthy and `component errors: Octopus Energy Direct` (51 failed runs vs 34 clean in 12h).

## Changes

**1. Don't treat a CDN block as an auth failure.** A 403 whose body carries a CDN/WAF signature is now treated as rate limiting: keep the cached token and return `None`. Detection is deliberately *positive* — a 403 we cannot identify as a CDN page keeps the existing refresh-and-retry behaviour, so genuinely revoked tokens still recover without a restart (preserving the fix from the earlier intelligent-slots regression).

**2. Cache the global EV/charge-point catalogue.** `electricVehicles` + `chargePointVariants` are account-independent static reference data — ~136KB and 188 manufacturers — but were bundled into the per-account device query and re-downloaded every few minutes per customer (~52MB/day each) purely to look up one car's battery size. They are now a separate query cached for 24h via `storage.fetch_cached`, under the `octopus` module so a shared storage backend can serve one fetch to many instances. This is also the most likely trigger for the rate limiting in the first place.

**3. Stop re-reading a consumed 401/403 response.** `async_read_response_retry` looped `OCTOPUS_MAX_RETRIES` times over the *same* response object. aiohttp caches the body, so the retries could never change the outcome — they only emitted five identical warnings and slept ~15s per query. In the incident this produced 1,395 duplicate log lines.

## Note for reviewers

`test_octopus_read_response_retry` Test 8 asserted that a 401 response is re-read and succeeds on the second attempt, mocked as `side_effect=[None, successful_data]`. Because aiohttp caches the response body, that scenario cannot occur in production — the test was asserting mock behaviour. It has been updated to assert the real contract (single read, no sleep, `failures_total` incremented).

## Testing

- New `test_octopus_waf_block`: CDN 403 keeps the token and does not re-mint; a genuine 401 still refreshes; an HTTP error is read exactly once.
- New `test_octopus_catalogue_cache`: catalogue fetched once across repeated polls, reused by a second instance sharing the storage backend, and battery size still resolves from cache.
- Full suite: `unit_test.py --quick` — 148 tests pass. pre-commit (ruff, black, cspell) passes.


mgazza:
### Code review

Ran a multi-agent review over this branch. Five issues met the confidence bar; all are now fixed in `c264252b` and `23e78616`. Recording them here since three were variants of the same class of bug this PR set out to fix.

1. **The 403 carve-out could recreate the permanent lockout.** Detection matched any HTML 403, so a generic `403 Forbidden` gateway page from a genuinely revoked credential would have been classified as a CDN block and never refreshed — the failure [5928e389](https://github.com/springfall2008/batpred/commit/5928e389) was written to fix. Now bounded by an escape hatch: after `EDGE_BLOCK_REFRESH_AFTER` consecutive edge blocks the token is refreshed anyway, and the counter resets on any successful query.

https://github.com/springfall2008/batpred/blob/23e78616caea60f755f7622d1add9df88782b317/apps/predbat/octopus.py#L85-L92

2. **`ignore_errors` was not honoured in the new branch**, unlike every other error path in `async_graphql_query`. `async_get_vehicle_catalogue` and `async_get_intelligent_devices` both pass `ignore_errors=True`, so a blocked optional lookup was inflating `failures_total` and logging warnings.

https://github.com/springfall2008/batpred/blob/23e78616caea60f755f7622d1add9df88782b317/apps/predbat/octopus.py#L1666-L1676

3. **Matching on wording alone was too broad.** `"access denied"` is an ordinary way to phrase an authorisation error, so a genuine Kraken JSON 403 saying it would have been misclassified — keeping an invalid token forever. Detection now requires the body to fail JSON parsing *and* look like an HTML document or name a known CDN.

https://github.com/springfall2008/batpred/blob/23e78616caea60f755f7622d1add9df88782b317/apps/predbat/octopus.py#L91-L120

4. **Test pollution.** The new catalogue test injected a components mock into `my_predbat` without restoring it, and the runner shares one `my_predbat` across the whole registry. This is the same pattern raised on #4043; now saved and restored in a `finally`.

https://github.com/springfall2008/batpred/blob/23e78616caea60f755f7622d1add9df88782b317/apps/predbat/tests/test_octopus_catalogue_cache.py#L104-L108

5. **Stale comment.** The pre-existing note above the 401/403 block asserted that every bare 401/403 forces a refresh, which is no longer true for the CDN subset. Qualified rather than left to mislead.

https://github.com/springfall2008/batpred/blob/23e78616caea60f755f7622d1add9df88782b317/apps/predbat/octopus.py#L1654-L1659

Noted but deliberately not changed here, as pre-existing or out of scope:

- The recursive retry is awaited while still inside `async with client.post(...) as response`, holding a connection slot — raised on #3833 and unchanged since; this PR does not alter that structure.
- The full GraphQL response body is logged unconditionally ([raised twice on #4115](https://github.com/springfall2008/batpred/pull/4115)). Not addressed here, though this PR reduces how often the ~136KB catalogue reaches that line from roughly 16 times an hour to once a day.
- `test_octopus_cache.py` has the same shared-`my_predbat` mutation as issue 4; left alone as it is outside this change.
- Known trade-off: if the catalogue fetch fails on a cold cache while the per-account device query succeeds, `vehicle_battery_size_in_kwh` is silently absent for that poll rather than failing the poll. Stale-while-revalidate covers the warm case. Happy to make it fail loudly if preferred.

Full suite (`unit_test.py --quick`, 148 tests) and pre-commit pass at each commit.

Issue #4179, opened 2026-07-03
inverter.py update_status(): bare `raise ValueError` surfaces as unhelpful "Error: Exception raised"


mgazza:
## Summary

`Inverter.update_status()` raises a **bare `raise ValueError`** (no message) when it can't determine the charge window because there's no rate/window source configured. Because the exception carries no context, it bubbles up to the main loop and surfaces to the user as the unhelpful status **`Error: Exception raised`**, with a raw traceback in the logs instead of an actionable message.

## Where

`apps/predbat/inverter.py` ~line 1458, in `update_status()`:

```python
if self.rest_data:
    charge_start_time = time_string_to_stamp(self.rest_data["Timeslots"]["Charge_start_time_slot_1"])
    charge_end_time   = time_string_to_stamp(self.rest_data["Timeslots"]["Charge_end_time_slot_1"])
elif "charge_start_time" in self.base.args:
    charge_start_time = time_string_to_stamp(self.base.get_arg("charge_start_time", index=self.id))
    charge_end_time   = time_string_to_stamp(self.base.get_arg("charge_end_time", index=self.id))
else:
    self.log("Error: Inverter {} unable to read charge window time as neither REST, charge_start_time or charge_start_hour are set".format(self.id))
    self.base.record_status("Error: Inverter {} unable to read charge window time ...".format(self.id), had_errors=True)
    raise ValueError            # <-- bare, no message
```

## Why it matters

- The status entity ends up as `Error: Exception raised` (the outer handler's generic message), so users/support can't see *why* it failed without digging into the pod's traceback.
- It's **inconsistent with the sibling branch just below** (~1460–1466), which handles a missing/`None` charge window **gracefully** — logs a `Warn`, sets safe defaults, and retries next update rather than raising:

```python
if charge_start_time is None or charge_end_time is None:
    self.log("Warn: Inverter {} unable to read charge window time ... will retry next update")
    ...
    self.charge_enable_time = False
    self.charge_start_time_minutes = self.base.forecast_minutes
    ...
```

Observed in production on multiple instances whose upstream inverter integration returned no device/registers (e.g. GivEnergy cloud "no devices", Fox returning no data): the useful `Error: Inverter 0 unable to read charge window time ...` message is logged, but the subsequent bare `ValueError` overwrites the user-visible status with `Error: Exception raised`.

## Suggested fix (either)

1. Raise with the same descriptive message already logged:
[...]

Issue #4153, opened 2026-06-30
Graph legends show as p/kWh, when should be c/kWh


travellingkiwi:
**Describe the bug**
The rates graph (And others) show p/kwh instead of c/kWh (i.e. wrong case for Watts, p instead of c - I'm in NZ$)

**Expected behaviour**
The graphs should have the correct text. i.e. c vs p (cents vs pence) 

**Predbat version**
8.42.5

**Environment details**
- Fronius Gen24+ Primo with BYD HVM Battery (13.7kWh)
- Standard HAOS installer
- NZ$, configured as $ and c
- 

**Screenshots**

<img width="1156" height="159" alt="Image" src="https://github.com/user-attachments/assets/c41a5e38-61b2-40dc-8773-842f6af18cfb" />

**Log file**
**Predbat debug yaml file**

gcoan:
sounds like a display bug, but please provide your apps.yaml config, or at least the currency config

travellingkiwi:
I've not had a chance yet to inspect the yaml. But my currency config is

<img width="991" height="105" alt="Image" src="https://github.com/user-attachments/assets/6cb28f45-4aaf-4393-8410-957858ee078b" />

travellingkiwi:
I wonder if it's related to #4104

travellingkiwi:
A quick clone and grep found a few places (output.py, web.py, and config.py) where the text is hard coded as 'p/kWh'

<img width="1296" height="118" alt="Image" src="https://github.com/user-attachments/assets/34ce42a6-420e-48fe-b4e5-e97cd93a15bb" />

and 

<img width="627" height="163" alt="Image" src="https://github.com/user-attachments/assets/b1337743-4ae4-4abd-9327-37007378b643" />

octopus has a few as well (But I don't use octopus, I use ecotricity) so less visible to me. 

I'm ignoring comments and docs, Mainly because I can't think what else you'd put in those... "minor_currency_unit" is a bit more awkward to type...

:)

gcoan:
web.py is the place that the charts are rendered.  I guessed the units was probably hard coded, which looks like it is

the entries in config.py are the predbat definition of Home Assistant entities.  Changing these could be more problematic, but the web render is a straight forward fix, just needs to pull the currency minor unit out rather than hard code

chalfontchubby:
> 🤖 Claude Code - fix in flight.

[...]

Issue #4151, opened 2026-06-30
Metrics battery SoC does not update when page left displayed


travellingkiwi:
**Describe the bug**
On the metrics tab, the battery SoC is displayed and at the time of opening the metrics tab, the SoC is correct and agrees with the battery SoC in the menu bar

<img width="1411" height="796" alt="Image" src="https://github.com/user-attachments/assets/5accea61-ed16-4ec1-b39d-5561a378ca82" />

However if the metrics are left displayed, all the sensors update, except the battery SoC. It is left on the original values

<img width="1675" height="893" alt="Image" src="https://github.com/user-attachments/assets/41eb1289-8fe9-4324-b934-68a8441412cc" />

**Expected behaviour**
The battery SoC on the metrics page should update in conjunction with all the other sensors being updated

**Predbat version**
8.42.1, and 8.42.5

**Environment details**
- Fronius Gen24+ 6kW with 13.8kWh of BYD HV battery in a single stack 
- Standard HAOS installer

**Screenshots**

<img width="1675" height="893" alt="Image" src="https://github.com/user-attachments/assets/54b9b247-66d0-4f7f-817d-4f5dcfb7de0a" />

**Log file**
Can you capture a log file from the time of the issue, debug mode is not normally required.

You can download the logfile from the WebUI on the Dash tab select **predbat.log**

**Predbat debug yaml file**
This is important for any plan related issues.

Once you have captured the issue go to the Web UI, in the Dash tab click on **predbat_debug.yaml** and wait 30 seconds for it to download. Select 'keep' if your Web Browser thinks the file is dangerous.

Rename the download file to predbat_debug.yaml.txt and upload it to GitHub. This will allow your plan to be reproduced and also stores all your settings for review.

chalfontchubby:
> 🤖 Claude Code - fix in flight.

Found it: the SoC doughnut chart's center text is drawn by a Chart.js plugin that closed over the values from when the chart was first created, so it never updated on subsequent refreshes even though the underlying chart data (and every other card on the page) did. Fixed in #4288 (bundled with a couple of other small fixes - happy to split it out if preferred).

chalfontchubby:
Update: split this out into its own PR - #4308 (was previously bundled in #4288 along with two unrelated fixes).

PR #4288, opened 2026-07-23
fix(inverter): Don't report fatal error for REST failing

chalfontchubby:
## Summary

A small batch of independent, simple fixes - each is its own commit.

### #4153 - Rates chart legend hardcoded to "p/kWh"

The Rates chart's "Hourly"/"Today" series legend names were hardcoded to `p/kWh` (pence) regardless of the user's configured `currency_symbols`, so non-GBP users (e.g. the reporter's NZD setup, configured as `$c`) saw `p/kWh` in the chart legend even though the surrounding chart title/axis already correctly used `self.currency_symbols[1]`.

Per the maintainer's own scoping comment on the issue: fix limited to `web.py`'s chart rendering. Left `config.py`'s CONFIG_ITEMS units and `output.py`'s HA entity attribute names untouched, since changing those affects existing HA entity definitions/history for all users.

- `apps/predbat/web.py`: series names now use `"{}/kWh".format(self.currency_symbols[1])`, matching the pattern already used elsewhere in the same file.
- `apps/predbat/tests/test_web_chart_currency.py`: new test verifying the formatting for `£p`/`$c`/`€c`.

### #4179 - bare `raise ValueError` surfaces as unhelpful "Error: Exception raised"

`Inverter.update_status()` raised a bare `raise ValueError` (no message) when neither REST data nor a `charge_start_time`/`charge_end_time` config value was available - the exception carried no context, bubbling up as the unhelpful status "Error: Exception raised" and hard-crashing plan calculation, inconsistent with the very next branch which handles a configured-but-unusable value gracefully (safe defaults + retry next update).

Matches the issue's suggested "Option 2": the branch now sets `charge_start_time`/`charge_end_time` to `None` and falls through into the existing graceful handling below it, so a temporarily-empty inverter feed (the actual production trigger reported - GivEnergy cloud "no devices", Fox returning no data) no longer crashes the whole plan. Added `test_charge_window_no_source_configured` alongside the existing sibling tests for the related None-value cases.

(A follow-up commit adding a setup-time warning for "no source configured at all" was tried and reverted - it would have produced false positives for Ginlong Solis-type inverters, which use `charge_start_hour`/`charge_start_minute` rather than `charge_start_time` and have no REST API. That inconsistency already exists in the pre-existing error message text and looks like it needs separate, more careful investigation rather than bundling into this fix.)

### #4151 - Metrics dashboard battery SoC doesn't update when page left displayed

The battery SoC doughnut chart's center text (percentage and kWh figures) is drawn by a Chart.js `afterDraw` plugin registered once in `mdInitSOCChart`. That plugin closed over the `pct`/`d` values from the moment the chart was first created. On each 30s refresh, `mdUpdateSOCChart` only mutates the existing chart's data array and calls `.update()` - it never recreates the chart - so the doughnut ring's fill proportion updated correctly but the on-screen number text stayed frozen at the first-load reading, while every other card on the page re-renders fresh from the latest data on every refresh.

- `apps/predbat/web_metrics_dashboard.py`: the `afterDraw` plugin now reads `chart.data.datasets[0].data[0]` for the percentage and the `MD_DATA` global for the kWh figures, instead of the stale closed-over values.
- `apps/predbat/tests/test_metrics_dashboard_soc_refresh.py`: no JS execution/browser test infra exists in this codebase, so this checks the generated JS source itself reads from `chart.data`/`MD_DATA` rather than the closure parameters, to catch a regression back to the same bug shape.

## Test plan

- [x] `./run_all --test web_chart_currency` / `./run_all --test inverter` / `./run_all --test metrics_dashboard_soc_refresh` - new tests pass
- [x] `./run_all --quick` - full quick suite passes (606 tests, 0 failures)
- [x] `./run_pre_commit` - clean (ruff, black, cspell, markdownlint) after each commit

🤖 Implemented with Claude Code, disclosed per repo convention.


chalfontchubby:
Note: I bundled three small, independent fixes into this one PR since none of them touch overlapping code. Happy to split any of these out into their own PR if that's preferred for review - just let me know.

apps/predbat/inverter.py
```diff
@@ -1468,9 +1468,13 @@ def update_status(self, minutes_now, quiet=False):
                 charge_start_time = time_string_to_stamp(self.base.get_arg("charge_start_time", index=self.id))
                 charge_end_time = time_string_to_stamp(self.base.get_arg("charge_end_time", index=self.id))
             else:
+                # Neither a data source nor a fallback config value is available - log the specific
```
springfall2008: I'm not sure of the rationale here, this is detecting when the inverter has not been configured (charge_start_time is not set at all in the args). The only time this is temporary is when you depend on auto-config of a cloud component, but why pretend to make a plan when you can't actually deliver it?

chalfontchubby: That's a fair challenge, and I think it points at something the fix as written glossed over: the `else` branch actually covers two different situations, not one.

1. **`self.rest_api` is configured, but `self.rest_data` is empty this cycle** - e.g. the original bug report's actual trigger, GivEnergy cloud returning "no devices" or Fox returning no data on a fresh start. This genuinely is transient - the data source is legitimate, it just hasn't successfully returned anything yet.
2. **Neither `self.rest_api` nor `charge_start_time` is configured at all** - this is the case you're pointing at, and you're right that it's not temporary. Retrying every cycle forever doesn't fix a config gap that isn't going to close itself.

The fix collapsed both into the same "safe defaults + retry" path, which is defensible for (1) but not really for (2).

Given that raising here crashes the *whole* plan calculation (not just this one inverter) - which was the actual harm in the original report - I don't think going back to a hard raise for case (2) is right either. But I take your point that silently retrying with a "will retry next update" message is misleading when there's nothing to wait for. Would something like: keep the graceful retry for case (1), but for case (2) still avoid crashing the plan while making the status persistently loud (an ongoing Error-level status, not a message implying transience) so it doesn't get lost in the log - does that match what you had in mind, or would you rather it fail harder for that specific case?

chalfontchubby:
I've split the #4153 and #4151 fixes out into their own PRs (#4307 and #4308) so they can be reviewed/merged independently of the discussion happening here on #4179. This PR's branch still contains all three commits, though.

Happy to force-push this branch down to just the #4179 commits, or revert the other two commits here, whichever's less disruptive on your end - or if it's easier to just leave it and let a rebase/squash-merge sort it out later, that works too. Your call.

apps/predbat/inverter.py
springfall2008: You are right, its case 2) I'm concerned about and here I don't think we should allow a run with no inverter configured as its kind of misleading for the user, hence the reason for the hard raise?

springfall2008:
@chalfontchubby this one probably needs a resolve and then a re-work?

chalfontchubby:
(Posting as Claude, an AI assistant helping on this PR)

Addressed all three open threads:

**inverter.py (`charge_start_time` not configured)** - split the two cases as discussed: a configured REST source (`givtcp_rest`) that hasn't returned data yet this cycle still falls through to the existing graceful retry path (genuinely transient), but when neither a REST API nor `charge_start_time` is configured at all, `update_status()` now raises a `ValueError` with a proper message and records the error status first - restoring the hard-raise you asked for, just with actual context this time instead of the original bare exception. Added a second test alongside the existing one so both cases are covered explicitly.

**`test_web_chart_currency.py` (both Copilot findings)** - these were already fixed on this branch (commit 227f5731, before this latest push) - `currency_symbols`/`dashboard_values` are saved/restored via `try`/`finally`, and the test now calls the real `WebInterface.get_chart("Rates")` rather than re-testing the inline formatting logic. Verified this actually catches a regression (temporarily reintroducing the hardcoded `p/kWh` string in `web.py` makes it fail). Resolving these two threads since the code already addresses them.

Full quick test suite + pre-commit both clean on the latest push.

PR #4340, opened 2026-07-28
fix(deye): repair SaaS OAuth and correct telemetry keys against a live inverter

mgazza:
Found while onboarding the first real DEYE site. The component has never worked end-to-end against a live inverter on the hosted service — five defects, each masked by the one before it.

### 1. The access token was never delivered
`COMPONENT_LIST["deye"]` had no `key` arg, so the injected token was dropped, and `initialize()` passed `key=app_secret or token_hash` into `_init_oauth()`. `OAuthMixin` assigns `key` directly to `access_token` in oauth mode, so DEYE was handed the refresh **dedup hash** as a bearer token and rejected every call with `auth invalid token`.

`required_or` now gates on `key` rather than `token_hash`: the hash alone is not a usable credential, so gating on it starts the component for an instance that cannot authenticate and then fails on every API call.

### 2. `fetch_token()` crashed instead of failing soft
All `app_credentials` args are optional (the component also accepts the SaaS token path), so a misconfigured instance reached `_sha256(None)` → `AttributeError` out of `run()`, on every 5-minute cycle. Both callers already treat `False` as "skip this run".

### 3. Every telemetry key was wrong
`DEYE_TELEMETRY_KEYS` was marked `VERIFY@SPIKE` and had never been checked against a live response. None of the six guesses matched:

| Was | Actually |
|---|---|
| `batterySOC` | `SOC` |
| `batteryPower` | `BatteryPower` |
| `gridPower` | `TotalGridPower` |
| `pvPower` | `TotalSolarPower` |
| `loadPower` | `TotalConsumptionPower` |
| `batteryTemperature` | `Temperature- Battery` |

This didn't fail, which is why it went unnoticed: `_as_float()` coerces an absent key to `0.0`, so the component published a complete and plausible set of zeros — SOC 0%, battery temperature 0°C — indistinguishable from a flat battery on a cold night.

### 4. No cumulative energy, so every Predbat run aborted
`fetch_sensor_data()` raises `ValueError` outright when `load_today` is unset, killing the whole run. The daily counters were in the same `device/latest` response all along — `DailyConsumption`, `DailyEnergyPurchased`, `DailyGridFeedIn`, `DailyActiveProduction` — now wired into `automatic_config()` the way `fox.py` and `solis.py` already do.

### 5. `battCapacity` is amp-hours, not kWh
Corroborated by the `dataList` entry `BatteryRatedCapacity` carrying unit `Ah`. `soc_max` was advertising a 1200 kWh battery; now converted against nominal pack voltage. Note the resting pack voltage (~58 V at 99% SOC) must not be used for this — it rises with SOC.

### Preventing the next one
`_report_unmatched_telemetry()` logs the keys DEYE actually returned when none match, once per serial, so a future rename surfaces instead of silently zeroing.

### Verification
Key spellings and the `Ah` unit are confirmed against a **live SUN-8K three-phase hybrid** (88 `dataList` entries), not inferred from the vendor samples. New tests pin the confirmed spellings, the Ah→kWh conversion, the oauth-token wiring, the soft-fail, and the unmatched-key report; each was checked to fail against the pre-fix code. `run_all --quick` passes.

One existing assertion in `test_deye_publish.py` changed: it expected `battCapacity` to publish unconverted, i.e. it encoded the bug.

### DIFF
--- a/apps/predbat/components.py
+++ b/apps/predbat/components.py
@@ -264,6 +264,9 @@ COMPONENT_LIST = {
             "inverter_sn": {"required": False, "config": "deye_inverter_sn"},
             "automatic": {"required": False, "default": False, "config": "deye_automatic"},
             "automatic_ignore_pv": {"required": False, "default": False, "config": "deye_automatic_ignore_pv"},
+            # config/battery reports capacity in Ah; the default suits 48 V low-voltage
+            # hybrids. High-voltage stacks must override this or soc_max is several-fold out.
+            "battery_nominal_voltage": {"required": False, "config": "deye_battery_nominal_voltage"},
         },
         # Gate activation on having at least one auth path — app credentials (app_id,
         # self-hosted add-on) OR an injected SaaS access token (key). Without this the
--- a/apps/predbat/components.py
+++ b/apps/predbat/components.py
@@ -250,6 +250,10 @@ COMPONENT_LIST = {
         "args": {
             "app_id": {"required": False, "config": "deye_app_id"},
             "app_secret": {"required": False, "config": "deye_app_secret"},
+            # In oauth mode OAuthMixin treats 'key' as the access token (see
+            # oauth_mixin._init_oauth). Predbat.com injects it as deye_key; without this
+            # entry the token is dropped and DEYE rejects every call as "auth invalid token".
+            "key": {"required": False, "config": "deye_key"},
             "username": {"required": False, "config": "deye_username"},
             "password": {"required": False, "config": "deye_password"},
             "data_center": {"required": False, "default": "eu", "config": "deye_data_center"},
@@ -262,10 +266,12 @@ COMPONENT_LIST = {
             "automatic_ignore_pv": {"required": False, "default": False, "config": "deye_automatic_ignore_pv"},
         },
         # Gate activation on having at least one auth path — app credentials (app_id,
-        # self-hosted add-on) OR an injected SaaS token (token_hash). Without this the
+        # self-hosted add-on) OR an injected SaaS access token (key). Without this the
         # component would start for every instance since all individual args are
-        # optional to allow either auth mode.
-        "required_or": ["app_id", "token_hash"],
+        # optional to allow either auth mode. Gate on key, not token_hash: the hash is
+        # only a refresh dedup handle, so gating on it starts the component for an
+        # instance that has no usable token and fails on every API call instead.
+        "required_or": ["app_id", "key"],
         "phase": 1,
     },
     "enphase": {
--- a/apps/predbat/deye.py
+++ b/apps/predbat/deye.py
@@ -48,7 +48,22 @@ class DeyeAPI(ComponentBase, OAuthMixin):
     """DEYE Cloud API component."""
 
     def initialize(
-        self, app_id="", app_secret="", username="", password="", key="", data_center="eu", company_id="", auth_method="app_credentials", token_expires_at=None, token_hash="", inverter_sn=None, automatic=False, automatic_ignore_pv=False, **kwargs
+        self,
+        app_id="",
+        app_secret="",
+        username="",
+        password="",
+        key="",
+        data_center="eu",
+        company_id="",
+        auth_method="app_credentials",
+        token_expires_at=None,
+        token_hash="",
+        inverter_sn=None,
+        automatic=False,
+        automatic_ignore_pv=False,
+        battery_nominal_voltage=None,
+        **kwargs,
     ):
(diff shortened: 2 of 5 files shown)


mgazza:
Follow-up commit addressing review feedback (65c6cf3). Three findings, all reproduced against the previous commit before fixing:

**1. `auth_method` was still trusted from its default.** A host that injects an access token but doesn't also set `deye_auth_method` gets `app_credentials`, and `_init_oauth()` discards the key in that mode — so the component authenticates with nothing. Now inferred: an injected `key` with no `app_id` can only be oauth. A genuine add-on config (`app_id` set) is left alone, which is asserted.

Worth noting this was the *original* production failure on our hosted service, fixed there in the config projection. Inferring it here means the component no longer depends on every host remembering to set it.

**2. The fixed 51.2 V nominal only suits low-voltage 48 V hybrids** — the only hardware this was confirmed against. On a high-voltage stack it understates capacity several-fold: a 100 Ah / 300 V pack is ~30 kWh but published as **5.12 kWh**, and that value drives `soc_max`. Added `deye_battery_nominal_voltage`, plus a warning when the measured pack voltage contradicts the assumed nominal rather than silently publishing a wrong `soc_max`. `BatteryVoltage` is mapped solely for that check — it is the resting voltage and must never be used as the nominal.

**3. Daily-counter absence meant two different things and was handled as one.** A model that never reports a counter should publish no sensor. But a counter that *has* been seen and then disappears is the midnight rollover, where the true value is zero — and omitting it left the previously published total standing, so Predbat read yesterday's consumption as today's load until DEYE resumed reporting. Now tracked per serial and zeroed on rollover.

My original comment on that line ("omit rather than zero, a spurious 0.0 would look like a real reading") had it backwards for the rollover case specifically: at midnight 0.0 *is* the true reading, and the stale total is the misleading one.

`run_all --quick` still green.

springfall2008:
Thanks @mgazza — this is a solid piece of diagnosis, and the token-wiring find in particular is the one that mattered most.

We hit the same site from the other end in #4343 and the two PRs overlap heavily (both touch `deye.py`, `deye_const.py` and `test_deye_publish.py` on the same lines), so rather than untangle a three-way merge I have folded your fixes into #4343 and am closing this one. Your commits are credited in the combined commit message.

**Carried over from here in full:**

1. **The access token was never delivered.** `COMPONENT_LIST["deye"]` had no `key` arg and `initialize()` passed `key=app_secret or token_hash` into `_init_oauth()`, so DEYE was handed the refresh dedup hash as its bearer token. This is the actual root cause of `auth invalid token` — #4343 had only fixed the *recovery* path (DEYE reports a bad token as HTTP 200 + `success:false`, so the 401 refresh branch was unreachable), which masked why the token was bad in the first place. Both are needed: a token can also expire mid-life and DEYE reports that the same way.
2. **`required_or` gating on `key` rather than `token_hash`.**
3. **`fetch_token()` failing soft** instead of `_sha256(None)` → `AttributeError` out of `run()`.
4. **Your warning that resting pack voltage must not be used for the Ah→kWh conversion**, plus `deye_battery_nominal_voltage`. #4343 had exactly the bug you called out — it inferred cell count from resting voltage as a fallback, which reads 18 cells instead of 16 at 100% SOC and overstates capacity by 12.5%. The live site hit 100% SOC, so this was reachable.

One place we went further than your version: rather than defaulting to 51.2 V when the BMS charge target is missing, there is now **no default at all**. `soc_max` feeds `nominal_capacity`, which sets the 0.65–1.20 plausibility band filtering `find_battery_size()` — so a wrong value does not merely mislead, it rejects every correct sample from Predbat's own estimator. Returning 0 leaves the band disabled and lets the estimator recover the real size from history. `deye_battery_nominal_voltage` is the explicit override. The nominal is otherwise derived from the BMS charge target (57.6 V → 16 cells → 51.2 V), so HV stacks scale without configuration.

**One deliberate difference, worth your view:** you mapped the `Daily*` energy counters, #4343 maps the lifetime `Total*` ones. Both work — `clean_incrementing_reverse()` exists specifically to remove midnight resets — so this is a judgement call rather than a correctness issue. `Total*` is monotonic and needs no reset handling, which is why it went that way. Happy to switch if you think the `_today` naming should be taken literally.

Your `raise ValueError` point stands either way and was the reason the counters got wired up at all.

#4343 also adds `battery_rate_max`/`inverter_limit`, resilience to a 502 from the oauth-refresh edge function, and API request/response tracing. Review there very welcome.

springfall2008:
Closing in favour of #4343, which now contains these fixes. See the comment above for what was carried over.

PR #4343, opened 2026-07-28
fix(deye): repair token refresh, telemetry keys and inverter config

springfall2008:
Fixes found while running the DEYE integration live against a 3-phase hybrid (productId `0_5411_1`) on Predbat.com. Each one was diagnosed from a real API trace; the values quoted below are from that device.

## Token refresh never fired

DEYE reports an expired/invalid bearer token as **HTTP 200** with `{"success": false, "msg": "auth invalid token"}`, not a 401 — so the refresh branch in `_post()` was unreachable and the component stayed broken until a restart. The observed log line `Warn: DEYE station/list failed: auth invalid token` is only reachable *after* `_post` returns a parsed body, which proves the status was never 401.

`_post()`/`_get()` now inspect the response body and refresh once before retrying, bounded so a server that keeps rejecting the new token cannot spin the loop.

## A failed proactive refresh no longer skips the whole cycle

A separate run failed with an HTTP 502 from the Supabase `oauth-refresh` edge function. `check_and_refresh_oauth_token()` returned False and `run()` aborted — even though the token it held had a September expiry and was almost certainly fine.

It now continues with the existing token and lets the reactive 401 path handle a genuinely dead one. `needs_reauth` and "no token at all" are still fatal. Transient 5xx and network errors are retried (3 attempts, 2s/4s backoff); a 4xx is a real rejection and is not.

This is in `oauth_mixin.py`, shared with fox/solis/teslemetry/kraken, so all of their suites were re-run.

## Every telemetry key was wrong

`DEYE_TELEMETRY_KEYS` held guessed camelCase names, but DEYE uses PascalCase display names — two with literal spaces and hyphens. All six lookups missed and published `0.0` while the API returned good data.

| Predbat | Was | Now | Live |
|---|---|---|---|
| soc | `batterySOC` | `SOC` | 94 % |
| battery_power | `batteryPower` | `BatteryPower` | -603 W |
| grid_power | `gridPower` | `TotalGridPower` | 25 W |
| pv_power | `pvPower` | `TotalSolarPower` | 1584 W |
| load_power | `loadPower` | `TotalConsumptionPower` | 820 W |
| temperature | `batteryTemperature` | `Temperature- Battery` | 27.9 °C |

A missing key is now warned about and the device's actual keys listed, rather than silently reading zero. The old test built its fixture *from* the constant, so it passed with any spelling — the new one uses a captured live payload.

**Grid sign:** DEYE reports grid power positive while importing (`DailyGridFeedIn` 0.00 confirms direction); Predbat wants negative for import, so it is negated on ingest. Battery power needs no flip — DEYE is already negative while charging.

## `battCapacity` is Ah, not kWh

`config/battery` returns `battCapacity: 1200` for the same pack `device/latest` reports as `BatteryRatedCapacity: 1200 Ah`. Published raw, that told Predbat the battery was **1200 kWh**.

Both sources are now scaled by the nominal pack voltage, inferred from the BMS charge target: 57.6 V → 16 cells → 51.2 V nominal → **61.44 kWh**. This answers an open `VERIFY@SPIKE` question in the design spec.

## Newly mapped inverter config

- **import_today / export_today / pv_today / load_today** from the lifetime `Total*` counters. The `Daily*` counters reset at midnight and would read as a large negative delta, and the docs require an incrementing series. The `pv_today` mapping is confirmed by DEYE's own daily figures balancing exactly: `DailyConsumption 12.80 = DailyActiveProduction 4.50 + DailyEnergyPurchased 7.00 + (DailyDischargingEnergy 4.10 - DailyChargingEnergy 2.80)` — which only closes if ActiveProduction is PV alone, excluding battery discharge.
- **battery_rate_max** from `maxChargeCurrent` (185 A) scaled to watts → 9472 W. This is the DC battery-side limit and legitimately exceeds the 8000 W AC rating; `inverter_limit` constrains the AC side separately.
- **inverter_limit** from `RatedPower` → 8000 W.

`export_limit` has no read endpoint on DEYE — `MAX_SELL_POWER` is write-only via `/order/sys/power/update` — so it stays a manual `apps.yaml` value.

`automatic_config` no longer points Predbat at sensors that were never published (which is what happened to `soc_max` when `config/battery` failed); it warns to set the value in `apps.yaml` instead.

## Supporting changes

- **API request/response tracing**, on by default while this beds in, credentials redacted and bodies truncated. This is what made all of the above diagnosable. Flip `DeyeAPI.api_debug` to False once stable.
- **`station/latest` and `device/measurePoints`** polled once on the first cycle, so the model's capabilities are visible in the log. Both are read-only and confirmed in the official sample repo.

## Testing

- 24 new tests across `test_deye_api.py`, `test_deye_oauth.py`, `test_deye_publish.py`, `test_oauth_mixin.py`, built from captured live payloads rather than from the constants under test.
- Each regression test was verified to **fail without its fix**.
- Full quick suite: **189 tests, all passing**. `pre-commit` clean.


springfall2008:
Good catch — valid, and fixed in aa8f3b1a.

Confirmed the consequence you described: `publish_data()` gates on `rated_power > 0` and `automatic_config()` gates on `all(... > 0)`, so a single incomplete payload stops both the `inverter_limit` sensor being published and the arg being mapped, despite the device having reported the rating earlier.

You also identified the right inconsistency. The energy counters immediately above guard with `if key in flat`, and `derive_battery_capacity()` returns early without touching its cache when the Ah rating is absent — `RatedPower` was the only one of the three that clobbered. It now follows the same pattern.

Worth adding that this is reachable rather than theoretical on this hardware: `config/battery` returned `2106001 config point not supported` on one cycle of the live site and succeeded on the next, so partial DEYE responses are demonstrated.

Added a regression test covering both halves — a rating captured then survived across an incomplete payload, and a device that has never reported one recording nothing rather than a bogus 0.0. Verified it fails against the previous code (`rating was clobbered by the incomplete payload, got 0.0`).

PR #4349, opened 2026-07-28
feat(plan): tap/focus disclosure for the "why" reason, not just hover

chalfontchubby:
## Summary
Fast-follow to #4311 (merged) - addresses the mobile/accessibility gap flagged in review on that PR (a hover-only `title=` attribute doesn't serve touch devices at all, since a tap can't trigger `:hover`).

- Reuses the existing `.dropdown-content`/`toggleForceDropdown`/`closeDropdowns` tap-to-toggle mechanism already used for the time/rate cell manual-override menus, kept alongside `title=` rather than replacing it (a touch interaction can't trigger `:hover` in the first place, so the two never fire together).
- Added `tabindex`, `onkeydown`, `role="button"` and a focus-visible outline so it's keyboard-operable too, not just tap.
- No changes to the reason data/JSON side (`output.py`) - purely a `web_helper.py` rendering change, since the data and rendering were already cleanly decoupled in #4311.

## Test plan
- [x] `./run_all --test plan_why_reason` - all pass (including two new cases for the tap/focus disclosure wiring)
- [x] `pre-commit` (ruff, black, cspell) run against the changed files, passed

### DIFF
--- a/apps/predbat/tests/test_plan_why_reason.py
+++ b/apps/predbat/tests/test_plan_why_reason.py
@@ -343,21 +343,52 @@ def run_test_plan_why_reason(my_predbat):
     if "title=" not in renderer_js:
         print("ERROR: expected renderStateCell to emit a title= attribute")
         failed = True
-    if "state2_color || '#FFFFFF'}${titleAttr}" not in renderer_js:
-        print("ERROR: expected the split (state2) cell to also carry the title= tooltip, not just the first half")
+    # Scoped to renderStateCell's own body (not a whole-file count) since the read-only path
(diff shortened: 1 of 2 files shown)


chalfontchubby:
Fixed - the click-outside handler only whitelisted `.clickable-time-cell`, so tapping the new reason cell opened the panel and the same click's bubble-phase then immediately closed it again via `closeDropdowns()`. Added `.clickable-state-cell` to the whitelist, plus a regression test asserting it (the existing tests in this file only checked `get_plan_renderer_js()`'s output, but the click-outside handler lives in the separate `get_plan_css()` function, which is why this slipped through).

PR #4353, opened 2026-07-28
fix(plan): tooltips on History views, split-slot first half, and an invalid escape

springfall2008:
Follow-ups to the "why" tooltips in #4311. All three were found by executing the shipped renderer JS against a real `raw_plan` under Node, rather than by reading it.

## History / "Yesterday Without Predbat" had no tooltips at all

`renderPlanTable` has two state-cell paths and only the editable one (the plan view, `editable = currentView === 'plan'`) called `renderStateCell`. The read-only branch built its `<td>` inline with no `title=`, so those views rendered zero tooltips. Both paths now share a `reasonTitleAttr()` helper.

Those datasets already carry `reasons` and `reason_templates` — they come from the same `publish_html_plan` — so this is a render-only fix, no new data.

## Templates were read from the wrong dataset

`renderReasonText` looked templates up in `window.planData.reason_templates` even though `renderPlanTable` was handed a `jsonData`. Harmless while only the plan view rendered tooltips; once History does too, it would be looking up its rows against the plan view's table. It now takes them from the dataset it is rendering.

## A split slot only explained one of its two halves

When an export window starts partway through a slot, the state cell splits into `arrow | Exp`. The pre-window arrow ([output.py](../blob/main/apps/predbat/output.py) ~L1302) was the only state that appended no reason, so `reasons` held just the export entry and the tooltip explained only the second half:

```
split: True    state_text: " ↘"    state2_text: "Exp↘"
reason codes: ['export_high_rate']          <-- one reason, two-part cell
```

It now emits `demand_before_export_{rising,falling,steady}`, deliberately worded without the "no charging or exporting is scheduled this slot" clause the plain demand reasons use — that would contradict the export sentence sitting beside it in the same tooltip.

## SyntaxWarning on import

```
/config/web_helper.py:6330: SyntaxWarning: invalid escape sequence '\{'
```

`renderReasonText`'s regex `/\{(\w+)\}/g` sits inside a plain (non-raw) triple-quoted Python string, so its backslashes needed doubling — the convention already used by every other JS regex in these blobs (e.g. `(\\d{2})` a few lines below). The emitted JS is **byte-identical** (Python passes invalid escapes through verbatim), so this changes no behaviour, but it becomes a `SyntaxError` in a future Python. It was the only such warning in the package.

## Verification

Running the real `get_plan_renderer_js()` output against a generated plan:

```
PLAN view (editable)    -> title= attributes: 48
HISTORY view (readonly) -> title= attributes: 48     (was 0)
with window.planData UNSET -> title= attributes: 48
```

New tests in `test_plan_why_reason.py` (15 total, all passing): split-slot reason pairing and non-contradictory wording, the read-only tooltip path, templates sourced from `jsonData`, and a guard that `web_helper.py` compiles with no `SyntaxWarning`. `./run_all --quick` and `./run_pre_commit` both pass.

### DIFF
--- a/apps/predbat/output.py
+++ b/apps/predbat/output.py
@@ -1305,15 +1305,18 @@ class Output:
                     start = self.export_window_best[export_window_n]["start"]
                     if start > minute:
                         soc_change_this = self.predict_soc_best.get(max(start - self.minutes_now, 0), 0.0) - self.predict_soc_best.get(minute_relative_start, 0.0)
-                        if soc_change_this >= 0:
-                            state = " &nearr;"
-                            reason_parts.append({"code": "demand_before_export_rising", "params": {}})
-                        elif soc_change_this < 0:
-                            state = " &searr;"
-                            reason_parts.append({"code": "demand_before_export_falling", "params": {}})
-                        else:
+                        # Same near-flat tolerance as the whole-slot demand arrow above - testing
+                        # soc_change_this >= 0 first would make the steady case unreachable and
+                        # render a flat pre-window period as rising
+                        if abs(soc_change_this) < 0.05:
                             state = " &rarr;"
                             reason_parts.append({"code": "demand_before_export_steady", "params": {}})
+                        elif soc_change_this >= 0:
+                            state = " &nearr;"
+                            reason_parts.append({"code": "demand_before_export_rising", "params": {}})
+                        else:
+                            state = " &searr;"
+                            reason_parts.append({"code": "demand_before_export_falling", "params": {}})
                         state_color = "#FFFFFF"
                         show_limit = ""
                         had_state = True
--- a/apps/predbat/output.py
+++ b/apps/predbat/output.py
@@ -32,12 +32,18 @@ REASON_TEMPLATES = {
     "demand_rising": "Battery level is expected to rise from solar generation; no charging or exporting is scheduled this slot.",
     "demand_falling": "Battery is expected to discharge to cover house demand; no charging or exporting is scheduled this slot.",
     "demand_steady": "Battery level is expected to stay steady; no charging or exporting is scheduled this slot.",
-    "freeze_charge_reserve": "Freeze charging — the battery holds at the reserve level rather than charging further this slot (import rate {rate}p/kWh vs. your {threshold}p/kWh threshold).",
+    # Used for the first half of a split slot where the export window only starts partway through -
+    # deliberately worded without the "nothing is scheduled this slot" clause of the plain demand
+    # reasons above, which would contradict the export reason sitting alongside it in the same slot.
+    "demand_before_export_rising": "Until the export window starts partway through this slot, the battery level is expected to rise from solar generation.",
+    "demand_before_export_falling": "Until the export window starts partway through this slot, the battery is expected to discharge to cover house demand.",
+    "demand_before_export_steady": "Until the export window starts partway through this slot, the battery level is expected to stay steady.",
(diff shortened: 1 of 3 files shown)


springfall2008:
I also changed some of the descriptions as one was misleading and removed some assumptions as to why (we don't know why)

springfall2008:
Good catch on the dead branch — fixed in 4e8c9737, though not quite as suggested.

`> 0 / < 0 / else` only reaches steady on an *exact* `0.0` float, which in practice never occurs for a subtraction of two predicted SoC values. It would also still disagree with the whole-slot demand arrow 130 lines above, which treats near-flat as steady via a tolerance:

```python
if abs(soc_change) < 0.05:
    soc_sym = "&rarr;"
elif soc_change >= 0:
    soc_sym = "&nearr;"
```

So the pre-export arrow now leads with that same `abs(...) < 0.05` test, making the two paths consistent and the steady branch genuinely reachable.

The new test uses a near-flat but *non-zero* rise (0.02kWh over the pre-window period), which is the case that actually separates the two fixes. Verified both ways — under the suggested strict comparison it fails:

```
ERROR: flat pre-export slot reasons unexpected: ['demand_before_export_rising', 'export_high_rate']
```

and passes under the tolerance.

Worth noting this also corrects the rendered arrow, not just the tooltip text: a near-flat pre-window period now shows → instead of ↗.

chalfontchubby (commented):
(Posting as Claude, an AI assistant helping review this, working with @chalfontchubby)

Reviewed the diff and ran it independently rather than just reading the description.

**The four stated fixes all check out:**
- History/Yesterday tooltips - verified, both render paths now share `reasonTitleAttr()`.
- Templates sourced from the wrong dataset - verified fixed, now reads `jsonData.reason_templates` instead of the global `window.planData`.
- Split-slot missing a reason for the pre-export half - verified, and the new `demand_before_export_{rising,falling,steady}` wording correctly avoids contradicting the export half's text in the same tooltip.
- Invalid escape sequence - independently scanned every file in `apps/predbat/*.py` for `SyntaxWarning` on compile (not just this one file) to check the "only one in the package" claim - confirmed zero remain anywhere.

The follow-up commit (near-flat tolerance for the pre-export arrow) is a good catch too - reusing the same `abs(...) < 0.05` tolerance the whole-slot arrow already uses a few lines up is the right fix, rather than the stricter `>0`/`<0` Copilot originally suggested (which would only ever reach `steady` on an exact `0.0` float).

Ran `./run_all --test plan_why_reason` myself on the actual branch - all 16 tests pass, including both new ones.

**One thing worth a quick answer before merging, not mentioned in the description**: the diff also rewords `charge_low_rate` and `export_high_rate`, dropping the causal "because the rate is low/high enough to be worth it" clause:

> old: *"Charging up to 80% because the import rate this slot (8p/kWh) is low enough to be worth it before a more expensive period."*
> new: *"Charging up to 80% at the import rate for this slot of (8p/kWh)."*

That's the "why" being dropped from a why-tooltip, and it's outside the scope the description covers - was that intentional? The `export_high_rate` rewording also reads a bit awkwardly ("at the export rate of (X) using stored energy back to the grid").

Not a blocker either way - the four fixes described are all correct and good to merge - just want to flag it before it lands, since it wasn't clear whether it was deliberate or a side effect of iterating on the diff.

chalfontchubby:
(Posting as Claude, an AI assistant helping review this, working with @chalfontchubby)

That makes sense, thanks - and it's actually the right call rather than a regression. Those causal clauses were never something the code verified, just a narrative on top of a decision the optimiser made for its own reasons. That's the same category of problem flagged in our own backlog for the `FrzExp`/`HoldExp` wording (a confident-sounding causal story the code can't actually back up) - good to see it addressed more broadly here rather than left as a one-off. No objection from me, good to merge.

PR #4360, opened 2026-07-28
fix(gateway): write entities for the bound control target, not just primary units

### LINKED ISSUE #4356: GWMQTT: stale gateway telemetry is served indefinitely as live — no staleness gate, component health never degrades, surfaces only as a misleading clock-skew warning
## Summary

When an ESP32 gateway stops publishing valid telemetry, the `GWMQTT` component keeps serving the
**last cached frame indefinitely**, presented as live data. Nothing logs, nothing alarms, and
component health stays green. PredBat continues to plan and issue inverter control against values
that may be over an hour old.

Observed in production: telemetry froze for **69 minutes**. Every field in `Inverter 0 SoC:` was
byte-identical for the whole window — SoC, battery power, grid power, load power and PV power all
unchanged, cycle after cycle — while the battery was in reality force-exporting at ~5.5 kW. The
only symptom that surfaced anywhere was PredBat's clock-skew warning.

## Why the clock warning is the only thing that escapes

On a `GWMQTT` site the `inverter_time` entity is **not the inverter's RTC**. It is derived from
`GatewayStatus.timestamp`, i.e. the gateway's own clock — `gateway.py` `_inject_entities()`:

```python
if status.timestamp > 0 and len(status.inverters) > 0:
    primary_inv = next((inv for inv in status.inverters if inv.primary), status.inverters[0])
    ...
    dt = datetime.datetime.fromtimestamp(status.timestamp, tz=self.local_tz)
    self.dashboard_item(f"sensor.{self.prefix}_gateway_{ts_suffix}_inverter_time", ...)
```

`status.timestamp` happens to be the one cached value whose staleness is self-evident, because it can
be compared against wall clock. So a frozen stream shows up as a skew that **grows linearly with real
time** (-37 → -42 → -47 → -52 → -57 → -62 → -68.9 min in one run), and gets reported to the user as
"please fix your inverter time" — which is misleading. The clock is fine; the feed is dead.

Every other cached value fails silently, because nothing about a stale SoC looks wrong on its own.

## Three gaps that compound

**1. Zero-inverter frames are dropped with no trace** — `_process_telemetry()`:

```python
if len(status.inverters) == 0:
    return
```

A gateway publishing frames with an empty inverter list is completely invisible. No log at any level,
no counter, no state change. This is also the silent path if a device-side topology/discovery churn
ever emits an empty list.

**2. The publish path re-emits the cached frame forever.** The entity refresh reads
`self._last_status` and republishes it every cycle:

```python
if self._last_status and self._last_status.timestamp:
    dt = datetime.datetime.fromtimestamp(self._last_status.timestamp, tz=self.local_tz)
    self.dashboard_item(f"sensor.{pfx}_inverter_time", ...)
```

There is no staleness gate on this path, so entities never go `unavailable` — they just keep
asserting stale values as current, indefinitely.

**3. Component health cannot detect it.** `update_success_timestamp()` is called at the end of every
successful housekeeping cycle, independent of whether any telemetry arrived:

```python
# Mark component as alive on successful housekeeping
self.update_success_timestamp()
```

Housekeeping succeeds fine with a dead feed (it publishes plans, checks read-only state, refreshes
tokens), so the component's liveness never degrades and the health/backoff machinery never fires.

`is_healthy()` *does* contain a freshness check against `_TELEMETRY_STALE_THRESHOLD` (120 s), but it
short-circuits before reaching it:

```python
if not self._gateway_online:
    # Gateway is offline but we're connected to broker — that's OK
    return True
```

`_gateway_online` is driven by the retained `/online` topic, which is itself unreliable — a clean
disconnect never overwrites it, and a broker that denies the last-will publish leaves it reading
`online` for a device that is gone. So this branch can return healthy in exactly the situation it
exists to catch.

Net effect: a 69-minute total telemetry outage produced no error log, no health transition, and no
unavailable entity — only a misleading clock warning.

## Suggested direction

- Gate the cached-republish path on `_last_telemetry_time`: past the staleness threshold, stop
  asserting values and let entities go `unavailable` rather than lie.
- Log (rate-limited) when a frame is dropped for `len(status.inverters) == 0`, and count it.
- Decouple "housekeeping ran" from "component has data". Telemetry freshness should drive liveness;
  a component with a dead feed is not healthy just because its timers still fire.
- Reconsider the `not self._gateway_online → return True` short-circuit, given the retained `/online`
  topic can assert `online` for a device that is not publishing.
- Consider distinguishing the two cases in the skew warning: a *constant* offset is a genuine clock
  problem, a *linearly growing* one means the feed has stopped. The current message sends users to
  fix a clock that isn't broken.

Happy to put a PR together for whichever shape you'd prefer.

mgazza:
Fixes #4356.

## The bug

On a multi-AIO site `automatic_config()` picks the Gateway/EMS as the control target (GivTCP rules) and binds **every** PredBat arg to its serial suffix:

```python
elif gateway_units and len(aios) > 1:
    inverters = gateway_units[:1]     # Gateway is the control point
...
suffix0 = inverters[0].serial[-6:].lower()
self.set_arg("inverter_time", [f"sensor.{self.prefix}_gateway_{suffix0}_inverter_time"])
```

But `_inject_entities()` only ever *wrote* entities for units flagged `primary` — and firmware never flags a Gateway or EMS `primary` (it is set on battery inverters). So on those sites **none** of the bound entities were ever written. They kept whatever value they last held and silently froze.

## Impact

Seen in production: a site froze for 69 minutes with byte-identical readings every cycle —

```
SoC: 20.52kW 76%, battery power 5467W, grid power 5336W, load power 348W, PV Power 217W
```

— while the battery was really force-exporting at ~5.5 kW. PredBat kept planning and issuing control against a stale 76% SoC.

The only symptom that escaped was the clock-skew warning, because `inverter_time` is the one bound value whose staleness is self-evident against wall clock. It grew linearly (−37 → −42 → … → −98 min) and told the user to *"fix your inverter time"* — misleading, since the inverter clock was fine. Everything else failed silently, because nothing about a stale SoC looks wrong on its own.

Verified against a live device: the gateway was publishing healthy frames throughout (current `timestamp`, all inverters present) and PredBat's broker session was receiving them. Nothing was wrong with the device or the transport — only the entity binding.

There is a second divergence mode with the same cause: `_needs_reconfigure()` only re-runs on a **new serial** or **new EV charger**, so if the `primary` flag moves between already-known units the binding goes stale the same way. This PR fixes the write side, which covers both.

## The fix

- Inject a unit when it is `primary` **or** when `automatic_config` bound args to it. `_suffix_to_serial` is already populated by `automatic_config` with exactly the control-target suffixes, so it is the authoritative record of what PredBat reads.
- Write `inverter_time` under the **bound** suffix rather than the primary's.
- Non-primary units that are *not* the control target stay skipped, preserving the existing guard against EMS/gateway units double-counting power readings on the dashboard.

## Tests

New `TestBoundEntitiesAreWritten` asserts the invariant directly — every entity `automatic_config` binds must actually be written by `_inject_entities` — using a Gateway + 2 AIO topology matching the field case.

Confirmed RED first. Before the fix it reported:

```
automatic_config bound inverter_time to sensor.predbat_gateway_15g357_inverter_time
but _inject_entities never wrote it; written entities were: [... only _35g421 and _32g070 ...]
```

After: all 251 gateway tests pass, and `unit_test.py --quick` is green. `flake8` output is unchanged from the `main` baseline (the remaining `C901`/`E402` are pre-existing); `interrogate` stays at 100%.

### DIFF
--- a/apps/predbat/gateway.py
+++ b/apps/predbat/gateway.py
@@ -59,6 +59,21 @@ _GATEWAY_BASE_TIME = datetime.datetime.strptime("00:00", "%H:%M")
 _GATEWAY_OPTIONS_TIME = [(_GATEWAY_BASE_TIME + datetime.timedelta(seconds=m * 60)).strftime("%H:%M:%S") for m in range(0, 24 * 60, 5)]
 
 
+def _serial_suffix(serial):
+    """Return the entity-name suffix for an inverter serial.
+
+    Entity IDs use the last six characters of the serial, lower-cased; serials of
+    six characters or fewer are used whole.
+
+    Args:
+        serial: Inverter serial string.
+
+    Returns:
+        str: The lower-cased suffix used in gateway entity IDs.
+    """
+    return serial[-6:].lower() if len(serial) > 6 else serial.lower()
+
+
 PLAN_MODE_AUTO = 0
 PLAN_MODE_CHARGE = 1
 PLAN_MODE_DISCHARGE = 2
@@ -719,6 +734,22 @@ class GatewayMQTT(ComponentBase):
         if self._needs_reconfigure(status):
             self.automatic_config()
 
+    def _is_bound_target(self, inv):
+        """Whether automatic_config bound PredBat's args to this inverter's suffix.
+
+        ``_suffix_to_serial`` is populated by ``automatic_config`` with exactly the
+        control-target suffixes, so it is the authoritative record of which entities
+        PredBat reads. Empty before the first auto-config, in which case no unit is
+        bound yet and the primary-only path applies.
+
+        Args:
+            inv: A ``predbat_InverterEntry`` from the gateway status.
+
+        Returns:
+            bool: True if this unit's entities are read by PredBat.
+        """
+        return _serial_suffix(inv.serial) in self._suffix_to_serial
+
     def _inject_entities(self, status):
         """Inject inverter entities into PredBat state cache.
 
@@ -735,10 +766,14 @@ class GatewayMQTT(ComponentBase):
             app="gateway",
         )
 
-        # Inverter time from gateway timestamp — use first primary inverter's serial
+        # Inverter time from gateway timestamp — write it under the suffix PredBat
+        # actually reads (the control target), not the primary's, or the bound
+        # inverter_time arg is never updated and silently freezes.
         if status.timestamp > 0 and len(status.inverters) > 0:
-            primary_inv = next((inv for inv in status.inverters if inv.primary), status.inverters[0])
-            ts_suffix = primary_inv.serial[-6:].lower() if len(primary_inv.serial) > 6 else primary_inv.serial.lower()
+            ts_inv = next((inv for inv in status.inverters if self._is_bound_target(inv)), None)
+            if ts_inv is None:
+                ts_inv = next((inv for inv in status.inverters if inv.primary), status.inverters[0])
+            ts_suffix = _serial_suffix(ts_inv.serial)
             dt = datetime.datetime.fromtimestamp(status.timestamp, tz=self.local_tz)
             self.dashboard_item(
                 f"sensor.{self.prefix}_gateway_{ts_suffix}_inverter_time",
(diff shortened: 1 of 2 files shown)


apps/predbat/gateway.py
```diff
@@ -748,12 +783,15 @@ def _inject_entities(self, status):
...
+            # bound PredBat's args to. On a multi-AIO site the control target is the
+            # Gateway/EMS, which firmware never flags primary — skipping it left every
+            # bound arg unwritten and frozen at its last value. Non-primary units that are
+            # NOT the control target stay skipped: they report overlapping power readings
+            # that would double up on the dashboard.
+            if not inv.primary and not self._is_bound_target(inv):
                 continue
-            suffix = inv.serial[-6:].lower() if len(inv.serial) > 6 else inv.serial.lower()
-            self._inject_inverter_entities(inv, suffix)
+            self._inject_inverter_entities(inv, _serial_suffix(inv.serial))
```
springfall2008, line 794: Why not just publish them all?

PR #4363, opened 2026-07-29
fix(deye): report run success, and never ask for SOC below the pack floor

springfall2008:
Two problems from the second live control run.

First, what that log confirms: the entity-write fix from #4354 works — the run made **four control-entity writes with zero failures** (`Wrote 9472 to charge_rate, successfully now 9472.0`, and likewise `charge_limit`, `reserve`, `discharge_rate`).

## 1. The component was permanently marked unhealthy

Every cycle ended with:

```
Error: Complete run status Demand with component errors: DEYE Cloud
```

despite nothing having failed. DEYE never called `update_success_timestamp()`, which ha, solis, kraken, octopus, enphase, ohme and gateway all do. [`components.is_alive()`](apps/predbat/components.py#L681-L684) requires a success within the last 60 minutes and returns False when no timestamp was ever recorded, so the component counted as "active but not alive" for the whole session and `record_final_run_status()` failed the run.

Reported at the end of a successful cycle only, so a cycle that bails (for example the first-poll deferral) still correctly looks unhealthy.

## 2. Slot SOC could sit below the inverter's own floor

The first control write of a cycle went out with slot SOC **0** on a pack whose installer floor is **14%**:

```
POST 0: slot socs = [0, 0, 0, 0, 0, 0]
POST 1: slot socs = [14, 14, 14, 14, 14, 14]
```

(Those two orders are a correction, not a duplicate — the second is the same schedule once Predbat had written the reserve.)

The cause is that the reserve control entity is published with `min: 0` and reads 0 until Predbat writes the real value, and nothing clamped the gap. `battLowCapacity` was already being read from `config/battery` and published as `battery_reserve_min`, but was never used as a floor.

It now is, the way [`fox.py`](apps/predbat/fox.py#L2070) has always clamped reserve with `max(value, fdsoc_min)`. Applied in three places so no path can bypass it:

- when the reserve enters `local_schedule` (both from the entity read and from a control event),
- when the entity is published — as the value *and* the entity `min`,
- finally in `build_dynamic_payload`, the last step before the payload goes out.

With no `config/battery` response there is no known floor, and the schedule passes through unchanged rather than a guessed default being imposed.

## Testing

Four new tests: the success timestamp being reported on a good cycle and *not* on a failed one, `battery_reserve_min` reading `battLowCapacity` (rejecting 0/negative/over-100), and slot SOC clamping — covering the idle case, an above-floor export target surviving untouched, and no clamping when the floor is unknown.

Both fixes verified to fail against the pre-fix code. Removing the clamp reproduces the live payload exactly (`slot SOC below the 14% floor: [0, 0, 0, 0, 0, 0]`); removing the timestamp gives `a successful cycle must call update_success_timestamp()`.

Full quick suite passes; `pre-commit` clean.

## Note on history

This change was briefly pushed directly to main as `fb369f6e` in error; that was reverted in `36bd8cb5` and re-raised here for review. The content is identical.

### DIFF
--- a/apps/predbat/deye.py
+++ b/apps/predbat/deye.py
@@ -455,7 +455,10 @@ class DeyeAPI(ComponentBase, OAuthMixin):
         """
         if sn not in self.device_battery_config:
             return 0
-        floor = int(self._battery_config_value(sn, "reserve_min", 0))
+        try:
+            floor = int(self._battery_config_value(sn, "reserve_min", 0))
+        except (TypeError, ValueError):
+            return 0
         return floor if 0 < floor <= 100 else 0
 
     def battery_rate_max(self, sn):
--- a/apps/predbat/deye.py
+++ b/apps/predbat/deye.py
@@ -446,6 +446,18 @@ class DeyeAPI(ComponentBase, OAuthMixin):
                 return round(configured_ah * volts / 1000.0, 2)
         return self.device_capacity.get(sn, 0.0)
 
+    def battery_reserve_min(self, sn):
+        """Return the inverter's own minimum SOC floor in percent, or 0 when unknown.
+
+        config/battery reports this as battLowCapacity (14 on the live 3-phase hybrid).
+        It is the floor the installer configured on the inverter, so Predbat must never
+        ask for a slot SOC below it.
+        """
+        if sn not in self.device_battery_config:
+            return 0
+        floor = int(self._battery_config_value(sn, "reserve_min", 0))
+        return floor if 0 < floor <= 100 else 0
+
     def battery_rate_max(self, sn):
         """Return the battery's maximum charge rate in W, for Predbat's battery_rate_max.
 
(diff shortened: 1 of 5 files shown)


springfall2008:
Good catch on the inconsistency — `publish_schedule_settings_ha()` clamped while `apply_reserve_live()` stored the raw value, so the two could disagree. Fixed in a53c4c80, but resolved the opposite way round from the suggestion, because I think clamping at entry does not fix the problem you describe.

If Predbat writes 4 and we store *and* publish 14, the read-back still returns 14 ≠ 4. `write_and_poll_value` compares what it wrote against what it reads, so it retries until it gives up regardless of which end is clamped — the exact `didn't complete got 0.0` failure this component was fixed for in #4354. Clamping at entry only makes the internal state self-consistent; it does not make the read-back match.

So the entity now echoes what Predbat wrote, everywhere — entity read, control event and published state all store and emit the raw value — and the floor is applied in exactly one place, `build_dynamic_payload`, immediately before the payload leaves for the API. That is what actually protects the battery, it cannot be bypassed by any caller, and it removes the mismatch by construction rather than by agreement between two call sites.

On reachability, which I checked rather than assumed: `adjust_reserve` already clamps to `self.reserve_percent`, sourced from the `battery_min_soc` sensor this very component publishes, so the automatic path can never write below the floor. A manually configured install that does not map `battery_min_soc` defaults it to 4.0 and could — narrow, but real, so worth closing.

The clamp now logs once per serial when it actually lifts a value: routine before Predbat has written the reserve, but a persistent complaint means Predbat is planning below the inverter's floor, which is better surfaced than silently corrected.

Added a test that the reserve entity echoes a below-floor write verbatim while the payload built from it is still lifted to the floor. Verified it fails against the clamped-publish version: `expected the entity to echo 4, got 14`.

PR #4368, opened 2026-07-29
fix(fetch): publish rate_import/rate_export atomically to remove a read race

springfall2008:
## Summary

Root-cause fix for the rate-dict read race noted in #4367: `update_pred` rebuilds `rate_import`/`rate_export` on an AppDaemon **worker thread**, while async components (the Teslemetry tariff builder, and any future reader) read them from the **asyncio event loop**, with no lock. The build:

- reset them to `{}` at the start,
- reassigned them per provider step, and
- **mutated them in place** (`load_saving_slot` / `load_free_slot` / `load_axle_slot` doing `self.rate_export[minute] += rate`).

So a concurrent reader could observe an **empty** dict (during the reset window) or a **half-merged** one (mid saving-session apply) — producing a transiently wrong result.

## Change

- Build each side into a **local** (`import_rates` / `export_rates`) through the entire fetch → replicate → slot pipeline, and assign `self.rate_import` / `self.rate_export` **exactly once** at the end. A reader now always sees either the previous complete dict or the new complete dict — never an intermediate.
- The three in-place slot mutators now take the working dict as a **parameter** (`load_saving_slot(..., rate_dict, ...)`, etc.) and operate on it, instead of touching `self.rate_*`. Call sites (5 in `fetch.py`) and their unit tests updated.

Only the working-variable references were rethreaded (a `\b`-anchored transform); the derived attributes (`rate_import_base`, `_replicated`, `_no_io`, `rate_export_base`, `_replicated`) are untouched and still assigned from the locals — so behaviour is preserved.

## Why not a per-consumer snapshot

#4367 added a defensive `dict()` snapshot in the Teslemetry component to dodge this race. This fixes it at the source for **all** consumers and closes the transient-empty window a per-read snapshot can't. Once this lands, that snapshot in #4367 becomes redundant and can be reverted (it's harmless until then). Note this PR does not itself remove that snapshot — it lives on the #4367 branch, not `main`.

## Test plan
- [x] `unit_test.py --quick` — full suite passes (identical to baseline, incl. the pre-existing intentional exception-handling tracebacks); the rate paths (octopus rates, saving/free/axle slots, overrides, manual rates, IO) directly assert the resulting `rate_import`/`rate_export` values, so byte-level behaviour is covered.
- [x] `load_free_slot`, `saving_session`, `axle` mutator tests updated for the new signature and pass.
- [x] pre-commit (ruff, black, cspell) clean.

### DIFF
--- a/apps/predbat/axle.py
+++ b/apps/predbat/axle.py
@@ -647,9 +647,9 @@ def fetch_axle_sessions(base):
     return axle_events_deduplicated
 
 
-def load_axle_slot(base, axle_sessions, export, rate_replicate=None):
+def load_axle_slot(base, axle_sessions, rate_dict, export, rate_replicate=None):
     """
-    Load Axle VPP session slot
+    Load Axle VPP session slot into rate_dict (in place)
     """
     if rate_replicate is None:
         rate_replicate = {}
@@ -679,11 +679,11 @@ def load_axle_slot(base, axle_sessions, export, rate_replicate=None):
                 base.log("Setting Axle VPP session in range {} - {} export {} pence_per_kwh {}".format(base.time_abs_str(start_minutes), base.time_abs_str(end_minutes), export, pence_per_kwh))
                 for minute in range(start_minutes, end_minutes):
                     if export:
-                        base.rate_export[minute] = base.rate_export.get(minute, 0) + pence_per_kwh
+                        rate_dict[minute] = rate_dict.get(minute, 0) + pence_per_kwh
                         base.load_scaling_dynamic[minute] = base.load_scaling_saving
                         rate_replicate[minute] = "saving"
                     else:
-                        base.rate_import[minute] = base.rate_import.get(minute, 0) - pence_per_kwh
+                        rate_dict[minute] = rate_dict.get(minute, 0) - pence_per_kwh
                         base.load_scaling_dynamic[minute] = base.load_scaling_free
                         rate_replicate[minute] = "saving"
 
--- a/apps/predbat/fetch.py
+++ b/apps/predbat/fetch.py
@@ -989,7 +989,7 @@ class Fetch:
             self.rate_import_no_io = {}
             self.log("Warning: No import rate data provided")
             self.record_status(message="Error: No import rate data provided", had_errors=True)
-        # Atomic publish: readers (e.g. async components) never see a half-built or empty rate_import.
+        # Atomic publish: readers (e.g. async components) never see a half-built or transiently-empty rate_import during rebuild.
         self.rate_import = import_rates
 
         # Replicate and scan export rates
--- a/apps/predbat/fetch.py
+++ b/apps/predbat/fetch.py
@@ -1902,8 +1902,12 @@ class Fetch:
         curr = self.currency_symbols[1]
 
         if print:
-            # Calculate minimum forward rates only once rate replicate has run (when print is True)
-            self.rate_min_forward = self.rate_min_forward_calc(self.rate_import)
+            # Calculate minimum forward rates only once rate replicate has run (when print is True).
+            # Use the `rates` argument, not self.rate_import: during fetch's atomic rebuild self.rate_import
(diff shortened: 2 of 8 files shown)


springfall2008:
## Review: correctness regression found

Nice root-cause fix for the read race, and the mechanical rethreading of `self.rate_import`/`self.rate_export` onto local variables (`import_rates`/`export_rates`) looks careful and correct throughout `fetch.py`. However, there's a bug that should be fixed before merging.

### `rate_min_forward` is now computed one cycle stale

In `rate_scan` (`apps/predbat/fetch.py`):

```python
def rate_scan(self, rates, print=True):
    self.rate_min, self.rate_max, self.rate_average, self.rate_min_minute, self.rate_max_minute = self.rate_minmax(rates)
    ...
    if print:
        self.rate_min_forward = self.rate_min_forward_calc(self.rate_import)   # <-- reads self.rate_import, not `rates`
```

This is called as `self.rate_scan(import_rates, print=True)` from `fetch_sensor_data`, **before** the new atomic publish `self.rate_import = import_rates` runs. `rate_minmax` correctly uses the `rates` parameter, but `rate_min_forward_calc` reads `self.rate_import` directly — at that point it still holds the *previous* cycle's value, since the publish is now deferred to the end of the block.

Previously this worked because `self.rate_import` was reassigned progressively throughout the old code, so by the time `rate_scan(self.rate_import, print=True)` ran, `self.rate_import` already *was* the current cycle's fully-processed data. Deferring the publish (correctly, for the race fix) broke this implicit dependency.

**Impact:** `rate_min_forward` feeds directly into charge/discharge economics in `plan.py` (`rate_min_forward.get(...)` at the window-scan and cost-comparison sites) and `output.py:1812`. Every 5-minute cycle now uses the *previous* cycle's forward-minimum rates instead of the just-computed ones — a persistent regression, and it runs counter to this PR's own goal that readers should only ever see a fully old or fully new state.

**Fix:** change `self.rate_min_forward_calc(self.rate_import)` → `self.rate_min_forward_calc(rates)` in `rate_scan`.

**Why existing tests don't catch it:** `test_rate_min_forward_calc.py` unit-tests `rate_min_forward_calc()` directly with hand-built inputs — it never exercises the `rate_scan()` call site inside `fetch_sensor_data()`, so this integration-level staleness is invisible to it. Worth adding a test that asserts `self.rate_min_forward` reflects current-cycle rates after a `fetch_sensor_data` rebuild.

### Minor, pre-existing (not introduced by this PR)

`rate_add_io_slots` in `octopus.py` (unchanged by this diff) has the same shape of bug: when `octopus_slot_low_rate` is `False`, it reads `self.rate_import.get(start_minutes, self.rate_min)` instead of using its own `rates` parameter, then writes that stale value back into `rates[minute]`. Already present on `main`, but this PR makes `self.rate_import` more stale at that call point (now strictly "previous cycle" instead of "partially-updated current cycle"). Low practical impact since `octopus_slot_low_rate` defaults to `True`, but probably worth a follow-up.

springfall2008:
Both addressed:

**rate_scan forward-min (the persistent regression)** — fixed in `c64ef16e`. `rate_scan()` now derives `rate_min_forward` from its `rates` argument instead of `self.rate_import`, which the atomic publish leaves as the previous cycle's data during the rebuild. Added rate_min_forward_calc **test 7**: scans fresh rates while `self.rate_import` holds distinct stale values and asserts the forward minimum reflects the scanned rates (RED before, green after).

**rate_add_io_slots (the minor/pre-existing one)** — fixed in `d8f78570`. With `octopus_slot_low_rate=False` it now reads the assumed price from the `rates` arg, not `self.rate_import`. On `main` these were the same object so it's behaviour-preserving there; it just avoids the stricter staleness this PR introduces. Added a regression test staging a stale `self.rate_import` distinct from the working rates.

Full `--quick` suite green (also caught and fixed a shared-state leak in the new test 7 so later optimiser tests are unaffected).

PR #4370, opened 2026-07-29
Auto-restart unhealthy ML Load Forecaster

nickgee31:
**Summary**
Automatically restart the ML Load Forecaster component when it becomes unhealthy outside of an active calculation. At the moment ML Load Forecaster can start before the Solis Cloud API component which leaves it in an error state until manually restarted.

**Changes**
• Added an auto_restart component setting, enabled only for load_ml.
• Added guarded asynchronous component restart handling to prevent duplicate concurrent restarts.
• Triggered automatic recovery when LoadML is active, unhealthy, and not calculating.
• Avoided restarting LoadML while it is performing training, prediction, or database work.
• Added tests covering automatic restart requests and ensuring calculating LoadML instances are not restarted.

**Behaviour**
Other components are unaffected. Genuine LoadML failures are still reported, but Predbat now attempts to recover the component automatically using the existing restart lifecycle.

### DIFF
--- a/apps/predbat/components.py
+++ b/apps/predbat/components.py
@@ -487,6 +487,7 @@ COMPONENT_LIST = {
         "class": LoadMLComponent,
         "name": "ML Load Forecaster",
         "event_filter": "predbat_load_ml_",
+        "auto_restart": True,
         "args": {
             "load_ml_enable": {"required_true": True, "config": "load_ml_enable", "default": False},
             "load_ml_source": {"required": False, "config": "load_ml_source", "default": False},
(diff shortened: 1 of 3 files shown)


springfall2008:
Is this the right fix or should we just place it in a phase after the cloud components?

nickgee31:
> Is this the right fix or should we just place it in a phase after the cloud components?

Yeah I can do that, whichever you think is best. I thought a restart would catch anything else as well as the order

springfall2008:
> > Is this the right fix or should we just place it in a phase after the cloud components?
> 
> Yeah I can do that, whichever you think is best. I thought a restart would catch anything else as well as the order

Sounds good, I don't like restarting on failure as it masks problems that need fixing

Issue #4379, opened 2026-07-30
Config validation status can go stale after a self-healed startup race


chalfontchubby:
**Describe the idea**

`validate_config()` only runs at startup and when config actually changes (init, or whenever `update_pending` is set) - never periodically. If a sensor a user has mapped in `apps.yaml` isn't populated yet at that exact moment (e.g. a slower-starting integration on an HA restart), Predbat logs a warning/error and the `config_valid`/`config_warnings` status reflects it - reasonably.

But if that sensor then comes good on its own a few seconds later (no restart, no config edit), there's nothing that re-runs validation, so the stale "config invalid" status and warning just sit there until the next restart or config change, even though `apps.yaml` is actually fine.

Ran into this myself: `sensor.sigen_plant_rated_energy_capacity` / `sensor.sigen_plant_daily_load_consumption` (Sigenergy integration) weren't populated yet at Predbat startup, logged 2 config errors, then cleared after a restart with no config changes.

**Possible direction** (not wedded to this, just the first thing that came to mind): re-run `validate_config()` periodically, or do one delayed retry a short time after startup, so a self-healed condition clears its own status rather than needing a manual restart.

Low priority - not something biting anyone hard, just noticed it and wanted to flag it for the backlog.

chalfontchubby:
(Posting as Claude, working with @chalfontchubby)

Implemented the "delayed retry" direction from the original report: PR #4394 adds a configurable retry sequence (default 2 retries, 1 minute apart) that only ever kicks in after an actual \`validate_config()\` failure - a clean validation never triggers anything. A retry that comes back clean clears the error status immediately; still-failing retries decrement and reschedule; once exhausted it gives up quietly until the next restart or config change, same as today. Both the retry count and interval are configurable in \`apps.yaml\` (\`validate_config_retries\`, \`validate_config_retry_minutes\`), and setting retries to 0 fully reverts to current behaviour.

chalfontchubby:
Fixed by #4394 (merged).

PR #4394, opened 2026-07-30
feat(config): retry validate_config() after a failure, self-healing (#4379)

chalfontchubby:
## Summary

`validate_config()` only ran at startup and when config actually changed - never periodically. If a sensor mapped in `apps.yaml` wasn't populated yet at that exact moment (e.g. a slower-starting integration during an HA restart), Predbat correctly logged a config error - but if that sensor came good on its own a few seconds later, nothing re-validated, so the stale "config invalid" status just sat there until the next manual restart.

- Adds `validate_config_schedule_retry()` (called right after every `validate_config()` call site with its error count) and `validate_config_check_retry()` (checked once per tick from the existing 15-second `update_time_loop()`).
- Retries only ever start following an **actual validation failure** - a clean validation never arms anything, matching the ask on #4379.
- Default: 2 retries, 1 minute apart. A retry that comes back clean clears the sequence immediately. A retry that's still failing decrements the counter and reschedules; once retries are exhausted it gives up quietly until the next restart or config change (the pre-existing behaviour).
- New apps.yaml-only settings (not HA-exposed, matching the shape of `balance_inverters_seconds`): `validate_config_retries` (default 2, set to 0 to disable and revert to the old behaviour) and `validate_config_retry_minutes` (default 1).
- No new scheduler primitive - this AppDaemon-shim codebase (`hass.py`) only has `run_every()`, no one-shot delayed-callback API, so this piggybacks on the existing 15-second loop rather than registering a new timer.

## Test plan

- [x] `./run_pre_commit` (via `coverage/run_pre_commit`) - all hooks pass
- [x] New `test_validate_config_retry` covers: clean validation never arms a retry; default retry count/interval; no premature retry before the due time; still-failing retries decrement and reschedule; exhausting retries gives up cleanly (and doesn't fire again afterwards); a successful retry clears the sequence immediately rather than just decrementing; `validate_config_retries: 0` disables the feature; custom retry count/interval are respected
- [x] Existing `test_validate_config` still passes unchanged
- [x] Full quick suite (`./run_all --quick`) passes

Addresses #4379

### DIFF
--- a/apps/predbat/config.py
+++ b/apps/predbat/config.py
@@ -2195,6 +2195,8 @@ APPS_SCHEMA = {
     "ge_cloud_automatic_split_pv": {"type": "boolean"},
     "num_inverters": {"type": "integer", "zero": False},
     "balance_inverters_seconds": {"type": "integer", "zero": True},
+    "validate_config_retries": {"type": "integer", "zero": True},
+    "validate_config_retry_minutes": {"type": "integer", "zero": True},
     "givtcp_rest": {"type": "string_list", "entries": "num_inverters"},
     "charge_rate": {"type": "sensor_list", "sensor_type": "float", "modify": True, "entries": "num_inverters"},
     "discharge_rate": {"type": "sensor_list", "sensor_type": "float", "modify": True, "entries": "num_inverters"},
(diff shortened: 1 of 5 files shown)


chalfontchubby:
Copilot changes have been applied

Issue #4397, opened 2026-07-30
WhatIf Annual Prediction run restarts the Predbat add-on before completing


mitulgovind:
Running the WhatIf Annual Prediction tool (Configure → Run simulations) causes the Predbat add-on to restart partway through the run, aborting it before it finishes. This happens consistently — reproduced multiple times, including after reducing samples_per_month from the default.

File /config/annual_work/apps.yaml was modified
Stopping Predbat due to file changes....
2026-07-30 23:23:14.261361: Stopping Predbat
2026-07-30 23:23:15.096032: Info: Web socket stopping
2026-07-30 23:23:15.096324: Info: Web Socket stopped
2026-07-30 23:23:15.097004: Info: HA interface stopped
2026-07-30 23:23:15.333634: Info: Components stopped
Web interface stopped
Shutdown, sleeping 20 seconds before restarting
[23:23:42] INFO: Predbat init script running
Running Predbat inside Add-on
Bootstrap Predbat
Startup
Predbat files are installed correctly for version v8.47.1
**** Starting Standalone Predbat ****

Steps to reproduce

1. Go to WhatIf → Configure, fill in the form.
2. Click Run simulations.
3. Run progresses partway (varies each time — seen anywhere from ~8% to ~25%).
4. Add-on restarts, run is aborted, progress is lost.

Expected behavior

The run should complete without the add-on restarting.

Environment

Predbat version: v8.47.1
Running as: Home Assistant Add-on
Home Assistant: Home Assistant Green

mitulgovind:
Update: This is more serious than just the run failing to complete — triggering the restart also reset my live Predbat config back to defaults, not just interrupting the annual job.

After one of the restarts, Predbat came back up in Monitor mode instead of my normal charge/discharge mode, and several tuned settings had reverted to stock defaults rather than what I'd configured. The log from the restart itself shows why:

New install detected
Set reset inverter true due to reset_inverter on item set_reserve_min
Set reset inverter true due to reset_inverter on item set_charge_low_power
Set reset inverter true due to reset_inverter on item set_export_low_power
Set reset inverter true due to reset_inverter on item set_export_freeze
Set reset inverter true due to reset_inverter_force on item set_read_only
Set reset inverter true due to reset_inverter_force on item mode
[...]

Issue #4396, opened 2026-07-30
Annual electricity cost prediction tool restarts Predbat


novocane1337:
**Describe the bug**
When running the Whatif simulation it restarts Predbat if I chose another Tariff from the drop down box and press save. If I just select the tariff and press Run simulation without saving it also restart Predbat.

It works fine on my default tariff.

**Expected behaviour**

Was hoping to compare against other custom tariffs or other tariffs is the list

**Predbat version**

v8.47.1

[predbat.log](https://github.com/user-attachments/files/30563399/predbat.log)

chalfontchubby:
Same root cause as #4397 - fix up in #4401.

chalfontchubby:
⚠️ Same warning as posted on #4397: until #4401 is merged, running/saving on the WhatIf page can restart your live Predbat instance and reset real settings back to defaults, not just interrupt the run. Recommend holding off on a live production instance until the fix lands.

PR #4401, opened 2026-07-31
fix(hass): stop the standalone watcher restarting on unrelated apps.yaml files

chalfontchubby:
## Summary

`hass.py`'s file watcher (used by the shipped add-on's runtime loop, not just dev hot-reload) walked its working directory tree and flagged any file whose basename was exactly `apps.yaml`, anywhere in that tree - never checking it against the actual configured apps.yaml path.

The new Annual prediction tool's headless work directory (`annual_work/`, inside `config_root` - `/config` in production) writes its own isolated `apps.yaml`-shaped scratch file on every WhatIf run, for legitimate reasons (an isolated headless PredBat instance needs its own config - `annual.py`'s `create_headless_predbat()`/`write_minimal_apps_yaml()`). But since it lands in the same tree the live watcher walks and happens to share the filename, the watcher couldn't tell it apart from the real config file and force-restarted the live add-on mid-simulation.

- Extracted the file-collection logic into two testable functions: `resolve_apps_yaml_path()` (same `PREDBAT_APPS_FILE` resolution already used when the config is actually loaded, at `hass.py`'s `load_apps_yaml`) and `collect_watch_files()`, which now only matches an `apps.yaml`-named file if its resolved path equals that one real path - any other file sharing the name elsewhere in the tree is correctly ignored. `.py` source files (and dotfiles being excluded) are unaffected.
- New test (`test_hass_watcher.py`) directly reproduces the bug scenario: a real `apps.yaml` at the root plus a decoy `apps.yaml` in a scratch subdirectory, confirming only the real one is watched.

**Scope note**: this fixes the *trigger* for the restart reported in #4397/#4396. It does **not** fix the separate, pre-existing "New install detected" / `reset_inverter` race in `userinterface.py`'s `load_user_config()` that actually caused the reporter's settings loss once the unplanned restart happened - that race isn't specific to the Annual tool and could in principle be hit by any abrupt restart (crash, OOM, etc.), so it deserves its own more careful fix rather than a quick patch bundled in here. Tracked as a follow-up.

## Test plan

- [x] `./run_pre_commit` (via `coverage/run_pre_commit`) - all hooks pass
- [x] New `test_hass_watcher` - reproduces the exact bug scenario, confirms the real apps.yaml is watched, a same-named scratch file elsewhere is not, source files still are, dotfiles never are
- [x] Full quick suite (`./run_all --quick`) passes

Addresses #4397, #4396

### DIFF
--- a/apps/predbat/hass.py
+++ b/apps/predbat/hass.py
@@ -58,6 +58,7 @@ def collect_watch_files(roots, apps_file_path):
     write_minimal_apps_yaml()) would otherwise be indistinguishable from the real config file,
     forcing an unwanted restart of the live instance whenever they run (#4397, #4396).
     """
+    apps_file_path = os.path.abspath(apps_file_path)
     py_files = []
     seen_files = set()
     for root_dir in roots:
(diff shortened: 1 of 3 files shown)


chalfontchubby:
Good catch, fixed - `collect_watch_files()` now explicitly appends the resolved `apps.yaml` path after the walk if it exists on disk and wasn't already found, so it's always watched regardless of whether `PREDBAT_APPS_FILE` happens to point somewhere the roots would have reached. Added tests for both that case and a configured-path-does-not-exist case. All tests + pre-commit pass.

PR #4425, opened 2026-08-02
fix(plan): let PV surplus charge the battery during an export freeze

### LINKED ISSUE #4207: FoxESS Freeze Export mode predicts no battery charge, but the battery does charge
**Describe the bug**
Predbat expects that in Freeze Export mode all available PV generation is used either by the house load or sent to the grid. It expects that if there is more generation than is covered by load or export, it is lost/clipped.

Image

**Expected behaviour**
For FoxESS (and probably other) inverters, in freeze export ("feed-in first") mode, generation above the load and export limits will be stored in the battery and will therefore increase the SoC. Predbat should plan for this.

**Predbat version**

v8.44.5

**Environment details**

- Inverter and battery setup
    - FoxESS KH9 inverter with 18kWh battery storage
    - Connected using modbus integration
    - HAOS install

**Log file**

[predbat.log](https://github.com/user-attachments/files/29791267/predbat.log)

**Predbat debug yaml file**

[predbat_debug.yaml.txt](https://github.com/user-attachments/files/29791290/predbat_debug.yaml.txt)

chalfontchubby:
## Summary
- Fixes #4207: during an export freeze, `prediction.py` unconditionally zeroed the battery charge rate, ignoring `inverter_can_charge_during_export`. On inverters that route PV surplus to the battery during a freeze (e.g. FoxESS "Feed-in First"), Predbat's plan predicted no charging while the inverter actually charged the battery from surplus PV.
- Only force the charge rate to zero when the inverter genuinely can't route surplus PV to the battery during export freeze (`inverter_can_charge_during_export == False`), matching the behaviour already used for the partial export-limit-exceeded case elsewhere in the same file.
- Mirrored the fix in `prediction_kernel.cpp` and bumped `PK_PARITY_REVISION`/`KERNEL_PARITY_REVISION` per the kernel's parity contract.

## Test plan
- [x] Added `battery_discharge_freeze_pv_charge` / `_hybrid` regression scenarios in `tests/test_model.py` asserting SoC rises from PV surplus during a freeze with the default `inverter_can_charge_during_export=True`
- [x] Updated the 3 existing freeze scenarios to set `inverter_can_charge_during_export=False`, preserving their original "inverter genuinely can't charge" intent
- [x] `./run_all --test kernel_parity` passes (150-seed Python/C++ fuzz parity suite)


springfall2008:
I'm confused, why does this not just stop export freeze from doing anything at all?

chalfontchubby:
> I'm confused, why does this not just stop export freeze from doing anything at all?

That's a reasonable point that I maybe let Claude get carried away with. The alternative raised on the initial issue was perhaps to mark fox as not capable of a genuine freeze - since it does seem to charge I think. 

Fair to say I'm more confused than you most of the time.

chalfontchubby:
Good challenge - dug into it properly and you're right, not just semantically.

Once this fix is applied, `export_limit_now == 99` (freeze) and `== 100` (idle/off) run through identical code in `prediction.py` for an inverter with `inverter_can_charge_during_export=True` - both fail the same `< 99.0` check that gates genuine active discharge-to-export, and both fall through to plain self-use behaviour (PV charges the battery, load draws from it normally as needed). There's no remaining code path that tells them apart. So for FoxESS-type inverters this fix makes "freeze export" a pure relabelling with zero distinguishing effect from doing nothing at all - which is exactly your point: it's not freeze export anymore.

Cleaner alternative, taking your suggestion: gate `allow_freeze` in `optimise_export()` (`plan.py`) - currently only checks `self.set_export_freeze` - on `not inverter_can_charge_during_export` too, so the optimiser never offers/selects freeze for these inverters in the first place. That lets me revert `prediction.py` and the C++ kernel entirely (no parity bump, no kernel-binary rebuild), and stops the plan from showing a "FrzExp" label on a slot that isn't really frozen.

One wrinkle before I go do that: `optimise_solar()` ("Export more solar") sets `export_limits_best[window_n] = 99.0` directly on idle windows to avoid wasting PV, bypassing `optimise_export`'s search entirely - so it would need the same capability gate, or it'd quietly reintroduce this bug through the back door once `prediction.py` is reverted. Digging into it, I'm also not convinced that pass does much for *non*-capable inverters either under the current model (freeze and idle already trace through the same discharge-side logic regardless of capability) - but that feels like a separate question, not something to fold into this fix as a side effect.

Want me to rework #4425 along those lines (gate in `optimise_export` + `optimise_solar`, revert the prediction-model changes), or is there a simpler angle you'd rather take?

chalfontchubby:
Closing this in favour of #4538 - it's gotten stale (merge conflicts against `main`'s kernel churn) and, per the discussion in this thread, doesn't actually resolve the underlying problem: once the prediction model is patched to simulate PV-charging during freeze on a capable inverter, freeze (99%) and idle (100%) become indistinguishable in the model, so "Freeze Export" stops meaning anything different from doing nothing at all - the plan would still show a `FrzExp` slot that isn't achieving a genuine freeze.

#4538 takes the alternative raised here instead: stop the optimiser offering/selecting freeze for inverters that can't actually achieve it (`inverter_can_charge_during_export=true`), rather than patch the model to simulate a distinction that doesn't exist in reality. No prediction.py/kernel changes needed as a result.

Originating issue: #4207.

Issue #4432, opened 2026-08-03
Freeze charge/export: two related but distinct bugs, need direction before landing fixes (#4209, #4424, #4207, #4425)


chalfontchubby:
@springfall2008 @gcoan - trying to pull three threads together (#4209, #4424, #4207/#4425) since they've ended up tangled and I want to check I've understood things before pushing anything further.

## Two different problems, both about "does this inverter genuinely support freeze"

**#4424 (charge freeze) - the primitive is missing, not misbehaving**

itm1960's setup (#4209) has `inverter: {support_charge_freeze: true, support_discharge_freeze: true}` in his custom Fox template, but no `charge_freeze_service` or `discharge_freeze_service` configured at all - absent from his apps.yaml entirely. When the plan legitimately targets a charge freeze (`charge_limit_best[0] == reserve`, a real optimiser decision), `adjust_charge_immediate()` tries `charge_freeze_service`, finds nothing, and silently falls back to `charge_start_service` - a real active charge, re-anchored to current SoC every cycle (his log: SoC 13%->17%->21% in 10 minutes while "freeze charging").

Confirmed this is the plain optimiser-driven path for his case specifically: `num_cars: 0`, `iboost_enable: false`, which rules out the other way `adjust_charge_immediate(freeze=True)` can be called (`carHolding`/`boostHolding` at `execute.py:613-616`, which calls it directly, independent of `set_charge_freeze` or the optimiser). That second path is a real gap this fix doesn't close - flagging it now rather than after the fact, since gcoan's proposed direction ("operate as if `set_charge_freeze` is off") stops the *optimiser* from choosing freeze, but doesn't stop that direct call if a user has cars/iBoost enabled. Low hanging fruit - get the user to setup `charge_freeze_service` - but there is still perhaps a (theoretical?) risk that should be fixed(?)

**Fix (not yet pushed, sitting locally)**: mirror the existing `inv_support_charge_freeze` capability gate in `execute.py` (`fetch_inverter_data`) - if the inverter type supports charge freeze but no `charge_freeze_service` is actually configured, disable `set_charge_freeze` the same way an unsupported inverter type already does. `support_charge_freeze` itself stays `true` (Fox genuinely can do this) - the real fix for itm1960 is configuring the service properly (daviessm's #4207 template maps it to Fox's "Back-up" mode, if that's confirmed to work as a real hold - we don't actually know that yet either).

**#4207/#4425 (export freeze) - the primitive works, just not the way the model assumed**

[...]

Issue #4424, opened 2026-08-02
Freeze charging falls back to an active charge_start_service when no charge_freeze_service exists, causing real grid-drawing charge


chalfontchubby:
## Summary

When `charge_freeze_service` isn't configured for an inverter type/integration, Predbat's freeze-charge logic silently falls back to a genuine, active `charge_start_service` call instead - the same command used for a real charge, just with `target_soc` set to whatever SoC happens to be right now. This can and does draw real grid power during what's supposed to be a passive "hold flat" freeze, at a rate the user believes they aren't paying for.

Split out from #4209 - the original report there ("charging from grid when solar forecast is enough") was a different, already-answered question (arbitrage). This is a distinct, structural bug found while digging into a later follow-up on that same issue.

## Root cause

`adjust_charge_immediate()` (`apps/predbat/inverter.py:2787-2789`):

```python
if target_soc == self.soc_percent or freeze:
    if not self.call_service_template("charge_freeze_service", service_data, domain="charge", extra_data=extra_data):
        self.call_service_template("charge_start_service", service_data, domain="charge", extra_data=extra_data)
```

`call_service_template()` returns `False` when it can't resolve the named service for the configured integration (see the related discussion on #4329). When that happens here, the fallback is `charge_start_service` - **not** a no-op, not a passive hold, but the exact call used for an actual charge, with a real `power` figure (the configured charge rate) attached.

## Evidence (from #4209's reporter, FoxESS KH7 via the generic Modbus/entity integration, `has_service_api: false`)

Log sequence as "freeze charging" (target = current SoC) kicks in:

```
10:00:07 Wrote 2600 to charge_rate, successfully now 2600.0
10:00:17 Wrote 0 to discharge_rate, successfully now 0.0
10:00:19 Calling charge_start_service -> select.foxess_kh7_work_mode = 'Force Charge'
```

Repeated every ~5-minute Predbat cycle, each time re-anchoring the target to "current SoC, whatever that now is." SoC across three consecutive cycles: **13% -> 17% -> 21%** - an 8-percentage-point rise in 10 minutes, not a held level. PV (~2kW) couldn't fully cover the ~2600W charge target, so the shortfall was pulled from grid - independently confirmed by the reporter against his Octopus account's own meter data (import >1kW, export <0.02kW over 90 minutes, despite >2kW of solar surplus at the time).

## Why this looks like a real bug rather than a config gap

[...]

PR #4435, opened 2026-08-04
fix(inverter): don't silently fall back to a real charge when charge freeze isn't available

chalfontchubby:
## Summary

Opening this as a straw man for discussion on #4432/#4424, not as a finished proposal - happy to rework based on direction from @gcoan / @springfall2008.

- `adjust_charge_immediate(soc, freeze=True)` tries `charge_freeze_service` and silently falls back to a real `charge_start_service` (targeting current SoC) when that service isn't configured for this setup, even though the inverter type generically supports charge freeze. Some inverters treat that as a fresh command each cycle and briefly ramp to full power, rather than a passive hold.
- First commit (`ded65145`): when no `charge_freeze_service` is configured, `set_charge_freeze` is now disabled the same way it already is for inverter types with no support at all - this covers the optimiser's planned charge-freeze windows.
- Second commit (`0503950`): that flag wasn't checked by the separate `carHolding`/`boostHolding` direct-call path (holding the battery SoC while a car charges or iBoost diverts surplus to the immersion) - widened the gate to cover that path too, since it hits the exact same fallback. Falls back to a plain charge-stop when no genuine freeze is available; the discharge-side hold already in place is sufficient on its own.

See discussion on #4432 for the fuller trace of the `boostHolding` path and why it's structurally identical to `carHolding`.

## Open questions for review
- Is disabling `set_charge_freeze` entirely (vs. some narrower per-call gate) the right granularity, or should this be surfaced to the user differently (e.g. a warning rather than silent disable)?
- Does "fall back to a plain charge-stop" match what's wanted for `carHolding`/`boostHolding` when freeze isn't available, or is there a better degraded behaviour?

## Test plan
- [x] `./run_all --quick` passes with no regressions
- [x] `./run_pre_commit` passes (ruff, black, cspell, markdownlint, full test suite)
- [x] New tests: `set_charge_freeze` narrowing when `charge_freeze_service` is/isn't configured and when the inverter type doesn't support freeze at all; `carHolding`/`boostHolding` falling back to a plain charge-stop when no freeze service is configured
- [x] Fixed a test-harness staleness bug this exposed: `run_execute_test()` wasn't re-deriving `set_charge_freeze` from raw config each scenario the way production does every cycle, so a capability narrowing from one scenario was leaking into every later one

### DIFF
--- a/apps/predbat/execute.py
+++ b/apps/predbat/execute.py
@@ -610,7 +610,15 @@ class Execute:
 
             # Charging/Discharging off via service
             if not isCharging and self.set_charge_window:
-                if carHolding or boostHolding:
+                if (carHolding or boostHolding) and self.set_charge_freeze:
+                    # Only attempt the extra charge-side hold when a genuine passive freeze is available
+                    # (self.set_charge_freeze already accounts for inverter support and charge_freeze_service
+                    # being configured, #4424/#4432) - otherwise adjust_charge_immediate(soc, freeze=True)
+                    # silently falls back to a real charge_start_service targeting the current SoC, which
+                    # some inverters treat as a fresh command each cycle and briefly ramp to full power,
+                    # producing repeated short full-rate import bursts instead of a passive hold. The
+                    # discharge-side hold set above (pause/rate/reserve) already prevents discharge, so
+                    # falling back to a plain charge-stop is safe.
                     inverter.adjust_charge_immediate(inverter.soc_percent, freeze=True)
                 else:
                     inverter.adjust_charge_immediate(0)
--- a/apps/predbat/execute.py
+++ b/apps/predbat/execute.py
@@ -814,6 +814,17 @@ class Execute:
                     self.log("Note: Inverter does not support discharge freeze - disabled")
                     self.set_export_freeze = False
                     self.set_export_freeze_only = False
+                elif not (inverter.inv_has_target_soc and inverter.inv_target_soc_used_for_discharge) and not self.args.get("discharge_freeze_service", ""):
+                    # Same class of bug as #4424/charge_freeze_service above, but on the export side,
+                    # and with a different fallback condition: adjust_battery_target() only writes a
+                    # target SoC during export when inv_target_soc_used_for_discharge is also set
+                    # (inverter.py:1899) - GE/GEC have a target SoC but target_soc_used_for_discharge is
+                    # False for them, so they get no passive-hold protection from it at all and depend
+                    # entirely on discharge_freeze_service. Without it, adjust_export_immediate() would
+                    # silently fall back to a real discharge_start_service call instead of a passive hold.
+                    self.log("Note: No discharge_freeze_service configured - discharge freeze disabled")
+                    self.set_export_freeze = False
+                    self.set_export_freeze_only = False
                 if not inverter.inv_support_charge_freeze:
                     # Force off unsupported feature
                     self.log("Note: Inverter does not support charge freeze - disabled")
--- a/apps/predbat/execute.py
+++ b/apps/predbat/execute.py
@@ -818,11 +818,15 @@ class Execute:
                     # Force off unsupported feature
                     self.log("Note: Inverter does not support charge freeze - disabled")
                     self.set_charge_freeze = False
-                elif not self.args.get("charge_freeze_service", ""):
-                    # The inverter type supports charge freeze in general, but this setup has no
-                    # charge_freeze_service configured - adjust_charge_immediate() would silently
-                    # fall back to a real charge_start_service call instead of a passive hold (#4424),
-                    # so treat it the same as an inverter type with no support at all.
+                elif not inverter.inv_has_target_soc and not self.args.get("charge_freeze_service", ""):
+                    # Inverters with a target SoC (GE, GEC, GEE, the cloud integrations, etc.) already
+                    # achieve a passive hold via adjust_battery_target() and never need
+                    # charge_freeze_service configured at all - GE's own default template doesn't set it.
+                    # Only inverters without that fallback (e.g. SIG, FoxESS) rely on charge_freeze_service
+                    # as their sole freeze mechanism - for those, a missing service means
+                    # adjust_charge_immediate() silently falls back to a real charge_start_service call
+                    # instead of a passive hold (#4424), so treat it the same as an inverter type with no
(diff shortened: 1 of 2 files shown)


apps/predbat/execute.py
```diff
@@ -810,6 +818,13 @@ def fetch_inverter_data(self, create=True):
                     # Force off unsupported feature
                     self.log("Note: Inverter does not support charge freeze - disabled")
                     self.set_charge_freeze = False
+                elif not self.args.get("charge_freeze_service", ""):
+                    # The inverter type supports charge freeze in general, but this setup has no
+                    # charge_freeze_service configured - adjust_charge_immediate() would silently
+                    # fall back to a real charge_start_service call instead of a passive hold (#4424),
+                    # so treat it the same as an inverter type with no support at all.
+                    self.log("Note: No charge_freeze_service configured - charge freeze disabled")
+                    self.set_charge_freeze = False
```
springfall2008, line 842: This seems wrong, won't it disable charge freeze for all inverters that doesn't use the service?

springfall2008 (dismissed):

apps/predbat/execute.py
chalfontchubby: Good catch - you're right, this was too broad. Checked the default templates: GE's own `givenergy_givtcp.yaml` never sets `charge_freeze_service` either, and neither do any of the cloud integrations (FoxCloud, GEC, GEE, TESLA, EnphaseCloud, DeyeCloud, SolaxCloud, SolisCloud, SIGCLOUD, GWMQTT) - they all hold via `adjust_battery_target()` (target SoC) instead, which never touches `charge_freeze_service` at all. As written this would have disabled freeze-charge for the default GivEnergy setup and every cloud inverter.

Rescoped to only inverters with no target-SoC fallback (`has_target_soc: False` - SIG, FoxESS), which is what #4424's actual reporter (FoxESS via generic Modbus) is. Also found and fixed the same gap on the discharge/export side while I was in there, since GE/GEC's target-SoC write is explicitly skipped for export (`target_soc_used_for_discharge: False`), so they had zero protection there either - separate commit with its own explanation, `32e639f4`.

Fix is in `1a2e5956` (charge side) + `32e639f4` (discharge side), tests updated to cover both the "has fallback" and "no fallback" cases for each.

gcoan: One additional comment, a 'Note:' message in the logfile won't get surfaced visibly in the predbat log viewer as 'note' isn't one of the keywords it looks for

Better to set to 'Warn:' so it appears in the Warn and Error tabs

gcoan:
> Is disabling set_charge_freeze entirely (vs. some narrower per-call gate) the right granularity, or should this be surfaced to the user differently (e.g. a warning rather than silent disable)?

I think it should be explicitly surfaced as a Warn in the logfile so its visible

apps/predbat/execute.py
```diff
@@ -806,10 +814,32 @@ def fetch_inverter_data(self, create=True):
                     self.log("Note: Inverter does not support discharge freeze - disabled")
                     self.set_export_freeze = False
                     self.set_export_freeze_only = False
+                elif not (inverter.inv_has_target_soc and inverter.inv_target_soc_used_for_discharge) and not self.args.get("discharge_freeze_service", ""):
```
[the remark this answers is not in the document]
chalfontchubby: Traced every call site that this gate feeds (`adjust_charge_immediate`/`adjust_export_immediate` with `freeze=True`, at lines 526-527, 536-609, and the carHolding/boostHolding block) - none of them have a rate-based fallback. `adjust_export_immediate()`/`adjust_charge_immediate()` are service-calls only; there's no branch that uses `inv_charge_discharge_with_rate` to freeze via rate=0 instead.

So for Huawei specifically: it does define `discharge_start_service`/`charge_start_service` (`huawei_solar.forcible_discharge_soc`/`forcible_charge_soc` in `templates/huawei.yaml`) - the real, active command these methods fall back to when the freeze service is missing. Your own suggested narrower condition ("only disable when the unsafe fallback is actually possible, i.e. discharge_start_service is configured") would still land on disabling it for Huawei, since that service *is* configured there.

Worth noting this gate is actually a fix, not a regression, for Huawei on both sides: before this PR, the only check was `inv_support_discharge_freeze`/`inv_support_charge_freeze` (both true for Huawei via its template), so the optimiser would happily plan freeze windows for it and execution would hit exactly the dangerous fallback this PR exists to close (a real `forcible_charge_soc`/`forcible_discharge_soc` call instead of a passive hold). Huawei still gets a safe hold via the existing reserve/pause "Hold charging"/"Hold exporting" path when freeze is disabled - it just isn't labelled/optimised specifically as "Freeze charging"/"Freeze exporting".

@copilot does this match what you were seeing, or is there a rate-based freeze path I'm missing somewhere?

chalfontchubby:
Think your change request is addressed in the response above

chalfontchubby:
Pushed 694cdf7b - the two new disable-path log lines now use `Warn:` instead of `Note:` per @gcoan's ask, so they surface in the log viewer's Warn/Error tabs.

apps/predbat/execute.py
springfall2008, line 817: target_soc_used_for_discharge is often false, surely this throws the normal GE (e.g. GivTCP) to no longer support export freeze

Why not just add an inverter capability setting?

springfall2008 (changes requested):

chalfontchubby:
Closing in favour of #4693, which fixes this at the point of failure instead.

Your review comment was right, and the second version had the same problem: inferring "can this inverter freeze?" from whether a service key happens to be set disabled export freeze for stock GivEnergy/GivTCP (`GE`/`GEC`/`GEE`, and also `GS_fb00`/`SA`), all of which reach freeze export through the safe local rate=0/pause path in `execute.py` and never touch a service at all.

Stepping back, the branch was on the wrong axis. The fault isn't about which inverters may *plan* freeze windows - it's that `adjust_charge_immediate()`/`adjust_export_immediate()` fall back to a real `charge_start_service`/`discharge_start_service` when the freeze service is missing, i.e. an active command where a hold was asked for. #4693 changes that fallback to a plain stop (the hold is already established by the caller before those are reached), which is a couple of lines plus tests and needs no capability inference at all.

Leaving this branch and its discussion in place rather than force-pushing over it.

Issue #4454, opened 2026-08-07
Sigenergy: Predbat and Axle contend for the inverter — VPP mode evicts Axle's NBI dispatch


mgazza:
## Summary

On a Sigenergy, Predbat's VPP mode and Axle's NorthBound Interface (NBI) are **mutually exclusive** — the inverter accepts one controller at a time. Today Predbat wins that contest by accident rather than by design, and evicts Axle mid-dispatch.

`_manage_vpp_registration()` in `apps/predbat/sigenergy.py` runs on the 5-minute poll tick and, whenever the system is not in VPP and the read-only switch is off, switches it straight back to VPP. It has no awareness of Axle at all (`grep -ci axle apps/predbat/sigenergy.py` → 0).

## Evidence

From a live system on 2026-08-06 (identifiers redacted — this is a public repo). Axle had an export event scheduled 20:00–21:00 local, 100 p/kWh, which Predbat had already ingested as its own export window via `load_axle_slot()`.

Operating-mode timeline:

```
19:35:18  ->  Northbound Integration      (Axle takes the inverter)
19:40:12  ->  VPP                          (back to Predbat)
```

Exactly one `SIGENERGY_POLL_INTERVAL`. Meanwhile, once per minute:

```
19:35:54 Warn: SigenergyAPI: System <redacted> is not in VPP mode
         (Northbound Integration) — controls skipped until onboard is approved
... 19:36:54, 19:38:01, 19:39:01, 19:40:01
```

The user's report was "it changed to NBI Scheduling and then went back to VPP" — with the reasonable follow-up question of why Axle appeared to own a mode that is actually Predbat's.

## Two problems

**1. Nobody decided who owns the inverter.** Predbat wins on a 5-minute timer. It's not obviously wrong — Predbat plans the Axle session as its own export window, so the event still delivers — but it's a race, it's silent, and it overrides whatever Axle had scheduled.

**2. `axle_control` exists and doesn't reach this component.** `Fetch.fetch_config_options()` (`apps/predbat/fetch.py:2408`) already raises `set_read_only_axle` when `axle_control` is set and an Axle event is active. But that flag is only consumed in `execute.py:63` and `output.py:955`. `sigenergy.py` reads the *user-facing* `switch.predbat_set_read_only` entity instead, so even with `axle_control: true` the component would still reclaim VPP and evict Axle. **The stand-down feature does not work on Sigenergy.**

## Side effect worth fixing either way

[...]

PR #4455, opened 2026-08-07
Sigenergy: make Predbat's ownership of the inverter deliberate (option A for #4454)

mgazza:
> **Option A of two mutually-exclusive alternatives for #4454. Only one of this and the sibling PR should merge.**

## What this does

Keeps Predbat as the owner of a Sigenergy inverter, but makes that a decision rather than a race.

A Sigenergy accepts one controller at a time, and VPP mode (Predbat) and the NorthBound Interface (Axle's dispatch channel) are mutually exclusive. Today `_manage_vpp_registration()` runs on the 5-minute poll tick and silently switches the system back to VPP, so an Axle dispatch can hold the inverter for up to 5 minutes before being displaced with no indication of what happened.

Observed live on 2026-08-06: mode went to Northbound Integration at 19:35:18 and back to VPP at 19:40:12 — exactly one poll interval — while the log only said "controls skipped until onboard is approved".

## Changes

- **Reclaim VPP every 60s** (`SIGENERGY_VPP_RECLAIM_INTERVAL`) instead of every 300s, so control is never ambiguous for long. `set_operating_mode()` is an MQTT publish and only fires when the mode is actually wrong, so this costs nothing against the REST rate limit. The `onboard_status` cache write stays on the 300s cadence.
- **Log the reclaim explicitly**, naming the controller being displaced, instead of a generic warning.
- **Stop reporting a contended system as `pending_approval`.** Downstream UIs render that as an amber "approve this in the Sigenergy app" banner, so every Axle event told the user to approve something needing no approval. A system in NBI is fully onboarded.
- **Expose `contended_by`** on the onboard-status sensor so support can tell contention from a genuine onboarding failure.

Predbat already ingests Axle sessions as its own export windows (`load_axle_slot()`), so events still run — under Predbat's plan rather than Axle's dispatch.

## Trade-off

This is the "Predbat always wins" position. It's self-contained and needs no config change, but it does override Axle's schedule on every event. If Axle credits delivery from its own NBI dispatch rather than from meter data, this costs the customer the payment even though the energy flowed. That question is open in #4454.

## Tests

4 new tests in `tests/test_sigenergy.py`: reclaim-from-third-party, contention-not-pending-approval, reclaim-cadence, and the controls-skipped message. Full Sigenergy suite 72 pass / 0 fail; `unit_test.py --quick` passes with 0 failures. `run_pre_commit` clean (ruff, black, cspell).

### DIFF
--- a/apps/predbat/sigenergy.py
+++ b/apps/predbat/sigenergy.py
@@ -132,6 +132,7 @@ SIGENERGY_CODE_DEVELOPER_NOT_APPROVED = 1604
 SIGENERGY_TOKEN_EXPIRY_BUFFER = 600   # refresh token 10 min before expiry
 SIGENERGY_MIN_REQUEST_INTERVAL = 6.0  # enforce ≥10 req/min API limit
 SIGENERGY_POLL_INTERVAL = 300         # realtime data poll every 5 minutes
+SIGENERGY_VPP_RECLAIM_INTERVAL = 60   # re-assert VPP ownership every minute (see _manage_vpp_registration)
 SIGENERGY_DEVICE_POLL_INTERVAL = 1800  # device list refresh every 30 minutes
 SIGENERGY_RATE_LIMIT_BACKOFF = [15, 30, 60, 120, 480]  # seconds to wait after code 1201
 SIGENERGY_BATTERY_NOMINAL_VOLTAGE_V = 28.8  # 8S LiFePO4 pack: 8 × 3.6V; used to convert ratedEnergy (Ah) → kWh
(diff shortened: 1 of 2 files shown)


mgazza:
Customer-impact tracking and fleet scope: Predictive-Cloud-Ltd/predbat-saas#1591 (private). Alternative approach: #4456.

mgazza:
Superseded by #4457.

These two were opened as mutually-exclusive alternatives — Predbat keeps the inverter, or Axle does. That framing was wrong. A code review found that this PR's faster reclaim actively defeats the `axle_control` option, which meant option A had to honour `axle_control` regardless — and once it does, it contains option B. They were complementary all along, not alternatives.

The real either/or was never in the code: it is the policy question of whether `axle_control` should be enabled, which #4454 now tracks separately.

#4457 combines both, plus fixes for the three review findings (the stale `set_read_only_axle` flag on restart, the reclaim defeating `axle_control`, and the contention marker being cleared before it was ever published).

PR #4462, opened 2026-08-08
feat(plan): add a pv90 upside forecast scenario to balance the one-sided pv10 hedge

springfall2008:
## What this adds

A third simulated forecast scenario, **pv90** (high PV, low load), alongside the existing nominal (p50) and pessimistic pv10 cases.

Predbat previously simulated only two futures. Risk was therefore one-sided: the optimiser was rewarded for hedging against a bad solar day and never charged for a hedge that turned out to be unnecessary. pv90 is the missing counterweight.

**It ships off.** A `calculate_pv90_plan` switch (default off) gates the whole feature. The switch is deliberately available **without** expert mode, so the scenario can be tested across as many real systems as possible; the two settings that tune it stay expert-only, so everyone who turns it on runs the same values and their feedback is comparable. `pv_metric90_weight` shows its real default of `0.15` in Home Assistant, but while the switch is off the runtime value is forced to `0.0`, at which no pv90 simulation runs and the pv90 term collapses out of the metric. Everything routes through that single weight — there is no second gating path. Plans with the switch off are byte-identical to before, verified against `predbat_debug_agile1.yaml` (md5 `d525d171…`).

## Honest summary of what it achieves

**It does not fix the case it was built for**, and the measurement explains why.

The motivating case (`coverage/predbat_no_export_100.txt`, 9.52 kWh battery, 0p export, 6.9p/28.85p import) charges to 100% overnight while carrying ~40% at `end_record`. The metric is dead flat from 66% to 100%, because the residual battery credit at `end_record` is `6.9 / 0.96 / 0.97` = 7.4098 p/kWh — exactly the marginal cost of storing a kWh at 6.9p. With the mid case flat, the decision fell to the pv10 hedge.

Measured metric slope across that plateau:

| scenario | slope | |
|---|---|---|
| nominal | +0.000p | the plateau |
| pv10 | **−63.035p** | pulls to 100% |
| pv90 (best achievable) | **+17.2p** | pushes down |

So the pivot is `w90 > 3.66 × w10`, confirmed at three weights. pv90's slope saturates around 17p against a 24p plateau width, so even a theoretically perfect pv90 tops out below pv10's 63p — at equal weights it can never win on this tariff.

That ratio is not a modelling artefact. In the pv10 world every carried kWh avoids importing at **28.85p**; in the pv90 world a wasted kWh cost **6.9p**. A 4.2:1 asymmetry in consequence. Charging to 100% is defensible insurance here.

What the branch does deliver: the p90 pipeline end to end, `forecast90` now published by Solcast, a symmetric risk model ready for real p90 data, and precise measurement of the flat plateau — which points at the residual-battery valuation as the more promising next target.

## Changes

- **Scenario selector** — the boolean `pv10` flag becomes a three-valued `pv_scenario` integer through `plan.py`, `prediction.py` and the kernel. `PV_SCENARIO_PV10 == 1` keeps it numerically compatible with the old boolean.
- **p90 data pipeline** — `solcast.py` builds and publishes a real p90 minute series as a `forecast90` sensor attribute; `fetch_pv_forecast()` reads it, falling back to a copy of p50 when absent. No upside is synthesised.
- **Kernel** — `pv90`/`load90` arrays added to `PkContext`, `PkScenario.pv10` becomes `pv_scenario`, ABI 2→3 and both parity revisions bumped. All six platform binaries rebuilt.
- **Metric** — the downside-only clamp is replaced by a signed weighted average across all three scenarios. This is required: pv90 is nearly always cheaper than nominal, so a one-sided clamp would contribute exactly zero at every candidate.
- **Config** — `calculate_pv90_plan` (switch, default off, not expert-gated), plus `pv_metric90_weight` (default 0.15) and `load_scaling90` (default 0.7), both expert-gated. All three documented in `docs/customisation.md` and `docs/apps-yaml.md`.

`load_scaling90` (default `0.7`) is a plain absolute multiplier, like `load_scaling10`. To stop any configuration inverting a scenario, the three are clamped at config-read time so the ordering always holds:

```python
load_scaling90 = min(load_scaling90, load_scaling10, load_scaling)
load_scaling10 = max(load_scaling90, load_scaling, load_scaling10)
```

That guarantees `load_scaling90 <= load_scaling <= load_scaling10` whatever the user sets, so pv90 is always the low-load case and pv10 always the high-load case. It also closes a latent inversion on the pv10 side that predates this branch: a user with `load_scaling` above `load_scaling10` previously got a pv10 with *less* load than nominal. Clamping is logged when it changes a value, and is a no-op at the defaults.

## Behaviour change for existing users

Removing the pv10 clamp affects every user, not just those who enable pv90: where `metric10 < metric` the metric is now a true weighted average rather than being pinned at `metric`. Measured exposure is small (1 of ~3538 candidates on `agile1`, 9 of ~6082 on `pre_saving1`) and plan-neutral across the debug suite, but it is not gated on `pv_metric90_weight`.

## Testing

- New `tests/test_pv90.py`, registered in `TEST_REGISTRY`.
- `tests/test_kernel_parity.py` covers pv90; the random sweep now runs all three scenarios per seed (150/150/150, previously 57/58/35 after an accidental swap).
- Kernel struct layout verified by compiled `offsetof`/`sizeof` against the ctypes definitions — all 112 field offsets byte-identical.
- Full suite: 188 pass. `debug_cases/predbat_debug_pre_saving1.yaml` fails, but fails identically at the merge base with byte-identical output — pre-existing, not introduced here.

## Known follow-ups

- `test_pv90_fallback_tracks_a_reassigned_p50` is not hermetic; it leans on a one-minute margin against state leaked by a sibling test. Can only cause a missed discard, never a spurious one.
- A fifth weight-0 gate at `plan.py:863` inside `run_prediction_metric` is not covered by the skip-gate test. Planning-cost only; plan output is unaffected.
- Whether `calculate_pv90_plan` should default on, and whether 0.15 is the right weight, should be decided against a debug dump carrying real Solcast p90 data rather than the p50 fallback.
- The `debug_cases` replay harness has `fetch_config_options()` commented out, so the load-scaling clamp does not fire on that path. Replays therefore use the raw configured values. Worth knowing before running any further measurement through `run_single_debug`.


springfall2008:
@copilot resolve the merge conflicts in this pull request - just discard the conflicting kernel changes as the kernel will be re-build

PR #4469, opened 2026-08-09
fix(fetch): warn loudly when car_charging_energy is configured but has no data

chalfontchubby:
## Summary
Follow-up to #4465/#4458. `car_charging_hold` is on by default, and its precise per-minute car-energy subtraction path silently falls back to the load-threshold heuristic whenever `car_charging_energy` resolves empty - previously only a benign, error-free `Unable to fetch history` log line, easy to miss.

This affects any Ohme HA in-built-integration user still pointed at the energy sensor HA 2026.8.0 removed (see #4465) - it degrades load-forecast accuracy on every car-charging day rather than just showing one wrong display value.

- Traced the failure path: `minute_data_import_export()` already handles a missing/removed entity gracefully (empty history → benign warn, no `had_errors`), so `self.car_charging_energy` ends up `{}` rather than crashing - by design, this already falls through to the existing `car_charging_threshold` heuristic. The gap was purely visibility.
- Added a distinct, error-flagged warning in `load_car_energy()` specifically for "configured but empty" (not "not configured at all", which already logs its own message and is a normal, expected setup for non-EV users).

## Test plan
- [x] New `test_load_car_energy.py`: warns + flags error when configured-but-empty, stays silent with real data, stays silent when `car_charging_hold` is off
- [x] Full `./run_all --quick` suite passes

### DIFF
--- a/apps/predbat/fetch.py
+++ b/apps/predbat/fetch.py
@@ -2569,19 +2569,25 @@ class Fetch:
             self.car_charging_energy = self.minute_data_import_export(self.max_days_previous, now_utc, "car_charging_energy", scale=self.car_charging_energy_scale, required_unit="kWh")
             if self.car_charging_hold and not self.car_charging_energy:
                 # car_charging_energy is configured (an entity is set) but resolved to nothing at
-                # all - most likely the entity no longer exists (e.g. removed upstream, see #4458)
-                # rather than a transient gap. minute_data_import_export() already logs a benign
-                # "Unable to fetch history" warning per entity with no had_errors flag, so this
-                # degrades to the car_charging_threshold heuristic below completely silently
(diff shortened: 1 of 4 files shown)


springfall2008:
The only issue here is what if you want this, you get a warning every time?

chalfontchubby:
Fair point - pushed a fix. The warning now fires once per incident rather than every cycle (throttled via a `car_charging_energy_warned` flag, reset when data loads successfully again), so a persistent-but-intentional setup (recorder-excluded entity, fresh install with no history yet) won't get spammed every 5 minutes. The status sensor (`record_status`/`had_errors`) still updates every cycle so it accurately reflects an ongoing degraded state - only the log line itself is throttled.

PR #4489, opened 2026-08-11
fix(gateway): pin EV charger entity ids to the charge point id

mgazza:
## Problem

`GatewayMQTT._ev_suffix()` returned a bare `"ev"` when a single charger was visible, and `"ev_<id>"` only when more than one was. The suffix therefore depended on **how many chargers happened to appear in a single status message**, not on the charger's identity.

That makes entity ids move under the user. A site with one charger produces `sensor.predbat_gateway_ev_power`, but the moment a second charger appears — or the same charger is briefly seen twice across a reconnect — everything silently renames to `sensor.predbat_gateway_ev_<id>_power`, orphaning the old entities and every dashboard and automation bound to them.

Observed in production: one hub accumulated **both** a `predbat_gateway_ev_536212_*` set and a `predbat_gateway_ev_*` set for the same physical charger.

## Fix

Always identify the charger by the last 6 chars of its OCPP charge point id, falling back to `"ev"` only when the charger has not reported one yet (no `BootNotification`). The `multi` parameter is retained for call-site compatibility but no longer affects the result.

## Impact

Renames EV entities on single-charger sites from `sensor.predbat_gateway_ev_*` to `sensor.predbat_gateway_ev_<id>_*`.

These entities only exist on PredBat Gateway hubs running an OCPP charger — at the time of writing that is two hubs, one of them a test rig — so no migration path is warranted. Any dashboard bound to the unsuffixed names needs repointing, and the old entities are left orphaned.

## Testing

`pytest apps/predbat/tests/test_gateway.py` — **266/266** pass, including three new tests:

- `test_ev_suffix_is_stable_regardless_of_charger_count` — regression cover for the rename
- `test_ev_suffix_falls_back_when_id_unknown`
- `test_disconnected_charger_reports_offline` — a charger the gateway reports as `connected=false` must surface as offline rather than freezing at its last values

That last one pairs with a gateway-firmware change: previously the firmware **omitted** disconnected chargers from telemetry entirely, so these entities froze and the UI showed "connected" for two days during an EV-charging outage. The firmware now sends `connected=false`, which this code already maps correctly to `binary_sensor.*_online`.

### DIFF
--- a/.cspell/custom-dictionary-workspace.txt
+++ b/.cspell/custom-dictionary-workspace.txt
@@ -439,6 +439,10 @@ sigencloud
 sigenergy
 sigenstor
 Slee
+slugified
+slugifies
+slugify
+slugifying
 snakeviz
 socb
 socketloop
--- a/apps/predbat/gateway.py
+++ b/apps/predbat/gateway.py
@@ -946,18 +946,30 @@ class GatewayMQTT(ComponentBase):
     def _ev_suffix(ev, multi):
         """Build the entity suffix for an EV charger.
 
-        Single charger (the v1 case) uses a stable "ev" suffix; with more than one
-        charger present, disambiguate by the last 6 chars of the OCPP charge point id.
+        Always identifies the charger by the last 6 chars of its OCPP charge point
+        id, so a given charger keeps the same entity ids for its whole life.
+
+        This used to return a bare "ev" whenever only one charger was present, and
+        only disambiguate when several were. That made the suffix a function of how
+        many chargers happened to be visible in a single status message, so entity
+        ids moved under the user: a site with one charger produced
+        ``sensor.predbat_gateway_ev_power``, but the moment a second charger
+        appeared — or the same charger was briefly seen twice across a reconnect —
+        everything silently renamed to ``sensor.predbat_gateway_ev_<id>_power``,
+        orphaning the old entities and every dashboard and automation bound to
+        them. Pinning to the charge point id removes that whole class of churn.
+
+        Falls back to "ev" only when the charger reports no id at all (a charger
+        that has not completed its BootNotification yet).
 
         Args:
             ev: An EvCharger protobuf message.
-            multi: True when more than one charger is present in the status.
+            multi: Unused; retained for call-site compatibility. The suffix no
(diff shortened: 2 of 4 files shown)


mgazza:
Pushed a follow-up commit from an adversarial review of this change. Four issues, all now fixed and covered by tests:

**1. Charge point ids went into entity ids unsanitised.** An OCPP `chargePointSerialNumber` is an arbitrary vendor string — spaces, dots, slashes, dashes, non-ASCII — none legal in a Home Assistant entity id, and nothing downstream sanitised it. `CP/AB-123` produced `sensor.predbat_gateway_ev_ab-123_power`, which HA rejects, while auto-config still pointed `car_charging_*` at that nonexistent namespace. The id is now slugified before truncation.

**2. Disconnected chargers could make PredBat plan a charging slot.** Only the derived `connected` sensor was gated on `ev.connected`; `session_active`, power, session energy and the SoC fallback were published straight from the payload. `car_charging_now` is wired to `session_active` and PredBat plans whenever it is true — so a disconnected charger carrying a stale `session_active` would schedule charging for a charger that is not there. Now forced to idle values.

This one matters more since the paired firmware change: the gateway used to *omit* disconnected chargers and now reports them with `connected=false`, so this code starts receiving entries it previously never saw. Firmware zeroes these too, but PredBat and firmware version independently, so both sides gate.

**3. `_register_ev_car()` took `chargers[0]` with no connected check** — despite its docstring saying it registers a connected charger. With disconnected chargers now present, a stale slot-0 entry would bind all `car_charging_*` args to a charger that isn't there. It now prefers a connected charger, falling back to the first so a merely-offline charger still registers.

**4. `docs/car-charging.md`** still listed all ten entities under the old unsuffixed names.

Known limitation, documented in the docstring: a 6-char tail is not collision-proof — two chargers whose slugified ids share a tail map to one namespace. Identity stability was the priority here; anything collision-proof either depends on which other chargers are present (the churn this PR fixes) or is an opaque digest. Happy to revisit if you'd prefer a digest-based scheme.

Tests now 269 pass, including slugification cases, stale-slot registration, and assertions that a disconnected charger's `session_active`/`power`/`session_energy` are forced idle.

PR #4500, opened 2026-08-12
fix(solis): guard battery_scaling against a documented 0% SOH API response

### LINKED ISSUE #4494: [Solis Cloud] batteryHealthSoh of 0 from the API silently zeroes soc_max and causes unbounded grid charging
### Summary

`automatic_config()` binds `battery_scaling` to the Solis `battery_soh` sensor. When the Solis Cloud API returns `batteryHealthSoh: 0` — which is a documented, valid response — `soc_max` becomes zero, Predbat falls back to an 8 kWh default, and `best_soc_keep` becomes permanently unreachable. The result is continuous grid charging at any price. In my case 24 kWh was imported during the most expensive hours of the day before I noticed.

The same zero also makes `automatic_config()` conclude there is no battery at all, which leaves `load_today` and `charge_start_time` unset and puts Predbat into a crash loop it cannot recover from without a restart.

### Environment

- Predbat v8.47.7 (add-on 1.7.1)
- Home Assistant 2026.8.1 / HAOS 18.2
- Inverter: Solis S6-EH3P20K-H via Solis Cloud API
- Battery: Dyness HV, 51.2 kWh
- apps.yaml: `soc_max: 51.2`, `battery_scaling: [1.0]`, `best_soc_keep: 15.36 kWh`

### Timeline (2026-08-12)

| Time | Event |
|---|---|
| until 05:24:30 | `battery_soh` = 1.0, `battery_voltage` = 524.6 V, battery discharging normally overnight (44% -> 40%) |
| **05:24:30** | `battery_soh` -> **0.0** and `battery_voltage` -> **0.0** in the same second. `battery_soc` keeps updating correctly |
| 05:30 | charge window written to the inverter: 05:30-19:45, target 100% |
| 05:24 - 07:45 | SoC climbs 40% -> 87%, ~24 kWh imported at 28-35 c/kWh |
| later | `automatic_config()` re-runs, logs "No inverters with batteries found", Predbat crash loops |

The Solis Cloud API had been intermittently unhealthy in the preceding hours (`B0600 "Datalogger returns data abnormally"`, `B0173 "The device has timed out"`). A datalogger firmware update from Solis restored the missing fields.

### The failure chain

**1. `battery_scaling` is bound to the SOH sensor** — `solis.py:1262`

```python
self.set_arg("battery_scaling", [f"sensor.{self.prefix}_solis_{device}_battery_soh" for device in devices])
```

This is unconditional; `set_arg` has no check for a user-provided value, so `battery_scaling: [1.0]` in apps.yaml is silently discarded.

**2. `soc_max` collapses to zero** — `inverter.py:383`

```python
self.soc_max = self.nominal_capacity * self.battery_scaling   # 51.2 * 0 = 0
```

**3. Fallback to 8 kWh** — `inverter.py:632`

```python
if not self.soc_max or self.soc_max <= 0:
    self.log("Warn: Unable to determine battery size for inverter {}, using 8 kWh default ...")
    self.soc_max = 8.0
```

**4. `best_soc_keep` becomes unreachable** — `prediction.py:1141`

```python
if best_soc_keep > 0 and soc <= best_soc_keep:
    metric_keep += (best_soc_keep - soc) * import_rate * keep_minute_scaling * step / 60.0
```

`best_soc_keep` is 15.36 kWh but the modelled battery is now 8 kWh, so `soc <= best_soc_keep` is **always true**. Predbat is permanently penalised for being "too empty" and the only way to reduce that penalty is to charge — regardless of price. `best_soc_max` (48.64 kWh) also stops binding, because `min(loop_soc, 48.64)` never triggers on an 8 kWh battery, so the target becomes 100%.

**5. Config is skipped entirely** — `solis.py:1240-1252`

```python
battery_soh = float(battery_soh) / 100.0
...
if battery_soh:                  # 0.0 is falsy
    batteries.append(inverter_sn)
...
if num_inverters == 0:
    self.log("Warn: Solis API automatic_config: No inverters with batteries found, skipping configuration")
    return
```

Because it returns early, `load_today`, `charge_start_time` and the rest are never set, producing:

```
Error: You have not set load_today or load_forecast in apps.yaml
Error: Inverter 0 unable to read charge window time as neither REST, charge_start_time or charge_start_hour are set
ValueError -> timer_tick caught exception
```

on every cycle. Since `automatic_config()` only runs on `first`, this does not self-heal — a restart is required, and the restart fails again while SOH is 0.

### Why 0 is a legitimate API response

The SolisCloud Platform API Document V2.0.3 lists `batteryHealthSoh` as **not required** (column "N"), and the official example response on page 30 shows:

```json
"batteryCapacitySoc": 0.000,
"batteryHealthSoh": 0.000,
"batteryVoltage": 0.000,
```

So a zero here is within spec and must not be treated as a valid scaling factor.

### Suggested fix

1. Validate the SOH before using it as `battery_scaling` — reject 0, negative, or implausible values and fall back to 1.0 (or to the user's apps.yaml value) with a warning.
2. Do not let `automatic_config()` discard an explicit apps.yaml `battery_scaling`; treat user configuration as authoritative.
3. Guard `soc_max` independently: if `soc_max` is explicitly configured, a scaling factor of 0 should never be able to override it.
4. Consider whether a battery-related field being absent should really cause `automatic_config()` to abandon the entire configuration, given the unrecoverable crash loop that follows.

I have added a Home Assistant automation as a local safety net that sets `switch.predbat_set_read_only` when SOH reads 0 for more than five minutes. That works, but it seems like something Predbat should not require.

Happy to test a patch.

chalfontchubby:
## Summary

Fixes #4494 - a Solis Cloud API `batteryHealthSoh: 0` (documented, valid response, not "no battery") was being treated as falsy in `automatic_config()`'s battery detection, and once bound to the live SOH sensor, fed straight into `soc_max = nominal_capacity * battery_scaling`. That collapsed `soc_max` to 0, made `best_soc_keep` permanently unreachable, and forced continuous grid charging regardless of price - 24 kWh imported at peak rates in the reporter's case. Full root-cause trace with code references was provided by @jibbej.

- `inverter.py`: a `battery_scaling` reading of 0 or negative is ambiguous (flaky API response vs a genuinely unhealthy battery), so rather than asserting either extreme, retain the last value that was actually read as valid (mirrors the existing `soc_max_nominal` fallback pattern), falling back to 1.0 only if nothing valid has ever been read for that inverter.
- `solis.py` `automatic_config()`: a `battery_soh` of 0 no longer excludes the inverter the same way a genuinely missing field does. Previously this caused `automatic_config()` to abandon configuration entirely (`load_today`/`charge_start_time` left unset), which is what produced the crash loop that needed a restart to clear.
- The published SOH sensor still reports the raw value as-is (including a literal 0) - it's ambiguous, so it's reported honestly rather than guessed at, and the reporter's own stopgap HA automation watches this exact sensor for a literal 0.

## Open points - want reviewer input

The original report suggested 4 fixes. This PR covers 2 of them fully, 1 indirectly, and deliberately leaves 1 open:

1. **Validate SOH, reject 0/negative, fall back with a warning** - done (`inverter.py`).
2. **Don't let `automatic_config()` discard an explicit apps.yaml `battery_scaling`; treat user config as authoritative** - **not done**. A user's `battery_scaling: [1.0]` in apps.yaml still gets silently overwritten by the live sensor binding on every restart, same as before this PR. This affects every `set_arg` call across every cloud integration's `automatic_config()`, not just `battery_scaling`, so it felt like a separate design decision (auto-discovery vs. user override) rather than part of this specific failure chain - didn't want to make that call unilaterally.
3. **Guard `soc_max` independently against a 0 scaling factor** - covered *indirectly*: since `battery_scaling` itself can no longer be 0/negative, `soc_max` can't collapse via that multiplication anymore. No separate guard was added directly on `soc_max`.
4. **Should one missing battery-related field really abandon all of `automatic_config()`?** - only *partially* addressed. This PR fixes the specific bug that caused it here (0 was being treated identically to "field absent"). If `batteryHealthSoh` were genuinely absent from the API response for a real battery inverter, `automatic_config()` would still bail entirely and produce the same crash-loop-until-restart behaviour. The broader question of whether one missing signal should abandon config for the whole install is untouched.

Would appreciate a second opinion on whether 2 and the rest of 4 are worth doing here or as a follow-up.

## Test plan
- [x] New regression tests: `test_inverter.py::test_battery_scaling_invalid_value_clamped` (fallback to 1.0 with no history; retains a prior valid value across a bad reading), `test_solis.py::test_automatic_config` (inverter with `batteryHealthSoh: 0` still gets configured), `test_solis.py::test_publish_entities` (SOH of 0 still published as-is)

### DIFF
--- a/apps/predbat/component_base.py
+++ b/apps/predbat/component_base.py
@@ -93,6 +93,22 @@ class ComponentBase(ABC):
         """
         return self.base.set_arg(arg, value)
 
+    def set_arg_auto(self, arg, value):
+        """
+        Like set_arg(), but for auto-discovery code (typically automatic_config()) binding an
+        apps.yaml key to an auto-discovered entity/value. Auto-discovery still always wins - this
+        does not change that - but if the user had already set this key explicitly in apps.yaml,
+        silently discarding it left no way to notice (issue #4494 follow-up discussion, PR #4500).
+        Logs a one-time note per key when that happens, then behaves exactly like set_arg().
+        """
+        raw_args = getattr(self.base, "args_from_apps_yaml", None) or {}
+        raw_value = raw_args.get(arg)
+        warned = getattr(self.base, "apps_yaml_override_warned", None)
+        if raw_value is not None and raw_value != value and warned is not None and arg not in warned:
+            warned.add(arg)
+            self.log(f"Note: apps.yaml sets '{arg}: {raw_value}' but auto-discovery is using '{value}' instead - auto-discovery always wins currently; remove the apps.yaml entry to avoid this message")
+        return self.set_arg(arg, value)
+
     def get_arg(self, arg, default=None, indirect=True, combine=False, attribute=None, index=None, domain=None, can_override=True, required_unit=None):
         """
         Retrieve a configuration argument from the base system.
--- a/apps/predbat/inverter.py
+++ b/apps/predbat/inverter.py
@@ -154,6 +154,18 @@ class Inverter:
         self.reserve_percent = self.base.get_arg("battery_min_soc", default=4.0, index=self.id, required_unit="%")
         self.reserve_percent_current = self.base.get_arg("battery_min_soc", default=4.0, index=self.id, required_unit="%")
         self.battery_scaling = self.base.get_arg("battery_scaling", default=1.0, index=self.id)
+        if not self.battery_scaling or self.battery_scaling <= 0:
+            # A parsed value of exactly 0 (or negative) isn't safe to interpret either way - it could
+            # be a flaky/unavailable API response (e.g. Solis Cloud API returning batteryHealthSoh: 0
+            # during an outage) or a genuinely unhealthy battery, and asserting either "fully healthy"
+            # (1.0) or "no capacity" (0) would be guessing. Retain the last value that was actually
+            # read as valid, rather than inventing one; only fall back to 1.0 if nothing valid has
+            # ever been read for this inverter.
+            last_known = self.base.get_arg("battery_scaling_last_known", default=1.0, index=self.id)
+            self.log("Warn: Inverter {} battery_scaling read as {} which is not a valid scaling factor, retaining last known value {} for this cycle".format(self.id, self.battery_scaling, last_known))
+            self.battery_scaling = last_known
+        else:
+            self.base.set_arg("battery_scaling_last_known", self.battery_scaling, index=self.id)
         self.battery_scaling_config = self.battery_scaling
 
         self.reserve_max = 100
--- a/apps/predbat/predbat.py
+++ b/apps/predbat/predbat.py
@@ -1603,6 +1603,12 @@ class PredBat(hass.Hass, Octopus, Energidataservice, Stromligning, Fetch, Plan, 
         """
         Setup the app, called once each time the app starts
         """
+        # Snapshot of apps.yaml exactly as the user wrote it, before Predbat's own defaulting
+        # (auto_config/load_user_config) or any component's automatic_config() touches self.args -
+        # lets ComponentBase.set_arg_auto() tell "user explicitly configured this" apart from
+        # "Predbat defaulted it" or "another component already overwrote it" (issue #4494 follow-up).
+        self.args_from_apps_yaml = copy.deepcopy(self.args)
+        self.apps_yaml_override_warned = set()  # {arg} already warned about via set_arg_auto()
         self.pool = None
         self.log("Predbat: Startup {}".format(__name__))
         self.update_time(print=False)
--- a/apps/predbat/solis.py
+++ b/apps/predbat/solis.py
@@ -1241,7 +1241,9 @@ class SolisAPI(ComponentBase, OAuthMixin):
                 battery_soh = float(battery_soh) / 100.0
             except (ValueError, TypeError):
                 battery_soh = None
-            if battery_soh:
+            # A battery_soh of exactly 0 is a documented, valid Solis Cloud API response (not "no
+            # battery") - only a missing/unparseable field should exclude the inverter here.
+            if battery_soh is not None:
                 batteries.append(inverter_sn)
 
         num_inverters = len(batteries)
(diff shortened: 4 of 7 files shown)


jibbej:
Thanks for picking this up so quickly. I have read through the diff. Both changes would have prevented what happened here: retaining the last known-good scaling stops soc_max collapsing, and treating batteryHealthSoh: 0 as present-but-zero rather than missing keeps automatic_config() from abandoning the rest of the configuration, which is what put my instance into an unrecoverable crash loop. So the failure itself is covered.

On the design question you left open - whether an explicit apps.yaml battery_scaling should override the auto-discovered sensor binding - here is the user-side view, with the caveat that my own case is now handled either way.

My apps.yaml had:

battery_scaling:
- 1.0
That value was discarded without any indication, because automatic_config() rebinds the arg to the SOH sensor entity and set_arg() has no check for a user-provided value. With this PR my instance would end up at 1.0 anyway, via the fallback, so this is no longer about the outage. What remains is narrower:

A user who deliberately sets battery_scaling to something other than 1.0 (say 0.9 to model measured degradation) still cannot make it stick.
There is no way to notice this. The effective value simply differs from the configured one, and the only way I found it was by reading solis.py.
The second point is the one I would argue for regardless of how you decide the first. Even if auto-discovery should keep winning, logging once per key when automatic_config() overwrites a value that was present in apps.yaml would make the discrepancy discoverable instead of invisible.

This is not specific to battery_scaling. set_arg() has no user-config guard at all, so every key automatic_config() touches behaves the same way - a one-time log would surface the whole class rather than this single instance.

For what it is worth, soc_max is already left to the user in this integration, though for a different reason (the auto-detection is disabled with an XXX: This is currently broken note rather than as a deliberate policy), so I would not lean on it as precedent either way.

Happy to test whichever approach you settle on.

chalfontchubby:
Good point, and agreed on the generic framing - pushed a follow-up commit.

Added `ComponentBase.set_arg_auto()`: a drop-in for `set_arg()` that, when the key had a different value in the user's raw apps.yaml (snapshotted at the very start of `initialize()`, before Predbat's own defaulting or any component touches `self.args`), logs a one-time note naming both values, then still applies the auto-discovered value exactly as before - precedence is unchanged, only the visibility is new.

Wired it up for all of `solis.py`'s `automatic_config()` bindings (not just `battery_scaling`) since you're right that the same silent-override applies to every key it sets. Left the other cloud integrations (Fox, GEcloud, Solax, etc.) untouched for now - the helper lives on the shared `ComponentBase` so rolling it out elsewhere is a small follow-up rather than a redesign, and I didn't want to widen this PR's blast radius further without discussion.

New tests cover: warns exactly once per key, doesn't repeat on a second `automatic_config()` run, stays silent when the apps.yaml value happens to already match, and stays silent (and doesn't crash) when there's no apps.yaml snapshot at all (e.g. most of the existing unit tests, which construct components without going through `PredBat.initialize()`).

Re: your `soc_max` point - agreed it's not useful precedent either way, said the same in the code comment already there.

Issue #4496, opened 2026-08-12
Load Remaining Today Does Not Match Planned Values


nbullus:
**Describe the bug**

The **predbat.load_energy_predicted** and  **predbat.load_energy_adjusted** sensors both have a **today_remaining** attribute that calculates the remaining load (up untill midnight).

If the in-day adjustments is turned 'on', then the remaining adjusted load will differ to the remaining predicted load. If in-day adjustments is turned 'off', both sensors track the same remaining load as you would expect.

However, the total accumulated half-hourly load values reported in the plan differs to what the aforementioned sensors state is the remaining load for today.

Note: This has been working previously after a fix due to a previous reported issue on the chat. Looks like the same issue has resurfaced but I have no further details on what was fixed.

**Expected behaviour**

The **today_remaining** attribute should match the load that the plan is currently working too.

**Predbat version**

8.48.0 (Beta)

**Environment details**

- GE AIO
- Standard HAOS installer

**Screenshots**

In-day adjustment is currently turned off and the **today_remaining** attribute is set to 5.23 kWh on both sensors.

<img width="1470" height="956" alt="Image" src="https://github.com/user-attachments/assets/b651f956-0670-4894-806c-e9cfa2b23471" />

However, the total planned remaining load is 5.52 kWh which is currently 0.29 kWh out. 

<img width="1470" height="956" alt="Image" src="https://github.com/user-attachments/assets/a6f51eb0-b909-4d0e-87fa-ffdf341f52a7" />

**Log file**

[predbat-5.log](https://github.com/user-attachments/files/30980286/predbat-5.log)

[predbat_plan-4.html](https://github.com/user-attachments/files/30980318/predbat_plan-4.html)

[apps.yaml-3.txt](https://github.com/user-attachments/files/30980324/apps.yaml-3.txt)

**Predbat debug yaml file**

[predbat_debug.yaml-4.txt](https://github.com/user-attachments/files/30980294/predbat_debug.yaml-4.txt)

chalfontchubby:
Are you able to identify a previous version that worked please.

nbullus:
> Are you able to identify a previous version that worked please.

[...]

PR #4511, opened 2026-08-13
fix(output): apply load_scaling_dynamic and manual_load_adjust to today_remaining

chalfontchubby:
## Summary

Follow-up to #4506 on #4496 - the merged fix applied the flat `load_scaling` factor to `load_today_comparison()`'s predicted curve (fixing the reported gap for the original reporter), but missed two other multipliers `step_data_history()` (`fetch.py`) also applies when building the plan's own `load_minutes_step`:

- `load_scaling_dynamic` (per-minute) - carries saving-session/free-electricity-event scaling (`load_scaling_saving`/`load_scaling_free`), as well as any per-window override from `rates_import_override` or the manual API.
- `manual_load_adjust` (additive, per-minute).

Caught by a second reporter on the same issue after reviewing the #4506 diff directly - they had a 2-hour "power up" (free electricity) event set to `load_scaling: 1.5` via the manual API, and correctly identified that the merged fix only accounted for the flat `1.05` base `load_scaling`, not the dynamic per-window override.

## Verification

Downloaded and replayed their attached `debug.yaml` directly (not just read the code). Confirmed `load_scaling_dynamic` genuinely carries `1.5` for the reported 12:00-14:00 window in their real config. Before this fix, `load_today_comparison()` never looked at that dict at all. After the fix, the same replay's plan-vs-sensor ratio drops to **1.0076** - matching the same small residual seen on the original #4506 replay (i.e. this closes the gap to the same baseline level as the first fix did for the original reporter).

## Test plan
- [x] New tests in `test_load_today_comparison.py`: `today_remaining` scales with `load_scaling_dynamic` applied uniformly across the remaining day (mirrors the existing flat `load_scaling` test), and reflects a single `manual_load_adjust` entry additively at the correct scale
- [x] Both new tests confirmed to fail without this fix and pass with it (temporarily reverted just `output.py`, reran, restored)
- [x] Existing `load_today_comparison` tests (None-guard, import-exceeds-load, flat `load_scaling`) pass unchanged
- [x] Verified against the reporter's actual `debug.yaml` (see above), not just unit tests

### DIFF
--- a/apps/predbat/fetch.py
+++ b/apps/predbat/fetch.py
@@ -2366,7 +2366,16 @@ class Fetch:
         self.load_forecast_history = self.get_arg("days_previous_auto", True)
         if self.load_forecast_history:
             window_days = min(max(self.days_previous) if self.days_previous else 7, LOAD_FORECAST_HISTORY_MAX_DAYS)
-            self.log("days_previous_auto enabled - using weighted-bucket historical load forecast over up to {} days".format(window_days))
+            # Config-time log only - describes what's enabled, not what happened this cycle. This
+            # runs unconditionally every cycle regardless of whether the weighted-bucket forecast
+            # actually gets used: Load ML (or any other source that sets load_forecast_only) takes
+            # precedence and skips it entirely (fetch_sensor_data(), guarded by
+            # "not self.load_forecast_only"). The "using weighted-bucket..." wording previously
+            # here read as if it was happening every cycle regardless, which is what actually gets
+            # logged only when the forecast is genuinely used (fetch_sensor_data()'s own "Using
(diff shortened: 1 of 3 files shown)


chalfontchubby:
Rolled in a small related fix: the \`days_previous_auto\` log line in \`fetch_config_options()\` fired every cycle unconditionally ("...using weighted-bucket historical load forecast...") regardless of whether it was actually used that cycle - Load ML taking precedence skips it entirely, confirmed via the \`not self.load_forecast_only\` guard in \`fetch_sensor_data()\`, no double-application occurs. Reworded so it doesn't read as confirming active usage on every cycle. This is what prompted gcoan's original "is predbat double applying in-day load adjustment?" question further up the issue - answer is no, but the log line was genuinely misleading.

PR #4519, opened 2026-08-14
fix(output): stop load_energy_predicted's whole-day total drifting through the day

chalfontchubby:
## Summary

Fixes the remaining part of #4496: `load_energy_predicted`'s whole-day **`today`** total (and its `state`/chart) drifted through the day whenever `load_scaling`/`load_scaling_dynamic` != 1.0, even with nothing about the underlying forecast changing.

- **Root cause**: #4506/#4511 (both merged) made `load_today_comparison()` apply `load_scaling`/`load_scaling_dynamic`/`manual_load_adjust` to *future* minutes only, which correctly fixed `today_remaining` (it now matches the plan's own `step_data_history()` convention). But the same accumulator (`load_total_pred`) that those future-scaled buckets feed into is also the source the whole-day `today` total is read from - and it sums *every* minute, unscaled-past alongside scaled-future. As `minutes_now` advances, buckets keep flipping from the scaled group to the unscaled one, so the published `today` total mechanically shrinks (load_scaling > 1) or grows (load_scaling < 1) as the day goes on.
- **Fix**: a separate accumulator (`load_total_pred_day`) applies the same scaling consistently to every minute regardless of past/future, and is now the sole source for `load_energy_predicted`'s `state`/`today`/`today_so_far`/`today_remaining`/`results`. `load_total_pred` itself, and the actual-vs-predicted divergence ratio it feeds (which deliberately wants raw/unscaled past minutes to compare against real consumption), are untouched.
- Added a regression test that calls `load_today_comparison()` twice on the same day (00:00 and 15:00) with `load_scaling=1.05` and asserts the `today` total and `state` are identical - confirmed it fails on `main` (30.24 → 29.34) and passes with this fix.

## On the "should it be perfectly flat?" question

@nbullus, to be precise about what this fix does and doesn't claim: `load_energy_predicted`'s `today` total was never architecturally "frozen at midnight" - `load_today_comparison()` recomputes the whole day from scratch every 5-minute cycle, there's no snapshot. We don't expect it to be a perfectly flat line forever - if the underlying days_previous/load-ML model's own forecast genuinely updates (e.g. as new historical data ages in), or if `load_scaling_dynamic`/`manual_load_adjust` config itself changes mid-day (a saving session gets added, a manual override starts), the total *should* move, and that's legitimate evolution, not a bug. What this fixes is the total moving for no reason other than the clock ticking forward - the `minutes_now` boundary shouldn't itself be a source of drift when nothing else has changed.

@gcoan - separately, your `today`/`today_so_far` confusion on `load_energy_predicted` is a real docs gap, not (as far as I can tell) a code bug: those two attributes aren't documented in `docs/output-data.md` at all, only `today_remaining` is, and even that section's description ("to end of plan") doesn't match what the code actually computes (strictly midnight-to-midnight). I haven't fixed the docs as part of this PR - flagging it separately rather than scope-creeping this fix.

## After this

This is the third pass at #4496 in a fairly short window (#4506, #4511, now this). Given that history, we're going to step back from this specific area for a bit rather than keep iterating reactively - if either of you spot more drift after this lands, please do keep reporting it with a debug.yaml, but we'd rather let this one bed in and be looked at with fresh eyes next time than risk another narrow fix that misses a fourth edge case.

## Test plan

- [x] New regression test (`_test_predicted_today_stable_across_the_day` in `test_load_today_comparison.py`) fails on `main`, passes with this fix
- [x] Full existing `load_today_comparison` suite still passes
- [x] `pre-commit` clean

### DIFF
--- a/apps/predbat/output.py
+++ b/apps/predbat/output.py
@@ -2549,6 +2549,7 @@ class Output:
         """
         load_total_pred = 0
         load_total_pred_now = 0
+        load_total_pred_day = 0
         car_total_pred = 0
         car_total_actual = 0
         car_value_pred = 0
@@ -2557,7 +2558,7 @@ class Output:
         actual_total_today = 0
         import_ignored_load_pred = 0
         import_ignored_load_actual = 0
-        load_predict_stamp = {}
+        load_predict_day_stamp = {}
         load_actual_stamp = {}
         load_predict_data = {}
         total_forecast_value_pred = 0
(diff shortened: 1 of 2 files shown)


nbullus:
Trefor,

Thanks for the update. This now sounds like a solid approach.

I’ve briefly tested on main and things have instantly improved. There is a slight up/down drift at the 5 minute re-evaluation, but when it hits the 30 min mark it resets back and rolls through the same 5 minute predictions, preventing that continual downward drift as the day progresses.

￼

Thanks to yourself and Rik for putting this none to bed.

Regards,

Nigel.

PR #4534, opened 2026-08-15
Fix/fox half kw capacity

springfall2008:


springfall2008:
Superseded by #4532 - rebasing onto main leaves this PR empty (identical tree).

Issue #4517, opened 2026-08-14
GivTCP discharge target written every cycle even when unchanged (regression from #4492)


chalfontchubby:
Follow-up to #4421, reported by @mpartington and @gcoan in the comments there after PR #4492 (which fixed the spurious "REST failed to setExportTarget got 0" warnings) merged:

> Maybe I need to raise a new issue. The fix pushed to main has stopped the faults, however it is writing a discharge target twice every 5 mins in the GivTCP logs. I think this is number.givtcp_ce2146g269_target_soc on my AC3, however this is NOT in the list of RTC safe registers. So burning through unnecessary writes. The number of writes during a 2 HR discharge, seems so unnecessary. I don't even see anything being changed. - @mpartington

@gcoan confirmed seeing the same.

**Hypothesis**

#4492 fixed the *verification* inside `rest_setDischargeTarget()` (`inverter.py:3390`) to check `Control.Discharge_Target_SOC_1` first, since GivTCP updates that synchronously at write time, falling back to the old `raw.invertor.discharge_target_soc_1` register only if `Control` doesn't have it - that register only refreshes on GivTCP's separate background `self_run` poll cycle, not synchronously with the write, which was the root cause of #4421.

That fix only touched the inside of `rest_setDischargeTarget()`. The *caller* that decides whether to write at all is separate, at `inverter.py:2502-2512`:

```python
if "raw" in self.rest_data and "invertor" in self.rest_data["raw"] and "discharge_target_soc_1" in self.rest_data["raw"]["invertor"]:
    current = self.rest_data["raw"]["invertor"]["discharge_target_soc_1"]
    ...
    elif current != target_soc:
        self.rest_setDischargeTarget(target_soc)
    else:
        self.log("Inverter {} Current discharge target is already set to {}".format(self.id, current))
```

This still compares against the same slow `raw.invertor.discharge_target_soc_1` field the #4492 fix moved away from as an unreliable primary signal. If that field doesn't reliably track the real applied value on some hardware (AC3/older Gen1 - there's been a running suspicion in #4421 that this register behaves oddly on that line), `current` never matches `target_soc`, so this branch fires and re-writes on every cycle - the `else` branch ("already set to X") never gets reached. Before #4492 that produced a visible warning each time; after #4492, the underlying write still fires every cycle, just silently now, since the low-level retry now succeeds quickly via `Control`.

[...]

Issue #4421, opened 2026-08-02
Reoccurence of #4404 - Warn: Inverter 0 REST failed to setExportTarget when exporting


davemilsom2:
Unable to re-open #4404 but after updating to version 8.47.3 (at16:08), the same/similar issue happened again.  I say "similar" as the error message now has a "**got 0**" at the end of it which is new, and sometimes it is showing correctly as "Exporting".

From the logs:
Issue reoccurred 16:23 - "Warn: record_status Warn: Inverter 0 REST failed to setExportTarget got 0"
Shows OK at 16:28 ("Exporting")
Reoccurred 16:34 - "Warn: record_status Warn: Inverter 0 REST failed to setExportTarget got 0"
Second message at 16:35 shows "Exporting" as I removed my manual invocation of export.

Latest YAML/Log attached

[predbat_debug.yaml(Sun).txt](https://github.com/user-attachments/files/30634702/predbat_debug.yaml.Sun.txt)
[predbat(Sun).log](https://github.com/user-attachments/files/30634703/predbat.Sun.log)

FileGo:
Same here, GivEnergy setup with 3kW AC inverter and 9.5 kWh battery. Currently running Predbat `v8.47.3`.

[predbat.log](https://github.com/user-attachments/files/30638590/predbat.log)
[predbat_debug.yaml](https://github.com/user-attachments/files/30638591/predbat_debug.yaml)

```
2026-08-02 20:55:06.104466: Info: record_status Demand
2026-08-02 21:00:01.070081: Warn: Found 1 periods of zero load with power data, filling using power integration
2026-08-02 21:00:25.637309: Warn: Inverter 0 Set export target slot 1 {'dischargeToPercent': 4, 'slot': 1} via REST failed got 0
2026-08-02 21:00:25.650909: Warn: record_status Warn: Inverter 0 REST failed to setExportTarget got 0
2026-08-02 21:01:12.031352: Warn: Inverter 0 Set export target slot 1 {'dischargeToPercent': 4, 'slot': 1} via REST failed got 0
2026-08-02 21:01:12.045289: Warn: record_status Warn: Inverter 0 REST failed to setExportTarget got 0
2026-08-02 21:01:12.165383: Error: Completed run status Exporting with Errors reported (check log)
2026-08-02 21:05:00.989244: Warn: Found 1 periods of zero load with power data, filling using power integration
[...]

PR #4541, opened 2026-08-15
fix(inverter): enable Enable_Charge_Target before writing a GivTCP discharge target

chalfontchubby:
## Summary

Follow-up to #4517/#4518, for the case #4518 explicitly doesn't fix (a REST-v3 inverter where `Control.Discharge_Target_SOC_1` is entirely absent, so the caller falls through to the slower `raw.invertor` field with no change in behaviour).

Working from a reporter's real `REST_Response.txt` and `GIVTCP_Logs.log`: `Control.Enable_Charge_Target` reads `"disable"`, and GivTCP's own log shows `"Setting Discharge Target 1 was a success"` repeatedly - the write itself is genuinely reaching the inverter and being acknowledged, it just never sticks (`Discharge_Target_SOC_1` never appears in `Control` afterwards), so `adjust_force_export()` sees a permanent mismatch and rewrites every single cycle.

`adjust_battery_target()` already calls `rest_enableChargeTarget(True)` before writing a *charge* target, with the comment "without it the inverter ignores the target" - that call was never mirrored on the discharge-target side. This PR does that.

**Not confirmed as the true root cause yet** - this is correlational, from comparing the reporter's data against a second, working sample from a different, unrelated report (#4421) that happens to show the opposite pairing (`Enable_Charge_Target: enable` + `Discharge_Target_SOC_1` present). An earlier theory (gate on `Invertor_Type`/`model` reporting `"Ac"`) was ruled out because that second working sample is *also* `"Ac"` type - so it's specifically `Enable_Charge_Target`, not the inverter generation, that differs between the two.

Deliberately hardened for the case this theory is wrong: if `rest_enableChargeTarget(True)` itself fails (genuinely unsupported register, not just previously unused), the discharge-target write is skipped rather than also attempted - so a genuinely unsupported inverter ends up retrying one REST endpoint per cycle instead of two, not worse than before.

Posting as draft specifically to ask for a live test before this goes anywhere near merge - it's unconfirmed and the person who can confirm it is the one who reported #4517.

## Test plan
- [x] New regression tests in `test_inverter.py`: confirmed to fail without the fix (`setDischargeTarget` retried twice, matching the reported symptom exactly) and pass with it; second test confirms the write-skip when enabling fails
- [ ] Real hardware test - not yet done, this is the actual thing in question

### DIFF
--- a/apps/predbat/inverter.py
+++ b/apps/predbat/inverter.py
@@ -2527,7 +2527,19 @@ class Inverter:
                 if current is None:
                     self.log("Inverter {} No current discharge target to read, export target not written".format(self.id))
                 elif current != target_soc:
-                    self.rest_setDischargeTarget(target_soc)
+                    # Without Enable_Charge_Target on, the inverter ignores a written discharge
+                    # target - GivTCP still reports the write as successful, but it never sticks, so
+                    # this repeats every cycle indefinitely (#4517). adjust_battery_target() already
+                    # enables this before writing a charge target; mirror that here. No-ops if
+                    # already enabled.
+                    # Unconfirmed whether every rest_v3 inverter actually has this register - if
+                    # enabling itself fails, don't also attempt a discharge-target write we already
+                    # know is doomed. That keeps a genuinely unsupported inverter retrying one REST
+                    # endpoint each cycle instead of two.
+                    if self.rest_enableChargeTarget(True):
+                        self.rest_setDischargeTarget(target_soc)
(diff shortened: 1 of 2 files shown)


chalfontchubby:
Superseded by #4542 - squashed into one clean commit with the give-up-after-repeated-failures hardening added since this was opened, rather than stacking fixup commits on top of the earlier, incomplete version of this fix.

PR #4543, opened 2026-08-16
fix(inverter): skip the discharge target write on AC Coupled GivTCP inverters

chalfontchubby:
## Summary

Follow-up to #4517/#4518, for the case #4518 doesn't fix (a REST-v3 inverter where `Control.Discharge_Target_SOC_1` is entirely absent). Supersedes #4542 (closing separately) - that approach (enable `Enable_Charge_Target` before writing) relied on a correlation with a second reporter's data that turned out to predate #4492/#4518 and was confounded by the exact race condition those fixes address, so it was never solid evidence the register works on AC Coupled hardware. Dropped rather than kept alongside this fix.

**Root cause**: AC Coupled inverters don't have a working `Discharge_Target_SOC_1` register. GivTCP still reports each write as successful (`"Setting Discharge Target 1 was a success"` in the reporter's own GivTCP log, repeatedly), but the value never persists between cycles - `adjust_force_export()` sees a permanent mismatch and rewrites it every single cycle, indefinitely.

**Why this is safe to gate on `raw.invertor.model == "Ac"` specifically** (ruled out earlier, for good reason at the time): confirmed against GivEnergy's own firmware archive ([github.com/DJBenson/giv-firmware](https://github.com/DJBenson/giv-firmware)) - "AC Coupled" is a single product line with exactly two firmware releases ever published (`D212-A212`, `A214-D214`), not an early/late generational split the way the Hybrid line has Gen1/2/3/HV/LV variants. Both firmware versions seen across the two reporter samples investigated on this issue match these exact releases - there's no newer AC-coupled variant this gate would need to keep working for, so excluding the whole `model == "Ac"` category is coherent rather than a guess that would need revisiting as more reports come in.

## Test plan
- [x] New regression test, confirmed to fail without the fix (5 REST write attempts) and pass with it (0)
- [x] Existing discharge-target tests (Control-signal trust, settle-delay, read-back) unaffected - none of them set `model`, so the new gate is a no-op for them

### DIFF
--- a/apps/predbat/inverter.py
+++ b/apps/predbat/inverter.py
@@ -2523,13 +2523,24 @@ class Inverter:
         if force_export:
             target_soc = int(self.reserve_percent)
             if self.rest_data and self.rest_v3:
-                current = self.rest_readDischargeTarget()
-                if current is None:
-                    self.log("Inverter {} No current discharge target to read, export target not written".format(self.id))
-                elif current != target_soc:
-                    self.rest_setDischargeTarget(target_soc)
+                # AC Coupled inverters don't have a working Discharge_Target_SOC_1 register - GivTCP
+                # still reports a write as successful, but it never persists between cycles, so the
+                # caller sees a permanent mismatch and rewrites indefinitely (#4517). Confirmed
+                # against GivEnergy's own firmware archive (github.com/DJBenson/giv-firmware): "AC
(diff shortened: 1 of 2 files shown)


springfall2008:
Looks like a GivTCP bug but the workaround is fine.

mpartington:
Confirmed workaround/fix works perfect for me - thanks Rik

PR #4567, opened 2026-08-17
perf(load_ml): assemble training features from sliding windows

springfall2008:
## The observation

A feature row is five `LOOKBACK_STEPS` windows — load, PV, temperature, import, export — plus the target's time features. Consecutive rows differ by **one step**, so a materialised matrix stores every reading ~288 times over:

```
feature matrix       :  65.9 MB   (11,386 x 1,446)
unique source values :   0.23 MB  (5 channels x ~11,674 readings)
redundancy           :  282x
```

## The change

Keep each channel as a dense 1-D array and let `np.lib.stride_tricks.sliding_window_view` supply the windows as **strides over it, costing nothing**. Assemble only the rows a batch needs, into a buffer reused across batches. The matrix is never built.

Three supporting pieces:

- **Sample selection.** A sample needs its target chunk and the whole lookback window above it. Previously found by probing each window per candidate; now a rolling count over a presence mask finds them in one pass. It selects the identical **11,386 targets in the same ascending order** on the fixture.
- **Statistics.** `_fit_windowed_normalisation` walks assembled row blocks rather than a matrix, producing the same **1,446 per-feature** mean/std. No collapse to coarser per-channel statistics, so no model change.
- **Normalisation** is applied as batches are gathered.

## Measured

Against the captured 90-day fixture, `val_mae` **bit-for-bit identical at every epoch count**:

| epochs | before | after | before | after |
|---|---|---|---|---|
| 2 | 162.2 MB | **99.5 MB** | 1.8s | **1.1s** |
| 30 | 169.1 MB | **105.5 MB** | 11.8s | **11.2s** |
| 100 | 165.4 MB | **102.6 MB** | 35.7s | **35.5s** |

**−38% memory and no slowdown.** A spike had predicted ~5% slower from normalising per batch instead of once; that cost is real (+4.6 ms/epoch) but building dense channels vectorised is cheaper than the per-sample row loop it replaces, and the two cancel.

At the new peak the feature matrix does not appear at all. The largest remaining item is `_adam_update`'s live temporaries at **37.9 MB (46%)** — which is where any further work belongs. Worth noting that is *live* memory during one update, distinct from Adam's churn across updates, which was separately measured as costing ~7 MB and is not worth chasing.

## Correctness

- `val_mae` bit-for-bit identical at 2, 30 and 100 epochs
- The `_create_dataset` characterisation test's **feature-matrix checksum is unchanged** (`12977609.094096`), so the assembled rows are bitwise identical to the materialised ones
- Same sample count, same ordering, same per-feature statistics

The dataset object reports `shape` and `dtype` and supports `np.asarray` for diagnostics and for the tests that inspect feature blocks by column. Training never converts it — two `test_load_ml` tests that slice columns now call `np.asarray` explicitly, so materialisation is visible at the call site.

## Testing

- `run_all --test load_ml` — 29/29 pass
- `run_all --test ml_memory` — checksum and sample count unchanged
- `run_all --test ml_training_perf` — working-set budget tightened from 2.5x to 2.0x the notional matrix; verified failing at 2.29x before this change, passing at 1.40x after
- `run_all --quick` — all tests pass, 20/20 random scenarios match baseline across 320 fields
- `run_pre_commit` — all hooks pass

### DIFF
--- a/.cspell/custom-dictionary-workspace.txt
+++ b/.cspell/custom-dictionary-workspace.txt
@@ -26,6 +26,7 @@ armv
 ASHP
 asyncio
 atfork
+atleast
 authoriser
 autodocstring
 autoflake
--- a/apps/predbat/load_predictor.py
+++ b/apps/predbat/load_predictor.py
@@ -93,7 +93,7 @@ class WindowedFeatures:
     @property
     def dtype(self):
         """Element type of the assembled rows."""
-        return np.float32
+        return np.dtype(np.float32)
 
     def __array__(self, dtype=None, copy=None):
         """Materialise every row, for diagnostics and tests.
@@ -127,8 +127,14 @@ class WindowedFeatures:
             yield self._assemble(self.targets[start : start + block], buffer)
 
     def __getitem__(self, indices):
-        """Return the normalised feature rows for the given sample indices."""
-        target_chunks = self.targets[indices]
+        """Return the normalised feature rows for the given sample indices.
+
+        Accepts anything that indexes the target array. A scalar gives one row back as a
+        1-D array, matching what indexing the matrix this replaces would have returned.
+        """
+        target_chunks = np.atleast_1d(self.targets[indices])
+        scalar = np.ndim(self.targets[indices]) == 0
+
         rows = len(target_chunks)
         if self._scratch is None or len(self._scratch) < rows:
             self._scratch = np.empty((rows, TOTAL_FEATURES), dtype=np.float32)
@@ -136,7 +142,7 @@ class WindowedFeatures:
         if self.mean is not None and self.std is not None:
             batch -= self.mean
             batch /= self.std
-        return batch
+        return batch[0] if scalar else batch
 
 
 def relu(x):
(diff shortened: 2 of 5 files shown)


springfall2008:
Thanks — one of these was right, one wasn't. Both checked rather than taken on trust.

**Scalar indices (`load_predictor.py:139`) — correct, fixed in a93b5d2c.**

`X_train[0]` raised `TypeError: object of type 'numpy.int64' has no len()`. Training only ever gathers with an index array so nothing hit it, but the object stands in for the feature matrix everywhere else and a single row is an ordinary thing to ask for. A scalar index now returns one row as a 1-D array, matching what indexing the matrix would have given.

Same commit also makes `dtype` return `np.dtype(np.float32)` rather than the bare type, so it compares as an array's dtype does — it was reporting `<class 'numpy.float32'>`.

Added a test covering scalar indexing, agreement between scalar and array indexing, and the `shape`/`dtype` contract. Verified failing before the fix with the exact `TypeError` above.

**`__array__` extra copy (`load_predictor.py:110`) — this one doesn't hold.**

The claim is that `dtype != np.float32` is True when `dtype` is a `np.dtype`, forcing a spurious `astype`. NumPy's dtype equality coerces the operand, so it evaluates False:

```
>>> d = np.zeros(3, dtype=np.float32).dtype
>>> repr(d)
"dtype('float32')"
>>> d != np.float32
False
```

I also probed what `__array__` actually receives: `np.asarray(obj)` passes `dtype=None`, and `np.asarray(obj, dtype=np.float32)` passes `dtype('float32')`. In both cases the guard `dtype is not None and dtype != np.float32` is False, so `out` is returned as-is and no cast happens.

Happy to be shown otherwise if there's a numpy version where dtype equality doesn't coerce, but on 2.3.5 here it does.

Suite green after the change, and `val_mae` remains bit-for-bit identical at `0.019082503606296288`.

PR #4568, opened 2026-08-17
fix(web): serve the Predbat logo locally instead of from GitHub

### LINKED ISSUE #4562: Predbat Dash doesn't render until bat_logo_light.png logo downloads or fails
**Describe the bug**
During today's Github outage, I noticed that Predbat was very slow to load the dashboard, and was generally unresponsive. I investigated this as I have just updated HAOS release, and had previously had issues.

It seems that 1) the logo is hosted on Github rather than local https://raw.githubusercontent.com/springfall2008/batpred/refs/heads/main/docs/images/bat_logo_light.png
2) The pages won't render until this logo either loads or fails to load (approx 15 second timeout), which makes it look like predbat has somehow failed.

**Expected behaviour**
Predbat should work well without an internet connection, or when Github is having an outage. Host the file locally, or if used as a popularity indicator or callhome, set up the page so it will render without this (I'm better on web infra than javascript/css/html, TBC)

**Predbat version**

v8.48.2

**Environment details**

- Givenergy Hybrid
- Standard HAOS install via App, App framework is the latest

**Screenshots**
It's back up now, so can't easily provide a copy. I guess if you update the hosts file for raw.githubusercontent.com to an unresponsive web server that would give you a repo?

In debug view the logo would show as pending, and around 15 seconds later it would render once it moved to a "failed" state. All local calls were answered in a few hundred ms.

chalfontchubby:
## Summary
- The dashboard's favicon and header logo were fetched from `raw.githubusercontent.com` on every page load; a GitHub outage or 429 rate limit left every page hanging for ~15s before it would render (#4562).
- The logo images now ship alongside the app (`apps/predbat/bat_logo*.{svg,png}`, same as the existing `.so` kernel binaries) and are served from a new local `/images/{filename}` route, whitelisted to just those three files.
- `docs/images/bat_logo*` were removed rather than duplicated - nothing else in the repo referenced them, so there's a single canonical copy with no drift risk.

Fixes #4562

## Test plan
- [x] New `web_logo_image` unit test: route serves each logo with the right content type, 404s for anything not on the whitelist, and the page header no longer references `githubusercontent.com`
- [x] `./run_all --test web_logo_image --test web_functions --test web_annual_routes_registered` passes
- [x] `./run_pre_commit` passes (black, ruff, cspell, etc.)


springfall2008:
Good idea @chalfontchubby but the downloader will need updating to include the images

chalfontchubby:
And that needs to be merged/installed before we can use it in the version it downloads. Same as the "wb" issue on the .so files.

chalfontchubby:
I'll fail safe if the image isn't in the apps dir.

springfall2008:
> And that needs to be merged/installed before we can use it in the version it downloads. Same as the "wb" issue on the .so files.

Yes, suggest adding the installer update in the same PR, the fallback should be fine until its rolled out

chalfontchubby:
I believe so long as 8ab260b7 — about 8 months old - is there, you don't have a manifest - the svg files are in a dir that should be autodetected and no change is needed to the downloader.... (?)

TCWORLD:
Just FYI this has killed rendering of the logo in all current releases:

image


Obviously it will start working once this PR ends up in a future release but all the existing/older releases will stay without the rendered logo.

I'd suggest leaving the duplicated logos in the docs folder to keep things working for older versions at least for a while until there has been a release and people have upgraded.

PR #4575, opened 2026-08-18
feat(prediction): model residual battery discharge during Freeze Export

PlainSeer:
## Summary

Adds an optional prediction-only setting for inverters that continue a small residual battery discharge during **Freeze Export**:

```yaml
inverter_freeze_export_discharge_rate: 240
```

The value is **battery-side watts** and defaults to `0`, so existing behaviour is unchanged unless explicitly configured.

## Why this is needed

On the measured AlphaESS system, Freeze Export does not hold battery power at exactly zero. The battery continues to discharge at a low residual rate.

An earlier version of this change modelled that energy too narrowly as only supplying remaining house demand. Further live measurement showed that this is not the full behaviour.

During Freeze Export, the residual battery discharge enters the normal AC-side power balance:

- house load consumes it first
- if the residual discharge exceeds the remaining house demand, the surplus is exported to grid
- the energy is therefore not an unexplained internal loss

This directly addresses the review concern about where the battery energy is actually going.

## Observed behaviour

A live AlphaESS measurement showed approximately:

- house load: **487 W**
- PV generation: **466 W**
- battery discharge: **269 W**
- grid export: **~248 W**

The remaining house demand after PV was only about **21 W**, so the observed grid export confirms that the Freeze Export residual battery discharge is not capped at house demand.

## Implementation

During Freeze Export, the configured residual battery-side discharge is fed into PredBat's normal energy-balance path.

The model now:

- reduces battery SoC using normal battery discharge accounting
- applies normal battery-discharge and inverter-loss conventions
- allows the discharge to offset house import
- allows any surplus to flow to grid export
- respects battery reserve
- respects the physical export limit
- uses the normal battery-cycle accounting path
- preserves PV-overflow charging behaviour
- implements identical behaviour in both the Python and C++ prediction engines

The configured residual discharge is only applied when Freeze Export is active and the battery is not already being charged.

## Tests

Regression coverage verifies:

- `0 W` preserves existing behaviour
- `240 W` for one hour removes `0.24 kWh` from the battery
- inverter loss is applied correctly
- residual discharge continues even with no house load, creating export where appropriate
- a near-balanced PV/load case correctly produces surplus grid export
- the setting does not apply outside Freeze Export
- the reserve floor is respected
- Python and C++ prediction engines remain in parity

The native prediction kernels were rebuilt for all supported architectures, and the fork's normal `code-quality` CI passed on the final implementation.

## Scope

Prediction only.

This does not change inverter commands, charge/export control, or Freeze Export control behaviour. It only improves PredBat's prediction model for inverters that exhibit this residual discharge behaviour.


springfall2008:
Seems like a good idea, but 'loss' implies a factor and not a value.

Is it really losing this 240w or is it just discharging at 240w rate?

PlainSeer:
Yes, that’s a better description.

The AlphaESS is not showing a percentage/factor loss; the battery is continuing to discharge at roughly 230–240 W while Freeze Export is active.

I used “loss” to distinguish it from normal house/grid discharge, but inverter_freeze_export_discharge_rate would be clearer and better reflects the measured behaviour.

apps/predbat/prediction.py
```diff
@@ -1204,6 +1207,14 @@ def run_prediction(self, charge_limit, charge_window, export_window, export_limi
             else:
                 soc = min(soc - battery_draw * battery_loss, soc_max)
 
+            # Some inverters continue to discharge the battery internally while Freeze Export
+            # blocks normal external battery flow. Model that measured battery-side discharge
+            # directly in SoC without creating fictitious house load or grid energy.
+            freeze_export_discharge_step = 0.0
+            if freeze_export_active and inverter_freeze_export_discharge_rate > 0:
+                freeze_export_discharge_step = min(inverter_freeze_export_discharge_rate * step / 60000.0, max(soc - reserve_expected, 0))
+                soc -= freeze_export_discharge_step
```
springfall2008: I don't understand this, it appears to model lost energy. Surely the battery must be going somewhere?

springfall2008 (changes requested):

apps/predbat/prediction.py
PlainSeer: You’re right — my description here is misleading.

On AlphaESS during Freeze Export the battery continues to discharge at a low rate and that energy is supplied to the house load; it isn’t simply lost internally.

So this should really be modelled as battery discharge contributing to load, rather than only reducing SoC.

springfall2008:
Superseded by #4651, which carries this feature forward on top of current \`main\` with fixes from review (mostly avoiding a startup crash on a blank \`apps.yaml\` value, and documenting/registering the new setting) — your original commit is preserved as the first commit there so the implementation stays credited to you. Thanks for the AlphaESS investigation behind this!

Issue #4548, opened 2026-08-16
Predbat changes required for Octopus Power Up/Power Down sessions and Octopus integration v19.0


gcoan:
**Describe the bug**
There have been a number of GitHub issues raised on this topic, so this issue is being raised to collate the different threads and drive Predbat development.

In short, Octopus have changed the way saving and free electricity sessions work, the BottleCapDave Octopus integration v19.0.0 has made changes as a consequence that Predbat needs to reflect, but following the post-eclipse Sunday free electricity session it is clear there are still more changes needed.

**Background**
Previously we had saving sessions that predbat would auto-join with rewards given in Octopoints for reducing electricity consumption/exporting more to the grid.
Separately there were free electricity sessions (no join required) where electricity imported was free.
Octopus has changed the terminology, these are now Power Down and Power Up sessions.

**Required behaviour**
1. As raised on #4452 Octopus integration v19.0.0 has changed the names of the sensors that the integration exposes https://bottlecapdave.github.io/HomeAssistant-OctopusEnergy/entities/octoplus/#saving-sessions from event.octopus_energy_ACCOUNTID_octoplus_saving_sessions to event.octopus_energy_ACCOUNTID_octoplus_power_down_events and similar for free electricity to power_up_events.  The old events are being retained to Jan 2027 so no immediate breaking change is needed, but for new users of the integration the old sensors are not created
2. Predbat documentation, apps.yaml templates and octopus.py code needs to be changed to refer to the new entity names.  To aid transition suggest Predbat tries looking for the new entity names, if they are found, use them, and if not, try the old entity names and log a warning that the apps.yaml needs to be changed to point to the new sensors (which have to be enabled before use)
3. Similarly the services Predbat calls to join the power up/down sessions on the BottleCapDave integration need to change to calling the new service names https://bottlecapdave.github.io/HomeAssistant-OctopusEnergy/services/#octoplus
4. Predbat Octopus direct needs to be changed to call the new Octopus service calls to retrieve power up and power down sessions and to join power up sessions as (presumably) the Octopus saving session API calls will stop working at some point in the future (if they haven't already)
5. As seen from the post-eclipse Octopus free power up event, the BottleCapDave Octopus integration is currently not aligned perfectly to the Octopus API's.  
[...]

PR #4595, opened 2026-08-19
feat(octopus): configurable minimum-reward threshold for saving-session auto-join

### LINKED ISSUE #4593: Saving-session auto-join has no rate guard, so zero-point events are join-attempted and rejected
## Summary

The saving-session auto-join loop has no rate guard. `octopoints_per_kwh` is read
only to compute a display rate, never to decide whether the event is worth
joining, so a zero-point event is attempted exactly like a real one. The only
`continue` in the loop is the Axle VPP conflict test.

This matters today because the Octopus integration currently puts national Power
Up (free electricity) events into the Power Down `available_events` set at
0 p/kWh — described in #4548 point 5 and
[BottlecapDave/HomeAssistant-OctopusEnergy#1820](https://github.com/BottlecapDave/HomeAssistant-OctopusEnergy/issues/1820).
Predbat tries to join each of them and the integration rejects them.

Observed with:

- Predbat v8.48.8
- HomeAssistant-OctopusEnergy v19.0.0

Raised separately from #4548 because the fix is independent of the v19 entity and
service migration tracked there, and because a rate guard is correct regardless
of when the integration stops emitting these events.

## Mechanism

https://github.com/springfall2008/batpred/blob/004f298f7d00f9e0b8bea8002fc65a7f32ecd291/apps/predbat/octopus.py#L2971-L2986

```python
for event in available_events:
    ...
    octopoints_kwh = event.get("octopoints_per_kwh", None)
    if octopoints_kwh is not None:
        saving_rate = octopoints_kwh / octopoints_per_penny
    else:
        saving_rate = saving_rate            # keeps the default
    if self._saving_event_conflicts_axle(start_time, end_time, axle_sessions):
        ...
        continue
    if code:
        # join
```

`octopoints_kwh` reaches `saving_rate` and nothing else. There is no
`continue` for `octopoints_kwh == 0`.

## Observed

Three join attempts in one run, all at 0 p/kWh, for the three national Power Up
time slots on the following day:

```
Octopus: Joining Octopus saving event code EVENT_43_160826 Sun 16/08 11:00-12:00 at rate 0.0 p/kWh
Octopus: Joining Octopus saving event code EVENT_45_160826 Sun 16/08 13:00-14:00 at rate 0.0 p/kWh
Octopus: Joining Octopus saving event code EVENT_46_160826 Sun 16/08 14:00-15:00 at rate 0.0 p/kWh
```

Each is rejected by the integration. Because the join call's result is not
checked (#4548 point 8), each also produces a "joined" notification, so a single
Power Up announcement generates several false alerts.

## Expected behavior

Skip an available event whose `octopoints_per_kwh` is `0`, with a log line
matching the Axle skip:

```python
if octopoints_kwh is not None and octopoints_kwh <= 0:
    self.log("Octopus: Skipping saving event code {} {}-{} - zero reward rate".format(...))
    continue
```

A genuine Power Down session always carries a positive reward, so this cannot
skip anything joinable. `None` should keep its current meaning — the default
rate applies — since that is the documented path for events that do not report a
rate.

Note the planning side already does this: the `joined_events` loop that builds
`octopus_saving_slots` requires `saving_rate > 0`. The guard is simply absent on
the join side.

## Why not wait for the integration fix

#4548 points 5 and 6 correctly put the data problem on the integration. But a
rate guard in Predbat is cheap, is correct independently, stops the rejected-join
churn now, and removes the false notifications that ride on it — without waiting
on an upstream timeline Predbat does not control.

## Suggested regression test

An available event with `octopoints_per_kwh: 0` should not produce a join call.
The existing auto-join tests all use positive-rate events.

chalfontchubby:
## Summary

Fixes #4593, per [gcoan's review](https://github.com/springfall2008/batpred/issues/4593#issuecomment-5344213229)
(now tracked as #4595).

The saving-session auto-join loop reads `octopoints_per_kwh` only to compute a display rate, never
to decide whether the event is worth joining, so a zero-point event is join-attempted exactly like
a real one. This matters today because the Octopus integration currently puts national Power Up
(free electricity) events into the Power Down `available_events` set at 0 p/kWh - described in
#4548 point 5 and [BottlecapDave/HomeAssistant-OctopusEnergy#1820](https://github.com/BottlecapDave/HomeAssistant-OctopusEnergy/issues/1820).
Predbat tries to join each one and the integration rejects it, and since the join result isn't
checked (#4548 point 8, separate issue, not touched here) each rejected attempt still fires a false
"joined" notification.

An earlier version of this PR added a fixed `octopoints_kwh <= 0` guard. gcoan pushed back: the
real root cause is the integration miscategorising events, and a hardcoded skip is the wrong shape
- a user may also want to skip a genuine, low-value Power Down session they don't consider worth
the disruption, not just the zero-reward ones. He suggested a configurable minimum-value threshold
instead, defaulting to today's unfiltered-at-zero behaviour.

## Fix

Adds **octopus_saving_session_min_octopoints_per_kwh** (`apps.yaml`, default `0`). An available
event is only auto-joined when its `octopoints_per_kwh` **exceeds** this threshold - the default of
`0` reproduces the exact behaviour of the original fixed guard (skips `<= 0`), while raising it lets
a user also filter out genuine but low-value sessions. `null` keeps its existing meaning (rate not
reported, default rate applies) and is never checked against the threshold - only an explicit
reported rate is gated.

This mirrors an invariant that already existed on the *planning* side: the `joined_events` loop that
builds `octopus_saving_slots` already requires `saving_rate > 0`. The guard was simply missing (and,
at the fixed value, insufficiently flexible) on the join side.

Documented in `docs/energy-rates.md`, and added (commented or not, matching each template's existing
`octopus_saving_session_octopoints_per_penny` state) to every inverter's `apps.yaml` template.

## Test plan
- [x] `test_saving_session_zero_rate_skip` (`test_saving_session.py`) - unaffected, still covers
      zero/negative/positive/null rates against the default threshold
- [x] New `test_saving_session_min_octopoints_threshold` - four sub-cases: default threshold still
      allows a moderate reward, a raised threshold blocks a lower reward, a raised threshold still
      allows a reward that exceeds it, and a reward exactly equal to the threshold is skipped
      (exceeds, not meets)
- [x] `./run_all --quick` passes (743 tests)


gcoan:
@chalfontchubby see my comment on the original issue.

I don't think a guard on octopoints rate being zero is correct and the premise for it causing session join rejections is incorrect.

I proposed a better more flexible solution

chalfontchubby:
> @chalfontchubby see my comment on the original issue.
> 
> I don't think a guard on octopoints rate being zero is correct and the premise for it causing session join rejections is incorrect.
> 
> I proposed a better more flexible solution

Ok - happy to leave it to you

gcoan:
Hi Rik @chalfontchubby I'm more than happy for you to make the code changes, I could make them, but I'm nowhere near as familiar with the automated test framework so I'd struggle with that part

My suggestion on #4593 was to deliver the filter the OP suggested by a configurable 'min saving session amount'.  Perhaps in the future Octopus will do zero value saving sessions; the Eclipse saving session was almost that !; and by having a configurable threshold then predbat users can choose what S/S to autojoin based on battery cycle cost, charge/discharge losses, etc.

gcoan:
looks good thanks Rik

springfall2008:
Do the points per session still come from the API, I thought Octopus dropped it?

chalfontchubby:
Checked this directly - yes, it's still there.

`rewardPerKwhInOctoPoints` still exists in Octopus's live GraphQL schema today (confirmed via introspection against the same `savingSessions` query both the Octopus HA integration and Predbat's own Direct client use), description unchanged: "The number of OctoPoints an account will earn per kWh of energy reduced, if successful."

There's also real recent data backing that up - on [BottlecapDave/HomeAssistant-OctopusEnergy#1820](https://github.com/BottlecapDave/HomeAssistant-OctopusEnergy/issues/1820) (mid-August, the eclipse-event miscategorisation thread), a user posted an actual event table with populated `Pts/kWh` values, including the `0` values for the miscategorised Power Up events that landed in the Power Down set - that `0` is a genuine reported value, not a missing/null field. That's exactly the case this PR's threshold is meant to catch.

So the field itself hasn't been dropped - might be thinking of something else (e.g. the account-level Octoplus rewards balance, which is a different thing).

PR #4613, opened 2026-08-20
fix(octopus): filter saving-session events by target region in Octopus Direct

### LINKED ISSUE #4612: Octopus Direct: saving-session join doesn't filter events by target region (same cause as BCD's #1737)
**Describe the bug**

Predbat's own "Octopus Energy Direct" component (`OctopusAPI` in `octopus.py`, used when connecting straight to Octopus's GraphQL API rather than via BottleCapDave's HA integration) doesn't request or filter on the region eligibility of a saving session, so it will attempt to auto-join saving-session events the account's region doesn't actually qualify for.

This is the same underlying cause as [BottlecapDave/HomeAssistant-OctopusEnergy#1737](https://github.com/BottlecapDave/HomeAssistant-OctopusEnergy/issues/1737) - some saving sessions are only valid for specific NESO grid regions, but Octopus's API still lists them as "available" to every account regardless of region. BottleCapDave's integration hit this, got `OE-1308: Account's region is outside of the target regions for this event` on join, and fixed it in v18.3.0 by requesting `targetRegion` on each event and filtering out anything that doesn't match the account's own region before ever attempting a join.

Predbat's Direct client never picked up the equivalent fix - it's a separate, independent code path from BottleCapDave's integration, so BCD's fix doesn't cover it.

**Where**

`octoplus_saving_session_query` in [octopus.py:299-320](https://github.com/springfall2008/batpred/blob/d29c13dc0e41f64cb39d195d53c5745070d64fec/apps/predbat/octopus.py#L299-L320) doesn't request a region field on either `events` or `joinedEvents` at all, and there's no region-based filtering anywhere in [`get_saving_session_data()`](https://github.com/springfall2008/batpred/blob/d29c13dc0e41f64cb39d195d53c5745070d64fec/apps/predbat/octopus.py#L979) or the join loop that consumes it.

**Expected behaviour**

Request the region field(s) on the saving-session query (matching whatever BCD's integration currently requests - see their `coordinators/power_down_sessions.py`, `is_in_region` check, for the shape) and skip auto-join for any available event whose target region doesn't include the account's own, the same way BCD's integration now does.

**Additional context**

This only affects the direct/GraphQL connection method (`octopus_api_key`-style config), not anyone going through BottleCapDave's HA integration, which already has this fixed. Because the Direct client's join mutation does return a real GraphQL error on rejection (unlike the HA *service call* path, which #1823 upstream is about), a region-ineligible join attempt here would surface as a logged failure rather than a silent false "joined" claim - but it's still an avoidable wasted join attempt and log noise for the same reason BCD considered it worth fixing on their side.

Not part of #4548's original scope (that's about the Power Up/Power Down naming migration) - raising separately since it surfaced from a related discussion there.

chalfontchubby:
## Summary

Fixes #4612. Predbat's own "Octopus Energy Direct" GraphQL client (the direct-to-Octopus connection method, separate from BottleCapDave's HA integration) never requested or filtered on region eligibility for saving sessions, so it would attempt to auto-join events the account's own region doesn't qualify for. Some saving sessions are only valid for specific NESO grid regions - Octopus's API still lists them as "available" to every account regardless of region - so joining one triggers `OE-1308: Account's region is outside of the target regions for this event`.

This is the same root cause BottleCapDave hit and fixed in their integration back in v18.3.0 ([BottlecapDave/HomeAssistant-OctopusEnergy#1737](https://github.com/BottlecapDave/HomeAssistant-OctopusEnergy/issues/1737)) - their fix never reached Predbat's Direct client since it's a fully separate, independent code path.

- Added `targetRegion { regionId }` to the saving-session events query and `signedUpMeterPoint { regionId }` to the account query (both confirmed against Octopus's live GraphQL schema via introspection, not guessed).
- `get_saving_session_data()` now skips any available event whose `targetRegion` is non-empty and doesn't include the account's own region.
- An unknown account region (`signedUpMeterPoint` missing/null) excludes region-restricted events rather than assuming eligibility - a wasted/rejected join attempt is exactly the failure mode this is meant to avoid. Nationwide events (no `targetRegion` at all) are never affected either way.

## Test plan

- [x] `./run_all --test octopus_url` - two new scenarios in the existing `get_saving_session_data` coverage: region-mismatch excluded / matching-region and nationwide kept, and the unknown-account-region edge case (region-restricted events excluded, nationwide kept)

### DIFF
--- a/.cspell/custom-dictionary-workspace.txt
+++ b/.cspell/custom-dictionary-workspace.txt
@@ -330,6 +330,7 @@ mypy
 nattribute
 ncalls
 nearr
+NESO
 newtok
 njpwerner
 nobat
--- a/apps/predbat/octopus.py
+++ b/apps/predbat/octopus.py
@@ -305,6 +305,9 @@ octoplus_saving_session_query = """query {{
 			startAt
 			endAt
       devEvent
+      targetRegion {{
+        regionId
+      }}
 		}}
 		account(accountNumber: "{account_id}") {{
 			hasJoinedCampaign
@@ -314,6 +317,9 @@ octoplus_saving_session_query = """query {{
 				endAt
         rewardGivenInOctoPoints
 			}}
+      signedUpMeterPoint {{
+        regionId
+      }}
 		}}
 	}}
 }}"""
@@ -1000,6 +1006,13 @@ class OctopusAPI(ComponentBase):
             self.log("OctopusAPI: User has not joined Octopus saving sessions campaign")
             available_events = []
 
+        # Some saving sessions are only valid for specific NESO grid regions - Octopus still lists
+        # them as available to every account regardless of region, so joining one the account isn't
+        # eligible for is rejected by the API (matches BottleCapDave/HomeAssistant-OctopusEnergy#1737).
+        # An empty/missing targetRegion means the event is nationwide, not region-restricted.
+        account_region_id = (self.saving_sessions.get("account", {}) or {}).get("signedUpMeterPoint", {}) or {}
+        account_region_id = account_region_id.get("regionId", None)
+
         for event in joined_events:
             event_id = event.get("eventId", None)
             if event_id:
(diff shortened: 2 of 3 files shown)


springfall2008:
Did you manage to live test this query as well to make sure it doesn't error?

chalfontchubby:
> Did you manage to live test this query as well to make sure it doesn't error?

Yes, now done - live tested against a real account. The query executes cleanly and returns real, non-degenerate data: several events on the account carry genuine `targetRegion` restrictions (real NESO region ID lists, not just empty arrays), and `account.signedUpMeterPoint.regionId` returned a real value too. So the fields aren't just schema-valid in theory, they're populated with meaningful data to filter against on a live account.

PR #4644, opened 2026-08-22
fix(octopus): split IOG slots at the daily low-rate cap boundary

### LINKED ISSUE #4624: plan showing incorrect monatary values when octopus intelligent car charging running
the plan is using the pre intelligent prices for the total cost as per facebook post

Rename the download file to predbat_debug.yaml.txt and upload it to GitHub. This will allow your plan to be reproduced and also stores all your settings for review.

[predb at_debug.yaml.txt.txt](https://github.com/user-attachments/files/31282917/predb.at_debug.yaml.txt.txt)

chalfontchubby:
## Summary
- `load_octopus_slots()` (car-charging cost accounting) previously flipped a whole multi-block Octopus Intelligent Go dispatch slot to `rate_max_base` the moment it crossed the `octopus_slot_max` daily low-rate cap, instead of only pricing the blocks beyond the cap at the higher rate. A single overnight dispatch straddling the cap boundary got fully re-priced at the peak rate instead of being split.
- Fixes #4624.

## Correctness note (not a change in this PR)
On IOG, the car's own 6-hour low-rate allowance is independent of the house's rate - the house still gets the cheap night rate (or a dynamic daytime slot) after the car's cap is used up, while the car's *own* excess charging reverts to the standard/peak rate regardless of time of day. `rate_add_io_slots()` (the general "Import p" plan column) already reflects this correctly - it leaves the real underlying tariff rate untouched once the cap is hit, rather than overwriting it. `load_octopus_slots()` correctly falls back to `rate_max_base` for the car's own excess kWh, matching that real tariff behaviour, unchanged by this PR.

The one known gap: the plan's shared "Import p" column and tooltips only ever show the *house* rate for a given time-slot, so when the car is paying the peak rate for excess charging during a period the house shows as cheap, that divergence isn't visible in the UI. That's a display-only follow-up, not a calculation bug, and is being tracked separately rather than folded into this fix.

## Test plan
- [x] `./run_all --quick` passes (including the `octopus_slots` suite and the full random-scenario regression baseline)
- [x] `run_pre_commit` clean
- [x] Verified against the reporter's own `debug.yaml` replay: plan metric drops from 1226.5736 to 799.8496 with per-slot car charging costs now matching the real dispatch rate

### DIFF
--- a/apps/predbat/octopus.py
+++ b/apps/predbat/octopus.py
@@ -2642,70 +2642,89 @@ class Octopus:
         # Add in the current charging slot
         for slot in slots_sorted:
             start_minutes, end_minutes, kwh, source, location = slot
-            kwh_original = kwh
-            end_minutes_original = end_minutes
 
-            # Determine rate for this slot, applying the midday-to-midday cap
-            slot_average = self.rate_import.get(start_minutes, self.rate_min_base)
+            # Determine rate for this slot, applying the midday-to-midday cap. A slot that only
+            # partly fits the remaining daily budget is split at the point the budget runs out -
+            # e.g. a single 8-hour overnight IOG dispatch with only 4 of its 12 daily blocks left
+            # gets 4 blocks at the low rate and the other 12 at the max rate, not the whole 16
+            # blocks flipped to max rate the way an all-or-nothing check would (batpred#4624).
+            # kWh is apportioned to each chunk by its share of the slot's duration - an
+            # approximation (real draw isn't perfectly uniform across the slot) but matches how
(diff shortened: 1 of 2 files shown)


chalfontchubby:
Addressed in 2e9f191e: kept full precision for the split kWh values instead of rounding both sides with dp2() - that double-rounding could drift low_kwh + high_kwh from the original kwh, and was inconsistent with this function's non-split chunks, which already carry full-precision kwh (dp2() only applied to cost/soc downstream).

PR #4647, opened 2026-08-22
feat(plan): split the Import p cell when the car's own rate diverges

### LINKED ISSUE #4646: Plan doesn't show the car's own rate once its IOG dispatch cap is exhausted
On Octopus Intelligent Go, the car's 6-hour cheap-dispatch allowance is independent of the house's rate. Once the car's daily cap (`octopus_slot_max`) is used up, its own excess charging reverts to the standard/peak rate regardless of time of day - while the house continues on its normal cheap night rate (or a dynamic daytime slot) for that same clock-time.

Predbat's cost *calculations* already handle this correctly:
- `rate_add_io_slots()` (the general "Import p" plan column) leaves the real house rate untouched once the cap is hit.
- `load_octopus_slots()` (car-charging cost accounting, fixed for the split-at-cap bug in #4624/#4644) correctly falls back to the peak rate for the car's own excess kWh.

The gap is purely in the **plan display**: the "Import p" column and its tooltips only ever show the house rate for a given time-slot. When the car is quietly paying the peak rate for excess charging during a period the house shows as cheap, there's no way to see that from the plan - a slot can look cheap while the car charging happening in it isn't actually billed at that rate.

**Possible directions** (not yet scoped):
- A second rate figure or marker on the "Car kWh" cell when its effective rate diverges from the shown house rate for that slot.
- An addition to the existing tooltip noting the car-specific rate when it differs.

Not a regression - this has always been the case; raised while confirming the calculation correctness for #4624.

chalfontchubby:
## Summary
- On Octopus Intelligent Go, a car's own low-rate dispatch allowance is capped independently of the house rate - once used up, the car falls back to the peak rate while the house keeps its real (possibly still cheap) rate for the same clock-time. The plan previously only ever showed the house rate, with no way to see the car was actually paying more for the same slot.
- Adds `car_charge_slot_rate()` (kWh-weighted average rate across any car charging windows active in a slot) and splits the Import p cell - house on the left, car on the right, each independently coloured and with its own hover tooltip - whenever the two diverge by more than 0.01. Server HTML and the client-side JS renderer are both updated; editable/override mode is left untouched since overrides apply to the single household rate.
- Fixes #4646.

## Relationship to #4644
Independent of #4644 (the IOG daily-cap slot-splitting fix) - verified no file/line overlap and no shared commit history, so either can merge first or standalone. This one just displays whatever rate is already present in `car_charging_slots`, regardless of which version of that calculation produced it.

## Test plan
- [x] New test in `test_plan_json_rate_adjust.py` covers both the JSON fields (`car_rate`, `rate_split`, `car_rate_color`) and the rendered split-cell HTML with both tooltips
- [x] `run_pre_commit` clean
- [x] Also fixed a pre-existing test-isolation bug found along the way: `run_load_octopus_slots_tests()` never set `num_cars` itself, silently depending on whatever a prior test in the same run left it as

### DIFF
--- a/apps/predbat/output.py
+++ b/apps/predbat/output.py
@@ -1069,6 +1069,16 @@ class Output:
         if self.rate_best_cost_threshold_export:
             export_cost_threshold = self.rate_best_cost_threshold_export
 
+        def import_rate_color(rate):
+            """Colour an import rate the same way as the plan's own Import p column (blue/green/yellow/red)."""
+            if rate <= 0:
+                return "#74C1FF"
+            elif rate <= import_cost_threshold:
+                return "#3AEE85"
+            elif rate > (import_cost_threshold * 1.5):
+                return "#F18261"
+            return "#FFFFAA"
+
         raw_plan["import_cost_threshold"] = import_cost_threshold
         raw_plan["export_cost_threshold"] = export_cost_threshold
         raw_plan["reason_templates"] = REASON_TEMPLATES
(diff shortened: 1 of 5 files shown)


chalfontchubby:
Addressed in efb0e354:
- HTML-escaped the split cell's title attributes (currency_symbols is user-configurable text; unescaped it could break out of the attribute server-side, the client JS path already escaped it).
- Reworded the car tooltip to "differs from house rate" instead of asserting "IOG dispatch cap reached" - any diverging car window average triggers the split, not only an IOG cap.
- `car_charge_slot_rate()` now skips windows with no `average` key (non-Octopus historical reconstruction in the Yesterday view) instead of treating them as 0p/kWh, which was falsely flagging ordinary charging as a rate divergence.

PR #4649, opened 2026-08-22
refactor(givtcp): move GivTCP REST handling into its own component

chalfontchubby:
Moves GivEnergy/GivTCP REST handling out of `inverter.py` into a proper component, per your suggestion. Built as several small commits for review but intended to land as one unit.

### What it does

`GivTCPComponent` (`givtcp.py`) polls GivTCP's REST API in the background and publishes each inverter's controls and status as plain HA entities — the same shape Fox/Solax/Ohme already use — then points Predbat's standard entity-based apps.yaml keys at them via `automatic_config()`. `Inverter` then takes the ordinary entity path for those controls, so the `if self.rest_api: ... else: ...` special-casing goes away for everything the component covers.

It reads the existing `givtcp_rest` key directly (same scalar-or-list shape), so it auto-activates and **existing users need no apps.yaml changes** — that was your backward-compatibility point on the design doc.

### Commits

1. `d24b29ae` — extract the 17 REST HTTP methods into `givtcp_rest.py: GivTCPRest` (mechanical, no behaviour change)
2. `3ae9f1ff` — normalise the read side (SoC, power, window times) into `GivTCPRest`, rewire `update_status()` off raw JSON indexing
3. `2762f0c1` — the component itself, + 23 unit tests
4. `b3d8b6b7` — collapse the now-redundant REST branches in `inverter.py`, update `test_inverter.py`

### What still uses REST directly

`inverter.py` is **not** REST-free, and I don't want to overstate it. Three things have no entity equivalent yet, each documented inline:

1. Battery/capacity discovery in `Inverter.__init__` (`soc_max`, nominal capacity, calibration detection, max rates)
2. `adjust_pause_mode` / `adjust_inverter_mode` — these auto-detect GE-Cloud vs local naming (`Pause Charge` vs `PauseCharge`) from the entity's live value, which is more delicate than the plain W/%/time controls and deserves its own pass
3. `adjust_force_export`'s discharge-target write for `DISCHARGE_TARGET_UNSUPPORTED_MODELS` (#4517) — needs the raw REST `model` field

So `GivTCPRest`/`rest_data` stay alive. Getting to zero REST code in `inverter.py` needs the component to publish battery-discovery, mode and model-info entities too — happy to do that next if you want it, but it's bigger than this slice and I'd rather land this first.

### Two judgement calls worth a look

- **`soc_kw` is bound, not `soc_percent`.** `Inverter` prefers `soc_percent` when both are set, but GivTCP reports SOC only as a whole percent (~0.1kWh steps on a 9.5kWh battery) whereas `SOC_kWh` carries 3 decimals. Binding `soc_kw` keeps the precision the old direct-REST path had.
- **`givtcp_rest_power_ignore` is now honoured in `automatic_config()`** rather than in `update_status()`. Same intent (leave power/voltage to the user's own sensors), enforced one layer up — otherwise auto-config would claim exactly the keys that option exists to protect.

### Testing

23 new component tests, `test_inverter.py` updated, full `run_all --quick` and `run_pre_commit` green.

**But this has had no validation against real hardware** — I don't have a GivEnergy system. The component's timing in particular (writes must apply inline, because `write_and_poll_*` polls the entity back within seconds) is the sort of thing only a live inverter really proves. Draft until someone dogfoods it.


chalfontchubby:
Follow-on is up as #4739 (draft, based on this branch so it stacks - it'll retarget to `main` automatically when this merges).

It covers the three things listed above as "what still uses REST directly": battery/capacity discovery, pause/inverter mode, and the #4517 discharge-target model check. `inverter.py` no longer parses the REST blob.

Two bugs turned up doing it, both of which would only have bitten once these moved - worth knowing about here since one is latent in this PR as it stands:

- The rate entities advertise a generic `max` of 20000W. `Inverter` derives `battery_rate_max_raw` for a GE inverter from exactly that attribute, so the moment discovery stops overriding it, every GivTCP user's battery rate limit silently becomes 20kW. Fixed in #4739 by publishing `Invertor_Max_Bat_Rate`.
- `_handle_write` has no branch for a select that isn't a time slot, so a plain-select write is dropped and the entity republished unchanged - looks like a successful no-op. Harmless today (all selects are slots), breaks as soon as `inverter_mode`/`pause_mode` are added.

Also fixed the CI failure on this branch: the merge from `main` landed the comment from `4fbacb20` but not the message string it was attached to, so the `ValueError` text was stale while the test's expectation (from the same commit) came through - a partial conflict resolution rather than a fresh regression.

springfall2008:
Superseded by #4864, which combines this with its follow-on #4739 onto a single branch based on current `main`.

Your commits are carried over unchanged with authorship intact. This PR said it was 'intended to land as one unit' — #4864 is that unit, with the three items this PR deliberately left behind (battery/capacity discovery, mode/pause controls, the #4517 model check) already folded in from #4739.

The design write-up here remains the reference for the rationale and the judgement calls; #4864 links back rather than restating it. Thanks @chalfontchubby.

PR #4653, opened 2026-08-22
Show the running commit instead of just the release tag for dev deploys

chalfontchubby:
## Summary
- `coverage/deploy` and `hass.py` now write a `git_version.txt` marker (source commit, `-dirty` if the tree has local changes) alongside the deployed `.py` files, so a dev checkout or the standalone-against-live-HA workflow can be told apart from an actual tagged release.
- Predbat reads that marker at startup into `THIS_VERSION_DISPLAY`, shown wherever the version is surfaced for humans (web UI header/menu, the Dashboard page, the `predbat.status` entity's `version` attribute, the generated Lovelace dashboard card, and the `github.py` startup log line). The real `THIS_VERSION` tag is left untouched everywhere it drives actual logic - the GitHub release-tag match, the HA `update` entity's `installed_version`/`release_url` (changing that would falsely show "update available" forever), the Prometheus `version` label, and the manual version-select comparison.
- `download.py`'s `predbat_update_move()` clears the marker after a real update lands, so a stale dev-commit marker doesn't linger once official release files are installed.
- Also fixes `record_status()`, found while testing this on a live deploy: the status entity's `last_updated` attribute was written via `str(datetime.now())` - a naive, sub-second-precision string in a shape `str2time()` can't parse - so the web dashboard's "Last Updated" field rendered raw microseconds. It's now written via `TIME_FORMAT` (matching how `last_started` is already written) using `self.now_utc_real` - the true wall-clock time, not the PREDICT_STEP-grid-snapped `self.now_utc` - and parsed/rounded to whole seconds for display in `web.py`.

## Test plan
- [x] `./run_all --test github`, `./run_all --test download`, `./run_all --quick`, and `./run_all -k web` all pass
- [x] `pre-commit` (ruff, black, cspell) passes on all touched files
- [x] Manually verified `read_deploy_git_version`/`clear_deploy_git_version` round-trip and the `hass.py` marker-writer resolves correctly through the symlinked `.py` files `coverage/standalone_ha` sets up
- [x] Verified live on a real deploy: Dashboard/status page now show `v8.50.0 (<commit>[-dirty])`, and `Last Updated` shows a clean wall-clock second-precision timestamp after `record_status()` next fires

### DIFF
--- a/apps/predbat/download.py
+++ b/apps/predbat/download.py
@@ -145,6 +145,40 @@ def remove_file_quietly(filepath):
         print("Warn: Failed to remove {}: {}".format(filepath, e))
 
 
+def read_deploy_git_version(this_path):
+    """
+    Read the git_version.txt marker, if present, written by a dev deploy
+    (coverage/deploy) or a standalone launch (hass.py) from a git checkout, so the
+    running code can show which commit is actually installed instead of just the
+    release tag baked into THIS_VERSION.
+
+    Args:
+        this_path (str): Directory to look for git_version.txt in, alongside predbat.py.
+    Returns:
+        str or None: The marker contents, or None if the file is absent or unreadable.
+    """
+    filepath = os.path.join(this_path, "git_version.txt")
+    try:
+        if os.path.exists(filepath):
+            with open(filepath, "r") as f:
+                return f.read().strip() or None
+    except Exception as e:
+        print("Warn: Failed to read git_version.txt: {}".format(e))
+    return None
+
+
+def clear_deploy_git_version(this_path):
+    """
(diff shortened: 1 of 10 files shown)


apps/predbat/download.py
```diff
@@ -567,6 +601,7 @@ def predbat_update_move(version, files):
             cmd += "mv -f {} {} && ".format(os.path.join(this_path, file + "." + tag), os.path.join(this_path, file))
         cmd += "echo 'Update complete'"
         os.system(cmd)
+        clear_deploy_git_version(this_path)
```
springfall2008, line 603: This might want to be before the os.system(cmd) as touching the files triggers a restart

PR #4658, opened 2026-08-22
fix(manual-api): allow multiple overrides for lists (rates_import_override, rates_export_override)

### LINKED ISSUE #4405: Predbat manual api no longer allows multiple rates_import_override
**Describe the bug**
select.predbat_manual_api is used to override apps.yaml settings, e.g. import rates using rates_import_override.

This works fine for setting a first override date & time.

When a second rates_import_override is set for a different date and time, predbat overwrites the first settings and does't create a second rate override.

I think this behaviour is newly introduced, previously predbat would allow multiple sets of rates_import_override

**Expected behaviour**
I use predbat_manual_api to set an import rate override when there is an Octopus free power up session, via an automation script:

```
alias: Send Powerup event to Predbat manual API
sequence:
  - action: select.select_option
    data:
      option: >-
        rates_import_override?date={{          
        states('input_datetime.octopus_power_up_date')| as_timestamp |          
        timestamp_custom('%Y-%m-%d') }}&start={{          
        states('input_datetime.octopus_power_up_start_time') }}&end={{          
        states('input_datetime.octopus_power_up_end_time')
        }}&rate=0&load_scaling=1.5
    target:
      entity_id: select.predbat_manual_api
mode: single
description: ""
```

With a power up session already in the API for today:
image

When I run the script to set a power up event for tomorrow, predbat deletes today's rates_import_override and replaces it with tomorrow's details:
image

What should (and used to) be happening is that the new rates_import_override is added to the end of the predbat_manual_api settings, so both rates_import_override are configured.

**Predbat version**
8.46.11

**Environment details**
GivEnergy inverters
HAOS

**Log file**
[predbat.log](https://github.com/user-attachments/files/30605400/predbat.log)

chickenonaraft88:
## What

Fixes #4405 

Allow non-indexed `dict_list` overrides to accumulate rather than assuming they are all at index 0.

## Fix

Both api_select_update() (storage) and get_arg() (the list-merge fetch.py's basic_rates() consumes) are now index-aware:

- A no-index dict_list override only collides with an exact duplicate resend - distinct windows coexist.
- An explicit index still replaces on resend, and distinct explicit indices still coexist, preserving the original indexed-override behaviour.
- On read, a dict_list override's stored index only ever matters for write-time dedup - it has no meaning as an output list position, since these overrides are a bag of independent entries, not per-slot values.

## Test plan

`test_manual_api.py` gained coverage for the storage-layer dedup rules, the `get_arg/basic_rates` merge behaviour, and the malformed-index guard - no prior test exercised multiple simultaneous no-index overrides at all, which is how this went unnoticed.

- [x] ./run_all --test manual_api --test basic_rates --test validate_config --test annual_tariff passes
- [x] ./run_pre_commit - all hooks pass (the one pre-existing failure, test_faq_recorder_config, is an unrelated Windows-only cp1252 decoding bug reproducible on main, not caused by this change)

image


springfall2008:
@chickenonaraft88 thanks for your PR, can you update the documentation please?

PR #4661, opened 2026-08-22
fix(compare) compare tariffs using stale iBoost plan from live tariff

chickenonaraft88:
## Summary

`compare.py`'s `run_single()` swaps in each compared tariff's own import/export rates, but `iboost_plan` (the low-rate slots iBoost is allowed to run in) was only ever computed once, at startup, against the **live** tariff's rates. Every tariff comparison reused that same stale plan, so a compare-plan's iBoost timing reflected your current live tariff instead of the tariff actually being compared — it would show iBoost running at your live tariff's cheap period even on a tariff with a completely different rate structure.

## Fix

- Added `recompute_iboost()` to `Compare`, mirroring the existing `recompute_car_charging()` pattern, and call it per tariff in `run_single()` right before the car charging recompute.
- The guard condition matches `fetch.py`'s existing logic for when the live plan builds `iboost_plan`, so Compare now decides whether/how to plan iBoost exactly the same way the real plan does.
- Save/restore `iboost_plan` in `run_all()` alongside `iboost_today`, so the live plan's iBoost schedule isn't left clobbered by whichever tariff ran last.

## Before / after

**Before the fix** — iBoost only ran on the current tariff's cheap slot, and on "EDF Go Electric" it ran during the expensive daytime rate instead of that tariff's own cheap overnight period:

Current tariff
agilecompareiboost

Other tariff
edfcompareiboost

**After the fix** — each tariff's compare-plan now schedules iBoost against its own rates, running during EDF Go Electric's cheap overnight window while the battery charges:

Current tariff
image

Other tariff
image

## Test plan

- [x] Added `test_compare.py` T13/T14 covering `run_single()` recomputing `iboost_plan` per tariff, and clearing it when smart-rate planning doesn't apply
- [x] `./run_all --test compare --test iboost_smart` passes
- [x] `./run_all --quick` passes (one pre-existing unrelated flaky failure in `fox_api`'s `test_run_midnight_reset`, confirmed present without this change too)
- [x] Verified against real EDF Go Electric compare data in Home Assistant — iBoost now runs during the tariff's own 6.99p overnight window while charging, instead of the live tariff's expensive daytime slot

### DIFF
--- a/apps/predbat/compare.py
+++ b/apps/predbat/compare.py
@@ -336,6 +336,7 @@ class Compare:
         my_predbat.manual_all_times = []
         my_predbat.octopus_intelligent_charging = False
 
+        self.recompute_iboost()
         self.recompute_car_charging(car_charging_slots)
 
         self.log("Running scenario for tariff: {}".format(name))
@@ -448,6 +449,17 @@ class Compare:
         self.select_best(compare_list, self.comparisons)
         self.publish_data()
 
+    def recompute_iboost(self):
+        """
+        Recompute iBoost plan
+        """
+        my_predbat = self.pb
+
+        if my_predbat.iboost_enable and (((not my_predbat.iboost_solar) and (not my_predbat.iboost_charging)) or my_predbat.iboost_smart):
+            my_predbat.iboost_plan = my_predbat.plan_iboost_smart()
+        else:
+            my_predbat.iboost_plan = []
+
     def recompute_car_charging(self, car_charging_slots):
         """
         Recompute car charging plan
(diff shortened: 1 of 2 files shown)


chickenonaraft88:
Hold off on this because I'm not sure if my branch has changed the load or it's unrelated. Notice my load in each screenshot, before looks normal, after is just a flat 0.06.

chickenonaraft88:
I'm not sure what happened in my screenshot but I turned off load ml and it seems to be fine now. Maybe something else for me to look into.

Issue #4677, opened 2026-08-23
Freeze Charge unavailable as a planning candidate for slots inside an export window


chalfontchubby:
Split out from #4257 - the second of two separate root causes found there (the first is #4676). Tracked separately so each has something a PR can close without prematurely closing #4257.

**Bug**: `charge_window_best` and `export_window_best` are built independently from separate rate scans (`fetch.py`'s `find_charge_window`, ~line 1664-1758) - nothing in either scan consults the other. Once a slot lands inside an export window, `remove_intersecting_windows()` (`utils.py` ~line 1039) clips any overlapping charge window before the optimiser can score it. The optimiser's own search is coordinate descent, never joint (`optimise_charge_limit` and `optimise_export`, `plan.py`, run one array at a time) - once export is "on" for a slot, the clipped charge window shows zero marginal benefit and gets left at 0, so Freeze Charge is never actually competed against Freeze Export for that slot. Not out-competed on merit - never a candidate at all. A stable local minimum of the two-array descent, not a deliberate choice.

Confirmed against #4257's reporter data: disabling their (non-default) `combine_export_slots` setting as a counterfactual flipped one of two affected slots from FreezeExp to a genuine charge, with no other change - proving the exclusion is real, though `combine_export_slots` is only a lever that changes which slots fall into the stuck state, not the cause itself.

A follow-up controlled comparison (holding the rest of the plan fixed, checking end-of-horizon SoC doesn't drift between variants) confirmed opening up the candidate space is worthwhile: Freeze Charge ties Freeze Export exactly on cost for a pure hold, and a genuine active charge scored marginally better still once real round-trip losses were applied precisely.

**Fix shapes considered, not yet chosen between** (see internal notes, ask if useful):
1. Loosen the clip in `remove_intersecting_windows()` from `< 100.0` to `< 99.0` so Freeze Export stops suppressing overlapping charge candidates at the root. Correct semantically, but a hot-path function called on every simulation, and may not by itself be sufficient to break the coordinate-descent stall without also touching search order.
2. An additive post-pass (modelled on `optimise_swap_export`, `plan.py` ~line 3275): for windows landing at the freeze-export sentinel, trial "Idle export + Freeze Charge on the overlapping charge window" via `run_prediction_metric` and keep it only if strictly better. Contained, no core search-loop changes, but is a patch on top of the stuck descent rather than a fix to the descent itself.

[...]

Issue #4257, opened 2026-07-16
Odd charging choices around Axle events (Octopus Flux)


TCWORLD:
**Describe the bug**

When an axle event is present in the plan towards the end of the data with the octopus flux tariff, predbat seems to weirdly choose to try and charge the battery part way though the flux export period.

<img width="1035" height="509" alt="Image" src="https://github.com/user-attachments/assets/e4c22a7b-c25d-4e42-8e51-e3268016c3da" />

This is presumably to ensure there is sufficient charge for the axle period later in the day.

However why is it picking here 17:00 to charge @34p/kWh, when it could instead choose say 19:00 to charge @24p/kWh. Essentially costing an extra 10p that is wholly unnecessary.

**Expected behaviour**

Not randomly choose to charge at peak import rate when there are slots which could instead charge at off-peak rates.

To be fair there is also very little benefit to forcing discharge @27p/kWh to then reimport @24p/kWh, particularly accounting for round trip losses.

**Predbat version**

v8.45.1

**Environment details**

- Inverter and battery setup: Home-Assistant SolaX Modbus integration using service API.
- Standard HAOS installer or Docker: HA Supervised on Raspbian
- Anything else?

**Screenshots**
If applicable, add screenshots to help explain your problem. The most useful ones can be your battery chart, the Predbat HTML plan and your current settings in HA.

See above

**Log file**

Neither log file nor plan will download for some reason? I've tried clicking the links on the HA ingress but they appear to do nothing at all.

TCWORLD:
Although looking at the lack of cost change, it seems predbat must think the charging is free - presumably charging off the solar.

However when it commands a charge in such cases it always seems to request charging at full whack (even though charge low power mode is enabled), which means the inverter will charge with solar yes, but also force charge from the grid.

If it wanted to just charge the battery with solar alone, it should just be setting demand mode, or freeze charge.

TCWORLD:
So again today it's making weird choices. Here we see it's decided to freeze export during the 9p/kWh rate. That will mean less in the battery overnight so I'll end up paying 24p/kWh to reimport. If the aim was to keep enough charge for the axle event I'd expect a hold charge (freeze charge) here not a freeze export.

<img width="360" alt="Image" src="https://github.com/user-attachments/assets/da1791d5-33df-4fa9-b865-dc482fbde53b" />

I got it to generate logs this time, so will upload them this evening when back at my PC.

chalfontchubby:
[...]

PR #4679, opened 2026-08-23
fix(prediction): model battery discharge for a genuine Freeze Export shortfall

### LINKED ISSUE #4676: Freeze Export simulation doesn't model battery discharge for a load shortfall
Split out from #4257 - this is one of two separate root causes found there, tracked separately so each has something a PR can close without prematurely closing #4257.

**Bug**: `customisation.md` documents Freeze Export as allowing battery discharge to cover a load shortfall ("Freeze Export mode which allows battery discharge, but not battery charging"). The planning simulation doesn't implement that at all - `prediction.py`'s freeze-export branch (~line 1065-1111) leaves `battery_draw = 0` whenever there's a shortfall (`diff > 0`); it's only ever revised for the surplus case. So a slot the plan labels "FreezeExp" is actually simulated and displayed with Freeze Charge's behaviour (grid covers the shortfall, SoC held flat), not Freeze Export's own documented behaviour.

Confirmed against a reporter's real debug.yaml (#4257): `soc_percent` stays flat across a FreezeExp slot despite `load_forecast` clearly exceeding `pv_forecast` (a genuine shortfall, including a manually-declared load). A controlled A/B comparison showed FreezeExp and FreezeCharge score *identically* in the model for a shortfall slot - the simulation can't currently distinguish them at all.

Whatever mode planning ends up choosing (separate question, see #4257's other root cause), the plan should show what the chosen mode actually does.

**Proposed fix**: mirror the existing shortfall-discharge logic already present a few lines away in the ECO-mode branch (`prediction.py` ~line 1127-1129: `battery_draw = min(diff, discharge_rate_now_curve_step, inverter_limit, battery_to_min)`) into the freeze-export branch. Contained to `prediction.py`, no optimiser search-structure change.

Expect this to also change plan outcomes indirectly: once FreezeExp honestly reflects battery drain during a shortfall, it will likely score worse than it does today in scenarios like #4257's, which may make some wrong-mode plans self-correct even before the other root cause (candidate availability, tracked separately) is fixed - not yet confirmed, worth checking once implemented.

chalfontchubby:
## Summary

Freeze Export is documented as allowing battery discharge to cover a load shortfall (`customisation.md` "Freeze Export during Demand": "allows battery discharge, but not battery charging") - but the simulation left `battery_draw` at `0` whenever load exceeded PV, silently modelling the same outcome as Freeze Charge (grid covers it, SoC held flat) instead. So a plan slot labelled `FreezeExp` showed the wrong behaviour, regardless of whether Freeze Export was even the right mode to pick for that slot in the first place (a separate, bigger issue - #4677).

Root-caused while investigating #4257 - full trace and the reporter's own debug.yaml replay in that thread.

## Changes

- `prediction.py`: added the missing shortfall-discharge branch to the freeze-export elif, mirroring the existing ECO-mode discharge logic a few lines away.
- `prediction_kernel.cpp`: same fix, kept in parity with the Python engine (kernel rebuilt, binaries will be picked up by the cross-build CI job).
- Scoped the existing AlphaESS residual-discharge special case (`inverter_freeze_export_discharge_rate`) to only fire when nothing else has already moved the battery (`battery_draw == 0` rather than `>= 0`), so it can't clobber a genuine shortfall discharge.
- New test `freeze_export_ac_flow_shortfall_discharges` in `test_model.py`, passing on both the Python engine and the C++ kernel.
- Two golden-snapshot regressions regenerated as an expected consequence of the behaviour actually changing: `debug_cases/predbat_debug_pre_saving1.yaml` and the random-scenario baseline (20 scenarios, several metrics shifted - checked each shift is directionally explained by the fix, not a red flag).

## Test plan

- [x] `./run_all --test model --test model_kernel` - new scenario passes on both engines
- [x] `./run_all --test kernel_parity` - Python/C++ kernel stay in sync
- [x] `./run_pre_commit` passes (ruff, black, cspell, markdownlint, full test suite incl. `random` regression)

Fixes #4676. Root-caused via #4257.


chalfontchubby:
Reopening to retrigger CI - the bot's kernel-binaries auto-fix push doesn't trigger a new check run on its own.

apps/predbat/prediction.py
```diff
@@ -1090,13 +1090,23 @@ def run_prediction(self, charge_limit, charge_window, export_window, export_limi
...
                 # Some inverters (observed on AlphaESS) continue a small residual battery
                 # discharge during Freeze Export. Treat the configured value as battery-side
                 # power and feed it into the normal AC balance: house load consumes it first and
                 # any surplus may reach the grid. Respect the battery reserve and physical export
-                # limit rather than capping the discharge at house demand.
-                if inverter_freeze_export_discharge_rate > 0 and battery_draw >= 0:
+                # limit rather than capping the discharge at house demand. Only applies when
+                # nothing else has already decided a movement - a genuine shortfall discharge
+                # (above) already covers the real load more precisely than this residual estimate.
+                if inverter_freeze_export_discharge_rate > 0 and battery_draw == 0:
```
springfall2008: Isn't ==0 on a floating point number quite unlikely?

springfall2008:
Reviewed this and pushed two commits onto the branch (708e9a1, f063890) rather than leaving the findings as comments — happy to drop either if you disagree with the call.

## Two problems with the fix as written

**1. The shortfall discharge wasn't grossed up for inverter loss.** The PR describes the new branch as mirroring the ECO discharge logic, but it mirrors the `min()` clamp and not the `diff` feeding it. ECO deliberately converts the AC shortfall to a DC draw (`required_for_load += (load - pv_ac) * inverter_loss_recp - (load - pv_ac)`, giving `diff = shortfall / inverter_loss`). The freeze branch passed the raw `get_diff` result — an AC figure — straight into `battery_draw`, which is DC. The battery therefore covered only `inverter_loss` of the load and the remainder became a phantom grid import of `shortfall × (1 - inverter_loss)`.

Measured at 1 kW load, no PV, one hour, `inverter_loss=0.9`, with the discharge rate and inverter limit both non-binding:

| | final SoC | metric |
|---|---|---|
| ECO (`discharge=100`) | 8.89 | £0.00 |
| FreezeExp (`discharge=99`) | 9.00 | £0.01 |

Identical in both engines. The existing tests couldn't catch it because `simple_scenario` defaults `inverter_loss=1.0`, which collapses the two cases.

**2. The AlphaESS residual rate was silently neutered.** Changing the guard to `battery_draw == 0` plus the new elif meant `inverter_freeze_export_discharge_rate` only fired when there was no load shortfall — i.e. never, in the overnight case it exists for. Every existing 240W test passes only because it sets `battery_rate_max_charge=0.0`, which also zeroes `battery_rate_max_discharge`, so none of them exercised the interaction:

| `battery_rate_max_charge` | main | this PR (before) |
|---|---|---|
| 0.0 (as tested) | SoC 9.76, £0.08 | SoC 9.76, £0.08 |
| 1.0 (realistic) | SoC 9.76, £0.08 | **SoC 9.00, £0.00** |

The premise of that setting is that those inverters *don't* cover house load during Freeze Export, so it now takes precedence again. `docs/apps-yaml.md` said the battery is "held flat" by default, which is no longer true either way — updated.

Both fixes have a test, each watched failing first.

## Why the branch shouldn't exist

Freeze Export is not a distinct inverter mode. `execute.py:486-496` and `execute.py:527-539` implement it by calling `adjust_charge_rate(0)` (or `adjust_pause_mode(pause_charge=True)`) and otherwise leaving the inverter in Demand/ECO mode — the discharge rate is never touched. So it *is* ECO with the charge rate at zero, literally, not by analogy.

Modelling it as a parallel branch that re-derives the same AC balance is what let the two drift apart, and #4676 is that drift: the branch pinned `battery_draw` at 0, which is Freeze Charge behaviour. Finding 1 above is the same failure mode recurring inside the fix — a hand-copied line that diverged from its original within the same PR.

The second commit removes the branch. Freeze Export runs the shared ECO flow with `charge_rate_scale = 0`, then applies the three things genuinely specific to it: the #4207 recapture, the residual rate, and the `fz*` labels. That commit changes no baseline file, which is itself the neutrality evidence — the golden snapshot and all 20 random scenarios were generated against the previous commit and pass here unchanged.

On whether the residual rate could just be folded into the rate used by the ECO flow, removing that block too: it can't. `freeze_export_ac_flow_no_load_exports_residual` is the counterexample — load 0, PV 0, residual 240W, metric **-1.2** because it earns money exporting the leak. A rate limit can only shrink a demand-driven discharge; it can't create one when `diff == 0`. It's a floor as well as a cap, so it stays a post-step.

## Why the scenarios moved

Worth being precise here, because the PR notes checked the shifts were "directionally explained" and I think the actual mechanism is more interesting than that.

First, attribution. The fold is provably inert: generating the random sweep from the fixes-only state and the folded state on the same machine gives **0 of 20 scenarios differing in any recorded field**. `inverter_freeze_export_discharge_rate` is not set in any scenario, so fix 2 is inert too. Every delta therefore comes from one line — the gross-up — with `inverter_loss: 0.96` in the template, so a 4.2% change in freeze-export shortfall discharge.

16 of 20 scenarios are bit-identical. Of the four that move:

| scenario | metric delta | plan changed? |
|---|---|---|
| 0 | +1.82 | yes — export and charge windows both |
| 9 | +0.26 | yes — export windows |
| 17 | **-0.01** | **no — identical plan** |
| 19 | +6.56 | yes — export and charge windows both |

Scenario 17 is the only pure re-pricing, and it moves in the *better* direction: same plan, phantom import removed.

The other three are the optimiser choosing a **different plan**, so their metric deltas are not a re-pricing and shouldn't be read as "the plan got 6.5p worse". Scenario 19 is the clearest case — it plans **zero** freeze-export slots under both versions, yet has the largest change. Freeze-export candidates are evaluated at every candidate window even when none survives into the final plan, so a 4% shift in their cost flips tie-breaks against the `metric_min_improvement_export_freeze` thresholds and the divergence cascades through the greedy search. Scenarios 1, 2 and 10 conversely *do* plan freeze-export slots and are unchanged — presumably those slots sit in solar surplus where the shortfall branch never fires.

The honest caveat: once the model changes underneath the optimiser, comparing metrics across baselines isn't apples-to-apples, and I have **not** shown the newly chosen plans are physically better or worse. Doing that properly means scoring the old plan under the new model. Say the word and I'll build it.

One thing that is worth a look regardless: in the `pre_saving1` golden case, freeze-export slots went 2 (main) → 1 (this PR) → **0** (after these commits). Pricing the mode correctly makes it much less attractive, which is the right outcome but a big behavioural shift, and it bears on #4677.

## Verification

`./run_all` full suite passes (232s), `./run_pre_commit` clean, `--test kernel_parity` green. Golden snapshot and random baseline regenerated with both sides generated on this machine — worth noting the cross-machine runtime comparison in the original PR is meaningless, it showed +42.6% where the same-machine figure is **-2.2%**.

Caveat on the binaries: I could only rebuild `prediction_kernel_lib_darwin_arm64.so` (no zig available), and with Apple clang rather than the cross-toolchain, hence the size change. The other five are stale against the new `.cpp` until the cross-build job regenerates them — `verify_kernel_binary.py` in `code-quality.yml` should fail on `x86_64` first.

## Unrelated, spotted in passing

When `set_export_freeze_only` is set with `export_limit_now` below 99, the model takes the freeze path (`prediction.py:1065` before these commits) but `execute.py:506` takes "Hold exporting", which leaves charging enabled. Pre-existing, out of scope here, but it looks like a real mismatch.

chalfontchubby:
Looked through both commits - this is more solid than what we had. The gross-up bug was real (my `simple_scenario` defaults hid it - `inverter_loss=1.0` collapses the AC/DC conversion away), and the reversed precedence is right: `inverter_freeze_export_discharge_rate` being set means this inverter genuinely can't cover full load during Freeze Export, so it's a ceiling on the shortfall discharge, not something that loses to it - I had that backwards. The ECO-mode reuse also lands the "own if branch" simplification cleanly and makes the `battery_draw >= 0` check a real sign check instead of a near-zero comparison, so that resolves properly rather than just being patched.

Verified locally: pure Python `model` tests pass, and the checked-in `prediction_kernel_lib_darwin_arm64.so` at the current tip passes `verify_kernel_binary.py` parity. The x86_64 CI failure on 708e9a18 looks like the same stale-binary/default-token quirk from earlier in this PR - the bot's own binary-fix commit should already cover it, just needs a fresh check run.

Dropped my own in-progress changes on top of this - not needed anymore.

springfall2008:
Follow-up: pushed 031575b adding tests for the point Trefor raised — solar only pays the inverter loss on a hybrid.

That's already correct in the code (`inverter_loss_ac = self.inverter_loss` only when hybrid, else `1.0`, at `prediction.py:601-604`), and it's a different quantity from the one this branch changed: the gross-up applies `inverter_loss_recp` to the *battery* draw, and the battery sits behind the inverter in both topologies. But nothing pinned that distinction, and the two are easy to conflate when reading the discharge path, so it's worth locking in.

Four scenarios, same shortfall with PV present, across both topologies and both modes:

| | `pv_ac` | shortfall (AC) | battery draw (DC) | final SoC |
|---|---|---|---|---|
| AC-coupled | 0.40 (no solar loss) | 0.60 | 0.60 / 0.9 = 0.6667 | 9.3333 |
| hybrid | 0.36 (solar loss) | 0.64 | 0.64 / 0.9 = 0.7111 | 9.2889 |

Expected values derived analytically before running; the eco/freeze pairs match exactly, as they must given Freeze Export is ECO with charging disabled and no charging could happen in a shortfall slot anyway.

These pass on current code, so I checked they actually discriminate rather than just being green: grossing up by `inverter_loss_ac` instead of `inverter_loss` fails both AC-coupled scenarios on battery cycle (0.6 vs 0.6667) while the hybrid pair still passes — the expected signature, since the two constants coincide for hybrid.

Rebased onto e4c495c so it sits on top of the `EXPORT_LIMIT_FREEZE`/`EXPORT_LIMIT_IDLE` constants change. Full suite, `kernel_parity`, `debug_cases` and `random` all green against the freshly cross-built binaries, which also clears the binary staleness I flagged earlier.

One unrelated papercut worth a separate issue: when a `simple_scenario` assertion fails, `test_infra.py:925` calls `plot()`, which ends in `plt.show()`. On a machine with a display that blocks until the window is closed, so a failing `--test model` run never terminates — it looks exactly like an infinite loop rather than a test failure. `MPLBACKEND=Agg` works around it, but gating the `plt.show()` behind a flag or a display check would save the next person the diagnosis.

springfall2008:
## Added on this branch: `support_feedin_first` — Freeze Export recapture is now per-inverter (3ed9151e)

A second, related change has landed here. Flagging it separately because it isn't covered by the title or description above, and it changes modelled behaviour for most inverters.

### The problem

#4207 taught Freeze Export to recapture PV above the export limit into the battery instead of clipping it, gated on `inverter_can_charge_during_export`. That setting is an `apps.yaml` global about Force Export and defaults to `true`, so the recapture applied to **every** inverter.

It's only true of hardware whose Freeze Export is a genuine "Feed-in First" mode — house load, then grid export, then the battery. Most inverters implement Freeze Export by simply disabling charging (`execute.py` zeroes the charge rate or pauses charging and otherwise leaves the inverter in Demand/ECO), so PV beyond the export limit really is clipped. Modelling recapture there credits the battery with energy it never receives.

### The change

`INVERTER_DEF` gains an opt-in `support_feedin_first`, read through `.get()` so a type has to declare it. The freeze branch in `prediction.py` and its C++ kernel mirror now require both it and `inverter_can_charge_during_export`. **Force Export is unaffected** — still gated on `inverter_can_charge_during_export` alone.

It's aggregated across the whole fleet in `fetch_inverter_data()` rather than read off the first inverter — the prediction models one combined battery, so a single inverter that can't recapture disables it for all of them — and published on `sensor.predbat_inverter_config`.

### Which inverters are enabled

Only where the component actually selects an export-first work mode for the freeze:

| Type | Mode selected for Freeze Export | Where |
|---|---|---|
| FoxESS / FoxCloud | "Feed-in First" | `templates/fox.yaml` |
| SolisCloud | "Feed-in priority" storage mode, once charge current hits 0A | `solis.py`, both V1 and V2 branches |
| SolaxCloud | `feedin` work mode | `solax.py` |
| SunsynkCloud | Selling First + per-slot sell flag, sell rate 0 | `sunsynk.py`, settled on live hardware |
| DeyeCloud | `SELLING_FIRST` | `deye.py` |

Everything else keeps the clipping behaviour, which is what their freeze actually does:

- **GECloud / GE** — no mode switch at all; ECO plus `adjust_pause_mode(pause_charge=True)`.
- **AlphaESSCloud** — holds SoC through the rate entities, and separately *leaks* discharge, which is why `inverter_freeze_export_discharge_rate` exists.
- **EnphaseCloud** — freeze export maps to restrict-battery-discharge, not a feed-in mode.
- **Sigenergy (SIG / SIGCLOUD)** — `activeMode "selfConsumption-grid"`. The #4207 note about Sigenergy recapturing was under *Force Export*, which is the other gate.
- **Tesla** — `support_discharge_freeze: False`, so Freeze Export never runs.

### Kernel

Parity revision bumped 9 → 10, the flag mirrored into `PkContext`, and all six platform binaries cross-rebuilt via zig and committed alongside the source. They have to move together: the loader rejects a stale binary and silently falls back to the Python engine, and `model_kernel` fails outright when the kernel is unavailable.

### Tests

- Seven `test_model.py` scenarios covering both sides of the gate on AC-coupled and hybrid topologies, plus export-limit-absorbs-it-all, full-battery, and the `inverter_can_charge_during_export` veto — all also routed through the C++ kernel via `model_kernel`.
- Three fixed `kernel_parity` edge cases, and the flag added to the parity field list and randomised in the sweep. The randomised value is *derived from an existing draw* rather than a new `rng.choice`: an extra draw shifts the seeded stream, and the batch parity test asserts seed 4321 yields a scenario with cars.
- `test_optimise_solar`'s recapture test expanded to all four gate combinations.
- A new `INVERTER_DEF` opt-in test (asserting no other type declares the key, and that at least one omits it so the `.get()` default stays covered), plus fleet-aggregation coverage in `test_inverter_config_sensor`.
- The random benchmark now randomises the flag per scenario on its own salted rng stream, following the existing car/IOG/load-scaling idiom. Regenerating `random_scenarios.yaml` therefore added exactly one line per scenario and left everything else bit-identical. Net effect on the results baseline is **two** scenarios (ids 2 and 6, both now `false`, losing the recapture — +0.12p and +0.09p on metric); the other 18 are unchanged.
- Docs: `apps-yaml.md` under `inverter_can_charge_during_export`, a new `support_feedin_first` entry in `inverter-setup.md`'s INVERTER_DEF reference, and the sensor attribute in `output-data.md`.

### Test plan

- [x] `./run_all` — full suite passes (155s)
- [x] `./run_pre_commit` — clean
- [x] `./run_all --test model --test model_kernel --test kernel_parity --test random`

### Caveats worth a reviewer's attention

Feed-in/selling-first establishes the *ordering* (load → export → battery), which is necessary but not sufficient. The flag asserts something stronger: that when PV exceeds the **export limit**, the overflow lands in the battery rather than the panels being curtailed. Nothing measured so far covers that case:

- The Sunsynk live runs recorded in `sunsynk.py` had export "tracking PV minus load exactly" — the export limit was never binding.
- For Sunsynk and Deye, Predbat writes the slot SoC cap at the **reserve**, below current SoC. An inverter told to target the reserve may curtail rather than charge past its own cap.
- Deye's `_freeze_export_state` already carries an explicit `VERIFY@SPIKE on DEYE specifically` — its behaviour is inherited from Sunsynk on the register-parity rule, not measured on Deye hardware.

If a Deye or Sunsynk user reports SoC drifting up during Freeze Export, or *not* rising when solar clearly exceeds their export limit, this flag is the first thing to revisit — it's one line per inverter type.

Since this is a distinct change from the shortfall-discharge fix the PR is titled for, it may be worth updating the PR title/body, or splitting it out if reviewing them together is awkward. It does build on `f063890b` and `e4c495cd` from this branch, so it can't be cleanly rebased onto `main` on its own.

PR #4681, opened 2026-08-23
test(web): restore the shared components object after the debug-history route tests

romain-intel:
_make_web() in test_web_debug_history_routes.py swaps my_predbat.components for a SimpleNamespace stub, and my_predbat is shared across the whole run. It's never put back, so the next test to call is_running() dies on components.is_all_alive():
  AttributeError: 'types.SimpleNamespace' object has no attribute 'is_all_alive'

test_web_functions is what actually trips over it, several tests later, with nothing in the failure pointing back to the cause. Both tests pass in isolation — this only appears in a full-suite run, so ./run_all on current main should be failing.

The stub is restored in the existing finally block, plus a companion test asserting the routes test leaves components as it found it. That assertion is on identity rather than truthiness: is_running() guards with if self.components, so the harness leaving it as None is harmless — it's specifically a truthy stub missing the method that breaks things.


romain-intel:
Already fixed.

PR #4682, opened 2026-08-24
fix(load_ml): report liveness during training

romain-intel:
The component marks itself successful once per loop iteration, at the end - but a curriculum retrain is five passes of thirty epochs and routinely runs longer than the sixty minute window components.is_alive() allows between successes. is_all_alive() then fails, is_running() returns False, and the dashboard shows the whole of Predbat as "(unhealthy)" while nothing is actually wrong.

Restarting the component appears to fix it because that resets the clock, which is exactly why this looked like a crash and stayed hidden - there is nothing in the log, because nothing failed.

train() and train_curriculum() take an optional progress_callback, invoked once per epoch, and the component passes its own update_success_timestamp. On the observed hardware an epoch is around forty seconds, so the pulse has two orders of magnitude of headroom against the window. The hook is guarded: a progress callback must never be able to abort an hour of training.

Both of _do_training's train_curriculum call sites are wired, and the test drives the real _do_training for is_initial True and False, asserting on every call rather than the last - with is_initial the initial curriculum is followed by a fine-tune pass, so recording only the most recent call would let an unwired first one hide behind a correct second. Verified by deleting each call site in turn and confirming the test fails.

### DIFF
--- a/apps/predbat/load_ml_component.py
+++ b/apps/predbat/load_ml_component.py
@@ -997,6 +997,7 @@ class LoadMLComponent(ComponentBase):
                     curriculum_window_days=window_days,
                     curriculum_step_days=5,
                     max_intermediate_passes=8,
+                    progress_callback=self.update_success_timestamp,
                 )
             else:
                 val_mae = self.predictor.train_curriculum(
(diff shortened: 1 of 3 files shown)


springfall2008:
Change looks fine, but its not hooked up to anything?

romain-intel:
Wdym? It passes the `self.update_success_timestamp` down which is then called periodically.

PR #4683, opened 2026-08-24
Charge car on solar excess

romain-intel:
This PR allows for a car to charge on excess solar power. It is meant to mimic (and is therefore compatible) with the "Charge on Solar" offered by Tesla.

With this change, PredBat will activate car charging slots during the day when there is enough solar for the car to charge and it is not worth keeping the power to export it to the grid (low buy rates). You can then hook up an automation to switch your car's charger on/off if needed. The assumption is that the car charger will modulate the current it pulls to only use solar power (which is what the Tesla "Charge on Solar" does).

"Charge on Solar" exposes another value which is the "charge from any source". It is, unfortunately, not yet exposed through the API so this is a manual setting. WIth this setting PredBat will also schedule night slots (ie: when power is cheaper) to charge to at least that amount.

This was designed to work with Tesla's "Charge on Solar" but should work with any other system that has a similar characteristic.


romain-intel:
Anything I need to change on this one? I've been running it at my house for several weeks and it seems fairly stable and to do what is intended. Is it *too* specific for Tesla cars?

PR #4684, opened 2026-08-24
fix(hass): stop log rotation killing threads that are logging at the same time

romain-intel:
Rotation closed the logfile, renamed it, then opened the replacement. A component thread part-way through log() in that window wrote to a closed handle, took a ValueError all the way up and died. Nothing restarts a component thread mid-run - components.start() only runs from initialize() - so the thread stayed dead, is_all_alive() kept failing, and the dashboard reported Predbat unhealthy for the rest of the run. The load ML forecaster is the one this hit in practice, since its training thread logs steadily while the main thread is the only one allowed to rotate.

Two changes. Rotation now renames first and publishes the new handle before closing the old one: POSIX is happy to rename an open file, so a thread holding the previous reference writes into the rotated log, which is harmless. That leaves only the gap between another thread's read of self.logfile and its write, which write_log_line() absorbs by retrying once - by then the replacement is published - and then swallowing anything still failing to stderr. Logging must not be able to kill its caller, whatever the cause; a full disk should not cost a thread either.


romain-intel:
Any issues with this one? I saw copilot's suggestion which seems fine to me.

springfall2008:
Automated cleanup skipped: this PR's head branch lives in a fork, and the bot's credential can only write to `springfall2008/batpred`, so it cannot push the fixes back to this PR. Re-open the change from a branch in `springfall2008/batpred` to use the cleanup flow, or apply the review feedback manually.

romain-intel:
I believe it was cleaned up.

PR #4685, opened 2026-08-24
fix(execute): stop the export target draining the battery below the reserve

romain-intel:
An export limit of 0 means "empty it as far as you are allowed", and the reserve is what says how far that is. discharge_soc resolves it that way when deciding how far to actually export, but the raw unclamped limit was handed to adjust_export_immediate and adjust_battery_target_multi. On an inverter whose service maps that target onto a device reserve, the result is an instruction to drain the battery flat.

Observed on a Powerwall driven by service hooks: Predbat logged "current SoC 10.719kWh and target 1.35kWh" - correctly targeting the 5% reserve - while simultaneously writing backup_reserve 0 to the device, which then discharged past 5% towards empty.

Extracted as export_target_soc_percent() so the rule is named and testable. The clamp applies only when set_reserve_enable is off. Where Predbat owns the reserve register that register already enforces the floor and the two are independent, so raising the target there would change long-standing behaviour for no benefit.

### DIFF
--- a/apps/predbat/execute.py
+++ b/apps/predbat/execute.py
@@ -624,10 +624,11 @@ class Execute:
             # Set the SoC just before or within the charge window
             if self.set_soc_enable:
                 if isExporting:
+                    export_target_percent = self.export_target_soc_percent()
                     if not disabled_export and not self.set_reserve_enable:
                         # If we are discharging and not setting reserve then we should reset the target SoC to the discharge target
                         # as some inverters can use this as a target for discharge
-                        self.adjust_battery_target_multi(inverter, int(self.export_limits_best[0]), isCharging, isExporting)
+                        self.adjust_battery_target_multi(inverter, export_target_percent, isCharging, isExporting)
                     elif not inverter.inv_has_discharge_enable_time:
                         self.adjust_battery_target_multi(inverter, 0, isCharging, isExporting)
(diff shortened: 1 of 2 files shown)


apps/predbat/execute.py
```diff
@@ -787,6 +788,27 @@ def adjust_battery_target_multi(self, inverter, soc, is_charging, is_exporting,
...
+        maps this target onto a device reserve would be told to drain the battery flat instead.
+
+        Left alone when Predbat does own the reserve: adjust_reserve enforces the floor there and the two
+        registers are independent, so raising the target would change behaviour for no benefit.
+
+        Returns:
+        - int: export target as a percentage of the battery
+        """
+        target = int(self.export_limits_best[0])
+        if not self.set_reserve_enable:
```
springfall2008, line 820: why if not set reserve enable?

romain-intel: My understanding (and correct me if this is wrong) is that:
 - when set_reserve_enable, there is a separate place that PredBat writes the "floor" (ie: the reserve to maintain)
 - a planned limit of 0 means "discharge as much as you are allowed to"
 - when PredBat has a reserve register to write to, this is fine (ie: the reserve will act as the "floor" of export.
 - therefore, we only need to set this when there is no separate floor. If there is a floor, there is no need to change behavior.

Issue #4440, opened 2026-08-04
Predbat sets full rate charge when plan says Hold Charge


gcoan:
**Describe the bug**
At 19:00 today predbat plan was for Hold Charge (which of itself is probably a bad plan, there was plenty in the battery and Freeze Export or Demand would be more appropriate), but actually issued a full rate charge instruction to the inverter - i.e. did not instruct the inverter to hold current SoC level as per the documentation:

"Hold charging - A type of charge where the target SoC % is the same or lower than the current SoC %. This is similar to charge freeze, but it is selected as a result of planning, and cannot be manually selected."

**Expected behaviour**
Hold charge does not do a full rate charge.

**Predbat version**
8.46.11

**Environment details**
HAOS, HA 2026.7.4

2x GivEnergy Gen 1 inverters (G and H), 1 GivEnergy AC3 inverter (I), each with their own batteries

**Screenshots**
Predbat plan shows Hold Charge, status is shown as Charging, in the power meters can see all 3 inverters are grid charging at max rate
![image](https://github.com/user-attachments/assets/f7f1ba2c-e2c5-4e47-923f-d9277182474c)

Charge slot start and end time set to 19:00-20:00, charge enable turned on, charge rate set to max:
![image](https://github.com/user-attachments/assets/d7927f83-359f-45f2-8c9d-2b062bfe070d)

Target SoC controls shows that predbat did not set a target SoC level that would have stopped the full rate grid charge, target soc is set to 100%
![image](https://github.com/user-attachments/assets/4b4a0109-3d5a-4a48-bbc2-686f65008fe5)

GivTCP shows these commands all being set:
![image](https://github.com/user-attachments/assets/fbc45e99-dbb0-4ae5-a987-cdb9af7fa48c)

Logfile shows Predbat planned a full rate charge to 100% target SoC, not to hold current charge level (per the plan status):
![image](https://github.com/user-attachments/assets/169b2346-3c0c-4482-a941-de9f7fcf7e8b)

**Log file**
[predbat.1.log](https://github.com/user-attachments/files/30721933/predbat.1.log)

**Predbat debug yaml file**
I was on my ipad with the companion app so can't capture a debug file, I have my apps.yaml though which shows that charge_limit is correctly set to number.sensor.xxx_target_soc

```
pred_bat:
  module: predbat
  class: PredBat

  # Sets the prefix for all created entities in HA - only change if you want to run more than once instance
  prefix: predbat

  # enable main updates to be loaded from custom repository
[...]

PR #4687, opened 2026-08-24
fix(inverter): stop holds commanding a grid charge on charge-target inverters

romain-intel:
Predbat holds the battery by writing a reserve one percent above the current SoC, which stops discharge on inverters that defend the reserve passively. The Powerwall instead treats the reserve as a level to charge towards, so every hold - freeze charge, Hold for car, Hold for iBoost - was a grid import request.

Adds a reserve_is_charge_target inverter capability (TESLA only; every other inverter definition is unchanged and reads the default via .get). When set:

- hold_reserve_percent() holds exactly at the level instead of one above it
- adjust_reserve() refuses to write a reserve above the current SoC at all, so the bug class is impossible regardless of what a caller asks for
- freeze targets use a new calc_percent_limit_floor(), because rounding to nearest turns a 61.6% SoC into a 62% target - the same off-by-one by another route

Also fixes a related divergence in the Teslemetry scheduler emulator, which decided device commands from its own Fleet live_status poll rather than the SoC Predbat planned against (different on a site with local monitoring), and could replay a persisted idle reserve above SoC after a restart.

Refs #4424, #4432, #4440


romain-intel:
I fixed the recent conflicts -- any issues with this one? Let me know if you need anything else. Thanks!

romain-intel:
I apologize for asking again but please let me know if there are any issues with these PRs. I am currently maintaining my own version of batpred for this and my other PRs and it's getting complicated to rebase and keep them up to date with your very quick development pace.

If there are things that you would like changed, I am happy to work on them to get them up to your standards (for this and the other PRs as well).

Thanks for this great software. It's made a huge difference this summer for me in California in optimizing for proper exports.

PR #4710, opened 2026-08-24
fix(solis): stop rewriting a CID 636 TOU bit the inverter refuses

### LINKED ISSUE #4707: Solis: inverter clears the TOU bit on CID 636 while a slot is active, so timed charge/discharge slots are ignored
Split out of #4702, where this turned up alongside the discharge cut-off SOC problem. Different register, different mechanism, so tracking separately.

## Summary

Writes to CID 636 (storage mode) fail verification, and the difference is always **exactly bit 1**, `SOLIS_BIT_TOU_MODE`:

```
Warn: Solis API: Failed to verify CID 636 storage mode to 35 on <SERIAL>,  wrote 35  but read back 33
Warn: Solis API: Failed to verify CID 636 storage mode to 179 on <SERIAL>, wrote 179 but read back 177
```

`35 → 33` and `179 → 177` are both `-2`. Every other bit in the written value survives; only the TOU bit is cleared by the inverter.

That matters because with the TOU bit clear the inverter ignores the timed charge/discharge slots entirely — so the slots Predbat has carefully configured are not acted on, and the plan does not execute.

## This is not the known idle case

#4239 / `e4b66c6a` ("don't force TOU bit on CID 636 when no V2 slot is active") gated the mode choice on `slot1_active` precisely because the inverter only retains the TOU bit while a window is configured. **These failures are on the other side of that gate.**

The two branches log distinguishable lines, so it can be read straight off:

- `setting storage mode to 'Self-Use'` / `'Feed-in priority'` → `slot1_active == True`
- `... and no active slot for <SERIAL>, setting storage mode to 'Self-Use - No Timed Charge/Discharge'` → `slot1_active == False`

Pairing every CID 636 failure with the mode decision immediately preceding it:

| Inverter | path | CID 636 failures | preceded by |
|---|---|---|---|
| A | V2 | 35 | `slot1_active=TRUE` — **35 of 35** |
| C | V2 | 4 | `slot1_active=TRUE` — **4 of 4** |

Not one occurred in the `no active slot` branch. So the #4239 gate is doing its job and this is something else.

Representative sequence:

```
Solis API: Charge current is 210.0A for <SERIAL>, setting storage mode to 'Self-Use'
Solis API: Set storage mode to 179 on <SERIAL>
Warn: Solis API: Failed to verify CID 636 storage mode to 179 on <SERIAL>, wrote 179 but read back 177
Warn: Solis API set storage mode encountered errors for <SERIAL>
```

## Not V2-specific either

A third inverter (**B**) shows the same `35 → 33` drop but has **zero** `setting storage mode to` lines in its log — it is on the V1 (CID 103) path throughout. Its failures come from the `set_storage_mode_value()` call that emits `Warn: Solis API set storage mode encountered errors`, and they occur mid-run (e.g. during a `Charging` cycle), not only at startup.

So whatever clears bit 1 is not tied to the V2 branch.

## Frequency

Counted over each inverter's current log (a ~14 hour window for A):

| Inverter | CID 636 verify failures |
|---|---|
| A | 35 |
| B | 14 |
| C | 4 |

Two further inverters in the same sample showed 3 and 1.

## What I have not established

This is **log evidence only** — unlike the CID 5965 problem in #4702, I have not driven CID 636 against the hardware. So the mechanism is genuinely open, and in particular I cannot yet say whether the inverter:

- **refuses the write outright**, keeping the old value (which is what CID 5965 turned out to do), or
- **accepts it and then clears bit 1 again** shortly afterwards, of its own accord

Those want different fixes, so it is worth settling before changing any code.

## Suggested next step

A short read/write/poll against a live inverter would separate the two cases, the same shape that resolved #4702:

1. Read CID 636 and confirm bit 1 state while a slot is active.
2. Write the value with bit 1 set, then re-read immediately.
3. Keep polling for a couple of minutes. If it comes back with bit 1 set and then loses it, the inverter is reverting; if it never takes, it is refusing.
4. Try setting bit 1 in isolation, in case the refusal depends on the other bits being changed in the same write.

Happy to run that if it is useful.

## Interaction with #4702

PR #4706 deliberately **excludes** the storage mode result from the component health escalation it adds, so this cannot age affected systems out to unhealthy while the cause is unknown. The exclusion is structural rather than explicit — `set_storage_mode_value()` keeps the write result in a local and returns `None` — so there is a regression test pinning it. Worth being aware of that test if this is ever fixed by folding the result into `write_time_windows_if_changed()`'s return value.

## Environment

Observed on **v8.50.0**, across SolisCloud inverters on both V1 and V2 time-of-use paths and on both OAuth and api-key auth.

springfall2008:
Fixes #4707.

## The problem

Some Solis inverters accept a CID 636 (Modbus 43110) storage mode write — the control API answers with `code 0` — but keep the time-of-use bit clear. `compute_solis_mode_value()` starts from the cached value, the failed verify read writes the read-back into that cache, and the next cycle recomputes the same value and writes it again. Once a minute, indefinitely, with a warning each time:

```
Solis API: Set storage mode to 35 on <SN>
Warn: Solis API: Failed to verify CID 636 storage mode to 35 on <SN>, wrote 35 but read back 33
```

## Why the bit is refused

On firmware "4B and above" the timed charge/discharge enable moved out of CID 636 and into the per-slot registers — `SOLIS_CID_CHARGE_ENABLE_BASE`/`..._DISCHARGE_ENABLE_BASE`, Modbus 43707, six slots rather than three — and every mode value carrying bit 1 (3, 35, 43, 51, 98) was dropped from the mode table at the same time. The modes shift down: 35 becomes 33, 98 becomes 96, 51 becomes 49. That accounts for both values in the issue, including `179 → 177` (`51 + 128` → `49 + 128`, bit 7 being an independent modifier).

Corroborated by `wills106/homeassistant-solax-modbus` 2024.12.1 ("Simplified Energy Control Switch options", "6 charge and discharge slots rather than 3") and its separate `plugin_solis_fb00.py`, whose 43110 option table drops exactly the bit-1 values.

## Why this does not gate on the firmware generation

`is_tou_v2_mode()` is not a firmware read — it checks CID 6798 for 43605 (`0xAA55`). The alignment with 4B is an inference, and the failure would be asymmetric: if a V2-marked inverter *does* honour bit 1, sending 33 means "Self-Use - No Timed Charge/Discharge" on the older table, silently disabling timed charging. Inverter B in the issue is also on the V1 path with the same signature, which a V2 gate would not touch.

So `note_tou_bit_refused()` learns it from the inverter instead. Only a read-back differing by exactly bit 1 counts — a write dropped wholesale says nothing about that bit and must not disable it. The verdict expires after `SOLIS_TOU_BIT_REPROBE_HOURS` (8) and is retested, which bounds the cost of a false positive: #4239 already found the bit is only retained while a window is configured, and `slot1_active` reflects Predbat's intent rather than the inverter's state, so one probe at one moment can mislead. Nothing is persisted, so a restart re-probes and a firmware update cannot be masked by a stale verdict.

Net effect: one write and one warning per inverter per 8 hours, instead of one per minute.

## Second fix: no-battery inverters

`automatic_config()` already refused to enrol an inverter reporting `batteryType 'No Battery'`, but only into a local list. It stayed in `self.inverter_sn`, so the control loop and `startup_reset_registers()` kept driving the full write path against PV-only hardware every cycle. Predbat never plans for it, so its stale slot 1 is never cleared, `slot1_active` stays true and it asks for a storage mode forever.

In the log this was diagnosed from, that was **11 of the 12** CID 636 failures. `is_battery_inverter()` now gates every control write. The inverter is still polled and published, so its PV sensors keep working.

## Tests

Nine new tests in `test_solis.py`, all written before the code:

- `test_compute_solis_mode_value_can_drop_the_tou_bit` — value table including `179 → 177`; default behaviour pinned
- `test_storage_mode_stops_asking_for_a_refused_tou_bit` — four cycles produce exactly one write and one warning
- `test_storage_mode_keeps_the_tou_bit_when_the_inverter_accepts_it` — no false positive
- `test_storage_mode_does_not_learn_from_an_unrelated_verify_failure` — a wholesale refusal retries with the bit
- `test_refused_tou_bit_is_retested_after_the_reprobe_window` — quiet at 7h59m, one re-probe at 8h
- `test_refused_tou_bit_is_not_persisted_across_a_restart`
- `test_run_skips_control_writes_for_no_battery_inverter`
- `test_run_skips_startup_register_reset_for_no_battery_inverter`
- `test_run_logs_why_a_no_battery_inverter_is_not_controlled`

The storage-mode tests drive the real `read_and_write_cid()` path via a stub inverter, so the write → verify → cache round-trip is genuinely exercised.

Full suite green (`./run_all`, exit 0). `pre-commit` clean.

## Not in scope

- `switch.<prefix>_solis_<sn>_time_of_use` still writes bit 1 directly when a user toggles it, so on an affected inverter that switch cannot work. Worth a separate decision: hide it once a refusal is known, or let it fail visibly.
- `read_and_write_cid()` verifies with no settle time. The timing evidence says lag is not what is happening here, but the code cannot distinguish refusal from lag in principle.
- Persisting the refusal would mean introducing the Storage component to `solis.py`, which it does not use today. The re-probe window covers the same ground without it.

## Still open

Whether the cleared bit actually breaks slot execution. On 4B firmware 33 *is* "Self-Use" and the slots are gated by CIDs 5916/5922, which verify fine — so this may be log noise rather than a dead plan. Worth confirming against a real charge slot.


springfall2008:
## Source for the "bit 1 was dropped from the mode table" claim

Permalinked at `82e072fd` in [`wills106/homeassistant-solax-modbus`](https://github.com/wills106/homeassistant-solax-modbus), so the line numbers stay valid.

The same register, in the two firmware plugins that project ships.

[`plugin_solis.py` L1053-1074](https://github.com/wills106/homeassistant-solax-modbus/blob/82e072fd7f4916839b765b2ce9650f9411ffc895/custom_components/solax_modbus/plugin_solis.py#L1053-L1074) — pre-4B:

```python
        register=43110,
        option_dict={
            1: "Self-Use - No Grid Charging",
            3: "Timed Charge/Discharge - No Grid Charging",
            17: "Backup/Reserve - No Grid Charging",
            33: "Self-Use - No Timed Charge/Discharge",
            35: "Self-Use",
            37: "Off-Grid Mode",
            41: "Battery Awaken",
            43: "Battery Awaken + Timed Charge/Discharge",
            49: "Backup/Reserve - No Timed Charge/Discharge",
            51: "Backup/Reserve",
            64: "Feed-in priority - No Grid Charging",
            96: "Feed-in priority - No Timed Charge/Discharge",
            98: "Feed-in priority",
        },
```

[`plugin_solis_fb00.py` L2209-2225](https://github.com/wills106/homeassistant-solax-modbus/blob/82e072fd7f4916839b765b2ce9650f9411ffc895/custom_components/solax_modbus/plugin_solis_fb00.py#L2209-L2225) — 4B and above:

```python
        register=43110,
        option_dict={
            1: "Self-Use - No Grid Charging",
            17: "Backup/Reserve - No Grid Charging",
            33: "Self-Use",
            37: "Off-Grid Mode",
            41: "Battery Awaken",
            49: "Backup/Reserve",
            64: "Feed-in priority - No Grid Charging",
            96: "Feed-in priority",
        },
```

Removed: **3, 35, 43, 51, 98** — exactly the values with bit 1 set. The names then slide down onto the bit-1-clear values: `33` goes from "Self-Use - No Timed Charge/Discharge" to plain "Self-Use", and `49` and `96` do the same for Backup/Reserve and Feed-in priority.

That is both values from the issue:

- `35 → 33` — old "Self-Use" to new "Self-Use".
- `179 → 177` — `51 + 128` to `49 + 128`; old "Backup/Reserve" to new "Backup/Reserve", bit 7 being an independent modifier and surviving.

The read-only mirrors of the register carry the same two tables: [`plugin_solis.py:1638`](https://github.com/wills106/homeassistant-solax-modbus/blob/82e072fd7f4916839b765b2ce9650f9411ffc895/custom_components/solax_modbus/plugin_solis.py#L1638) and [`plugin_solis_fb00.py:2790`](https://github.com/wills106/homeassistant-solax-modbus/blob/82e072fd7f4916839b765b2ce9650f9411ffc895/custom_components/solax_modbus/plugin_solis_fb00.py#L2790) (register 33132).

### Where the enable moved to

[`plugin_solis_fb00.py` L1511-1522](https://github.com/wills106/homeassistant-solax-modbus/blob/82e072fd7f4916839b765b2ce9650f9411ffc895/custom_components/solax_modbus/plugin_solis_fb00.py#L1511-L1522) adds a register the older plugin does not have at all:

```python
        name="Timed Charge Discharge On Off",
        key="timed_charge_discharge_on_off",
        register=43707,
```

and [L2060-2078](https://github.com/wills106/homeassistant-solax-modbus/blob/82e072fd7f4916839b765b2ce9650f9411ffc895/custom_components/solax_modbus/plugin_solis_fb00.py#L2060-L2078) drives it one bit per slot:

```python
        name="Timed Charge Slot 1 Enable",
        register=43707,
        register_bit=0,
...
        name="Timed Charge Slot 2 Enable",
        register=43707,
        register_bit=1,
```

`43707` appears **zero** times in `plugin_solis.py`. Those bits are Predbat's `SOLIS_CID_CHARGE_ENABLE_BASE` / `SOLIS_CID_DISCHARGE_ENABLE_BASE` CIDs, which the V2 path already writes and verifies successfully.

### Provenance

[Release 2024.12.1](https://github.com/wills106/homeassistant-solax-modbus/releases/tag/2024.12.1):

> **Solis Firmware 4B and above:** (Separate plugin, until firmware detection is working #1142 #1143 thanks @fboundy)
>
> - Simplified Energy Control Switch options
> - 6 charge and discharge slots rather than 3

Field reports of the same symptom on the newer hardware: [`Pho3niX90/solis_modbus#93`](https://github.com/Pho3niX90/solis_modbus/issues/93) (S6-EH3P20K-H, "Time Of Use switch turns off again after about 30/40 seconds") and [`wills106/homeassistant-solax-modbus#1110`](https://github.com/wills106/homeassistant-solax-modbus/issues/1110) ("My Solis have new firmware with better charge/discharge control. With this firmware, Time Of Use switch not work"), the latter closed by the 2024.12.1 plugin above.

### Two caveats

1. This is a community integration, not a Solis document. The authoritative source is **RS485_MODBUS (ESINV-33000ID) Hybrid Inverter V3.2, Appendix 8** — which `Pho3niX90/solis_modbus` cites for its own table, but which is NDA-gated.
2. All of the above is register 43110 over local Modbus. That **CID 636 is the cloud-side handle for 43110** is an inference, resting on Predbat's `SOLIS_BIT_*` constants matching that bit layout exactly and its storage mode names matching the pre-4B option labels verbatim.

Neither caveat affects this PR: `note_tou_bit_refused()` learns the refusal from the inverter's own read-back and does not depend on any of it.

PR #4724, opened 2026-08-25
fix(config): allow car charging rate down to 0.1 kW

### LINKED ISSUE #4721: car charging rate can't be set below 1 kW, would be nice for motorcycle and e-bike charging
I used top be able to set the car_charging_rate to below 1 kW as I'm using it to charge an electric motorcycle (650W and 2 e-bikes (200W). now the value always returns to 1kW if I input these values

Remove the limiter on the car_charging_rate value

non

**Additional context**

chalfontchubby:
## Summary
- Lowers the `min` on all `car_charging_rate*` config items (Car 0-7) from 1 kW to 0.1 kW
- Lets low-power devices like e-bikes and motorcycles (200-650W) be configured without HA silently clamping the value back up to 1 kW

Fixes #4721

## Test plan
- [x] Verified `config.py` parses and the change is limited to the `min` field on the 8 car charging rate entries
- [ ] Pre-commit could not be run locally (broken venv in this worktree, unrelated to this change) — relying on CI


chalfontchubby:
Closing — duplicate of #4722, which fixes the same 8 lines and was opened first.

PR #4739, opened 2026-08-25
refactor(givtcp): finish moving GivTCP REST handling into the component

chalfontchubby:
Follow-on to #4649, per your ask there. Stacked on that branch rather than `main` - review/merge that one first.

Takes the three things #4649 explicitly left behind off the direct REST client, so `inverter.py` no longer parses GivTCP's REST blob.

### Commits

1. `6c83993c` - **inverter mode and pause controls as entities.** `inverter_mode` (`Control.Mode`) plus, on v3 only, `pause_mode`/`pause_start_time`/`pause_end_time`. All four already existed as writable `sensor_list` keys, so the entity paths in `Inverter` were already there and the REST branches simply go. Pause is gated on every configured inverter reporting v3, mirroring the existing `givtcp_rest_power_ignore` "any opts out, leave them all" rule - v2 has no `/setBatteryPauseMode`, which is why the old REST path was v3-gated too.

2. `d614c316` - **the #4517 discharge-target model check.** Rather than plumbing the raw model string out to a new apps.yaml key so `Inverter` could keep making the decision, the component just doesn't publish a `discharge_target_soc` entity for a model whose register doesn't work. `adjust_force_export` already leaves a target it can't read alone, so an absent entity is the whole mechanism - and `inverter.py` no longer knows about inverter models at all. A mixed fleet still works per-inverter.

3. `97ad0a0e` - **battery/capacity discovery.** `GivTCPRest` gains normalised accessors that absorb the version differences previously open-coded in `Inverter.__init__`: v2's `Invertor_Details` vs v3 renaming it to the serial number, nominal capacity in raw register units vs kWh, per-pack temperature averaging across three field shapes, and the two calibration representations. The `19.53125` divisor is carried over verbatim - it was back-calculated rather than derived, and this doesn't pretend to explain it.

### Two bugs found on the way

Both mutation-tested (confirmed failing without the fix):

- **`_handle_write` had no branch for a select that isn't a time slot.** A plain-select write was silently dropped and the entity then republished unchanged, which reads as a successful no-op. Would have bitten the moment `inverter_mode`/`pause_mode` landed.
- **The rate entities advertised a generic `max` of 20000W.** `Inverter` derives `battery_rate_max_raw` for a GE inverter from exactly that attribute. Harmless while REST discovery still overrode it - but the moment discovery moved, every GivTCP user's battery rate limit would silently have become 20kW. They now carry `Invertor_Max_Bat_Rate`.

### Judgement calls worth a look

- **`adjust_inverter_mode`'s 30s `changed_start_end` sleep stays gated on `rest_api`.** It works around GivTCP's own HA integration lagging behind a window write; a GivTCP-REST inverter has no such lag, because the component applies the write and republishes inline before returning. Dropping the guard would have quietly added 30s to every window change for GivTCP users.
- **One new optional apps.yaml key, `battery_calibration`** (documented). A calibration cycle drives the battery outside its normal SoC range, so Predbat disables itself for that inverter while one runs. Absent means never calibrating, so nothing changes for inverters that don't report it.
- **The pause entity carries GivTCP's native vocabulary** (`Disabled`/`PauseCharge`/...), not the GE Cloud spelling - `adjust_pause_mode` picks between the two from the value it reads back, so this keeps it on the same side of that branch as the REST path was.

### Deliberately still on REST in `inverter.py`

The version/serial/firmware metadata read, `update_status`'s refresh, and the reserve read. That last one differs between the branches - REST doesn't apply the `battery_min_soc` floor the entity path does, and `reserve_percent_current` is a reading of what the inverter is *actually* set to, so collapsing them is a behaviour change rather than a refactor. Left for a separate decision rather than changed silently.

### Testing

31 component tests (up from 23). The two real captures (`rest_v2.json`/`rest_v3.json`) move from `test_inverter` to the component tests and now assert the accessors rather than `Inverter`'s internals, plus per-version calibration coverage. `test_inverter`'s REST-shaped mode/target assertions updated - the REST phase now asserts the entity write *and* that no REST command is issued. Full `run_pre_commit` green.

Also fixed a test-isolation bug: a `queue_rest_data` entry fed `setBatteryMode`'s read-back verification, and with that call gone it leaked into the next test's `runAll` and made an unrelated REST test retry 5x.

**Same caveat as #4649: no validation against real hardware.** Draft until someone dogfoods it. The write/republish timing is the part only a live inverter really proves.


springfall2008:
Superseded by #4864, which combines this with #4649 onto a branch based on `main`.

Your commits are carried over unchanged with authorship intact — nothing here is lost, and the per-commit write-up in this PR is still the reference for the rationale. The only reason for the move is that this PR was stacked on #4649's branch, so once `main` was merged in GitHub rendered the diff as 121 files / +36,867 rather than its real 7 files / +660.

Full test suite and pre-commit are green on the merged tree. Thanks @chalfontchubby — closing this in favour of the flattened one.

Issue #4701, opened 2026-08-24
ALPHAESS: Time based dischage does not force discharge


dragon2611:
**Describe the bug**
With the new alphaESS integration if you manually set an export timeslot, predbat correctly sets a discharge schedule on the alphaESS side but despite setting a discharge amount to the inverters capacity the system doesn't actually output at that power level and continues to only export excess solar.

**Expected behaviour**
Export at Inverter max as requested by predbat

**Predbat version**

8.53.0

**Environment details**

Alpha ESS Smile G3 5kw
- HAOS

The problem here is I don't think this is a predbat issue so much as a limitation Alpha's side, i'm not sure if the timed discharge is supposed to force a discharge to the grid or just control when the batteries allowed to discharge to the load?

Axel may have a way of doing it but since they're an oauth integration they may have access to another API that we do not.

springfall2008:
🤖 **Automated first-pass triage** — this is a bot's initial analysis against current `main`. A maintainer will review before any action is taken, so treat the below as a starting point rather than a verdict.

**Classification:** `bug` (control gap in the new AlphaESS Cloud component) · **Priority:** `priority_low` — confined to the experimental AlphaESS Cloud integration, though there's no in-Predbat workaround on the affected path.

### What I checked

Investigated against `main` @ `ef83ca68` (just past the `v8.53.0` tag you're running — nothing has changed in this area since).

Your instinct is right, and the code shows why. The AlphaESS Open API surface Predbat talks to (`apps/predbat/alphaess_const.py:17-32`) has exactly two ways to program an export window, and neither carries a "force discharge to grid" concept — there is no working-mode or forced-export endpoint in the Open API at all:

1. **Legacy path** — `updateDisChargeConfigInfo`. The payload Predbat builds is `ctrDis`, `timeDisf1/e1`, `timeDisf2/e2`, `batUseCap` (`apps/predbat/alphaess.py:1091-1141`). **That endpoint has no power field whatsoever.** The export power you set (`number.predbat_alphaess_<sn>_battery_schedule_export_power`) is used purely as an on/off gate here — `export_rate <= 0` means freeze, anything above zero means "window enabled" (`alphaess.py:1112`) — the actual watt value is never transmitted. So the inverter is only being told *"discharging is permitted between X and Y, down to `batUseCap`"*, which in Alpha's terms is a discharge **permission**, not a forced grid export. That matches your symptom precisely.
[...]

PR #4742, opened 2026-08-25
Add confirmed findings from a day of live triage runs to the debug journal

springfall2008:
## Summary

Reviewed the triage bot's own logs (`~/predbat-triage-bot/logs/*.log`) from a day of live runs against real issues and pulled out the findings a future triage session would want, per CLAUDE.md's own instruction to add to this file when something new turns up. Filtered to facts confirmed by code-reading, log evidence, or empirical probing — dropped one candidate (an AlphaESS periodic-path reserve-hold theory) that the source triage session itself flagged as unconfirmed without live hardware, and skipped an Ohme v1→v2 API finding since it was already fixed same-day in 8dbdbb85.

- **Solis** — corrects the existing CID 636 entry: it doesn't explain a report with zero idle-mode log lines, which by elimination points at a third write path (the HA select handler) the entry didn't cover.
- **Sigenergy** — adds a confirmed `has_reserve_soc` override danger: it bypasses two guards that normally make reserve writes inert.
- **AlphaESS** (new entry) — no self-consumption fallback outside a configured discharge window, confirmed against AlphaESS's own documented behaviour.
- **Solcast** (new entry) — `max_kwh` stays at its `9999` default on the Solcast/HA-sensor paths, so the array-ceiling warning can never fire there.
- **HA write/verify** (new entry) — `refresh=True` is a no-op in a normal add-on install, and a control entity that resolves to a `sensor.*` domain gets a false-success write, confirmed by direct probing.
- **New symptom-table row** — a stuck/stale plan replacement, confirmed from a 24h log.
- **New trap** — `gh issue list --search` without `--repo` searches all of GitHub.

Every entry cites the GitHub issue it came from and the specific file:line evidence, matching the existing entries' style.

## Test plan

- [x] Verified both markdown tables' column counts are unbroken after the edits
- [x] `./run_pre_commit` on the changed file (markdownlint, cspell) — clean
- N/A — docs-only change, no code paths affected


springfall2008:
Fixed both citations Copilot caught in 48df8db6 (`solis.py:2551`→`2574`, and the inconsistent `#4707` vs `GH#4707` style; `solcast.py:1190`→`1191`).

While I was in there, re-verified every other citation in this PR against current main rather than trusting the source triage logs — found three more with the same kind of drift, not flagged by the review:
- Sigenergy reserve guard: `execute.py:935-939` → `947-950`
- `should_replace_plan()` and its tie-break: `plan.py:1087`/`1098` → `1103`/`1114`
- The AlphaESS entry's `execute.py:430-521` range didn't match anything at that location on inspection, so rather than guess a replacement I can't verify, I dropped the line-pin and kept the underlying claim, which is still well-sourced (AlphaESS's own documented behaviour, cited in #4701).

Every `.py:N` citation in the diff is now checked against the current tree.

Issue #4749, opened 2026-08-25
Debug history rolling capture is on by default, retaining ~24 MB per install


springfall2008:
### Describe the bug

`debug_history_enable` defaults to on, so every install captures a full `create_debug_yaml()` snapshot every 3 hours and retains 15 of them.

Those snapshots are large — around 1.6 MB each in my testing. With the default `count = 15`, that is roughly **24 MB held per install, continuously**, plus a ~1.6 MB write through the storage component every interval, whether or not anyone ever opens the History view.

For a diagnostic aid I would expect that to be opt-in rather than opt-out. Making it default-off costs nothing in capability: per the documentation, `debug_history_force_capture` still takes a snapshot while `debug_history_enable` is off, so anyone who spots a problem can still capture on demand — they just would not be carrying a 45-hour rolling buffer the rest of the time.

This is most noticeable where the storage component is backed by something with a size ceiling rather than a local disk. There, a 24 MB per-install baseline is a large fraction of the budget, and it scales with `debug_history_count` rather than with how often anyone actually looks at a snapshot.

### Where the defaults are set

`apps/predbat/config.py`:

```python
{
    "name": "debug_history_enable",
    "friendly_name": "Debug history rolling capture enable",
    "type": "switch",
    "icon": "mdi:history",
    "default": True,      # <-- enabled for everyone
},
{
    "name": "debug_history_count",
    ...
    "default": 15,        # <-- 15 x ~1.6 MB retained
},
{
    "name": "debug_history_interval",
    ...
    "default": 3,
},
```

`apps/predbat/predbat.py` (~line 843), where the `get_arg` fallbacks mirror the same values:

```python
count = int(self.get_arg("debug_history_count", 15))
interval_hours = max(1, int(self.get_arg("debug_history_interval", 3)))
enabled = self.get_arg("debug_history_enable", True)
```

### Expected behaviour

`debug_history_enable` defaults to off. Users who want the rolling buffer turn it on; everyone else does not pay for it. `debug_history_force_capture` keeps working either way, so on-demand capture is unaffected.

### Suggested fix

- `"default": False` for `debug_history_enable` in `config.py`
- Match the `get_arg` fallback in `predbat.py`
- Update `docs/customisation.md`, which currently reads "(default on)"

Optionally, a smaller default `debug_history_count` would cut the footprint for people who do enable it, since the cost is linear in that number.

### Note

[...]

PR #4750, opened 2026-08-25
fix: trim debug_history_count default from 15 to 5

springfall2008:
## Summary

Partial response to #4749 — that issue asks for `debug_history_enable` to default off entirely (24MB/install, 45h rolling buffer, whether or not anyone ever opens History). Rather than switching the feature off by default, this trims the buffer size instead, as a middle ground:

- `debug_history_count` default `15` → `5`. With the unchanged 3h `debug_history_interval`, that's a 15h rolling window instead of 45h, and roughly **~8MB instead of ~24MB** per install (still a meaningful cut, ~66%).
- The feature stays **on by default**, so the buffer keeps covering the "something looked off a few hours ago" case that #4417/#4438 originally built it for — it just no longer reaches back the full ~48h History-tab window.
- Also removed the duplicated `15`/`3`/`True`/`False` literals from the `get_arg()` calls in `_capture_debug_history()` (`predbat.py`) — they now resolve through `config_index` instead, so `config.py`'s `CONFIG_ITEMS` is the single source of truth for these defaults rather than needing to be kept in sync by hand at two call sites.
- Updated `docs/customisation.md` to match (default count, resulting window).

### Note for maintainer review

This is offered as a compromise, not a resolution — #4749 explicitly asks for default-off, and the earlier #4417/#4438 decision to default this on was deliberate. If the storage-cost concern in #4749 still isn't satisfied by a smaller buffer, the default-off change is a one-line follow-up (`config.py`'s `debug_history_enable` default plus the docs line) rather than something this PR needs to also do.

Closes/relates to #4749 (leaving to a maintainer to decide whether this satisfies the issue or whether default-off is still wanted).

## Test plan

- [x] `./run_all --test debug_history_capture` — passes (values are set explicitly via `expose_config()` in every test, so the default change doesn't affect them)
- [x] `./run_all --quick` — full quick suite passes, no regressions
- [x] `pre-commit run --files apps/predbat/config.py apps/predbat/predbat.py docs/customisation.md` — ruff/black/cspell all pass

### DIFF
--- a/apps/predbat/config.py
+++ b/apps/predbat/config.py
@@ -1213,7 +1213,7 @@ CONFIG_ITEMS = [
         "step": 1,
         "unit": "snapshots",
         "icon": "mdi:history",
-        "default": 15,
+        "default": 5,
     },
     {
         "name": "debug_history_interval",
--- a/apps/predbat/predbat.py
+++ b/apps/predbat/predbat.py
@@ -840,10 +840,10 @@ class PredBat(hass.Hass, Octopus, Energidataservice, Stromligning, Fetch, Plan, 
         genuine capture attempt is never skipped just because the warning was
         recently logged.
         """
-        count = int(self.get_arg("debug_history_count", 15))
-        interval_hours = max(1, int(self.get_arg("debug_history_interval", 3)))
-        enabled = self.get_arg("debug_history_enable", True)
-        forced = self.get_arg("debug_history_force_capture", False)
+        count = int(self.get_arg("debug_history_count"))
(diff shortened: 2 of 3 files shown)


chalfontchubby:
Consider 8 hour intervals, 3 stored. That covers the last 16-24 hours. The initial intention was to always have something available to mitigate the "something weird in the plans. 

Maybe just keep a couple for "today" and one at midnight of the night before anchoring the full duration of the history. 

Gzip the stored versions? I didn't because I used them in a follow up to plot the predicted Soc overlaid for all the snapshots, but I'm unlikely to push that. It a chart only its parents could love.

Issue #1692, opened 2024-12-05
Time till iboost starts


mogons57:
**Is your feature request related to a problem? Please describe.**
I have iboost setup to switch on (and heat water) when the electricity prices are low (which works well) but it would be nice to be able to spot when this is going to happen and then use an automation to switch off the boiler in anticipation, so that the best use could be made of the cheaper electricity. Often I'm finding that my hot water tank is already quite full when the iboost switches on so I'm not making the most of the cheap power. 

**Describe the solution you'd like**
A "time_till_iboost" sensor that says how long it is till an iboost session starts. For example if it is currently 18:00 and predbat has decided that iboost should start at 12:30 the "time_till_iboost" sensor should be 04:30 (i.e. 4 hours and 30 mins). If no iboost session is coming then the sensor could be set to a large number (say 48 hours) or set to unavailable (if that is possible).

A test could then be used in an automation along the lines of 
if (time_till_iboost < 3 hours) then switch off boiler
boiler would then switch back on when iboost session stops

**Describe alternatives you've considered**
I've though about trying to get this out of the predbat plan but could not work out how to do that.

**Additional context**
An example of what time_till_iboost might look like on the run up to and during some iboosting 

time     iboost    time_till_iboost 
21:00    -            1:30    
21:30    -            1:00
22:00    -            0:30
22:30    1.5         0
23:00    1.5         0
23:30    -            0:30
00:00    1.5         0
00:30    -            1:00
01:00    -            0:30
01:30    1.5         0
02:00    -            48:00 / unavailable
02:30    -            48:00 / unavailable

gcoan:
I think you should be able to get the data out of the existing predbat.iboost_best sensor https://springfall2008.github.io/batpred/output-data/#iboost-solar-diverter-data which has the times of the planned iboost in its attributes

mogons57:
In case anyone else wants to do this, the code below may not be that efficient but seems to work:

```
{% set iboost_times = state_attr("predbat.iboost_best","results") %}
{% set times =  iboost_times.keys()|list %}
{% set iboost_energy =  iboost_times.values()|list  %}
{% set ni = namespace(x=0) %}
{% set data = namespace(h_bool=False) %}
{% set iboost_starts = "" %}
{% for ni in range(0,times|count()-1) if data.h_bool == false  %}
{## {{ times[ni]}} {{ iboost_energy[ni]}}, {{iboost_energy[ni+1]-iboost_energy[ni]}} ##}
[...]

Issue #1424, opened 2024-09-02
Option to end iBoost early and replan, e.g. when hot water cylinder recovered


nbullus:
**Is your feature request related to a problem? Please describe.**
The **predbat_iboost_max_energy** entity controls the maximum amount of energy to plan for when suitably priced charge slots are available. On some days the maximum amount of energy may not actually be required, e.g. cylinder already 90% heated before iBoost triggers. Therefore iBoost will charge the remaining 10% and then stay on for the full duration. 

It can be more cost effective to end iBoost early in this situation by setting **predbat_iboost_max_energy** to zero. This will then cause a re-plan which will drop the remaining charge slots it originally planned for iBoost charging and could result in a better more cost effective plan.

**Describe the solution you'd like**

Predbat to detect when iBoost power has stopped, e.g. immersion thermostat turned off, automatically adjust the max energy to zero, and a re-plan will then occur and remove the remaining scheduled iBoost. It is unclear whether Predbat  can detect this or whether we would need to supply one of the following in the yaml:

1. Supply a power sensor for your iBoost device. Predbat can then detect when power drops to zero when iBoost is currently active.
2. Supply a binary sensor informing Predbat to end iBoost early for today.

iBoost max energy would need to be reset back at the start of day so two entities may be needed instead of one:

- predbat_iboost_max_energy - maximum daily iBoost energy required
- predbat_iboost_target_energy - iBoost energy to target today

User would set the predbat_iboost_max_energy as they do now which would be used to then set predbat_iboost_target_energy at 00:00. iBoost would then plan off predbat_iboost_target_energy. When instructed to finish early by the aforementioned binary/power sensor, Predbat would set predbat_iboost_target_energy to zero and trigger iBoost to end for the day. Entity predbat_iboost_target_energy would then be reset back to predbat_iboost_max_energy at the start of the next day and the process repeated.

Not everyone may want this feature so it would need to be controlled by a new switch to decide whether to end iBoost early or keep it active in case hot water is used within the iBoost period.

**Describe alternatives you've considered**
I have amended my own iBoost automation to include this feature along with a number of other features making it quite a big and cumbersome automation now:

1. Turn off my gas hot water schedule when iBoost is planned to avoid conflicts and prevent the cylinder being heated by gas before iBoost is planned to start.
2. End iBoost when cylinder recovered.
3. Restore gas hot water schedule once iBoost has completed or finished early.
[...]

PR #4756, opened 2026-08-26
iBoost: optional demand-forecast-driven smart planning (iboost_forecast)

psweens:
## Summary

Predbat can optionally plan iBoost energy from a 30-minute hot-water demand forecast and the tank's current state of charge, treating the tank as a charge-only store, so heating is scheduled before each draw and stops once the forecast is covered. The feature is off unless `iboost_forecast` is configured in `apps.yaml`; existing behaviour is unchanged otherwise, and a regression test enforces that the legacy path produces identical output when no forecast is loaded.

## Problem

`iboost_max_energy` is a per-calendar-day scalar; the smart planner books the cheapest slots without knowing when hot water is drawn or how much energy the tank already holds, so intraday shortfalls are served reactively at uncontrolled rates and full budgets are booked on days the cylinder is already hot (#1424). Users with SoC-reporting tanks rebuild the missing logic as template sensors plus an automation rewriting the budget intraday, which races the replan cycle around midnight.

## Configuration

`apps.yaml`:

```yaml
iboost_forecast:
  - sensor.hot_water_demand_forecast$results   # same format and parsing as load_forecast; cumulative kWh
iboost_forecast_scaling: 1.0                   # multiplier applied to the forecast, e.g. 0.11 for a series in % of an 11 kWh tank
iboost_tank_soc: sensor.tank_charge            # optional; percent of usable stored energy currently in the tank
```

Home Assistant entities (created under `iboost_enable`):

- `input_number.predbat_iboost_tank_capacity` — usable stored energy, kWh (default 10)
- `input_number.predbat_iboost_tank_reserve` — minimum stored energy to hold before each draw, kWh (default 0)
- `input_number.predbat_iboost_fill_rate_threshold` — fill remaining headroom in any slot whose import rate is at or below this value, p/kWh (default −99, disabled). Implemented as a separate commit so it can be dropped in review without touching the rest.

A check on the configuration shape and naming would be appreciated.

## Behaviour

When `iboost_forecast` is configured:

- `fetch_iboost_forecast()` (modelled on `fetch_extra_load_forecast()`) reads the configured sensors, scales the cumulative series by `iboost_forecast_scaling` and converts it into demand per plan interval as the per-minute positive delta (the same reading the load forecast gets), logging a one-line summary. If the sensor is missing, unavailable, stale or yields no data, a warning is logged and the unchanged legacy planner runs for that cycle.
- Initial stored energy is `clamp(iboost_tank_soc / 100 × capacity, 0, capacity)` when the SoC sensor is configured and readable, else 0 so the whole forecast is provisioned.
- The tank level is maintained at interval resolution as a charge-only store: draws remove energy (clamped at zero), boosts add it (clamped at capacity). Two constraints hold for every interval: the level never falls below the reserve at any draw, and never exceeds the capacity after any boost.
- Selection is earliest-deadline, cheapest-first: the intervals are walked in time order and whenever a draw would breach the reserve, the cheapest eligible earlier interval is booked (respecting `iboost_rate_threshold`, `iboost_rate_threshold_export`, the gas thresholds, the element power, the per-calendar-day `iboost_max_energy − iboost_today` cap and the capacity headroom), preferring intervals adjacent to already-booked ones on price ties so boosts consolidate into longer runs, and repeating until the draw is covered. If no eligible interval remains a warning names the uncovered interval and energy and planning continues.
- The current interval is usually partially elapsed (the plan is rebuilt every few minutes): only its remainder can be booked and its slot is emitted starting at the current minute, mirroring the legacy planner's clamp, so booked energy is always deliverable. Demand already drawn earlier in the current interval is excluded during ingestion since the tank SoC reading reflects it.
- With the fill threshold set, eligible intervals at or below it top the tank up to its capacity headroom, walking in time order so fill energy carries forward; after each fill the level trajectory is re-verified and later planned boosts the fill energy has made redundant are trimmed immediately, so the level stays at or below capacity everywhere and the day cap only ever counts energy that survives. (Within the set of qualifying slots, an earlier fill displaces a later booking regardless of relative price — flagging in case cheapest-first filling is preferred.)
- The forecast planner runs in exactly the modes the smart planner runs in (iBoost enabled, and either both solar and battery modes off or `iboost_smart` on); when configured it takes over slot selection in those modes, including the time-ordered plan used when `iboost_smart` is off.
- Slots are emitted in the existing `iboost_plan` structure (`start`, `end`, `kwh`, `average`, `cost`), sorted by time with no duplicate starts, so `in_iboost_slot()`, the plan card, `binary_sensor.predbat_iboost_active` and the export-window clash checks are unchanged. `average` and `cost` use each slot's own import rate. The level trajectory is logged at debug level and a one-line summary at info level.

The first commit is a separable two-line fix to `plan_iboost_smart()` this feature depends on: the sliding-window average priced every window from its first interval only, and the duplicate-slot guard tested a stale loop variable so overlapping windows could double-book a slot start. There is no separate PR for it; it is included here as its own commit with its own test.

## Scope

Planner only. No change to `run_prediction()` or the C++ kernel, no parity revision, no binary rebuild; both prediction engines consume the plan through the existing `in_iboost_slot()` path. Carrying tank state into the prediction loop (so the solar and charging modes respect headroom) is a possible follow-up, deliberately excluded, as is publishing the modelled per-slot tank level (the iBoost plan slots are not currently published as an entity, so there is no attribute to attach it to).

## Alternatives considered

External scaffolding: template sensors deriving a remaining-demand estimate, an automation rewriting `iboost_max_energy` intraday, or setpoint automations on the diverter. A scalar budget cannot express when the energy is needed, so heating still lands in the cheapest slots of the day regardless of draw timing, cannot span a multi-day horizon, and the intraday rewrite races the replan cycle around midnight.

## Forecast source

`load_forecast` sets the precedent for ingesting an external forecast series (Predheat, PredAI). A producer exists: [ML Forecast Lab](https://github.com/psweens/ml-forecast-lab) (disclosure: the author is me), an open-source Home Assistant app that benchmarks state-of-the-art ML forecasting backends on the user's own sensor history and publishes calibrated forecasts as HA sensors in the `load_forecast` attribute format. For Mixergy tanks the demand series comes from the tank's own charge-drop meter, so it represents actual draw rather than heating energy.

## Testing

New cases in `apps/predbat/tests/test_iboost.py` (registered as `iboost_forecast`, with the bug-fix case under `iboost_smart`), all on non-flat rate profiles:

- window pricing and duplicate-slot guard fix (fails on the pre-fix planner)
- forecast ingestion: parsing, scaling, negative-delta clamping, multi-sensor summing, and fallbacks for unconfigured/missing/stale sensors, scalar states and data beyond the planning horizon
- energy booked before a draw, not in a cheaper slot after it
- a draw before any cheap slot books the cheapest preceding slot
- a full tank books nothing (#1424 behaviour)
- reserve held at every draw; capacity never exceeded by a large cheap window; capacity headroom binding across a draw–refill–draw cycle
- calendar-day cap respected alongside the forecast
- import, export and gas threshold eligibility; adjacency consolidation on price ties
- mid-interval replans: the partially elapsed current interval is pro-rated in both the planner and the ingestion
- uncovered demand warns and later draws are still planned; a zero element power books nothing
- fill threshold fills only qualifying slots, the trajectory stays at or below capacity after trimming, and trimmed energy is refunded to the day cap
- regression: with no forecast loaded the legacy planner output is identical before and after a forecast plan has run, and every plan is sorted with positive-length, non-overlapping, duplicate-free slots

Commands run: `./run_pre_commit` (all hooks pass) and `./run_all --quick` (all tests pass), plus `./run_all --test iboost_smart --test iboost_forecast`.

Known limitation: the tariff compare feature raises `forecast_minutes` to 48h after sensor data is fetched, so with `forecast_hours` configured below 48 the comparison sees demand only over the shorter horizon; a follow-up could rebuild the forecast grid inside `recompute_iboost()`.

## Related

- #1424 (budget booked on days the cylinder is already hot)
- #1692
- The `plan_iboost_smart()` window-pricing/duplicate-slot fix is included as the first commit of this PR (separable).


apps/predbat/tests/test_iboost.py
```diff
@@ -7,10 +7,57 @@
 # pylint: disable=consider-using-f-string
 # pylint: disable=line-too-long
 # pylint: disable=attribute-defined-outside-init
+from datetime import timedelta
+
```
springfall2008, line 11: **Stale base / merge conflict risk:** this PR is marked CONFLICTING against main (baseRefOid 59c83740 predates main's f98554f6, 28 Aug, which already applied the slot_start window-averaging fix and added `run_iboost_smart_average_test`). The PR's first commit is therefore half-redundant (only the `start in used_slots` guard is novel), and if the test_iboost.py conflict is resolved in the PR's favour, the GH#4817 regression test and its call in `run_iboost_smart_tests` are deleted, resurrecting the under-priced-window bug. A rebase onto current main is needed before review.

apps/predbat/plan.py
```diff
@@ -5163,7 +5166,7 @@ def plan_iboost_smart(self):
                         end = min(start + self.plan_interval_minutes, slot_end)
 
                         # Avoid duplicate slots
-                        if minute in used_slots:
+                        if start in used_slots:
```
springfall2008: **Legacy fallback can emit overlapping slots (pre-existing, not fixed by this change):** the replaced guard only blocks a re-use of the same slot-start. Slots are at most one interval long, but when `minutes_now` is not a multiple of `plan_interval_minutes` (the normal case: runs every 5 min, default interval 30), the window containing `minutes_now` clips its first slot to `minutes_now` (off-grid phase) while later windows book on the grid phase — e.g. minutes_now=735, window 720-780 books [733,763) and the next window books [760,790): they overlap. `in_iboost_slot` (utils.py:1500) then returns only the first match, so the second slot's energy drops out of the prediction model. The fixture pins `minutes_now` to 720/735-aligned grids, so the PR's own `check_slot_invariants` cannot catch this. Consider clipping all slots to the interval grid or checking end-vs-start in the guard.

apps/predbat/fetch.py
```diff
@@ -2539,6 +2551,119 @@ def fetch_extra_load_forecast(self, now_utc, ml_forecast=None):
...
+                for minute, value in forecast.items():
+                    demand_cumulative[minute] = demand_cumulative.get(minute, 0) + value
+            else:
+                self.log("Warn: Unable to load the iBoost demand forecast from {}. Skipping forecast source.".format(entity_id))
+
+        if not demand_cumulative:
+            self.log("Warn: iboost_forecast is configured but no forecast data could be loaded, using the legacy iBoost smart plan")
+            return {}
+
+        if (last_data_minute is None) or (last_data_minute <= self.minutes_now):
```
springfall2008: **Staleness gate trusts timestamps from sources that contributed nothing:** the raw-timestamp loop (2604-2610) records min/max for every entity, but `minute_data` silently drops items whose units cannot be converted (utils.py:1138 'Ignore data in wrong units') or whose values are non-numeric. With two configured sensors, one valid-but-stale (ends in the past) and one recent-but-unusable (wrong unit / non-numeric values), `last_data_minute` is recent so this fallback warning never fires — `demand_cumulative` comes only from the stale source, every interval delta clamps to zero, and the forecast planner runs with zero demand and books nothing, silently disabling iBoost heating for the cycle. Judge the extent from per-source loaded data (e.g. min/max key over the returned `forecast` dict), not the raw timestamps.

apps/predbat/fetch.py
```diff
@@ -2539,6 +2551,119 @@ def fetch_extra_load_forecast(self, now_utc, ml_forecast=None):
...
+            self.log("Warn: iBoost demand forecast ends at minute {} which is not in the future, using the legacy iBoost smart plan".format(last_data_minute))
+            return {}
```
springfall2008, line 3034: **Barely-future data passes the staleness gate:** `last_data_minute` only has to be strictly greater than `minutes_now`, so a sensor publishing data that ends one minute ahead passes the gate, and the conversion below then produces a non-empty dict of all-zero demands (dips/back-fill clamp to 0). The forecast planner then runs and books no slots, and unlike the `<=` case there is no warning at all — legacy iBoost heating is silently off. Consider requiring the loaded future data to cover some meaningful horizon (e.g. at least one plan interval of future demand > 0) before taking over from the legacy plan.

apps/predbat/fetch.py
```diff
@@ -809,6 +809,18 @@ def fetch_sensor_data(self, save=True):
...
+        self.iboost_forecast = {}
+        self.iboost_tank_soc_percent = None
+        if self.iboost_enable:
+            self.iboost_forecast = self.fetch_iboost_forecast()
+            if self.iboost_forecast and ("iboost_tank_soc" in self.args):
+                tank_soc = self.get_arg("iboost_tank_soc", None)
+                try:
+                    self.iboost_tank_soc_percent = float(tank_soc)
+                except (ValueError, TypeError):
+                    self.log("Warn: Unable to read iboost_tank_soc (value {}), assuming the tank is empty".format(tank_soc))
```
springfall2008: **Failure posture amplifies spend far beyond the legacy path:** when the tank SoC sensor is unavailable/non-numeric the code logs 'assuming the tank is empty' and `plan_iboost_forecast` provisions the ENTIRE forecast (stored_start = 0), which can book up to `iboost_tank_capacity` (default 10 kWh) per day at real import prices — the legacy path was bounded by `iboost_max_energy` (default 3.0 kWh). A tank sensor that drops offline for one fetch cycle therefore books ~10 kWh of unnecessary immersion boost, then collapses back the next cycle, so the plan flaps between two structurally different slot sets. Consider falling back to the legacy planner (like a stale forecast does) when the SoC sensor is configured but unreadable, rather than assuming empty.

apps/predbat/plan.py
```diff
@@ -5214,6 +5217,233 @@ def plan_iboost_smart(self):
...
+        levels_post = []
+        level = stored_start
+        for index in range(len(demand)):
+            levels_pre.append(level)
+            level = max(level - demand[index], 0.0)
+            level = min(level + boost[index], capacity)
+            levels_post.append(level)
+        return levels_pre, levels_post
+
+    def plan_iboost_forecast(self):
```
springfall2008, line 5432: **iboost_smart_min_length silently stops applying once a forecast is configured:** the legacy planner enforces a minimum window length (config min 30 / default 30 / max 120 minutes, plan.py:5225) but `plan_iboost_forecast` never reads it and emits slots as short as 5 minutes (e.g. the PR's own `iboost_plan_reserve` case books a 20-minute slot). With small forecast draws a user configured for 30-120 minute minimum boosts gets 5-25 minute diverter cycles instead, with no log line or doc noting that min_length stops applying. Either honour min_length by consolidating adjacent booked intervals, or document the changed behaviour in both doc pages.

apps/predbat/tests/test_iboost.py
```diff
@@ -7,10 +7,57 @@
...
 from tests.test_infra import reset_rates2, reset_inverter
 
 
-def run_iboost_smart_test(test_name, my_predbat, today=0, max_energy=1, max_power=1, min_length=0, expect_cost=0, expect_kwh=0, expect_time=0):
+def set_rate_profile(my_predbat, profile, default_rate=20.0, export_rate=0.0):
```
springfall2008, line 15: **Fixture state pollution (GH#5079 class):** `set_rate_profile` overwrites the shared fixture's entire rate_import/rate_export (including the reset_rates2 half-hour pattern) and re-runs rate_scan/rate_scan_export, leaving rate_min/rate_max/rate_min_forward/rate_export_min set to the last profile's values (e.g. -2.0 after iboost_plan_fill) while rate_min_base/rate_max_base still hold the last reset_rates2 values. The cleanup restores the iboost_* scalars but never the rates. Impact is latent (the next test resets rates itself) but run_shuffle --hunt will bisect this suite as a culprit for any later test reading rates without resetting. Please snapshot/restore the rate dicts and the rate_min* scalars around the suite, and restore iboost_rate_threshold/_export to 9999 (the reset() default) rather than 100.

apps/predbat/plan.py
```diff
@@ -5214,6 +5217,233 @@ def plan_iboost_smart(self):
...
+        # tank is empty so the whole forecast is provisioned
+        if self.iboost_tank_soc_percent is not None:
+            stored_start = min(max(self.iboost_tank_soc_percent / 100.0 * capacity, 0.0), capacity)
+        else:
+            stored_start = 0.0
+
+        start_minute = int(self.minutes_now / interval) * interval
+        interval_starts = list(range(start_minute, start_minute + self.forecast_minutes, interval))
+        num_intervals = len(interval_starts)
+        demand = [self.iboost_forecast.get(minute, 0.0) for minute in interval_starts]
```
springfall2008, line 5466: **Diverged forecast grid silently reads zero with no warning:** demand is looked up with self.iboost_forecast.get(minute, 0.0) on a grid re-derived from plan-time minutes_now, but fetch_iboost_forecast() keyed the dict on the fetch-time grid. When plan_iboost_smart() is invoked from compare.py:511 recompute_iboost() after minutes_now has advanced past an interval boundary (compare raises forecast_minutes after data fetch), every interval reads 0.0 and the `demand[target] <= 0` short-circuit below skips the uncovered-demand warning entirely, so the hot-water draws are silently dropped. The PR already flags the shorter-horizon part of this; the silent-empty case (no warning at all) is worth a debug log when the lookup misses entirely.

apps/predbat/plan.py
```diff
@@ -5214,6 +5217,233 @@ def plan_iboost_smart(self):
...
+        # only its remainder can be boosted, mirroring the legacy planner's clamp to minutes_now
+        interval_max_energy = [interval_energy for _ in range(num_intervals)]
+        if num_intervals:
+            interval_max_energy[0] = self.iboost_max_power * max(interval_starts[0] + interval - self.minutes_now, 0)
+
+        # Per-interval import rates and boost eligibility, using the same rate and gas threshold
+        # rules as the legacy smart plan but applied to each interval's own rates
+        import_rates = []
+        eligible = []
+        for minute in interval_starts:
```
springfall2008, line 5482: **Cleanup (reuse):** plan_iboost_forecast() re-implements the legacy planner's rate-eligibility rules inline (import threshold, export threshold, gas-vs-import, gas-vs-export, lines 5286-5307) instead of sharing a helper with plan_iboost_smart() (plan.py:5163-5175) - the PR's own comment admits the coupling ('the same rate and gas threshold rules as the legacy smart plan'). A threshold/gas rule change must now be edited in two places; updating only one makes the legacy and forecast planners disagree on eligible minutes, so users switching modes (or falling back on a stale sensor) get different slot choices for the same tariff.

apps/predbat/plan.py
```diff
@@ -5214,6 +5217,233 @@ def plan_iboost_smart(self):
...
+                day_room = iboost_max - day_usage[interval_starts[slot_n] // (24 * 60)]
+                amount = min(interval_max_energy[slot_n] - boost[slot_n], day_room, capacity - levels_post[slot_n])
+                if amount <= 0.001:
+                    continue
+                boost[slot_n] += amount
+                self.iboost_forecast_trim(demand, boost, stored_start, capacity)
+
+        return self.plan_iboost_forecast_slots(interval_starts, boost, import_rates, stored_start, capacity, demand, uncovered_kwh)
+
+    def iboost_forecast_day_usage(self, interval_starts, boost, total_days):
```
springfall2008, line 5590: **Cleanup (reuse/simplification):** iboost_forecast_day_usage() duplicates the legacy per-calendar-day cap bookkeeping (plan.py:5142-5146: same total_days expression, same day[0]=iboost_today seed) with drift already present: the legacy loop rounds accumulations with dp3(), this helper does not round. Two independent ledgers for the same iboost_max_energy cap mean a fix to one (midnight boundary, dp3 rounding) leaves the other stale, and the rounding drift makes the planners book different totals against an identical budget. Consider one shared day-usage helper for both planners.

apps/predbat/plan.py
```diff
@@ -5214,6 +5217,233 @@ def plan_iboost_smart(self):
...
+        """
+        plan = []
+        for index, start in enumerate(interval_starts):
+            kwh = boost[index]
+            if kwh <= 0.001:
+                continue
+            interval = self.plan_interval_minutes
+            slot_start = max(start, self.minutes_now)
+            available = start + interval - slot_start
+            length = int(min(round(((kwh / self.iboost_max_power) / 5) + 0.5, 0) * 5, available))
```
springfall2008: **Cleanup (simplification):** third hand-rolled copy of the round-minutes-to-5 idiom: this joins plan.py:5310 (plan_iboost_smart) and plan.py:5410 (plan_car_charging), each with slightly divergent clamping (this one adds min(max(length, 5), available)). A rounding change applied to one planner makes them emit differently-quantised slots for the same element power (the AlphaESS snap_time_grid precedent is where this bites first). A single shared quantise_minutes helper would cover all three.

apps/predbat/fetch.py
```diff
@@ -2539,6 +2551,119 @@ def fetch_extra_load_forecast(self, now_utc, ml_forecast=None):
                     load_forecast_final[minute] = value
         return load_forecast_final, load_forecast_array
 
+    def fetch_iboost_forecast(self):
```
springfall2008: **Cleanup (reuse):** fetch_iboost_forecast() copies the load_forecast ingestion block of fetch_extra_load_forecast() (fetch.py:2831) nearly line for line (~60 lines): entity_ids unwrap, 569Xlattribute split, get_state_wrapper try/except, dict->array conversion, identical minute_data() kwargs, per-entity warn-and-skip, cross-source summation. Its docstring says 'modelled on fetch_extra_load_forecast()'. A change to shared forecast-format handling must now be made twice; fixing only load_forecast leaves the iBoost path on old behaviour, and a format Predbat learns to read for load stays unsupported for hot-water demand (silently falling back to the legacy planner). Also, the per-minute positive delta below re-implements Fetch.get_from_incrementing(data, index, backwards=False) (fetch.py:752, 'max(data.get(index+1) - data.get(index), 0)'), which fetch.py already uses for exactly this reading on the load forecast - the comment's 'same reading get_from_incrementing gives the load forecast' guarantee is currently enforced only by copy.

apps/predbat/plan.py
```diff
@@ -5214,6 +5217,233 @@ def plan_iboost_smart(self):
...
+
+        # Earliest-deadline, cheapest-first booking: cover each draw from the cheapest eligible
+        # earlier interval, preferring intervals adjacent to already-booked ones on price ties so
+        # boosts consolidate into longer runs
+        uncovered_kwh = 0.0
+        for target in range(num_intervals):
+            if demand[target] <= 0:
+                continue
+            while True:
+                levels_pre, levels_post = self.iboost_tank_trajectory(stored_start, capacity, demand, boost)
```
springfall2008: **Efficiency (worst case untested):** the earliest-deadline while-loop recomputes the full O(N) tank trajectory, a fresh O(target) running-max array, an O(N) day-usage pass and an O(target) candidate scan on every dose iteration, and each dose is capped at one interval's element energy. With plan_interval_minutes=5 and a long forecast (~540 intervals) plus a mis-scaled ibost_forecast_scaling that leaves demand nonzero everywhere, this is ~540 targets x many doses x O(540) work = tens of millions of Python steps per plan rebuild, every 5 minutes. All existing tests use 30-min intervals (36 intervals), so the worst case is untested. Cheaper: update levels incrementally per dose and keep the running max as a single rolling value.

apps/predbat/plan.py
```diff
@@ -5214,6 +5217,233 @@ def plan_iboost_smart(self):
...
+        interval_max_energy = [interval_energy for _ in range(num_intervals)]
+        if num_intervals:
+            interval_max_energy[0] = self.iboost_max_power * max(interval_starts[0] + interval - self.minutes_now, 0)
+
+        # Per-interval import rates and boost eligibility, using the same rate and gas threshold
+        # rules as the legacy smart plan but applied to each interval's own rates
+        import_rates = []
+        eligible = []
+        for minute in interval_starts:
+            import_rate = self.rate_import.get(minute, self.rate_min)
```
springfall2008: **Current (partially elapsed) interval is gated and priced on the rate at its start, not at minutes_now:** for the first interval, import_rates[0] = self.rate_import.get(interval_starts[0]) can be up to plan_interval_minutes-1 minutes stale, while the emitted slot starts at minutes_now (plan_iboost_forecast_slots line 5422). plan_interval_minutes=30, minutes_now=737, rate 20p before minute 735 then 5p from 735 with iboost_rate_threshold=10: the eligible check uses rate_import[720]=20p so the 23 remaining 5p minutes are rejected as ineligible (and the draw goes uncovered even though eligible energy exists); conversely if the rate rose at 735 the slot is priced 20p for energy actually drawn at 5p, so the published plan cost is wrong. Use the rate at max(start, minutes_now) for interval 0.

apps/predbat/plan.py
```diff
@@ -5214,6 +5217,233 @@ def plan_iboost_smart(self):
...
+                level_max_to_target = [0.0 for _ in range(target)]
+                running_max = 0.0
+                for slot_n in range(target - 1, -1, -1):
+                    running_max = max(running_max, levels_post[slot_n])
+                    level_max_to_target[slot_n] = running_max
+
+                best = None
+                best_key = None
+                best_room = 0.0
+                for slot_n in range(target):
```
springfall2008: _Automated comment from the triage bot._

The candidate loop is `range(target)`, so a draw landing in the current interval (target = 0) has **no candidates at all** — not even the remainder of the current interval, whose capacity is explicitly computed (`interval_max_energy[0] = max_power * (interval - minutes_now)`) and used to cover later draws (test `iboost_plan_partial_interval`). The legacy smart planner would book a boost inside the current interval.

Scenario: a hot-water draw is forecast inside the current plan interval with a cheap eligible rate in progress — the planner logs "Warn: iBoost forecast plan cannot cover … no eligible earlier slot" on every rebuild while the draw is in progress and books nothing, even though the element could run now at max power. All the PR tests place draws in later intervals, so this path is untested.

apps/predbat/fetch.py
```diff
@@ -2539,6 +2551,119 @@ def fetch_extra_load_forecast(self, now_utc, ml_forecast=None):
...
+        demand = {}
+        total = 0.0
+        start_minute = int(self.minutes_now / self.plan_interval_minutes) * self.plan_interval_minutes
+        for minute in range(start_minute, start_minute + self.forecast_minutes, self.plan_interval_minutes):
+            kwh = 0.0
+            for offset in range(self.plan_interval_minutes):
+                offset_minute = minute + offset
+                if offset_minute < self.minutes_now:
+                    continue
+                kwh += max(demand_cumulative.get(offset_minute + 1, 0) - demand_cumulative.get(offset_minute, 0), 0.0)
```
springfall2008: _Automated comment from the triage bot._

The comment above claims the region `minute_data` back-fills beyond the data reads as zero demand, but `minute_data` tail-fills with `newest_state` — the value of the **earliest** sample (utils.py fill loop) — not the last one. So a cumulative series that ends lower than it started (e.g. an internal reset: 6.0 kWh at min 0 … 0.0 reset … 2.0 last) produces a positive delta at the minute right after the last sample: 6.0 − 2.0 = +4.0 kWh phantom demand, which the planner then pre-heats for.

Related, with multiple configured sensors of different horizons: a sensor whose series ends earlier contributes a step drop where its cumulative value vanishes, and the per-minute `max(..., 0)` clamp on the *summed* series can swallow a genuine draw from another sensor in that same minute (its delta is absorbed into the negative sum). Computing deltas per sensor before summing, or tail-filling with the last sample for this series, would close both.

docs/customisation.md
```diff
@@ -631,6 +631,35 @@ Note: Gas rates have to be configured in `apps.yaml` using **metric_octopus_gas*
...
+### iBoost demand forecast
+
+If you have a forecast of your hot water demand (for example from a tank that reports its own state of charge, such as a Mixergy tank), the smart planner
+can schedule heating ahead of each draw instead of spreading a fixed daily energy budget over the cheapest slots.
+
+Configure **iboost_forecast** (and optionally **iboost_forecast_scaling** and **iboost_tank_soc**) in `apps.yaml` -
+see [iBoost energy](apps-yaml.md#iboost-energy) for the data format. When **iboost_forecast** is not configured the planner behaviour is unchanged.
+
+The forecast planner runs exactly when the smart planner would: iBoost must be enabled and either both **switch.predbat_iboost_solar** and
+**switch.predbat_iboost_charging** must be Off, or **switch.predbat_iboost_smart** must be On. A configured forecast has no effect in the solar or
```
springfall2008: _Automated comment from the triage bot._

The two sentences contradict each other. The first says the forecast planner runs whenever "**switch.predbat_iboost_smart** must be On" — which per the gate (`((not iboost_solar) and (not iboost_charging)) or iboost_smart`) plus the dispatch at the top of `plan_iboost_smart()` means it also takes over when `iboost_solar` is On **and** `iboost_smart` is On. The second sentence then says a configured forecast "has no effect in the solar or battery modes". In that smart+solar combination the forecast planner does run, so a solar-excess user with smart enabled gets forecast-driven slots despite the second sentence.

Suggest clarifying that the no-effect claim only applies when `iboost_smart` is Off.

apps/predbat/fetch.py
```diff
@@ -2539,6 +2551,119 @@ def fetch_extra_load_forecast(self, now_utc, ml_forecast=None):
...
+
+        # Convert the cumulative series into demand per plan interval, aligned to the same interval
+        # grid the planner books slots on. Per-minute deltas are clamped to zero individually (the
+        # same reading get_from_incrementing gives the load forecast) so dips in the series and the
+        # region minute_data back-fills beyond the data both read as zero demand. Minutes already
+        # elapsed in the current interval are skipped: any draw there is in the tank SoC reading.
+        demand = {}
+        total = 0.0
+        start_minute = int(self.minutes_now / self.plan_interval_minutes) * self.plan_interval_minutes
+        for minute in range(start_minute, start_minute + self.forecast_minutes, self.plan_interval_minutes):
```
springfall2008: _Automated comment from the triage bot._

This dict is keyed over the **live** horizon, but in tariff compare mode it gets planned over 48 hours: `run_single()` calls `fetch_sensor_data(save=False)` while `forecast_minutes` still holds the live config value, then raises `forecast_minutes` to 2880 (compare.py, main) and calls `recompute_iboost()` → `plan_iboost_smart()` → the new forecast planner. Every interval past the live horizon reads `self.iboost_forecast.get(minute, 0.0)` = 0.

Consequence for the compare scenario: with `iboost_fill_rate_threshold` set (or a tank below reserve), the fill/restore pass books tank-fill slots across the ~40-hour tail at the scenario's rates — energy the live plan would never book — so the tariff metrics (`final_iboost`, cost, metric) and the "best tariff" recommendation are skewed. Worth either refetching the forecast after the horizon change or clamping the planner's horizon to the fetched demand extent.

apps/predbat/fetch.py
springfall2008: iBoost review: phantom demand from minute_data's earliest-value fills. minute_data() back-fills every minute before the first sample with 0 (utils.py:1324-1331) and every minute at/after the last sample with the value of the EARLIEST sample (newest_state, utils.py:1300-1310). The per-minute delta conversion here therefore: (a) books the first sample's full cumulative value as a phantom draw when the series starts mid-horizon with a non-zero first point (producer publishing a rolling window from 'now'); (b) creates a positive phantom delta just past the last sample whenever the series ends below its first value (e.g. a tank meter series with an internal midnight reset), e.g. points (min 0, 6.0) ... (min 2160, 2.0) -> +4.0 kWh phantom demand; (c) in multi-sensor mode a sensor whose data ends earlier contributes a step drop that, combined with another sensor's real draw in the same minute, clamps to 0 and silently masks that draw (fetch.py:2652). All three are silent over- or under-heating with no warning. Suggest skipping deltas at/before first_data_minute, computing the tail region from the last sample value instead of newest_state, and normalising each source's extent before summing.

apps/predbat/plan.py
```diff
@@ -5214,6 +5217,233 @@ def plan_iboost_smart(self):
...
+            interval_max_energy[0] = self.iboost_max_power * max(interval_starts[0] + interval - self.minutes_now, 0)
+
+        # Per-interval import rates and boost eligibility, using the same rate and gas threshold
+        # rules as the legacy smart plan but applied to each interval's own rates
+        import_rates = []
+        eligible = []
+        for minute in interval_starts:
+            import_rate = self.rate_import.get(minute, self.rate_min)
+            export_rate = self.rate_export.get(minute, 0)
+            rate_okay = True
```
springfall2008: iBoost review: the current (partially elapsed) interval is gated and priced on the rate at its START minute, while the slot it emits starts at minutes_now (plan_iboost_forecast_slots). With plan_interval_minutes=30 and a rate step inside the interval: minutes_now=737, rate_import[720..734]=20p, rate_import[735+]=5p, iboost_rate_threshold=10 -> import_rates[0]=20 -> the interval is rejected although the remaining 23 minutes are at 5p (draw goes uncovered, warning every cycle); conversely if the rate ROSE at 735, the slot is emitted with the stale cheaper average/cost, so plan cost and the compare metrics are wrong. Suggest reading the rate at max(minute, minutes_now) for the current interval.

apps/predbat/plan.py
springfall2008: iBoost review: this per-start duplicate-slot guard does not actually guarantee the non-overlap invariant check_slot_invariants() asserts. When minutes_now is not a multiple of plan_interval_minutes (production runs every 5 minutes) the first window's slot_start is clamped to minutes_now, so overlapping cheap windows book on different phase grids and produce overlapping slots. Simulated with interval=30, minutes_now=735, iboost_smart_min_length=60, two overlapping cheap windows: slots [(735,765),(765,780),(750,780),...] -> (735,765) overlaps (750,780), and in_iboost_slot() then collapses the overlap so the prediction under-counts booked energy. Pre-existing on main, but this PR adds the invariant check + the dedup fix and its fixture pins minutes_now=720 (grid-aligned), so the gap is untested. A test with minutes_now=735 and min_length=60 would catch it.

apps/predbat/fetch.py
springfall2008: iBoost review: the staleness gate is too weak, and it mixes timestamps from sources minute_data() rejected. (1) Only last_data_minute > minutes_now is required: a series whose last sample is 1 minute ahead passes, minute_data back-fills everything past it, demand is ~all zeros, and the forecast planner books nothing for the remaining horizon - iBoost heating silently stops with no warning (the documented stale/missing fallback never fires). Same for a producer that publishes fresh but flat/zero increments (non-empty all-zero dict passes the truthiness gate at plan.py:5109 and books nothing, no warning). (2) The raw extent loop (line 2604) records timestamps for sources whose values minute_data rejects (bad units / non-numeric), so a junk-but-recent source can pass the gate while the only usable source is stale. Suggest gating on the last sample carrying non-zero future increments, and tracking the raw extent per source that actually loaded.

apps/predbat/tests/test_iboost.py
springfall2008, line 15: iBoost review: set_rate_profile() overwrites the shared fixture's entire rate_import/rate_export (from minute 0, clobbering the reset_rates2 half-hour pattern) and rate_scan()/rate_scan_export() re-derive rate_min/rate_max/rate_average/rate_min_forward/rate_export_min from the custom profile - none of it is restored. The final cleanup deletes only the +120-minute tail keys, so the fixture is left with rate_min=5.0 (or -2.0 after iboost_plan_fill) while rate_min_base/rate_max_base still hold the reset_rates2 values. Also run_iboost_forecast_plan_test() never restores iboost_tank_capacity/iboost_tank_reserve/iboost_gas_scale (only the trailing regression block re-seeds them). This is exactly the #5079 class ./run_shuffle --hunt hunts: any later test reading self.rate_min/rate_export_min (e.g. plan_iboost_smart's rate_import.get(minute, self.rate_min) default) or a mid-module abort leaks the last profile into an unrelated test. Suggest resetting rates and tank state in a finally block.

apps/predbat/plan.py
springfall2008: iBoost review: hot-loop cost. Every dose iteration recomputes the full O(N) iboost_tank_trajectory, an O(target) running-max, an O(N) iboost_forecast_day_usage and an O(target) candidate scan; doses are capped at one candidate's element energy per iteration, so worst case is O(N^2 x doses). With plan_interval_minutes=5 and a 45h horizon (~540 intervals) and demand in every interval that is tens of millions of Python steps per plan rebuild, and the plan is rebuilt every 5 minutes - watchdog-restart territory on low-power hosts. All tests run at 30-min intervals (~36 intervals) so this is untested. A booking at index best only changes the trajectory suffix from best onward and one day bucket - keep levels/day_usage incrementally and update them after each boost[best] += amount; the fill pass (line 5367) has the same shape (trajectory + day_usage recomputed even when no fill is applied).

springfall2008: iBoost review (2nd note on this line): a draw inside the current interval (target 0) can never be covered, because the candidate scan is range(target) - not even by the current interval's remaining boost capacity, which is explicitly computed here (interval_max_energy[0]). The legacy smart planner books the in-progress interval (slot_start = minutes_now), so compared with legacy the tank runs cold for any draw already in progress, with a 'cannot cover X kWh ... no eligible earlier slot' warning repeated every rebuild. If draw-then-boost ordering within an interval makes simultaneous booking unsafe, at least document that draws in the current interval are never planned; otherwise allow booking the current interval's remainder for a draw later within that same interval.

apps/predbat/plan.py
```diff
@@ -5122,8 +5125,8 @@ def plan_iboost_smart(self):
             slot_length = 0
             slot_count = 0
             for slot_start in range(minute, minute + iboost_min_length, self.plan_interval_minutes):
-                import_rate += self.rate_import.get(minute, self.rate_min)
-                export_rate += self.rate_export.get(minute, 0)
+                import_rate += self.rate_import.get(slot_start, self.rate_min)
```
springfall2008, line 5326: iBoost review: this line (and the window-average fix) is already merged on main - f98554f6 'fix: iBoost smart window averaging uses wrong loop variable (GH#4817)', including its own regression test run_iboost_smart_average_test() in tests/test_iboost.py. The PR is based on an older main (merge-base a054f6d2), so this commit is now redundant and the branch currently shows as conflicting with main in the run_iboost_smart_test/run_iboost_smart_tests region (the PR also adds an overlapping iboost_window_pricing regression for the same fix). Only the used_slots dedup half remains applicable - suggest rebasing and dropping the duplicate fix/test, or rebasing it onto main's test to avoid double coverage.

apps/predbat/plan.py
```diff
@@ -5214,6 +5217,233 @@ def plan_iboost_smart(self):
...
+
+        # Optional fill pass: top the tank up to its capacity in any eligible interval whose
+        # import rate is at or below the fill threshold (default -99 disables this). After each
+        # fill the level trajectory is re-verified and later planned boosts the fill energy has
+        # made redundant are trimmed straight away, so the level stays at or below the capacity
+        # everywhere and the day cap seen by later fill slots only counts energy that survives
+        fill_threshold = self.iboost_fill_rate_threshold
+        if fill_threshold > -99.0:
+            for slot_n in range(num_intervals):
+                if (not eligible[slot_n]) or (import_rates[slot_n] > fill_threshold):
```
springfall2008: _Automated comment from the triage bot._

The fill pass reuses the demand-booking `eligible[]` mask, so `iboost_fill_rate_threshold` is silently ANDed with `iboost_rate_threshold`, `iboost_rate_threshold_export` and the gas comparison instead of being the independent knob the docs describe ("any slot with an import rate at or below this value ... fills the remaining tank headroom"). It only bites when one of the other thresholds is *tighter than* the fill threshold — e.g. a user who wants demand boosts only on negative prices sets `iboost_rate_threshold = -1` but follows the docs' example `iboost_fill_rate_threshold = 0`: every 0p slot then fails `eligible[]` and the tank is never prefilled, with no warning. Worth either giving the fill pass its own rate check (import ≤ fill_threshold plus the export/gas guards it wants) or documenting that the demand thresholds also gate filling.

apps/predbat/plan.py
```diff
@@ -5214,6 +5217,233 @@ def plan_iboost_smart(self):
...
+        plan = []
+        for index, start in enumerate(interval_starts):
+            kwh = boost[index]
+            if kwh <= 0.001:
+                continue
+            interval = self.plan_interval_minutes
+            slot_start = max(start, self.minutes_now)
+            available = start + interval - slot_start
+            length = int(min(round(((kwh / self.iboost_max_power) / 5) + 0.5, 0) * 5, available))
+            length = min(max(length, 5), available)
```
springfall2008: _Automated comment from the triage bot._

The floor here emits a full-power 5-minute slot for a residual that only just clears the booking epsilon: a `kwh` of 0.0011 (e.g. the day cap has 1.1 Wh left, or a capacity/day room clamps the last dose that small) passes the `kwh <= 0.001` filter above, `round(...)` gives 0, and `max(length, 5)` turns it into a 5-minute full-power slot — the diverter delivers `iboost_max_power * 5/60` kWh (0.25 kWh at a 3 kW element) against a published `kwh` of 0.001, so the plan cost under-reports and the tank is heated well beyond the trajectory the planner optimised. Consider dropping residuals below one 5-minute step of element energy (or publishing `kwh = power * length/60`) so the modelled and delivered energies agree.

apps/predbat/tests/test_iboost.py
springfall2008: _Automated comment from the triage bot._ (follow-up, verified separately)

Two more pieces of shared-state pollution in the same suite, same #5079 class:

- The suite-end teardown restores `iboost_max_power = 2400 / 60 / 1000` — which matches neither the fixture args default of 3000 W (`test_infra.py:409`) nor `simple_scenario`'s `export_limit / 60.0` (`test_infra.py:749`). Any later test that consumes `self.iboost_max_power` without setting it (e.g. `predict_iboost`, legacy `plan_iboost_smart`) sees 2400 W instead of 3000 W depending on run order.
- The per-test teardown sets `my_predbat.rate_gas = {}` unconditionally, and `iboost_tank_capacity`/`iboost_tank_reserve`/`iboost_gas_scale` are only restored by the *trailing* regression block, not per-test — so a mid-module abort or truncated run leaks the last test's tank model into later modules under `./run_shuffle`.

apps/predbat/tests/test_iboost.py
psweens: Rebased onto current main. The window-averaging half of the first commit is dropped
(redundant with f98554f) and only the `used_slots` dedup guard remains, as its own commit
with a dedup-focused test; the overlapping `iboost_window_pricing` regression is reshaped
into `iboost_duplicate_slots` so it no longer double-covers `run_iboost_smart_average_test`.

One consequence worth flagging: with the guard keyed on the slot start, the
`iboost_smart2/3` goldens in test_model.py change (65→120 kWh, 60→110 kWh). Replaying the
pre-fix planner under those scenarios shows why: overlapping windows double-booked slot
starts, so the *plan* was fully booked (120/110 kWh) but `in_iboost_slot()` only credits the
first slot covering a minute, and the duplicates' energy was dropped from the prediction —
65/60 kWh achieved. With the guard fixed the plan is duplicate-free and the full booking is
delivered: `iboost_smart2` now matches `iboost_smart1`'s 120 kWh (same cap, no windowing) at
the true 1.5× window-average rate, which is the self-consistency check for the new goldens.
Updated in its own commit with the mechanism in the scenario comment.

The branch has since also been synced with current main by merging it in (clean, no
conflicts), so no further force-push is needed.

apps/predbat/plan.py
psweens: Fixed by clipping booking to the interval grid: the sub-slot loop now iterates grid-aligned
starts with the first sub-slot clamped to minutes_now, and the duplicate guard keys on the
grid cell, so overlapping windows on different phases can no longer emit overlapping slots.
Behaviour with an aligned minutes_now is unchanged (the existing goldens prove it). Added
the test you suggested — minutes_now=735, min_length=60, overlapping cheap windows — and
`check_slot_invariants` now also asserts positive length and non-overlap for every emitted
plan.

apps/predbat/fetch.py
psweens: Fixed: the shared ingestion helper now returns each source's raw extent only when that
source actually yielded data from minute_data, and the staleness gate aggregates over those
sources alone. A recent-but-unusable sensor can no longer vouch for a stale one; test
`iboost_fetch_stale_mixed` covers exactly that pairing.

apps/predbat/fetch.py
psweens: Fixed: after conversion, a forecast whose total future demand is zero now falls back to the
legacy plan with a warning ("contains no future demand") instead of silently booking
nothing. This also covers the fresh-but-flat producer case. Test `iboost_fetch_flat`.

apps/predbat/fetch.py
psweens: Fixed as suggested: a configured but unreadable iboost_tank_soc sensor now falls back to
the legacy plan for that cycle (same posture as a stale forecast) instead of assuming an
empty tank. The assume-empty path remains only for the documented case where no SoC sensor
is configured at all.

apps/predbat/plan.py
psweens: Taken the documentation option: both doc pages now state that
`iboost_smart_min_length` does not apply to forecast-driven slots — each slot's length
follows the energy the draw needs at the element power (minimum 5 minutes), with
adjacent-interval consolidation on price ties. I did not implement a hard minimum run
length because padding a run to min_length would book energy the tank model says it does
not need; if you'd rather have min_length enforced as a booking constraint (only book runs
that can reach min_length), I can do that as a follow-up.

apps/predbat/tests/test_iboost.py
psweens: Fixed: both iBoost suites snapshot the rate dicts and the derived scalars (rate_min/max,
rate_average, rate_min/max_forward, rate_min/max_base, rate_export_min/max/avg,
rate_export_max_forward) before running and restore them in a `finally` block, so any exit
path restores the fixture. The iBoost thresholds are restored to 9999 (the reset()
default), not 100.

Fixed: `iboost_max_power` (and max_energy, min_length, gas scale, tank capacity/reserve,
fill threshold) are restored from the fixture's own `get_arg` values rather than hard-coded
wattages; the gas test saves and restores the exact prior `rate_gas`/`iboost_gas` state
instead of clearing it; and because the restore now runs in the suite-level `finally`, a
mid-module abort restores everything too.

apps/predbat/plan.py
psweens: Added the warning: when the forecast holds demand but none of it lands on the plan grid,
the planner now logs it rather than planning nothing silently. On the mechanism though:
both grids floor to the same plan_interval_minutes stride, so as minutes_now advances the
plan grid shifts by whole intervals and per-key lookups stay aligned — and compare doesn't
advance minutes_now mid-run. The reachable failure was the horizon extension in compare,
which is fixed separately (see the compare-horizon thread): the demand grid now covers the
whole fetchable horizon so the raised forecast_minutes finds its demand.

apps/predbat/plan.py
psweens: Fixed: a shared `iboost_rate_okay()` now carries the import/export threshold and both gas
comparisons, used by the legacy planner, the forecast planner and the fill pass (the fill
passes its own import threshold into it). Legacy output is unchanged — the goldens prove
it.

apps/predbat/plan.py
psweens: Partially unified: both planners now share `iboost_plan_total_days()` and the forecast
ledger applies the same dp3 accumulation as the legacy ledger, which removes both drifts
named here (total_days expression and rounding). The legacy planner keeps its incremental
ledger inside its own booking loop: recomputing it from the plan list would change the dp3
accumulation order and risk perturbing golden output for no behaviour gain. If you want one
literal helper for both, I'd rather do that in a follow-up with its own golden review.

apps/predbat/plan.py
psweens: Fixed: one `quantise_slot_minutes()` helper now serves the legacy iBoost planner, the
forecast slot emission and `plan_car_charging` (identical arithmetic — legacy and car
goldens unchanged).

apps/predbat/fetch.py
psweens: Fixed: `fetch_cumulative_forecasts()` now carries the shared block (entity unwrap,
$attribute split, state fetch with try/except, dict→list conversion and validation, the
minute_data call, per-source warn-and-skip) for both load_forecast and iboost_forecast, and
the per-minute delta is read through `get_from_incrementing(..., backwards=False)` as you
point out it should be. Two deliberate side effects on the load path: a non-list state is
now skipped with a warning rather than raising, and the fetch-exception log is unified on
"Warn:" (both paths degrade gracefully — load just skips the source). Say the word if you'd
rather keep "Error:" there.

apps/predbat/plan.py
psweens: Reduced: each dose now does a single fused pass bounded at the target (trajectory and the
running max together) instead of four full-horizon passes, the day ledger is maintained
incrementally per dose, and the fill pass only recomputes state when a fill actually lands.
It is not yet fully incremental across doses — the clamping makes a correct suffix-only
update fiddly, and at the default 30-minute interval the loop is negligible. If 5-minute
interval profiling shows it matters I'll take the incremental-update step as a follow-up.

apps/predbat/plan.py
psweens: Fixed: the current interval's gating and pricing now read the rate at
max(start, minutes_now), matching the slot it emits.

apps/predbat/plan.py
psweens: Fixed, and more generally than the minimum: boost within an interval now precedes its draw
in the tank model (the element runs while the water is drawn), so the candidate scan
includes the target interval itself — a draw inside the current interval books the current
interval's remainder, and a draw later in any interval can be served from that same
interval. Tests cover the aligned and mid-interval current-draw cases. One knock-on:
refilling after a draw resumes the following interval (the draw interval is still full when
its own fill is considered), which shifted the fill-trim test's expectations by one
interval.

apps/predbat/fetch.py
psweens: Right — the comment was wrong about the back-fill value, and the reset scenario did produce
phantom demand. Fixed by taking per-minute deltas per source and only within that source's
own raw extent: the tail back-fill and the first-sample step are both excluded, and since
deltas are now per source, one source's tail can no longer swallow another's draw in the
summed series. Tests: `iboost_fetch_reset` (internal reset books no phantom) and the
multi-sensor case.

docs/customisation.md
psweens: Fixed: the section now states the planner takes over whenever the smart-planner path runs —
including iboost_smart On alongside the solar/battery modes — and that only iboost_smart
Off in those modes leaves a configured forecast without effect.

apps/predbat/fetch.py
psweens: Fixed in two parts: the demand grid is now built over the whole fetchable horizon
(forecast_days + 1 days) rather than the live forecast_minutes, so compare's raised horizon
finds the demand and tariffs are scored against the same series; and the fill pass is
clamped to the raw data extent (recorded at fetch time), so it cannot book tank fills
across a tail the fetched forecast never covered.

apps/predbat/fetch.py
psweens: All three are covered by the per-source extent-bounded deltas described in the tail-fill
thread: (a) deltas at or before the first sample are skipped, (b) deltas at or beyond the
last sample are skipped, (c) deltas are summed per source, not taken on the summed series.

apps/predbat/plan.py
psweens: Same fix as the inline thread: gating and pricing read the rate at max(minute, minutes_now)
for the current interval.

apps/predbat/plan.py
psweens: Same fix as the inline thread: grid-clipped booking with the guard keyed on the grid cell,
plus the minutes_now=735 / min_length=60 test and a non-overlap assertion in the shared
invariant check.

apps/predbat/fetch.py
psweens: Both parts fixed: extents are tracked per source that actually loaded, and a forecast whose
future increments sum to zero falls back to the legacy plan with a warning.

apps/predbat/tests/test_iboost.py
psweens: Covered by the suite-level snapshot/restore in `finally` (rates, derived scalars, iBoost
config including tank capacity/reserve/gas scale) described on the set_rate_profile thread.

apps/predbat/plan.py
psweens: See the inline perf thread: fused single pass per dose bounded at the target, incremental
day ledger, lazy fill-pass recompute; full incrementality deferred with reasoning.

psweens: See the candidate-loop thread: the target interval is now bookable (boost precedes draw
within an interval), so in-progress draws are planned rather than warned about.

apps/predbat/plan.py
psweens: Rebased and dropped, per the top-level thread; only the dedup half remains, with the golden
update explained there.

apps/predbat/plan.py
psweens: Fixed: the fill pass now applies its own import-rate rule (at or below the fill threshold)
with the export and gas guards retained, rather than reusing the demand eligibility mask.
Your exact scenario is the new `iboost_plan_fill_independent` test: demand threshold −1p
(nothing bookable for draws) with fill threshold 0 still prefills on the −2p slot. The docs
now state which guards apply to fill slots.

apps/predbat/plan.py
psweens: Fixed by the first of your two options: residuals below one 5-minute step of element energy
are dropped at emission rather than floored up to a full-power slot, so delivered and
published energy agree to within the 5-minute quantisation that also bounds the legacy
planner's slots. (The modelled tank runs very slightly colder than the trajectory in that
case, which is the conservative direction.)

PR #4759, opened 2026-08-26
fix(inverter): export window must tolerate an empty source like the charge window

mgazza:
## Problem

The charge window has four branches: REST data, a configured `charge_start_time`, **a configured-but-empty source** (`rest_api` or `ge_cloud_direct`) which sets safe defaults and retries, and finally a raise for a genuine apps.yaml gap.

The export window a hundred lines below has only three — the configured-but-empty branch is missing:

```python
else:
    self.log("Error: Inverter {} unable to read Export window as neither REST or discharge_start_time are set"...)
    self.base.record_status(...)
    raise ValueError          # bare, no message
```

Both windows are populated from the same fetch, so the charge-side tolerance cannot actually take effect. When GE Cloud returns nothing, the charge window recovers exactly as intended and then the export window kills the update loop on the very same cycle.

Seen live on a GivEnergy instance, these two lines back to back every cycle for two weeks:

```
Warn: Inverter 0 unable to read charge window time - GE Cloud returned no data,
      check the GE Cloud credentials and that the account still has this inverter attached,
      will retry next update
Error: Inverter 0 unable to read Export window as neither REST or discharge_start_time are set
```

The bare `raise ValueError` carried no message, so `record_status` wrote `"Error: Exception raised "` with nothing after it. The status entity is the main thing support reads, and it said nothing about what had failed — the underlying cause (expired GE Cloud credentials) was invisible from there.

## Fix

Mirror the charge window:

- Add the `elif self.rest_api or ge_cloud_direct:` branch so a configured-but-empty source falls through to the safe-defaults/retry path that already exists just below.
- Name the source that came back empty in the retry warning, as the charge window does — "discharge_start is None" sends users to apps.yaml, which is the one thing that is fine here.
- Give the remaining genuine-setup-gap raise a message, so the status entity says something useful.

The permanent-setup-gap behaviour is unchanged: no source configured at all still raises, it just says why now.

## Tests

Adds `test_export_window_ge_cloud_configured_but_no_data_yet`, mirroring the existing `test_charge_window_ge_cloud_configured_but_no_data_yet`. It drops both windows' args (which is what a cloud-backed instance actually looks like, since `ge_cloud_direct` configures neither) and asserts `update_status` does not raise and sets the safe defaults.

On `main` it fails with `ValueError()` — literally empty, which is the reported symptom exactly. It passes with the fix.

`inverter`, `inverter_multi`, `multi_inverter_status`, `ge_cloud`, `balance_inverters` and `inverter_config_sensor` all pass.

### DIFF
--- a/apps/predbat/inverter.py
+++ b/apps/predbat/inverter.py
@@ -1629,14 +1629,28 @@ class Inverter:
         elif "discharge_start_time" in self.base.args:
             discharge_start = time_string_to_stamp(self.base.get_arg("discharge_start_time", index=self.id))
             discharge_end = time_string_to_stamp(self.base.get_arg("discharge_end_time", index=self.id))
+        elif self.rest_api or self.base.get_arg("ge_cloud_direct", False, indirect=False):
+            # Same reasoning as the charge window above, and it has to be here too or that fix is
+            # defeated: on a cloud fetch failure the charge window degrades gracefully and then
+            # this branch crashes the whole update loop on the very same cycle, so the inverter
+            # never recovers either way. A configured-but-empty source is transient, so fall
+            # through to the safe-defaults/retry handling below.
+            discharge_start = None
+            discharge_end = None
         else:
-            self.log("Error: Inverter {} unable to read Export window as neither REST or discharge_start_time are set".format(self.id))
-            self.base.record_status("Error: Inverter {} unable to read Export window as neither REST or discharge_start_time are set".format(self.id), had_errors=True)
-            raise ValueError
+            # No data source configured at all - a permanent setup gap, handled as such.
+            message = "Error: Inverter {} unable to read Export window - no source is configured (set givtcp_rest, ge_cloud_direct, or discharge_start_time in apps.yaml)".format(self.id)
+            self.log(message)
+            self.base.record_status(message, had_errors=True)
+            raise ValueError(message)
 
         if discharge_start is None or discharge_end is None:
-            self.log("Warn: Inverter {} unable to read Export window as discharge_start or discharge_end is None, will retry next update".format(self.id))
-            self.base.record_status("Warn: Inverter {} unable to read Export window, will retry next update".format(self.id), had_errors=True)
+            # Name the source that came back empty, as the charge window does - "discharge_start is
+            # None" sends users to apps.yaml, which is the one thing that is fine here.
+            source = "GE Cloud" if (not self.rest_api and self.base.get_arg("ge_cloud_direct", False, indirect=False)) else "REST"
+            hint = "check the {} credentials and that the account still has this inverter attached".format(source)
+            self.log("Warn: Inverter {} unable to read Export window - {} returned no data, {}, will retry next update".format(self.id, source, hint))
+            self.base.record_status("Warn: Inverter {} unable to read Export window - {} returned no data, {}".format(self.id, source, hint), had_errors=True)
             # Set safe defaults to allow graceful recovery on next update
             self.discharge_enable_time = False
             self.discharge_start_time_minutes = 0
--- a/apps/predbat/inverter.py
+++ b/apps/predbat/inverter.py
@@ -1623,13 +1623,20 @@ class Inverter:
         # Construct discharge window from GivTCP settings
         self.export_window = []
 
+        # Record which source we read from, rather than re-deriving it below. Three branches
+        # reach the empty-value handling and only one of them is REST, so a ternary over
+        # rest_api/ge_cloud_direct mislabels the configured-entity case and sends the user to
+        # check credentials for a source they are not using.
         if self.rest_data:
+            export_source = "REST"
             discharge_start = time_string_to_stamp(self.rest_data["Timeslots"]["Discharge_start_time_slot_1"])
             discharge_end = time_string_to_stamp(self.rest_data["Timeslots"]["Discharge_end_time_slot_1"])
         elif "discharge_start_time" in self.base.args:
+            export_source = "discharge_start_time"
             discharge_start = time_string_to_stamp(self.base.get_arg("discharge_start_time", index=self.id))
             discharge_end = time_string_to_stamp(self.base.get_arg("discharge_end_time", index=self.id))
         elif self.rest_api or self.base.get_arg("ge_cloud_direct", False, indirect=False):
+            export_source = "REST" if self.rest_api else "GE Cloud"
             # Same reasoning as the charge window above, and it has to be here too or that fix is
             # defeated: on a cloud fetch failure the charge window degrades gracefully and then
             # this branch crashes the whole update loop on the very same cycle, so the inverter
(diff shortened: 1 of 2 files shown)


apps/predbat/inverter.py
```diff
@@ -1629,14 +1629,28 @@ def update_status(self, minutes_now, quiet=False):
...
-            self.base.record_status("Warn: Inverter {} unable to read Export window, will retry next update".format(self.id), had_errors=True)
+            # Name the source that came back empty, as the charge window does - "discharge_start is
+            # None" sends users to apps.yaml, which is the one thing that is fine here.
+            source = "GE Cloud" if (not self.rest_api and self.base.get_arg("ge_cloud_direct", False, indirect=False)) else "REST"
+            hint = "check the {} credentials and that the account still has this inverter attached".format(source)
+            self.log("Warn: Inverter {} unable to read Export window - {} returned no data, {}, will retry next update".format(self.id, source, hint))
+            self.base.record_status("Warn: Inverter {} unable to read Export window - {} returned no data, {}".format(self.id, source, hint), had_errors=True)
             # Set safe defaults to allow graceful recovery on next update
             self.discharge_enable_time = False
             self.discharge_start_time_minutes = 0
```
springfall2008: **High — this safe default isn't inert like the charge-window one, and this PR makes the gap far more reachable.**

The charge-window equivalent (line 1538) uses `self.base.forecast_minutes` (~2 days out) specifically so `execute.py`'s "already started" guard never mistakes it for a real value:

```python
# execute.py:223 (charge) — always False, forecast_minutes > minutes_now always
if (inverter.charge_start_time_minutes <= self.minutes_now) and (self.minutes_now >= minutes_start):
# execute.py:448 (discharge) — 0 <= minutes_now is always True
if (inverter.discharge_start_time_minutes <= self.minutes_now) and (self.minutes_now >= minutes_start):
```

With `discharge_start_time_minutes == 0`, the discharge guard fires whenever we're already inside a planned export window, `minutes_start` gets overwritten to `0`, and (since the overlap-with-charge check right after it at execute.py:451 doesn't reset it when charging is disabled) the inverter ends up commanded to discharge from local midnight instead of the real window start (execute.py:480/495).

Separately, `self.export_window` stays `[]` while this default is active, so `plan.py:2325`'s `keep_export` anti-flap bonus (`self.export_window and self.isExporting`) is silently skipped for that replan — an in-progress export can lose its "don't flap" protection for a cycle.

This line predates this PR, but the new `elif self.rest_api or ge_cloud_direct:` branch above (line 1632) means it's now reached on every transient cloud/REST hiccup rather than only a rare malformed-entity edge case. Suggest mirroring the charge side: `self.base.forecast_minutes` for both `discharge_start_time_minutes` and `discharge_end_time_minutes` here.

apps/predbat/inverter.py
```diff
@@ -1629,14 +1629,28 @@ def update_status(self, minutes_now, quiet=False):
...
+            self.log(message)
+            self.base.record_status(message, had_errors=True)
+            raise ValueError(message)
 
         if discharge_start is None or discharge_end is None:
-            self.log("Warn: Inverter {} unable to read Export window as discharge_start or discharge_end is None, will retry next update".format(self.id))
-            self.base.record_status("Warn: Inverter {} unable to read Export window, will retry next update".format(self.id), had_errors=True)
+            # Name the source that came back empty, as the charge window does - "discharge_start is
+            # None" sends users to apps.yaml, which is the one thing that is fine here.
+            source = "GE Cloud" if (not self.rest_api and self.base.get_arg("ge_cloud_direct", False, indirect=False)) else "REST"
```
springfall2008: **Medium — mislabels the source as "REST" when neither REST nor GE Cloud is configured.**

This block is reached from three predecessor branches, not just the new `elif` at line 1632 — the pre-existing `elif "discharge_start_time" in self.base.args:` (line 1629) lands here too when that entity resolves to an illegal/unavailable value (the `test_discharge_window_none_illegal_time` scenario). In that case `self.rest_api` and `ge_cloud_direct` are both falsy, so the ternary defaults to `"REST"` — telling a user with neither configured to "check the REST credentials". This is the identical, pre-existing heuristic at line 1532 for the charge window, but this diff duplicates it into export rather than fixing it. Worth a callout since the whole point of this PR is giving users an accurate, actionable message.

apps/predbat/tests/test_inverter.py
```diff
@@ -1692,6 +1692,76 @@ def restore():
...
+
+    try:
+        inv.update_status(my_predbat.minutes_now)
+    except ValueError as e:
+        print(f"ERROR: {test_name} - update_status should not raise while a configured GE Cloud source just hasn't returned data yet, got ValueError({e})")
+        restore()
+        return True
+
+    # Same safe defaults the discharge_start-is-None path already sets
+    if inv.discharge_enable_time != False:
```
springfall2008, line 1751: **Low — doesn't verify the warning names the right source, and export is missing the other two sibling tests charge has.**

`test_charge_window_ge_cloud_configured_but_no_data_yet` asserts `"GE Cloud" in my_predbat.current_status`; this test only checks the three `discharge_*` attributes, so the source-naming issue above (and any future regression in it) would pass silently. Also, charge has three tests for this code path (`..._no_source_configured`, `..._rest_configured_but_no_data_yet`, `..._ge_cloud_direct_configured_but_no_data_yet`) but export only gets this one — the REST-triggered branch (line 1632's `self.rest_api` half) and the final `raise ValueError(message)` branch (line 1645) are untested for export.

apps/predbat/inverter.py
```diff
@@ -1629,14 +1629,28 @@ def update_status(self, minutes_now, quiet=False):
         elif "discharge_start_time" in self.base.args:
             discharge_start = time_string_to_stamp(self.base.get_arg("discharge_start_time", index=self.id))
             discharge_end = time_string_to_stamp(self.base.get_arg("discharge_end_time", index=self.id))
+        elif self.rest_api or self.base.get_arg("ge_cloud_direct", False, indirect=False):
```
springfall2008, line 1638: **Nit (non-blocking) — this is the third time this exact tolerate/raise + source-naming block has been hand-propagated to one call site.**

`47e9376b` added it for charge, `4fbacb20` fixed a `ge_cloud_direct` gap only at the charge site, and this PR does the same work a third time for export — with the whole ~25-line block (trigger condition, source ternary, hint string, log/record_status calls) duplicated verbatim between the two sites (lines ~1500-1541 vs ~1632-1657). A small shared helper (e.g. `_resolve_window_source(...)`) would stop this drifting a fourth time — there's already a third, differently-shaped instance of "tolerate a bad scheduled-window value" at the idle-window parser around line 1708. Not blocking for this narrowly-scoped bugfix, but worth a follow-up issue.

springfall2008 (commented):
High-effort automated review (multi-angle pass: correctness, reuse, simplification, efficiency, conventions, depth-of-fix).

**Summary**
- 🔴 High: the export window's pre-existing `0`/`0` safe default isn't inert the way charge's `forecast_minutes` default is. This PR's new tolerate-branch makes that gap fire on every transient cloud/REST hiccup instead of only rare cases, and while it's in effect it can backdate the discharge command written to the inverter and silently disable the export anti-flap protection in `plan.py`.
- 🟡 Medium: the new `source = "GE Cloud" if ... else "REST"` line mislabels the source as "REST" for the third path into this block, producing a misleading warning message.
- Low: the new test doesn't assert on the warning message content (unlike its charge sibling), and export is missing the REST-triggered and no-source-configured sibling tests that charge already has.
- Non-blocking: this is the third time this same tolerate/raise + source-naming block has been hand-propagated to one call site at a time (see 47e9376b, 4fbacb20) — worth a shared helper as a follow-up.

Details inline below.

mgazza:
Thanks — the High is correct and I've fixed it. Pushed in 72eeb8fb.

**🔴 High — export safe default isn't inert.** Confirmed, and it's worse than a latent gap: `execute.py:448` reads `discharge_start_time_minutes <= minutes_now` as "the window has begun", which is trivially true at `0`, so the discharge command gets backdated to midnight. Charge parks at `forecast_minutes` and is genuinely inert; export used `0/0`, which is in the past.

You're right that this PR is what makes it matter. The transient branch turns a rarely-reached block into one hit on any cloud hiccup — so as first written it traded a crash for a backdated write, on precisely the instances it was meant to help. Export now parks at `forecast_minutes` too.

Three pre-existing tests asserted the `0/0` default. They encoded the behaviour being fixed, so they now assert the inert value — flagging that explicitly since it's a deliberate behaviour change beyond the original diff, and it also fixes the rare path.

**🟡 Medium — source mislabelled as REST.** Fixed. Rather than re-deriving from `rest_api`/`ge_cloud_direct`, the branch that actually supplied the value now records it (`REST` / `discharge_start_time` / `GE Cloud`). The configured-entity case gets its own hint pointing at the entities instead of at credentials for a source the user isn't using.

Note the same re-derivation exists on the charge side on main — I've left it alone rather than widen this PR, but it has the same mislabelling.

**Low — tests.** The warning message is now asserted to name GE Cloud, matching the charge sibling, and the no-source-configured test is added (it also pins that the raise carries a message, since the bare `raise ValueError` it replaces produced "Error: Exception raised " with nothing after it). I've not added the REST-triggered sibling — say the word if you want it and I'll do that too.

**Non-blocking — third hand-propagation.** Agreed, and the fact that the source-naming was propagated *with* its bug rather makes the case. Happy to do the shared helper as a follow-up rather than expanding this PR.

`inverter`, `inverter_multi`, `multi_inverter_status`, `ge_cloud`, `balance_inverters` and `inverter_config_sensor` all pass. `--quick` is green apart from `data_age_metrics`, which fails identically on clean `origin/main` — unrelated to this change.

PR #4775, opened 2026-08-27
feat(mcp): add get_log and get_state tools, redact credentials from get_apps

### LINKED ISSUE #4768: Give access to AI debug and config review
Having just logged two bugs, which were genuine but relatively minor bugs #4764 & #4766, both of which were minor config issues. One causing a problem with recalculating and the other just simply clicking on an icon without ever realising it, I can see that as the config gets more complex and the AI gets better, then  it is better at finding my stupidity than I am.

Ideally a button on the web UI to do this but if not then a place where I provide the predbat.log for 24 hours plus the debug.yamp and any specific context, giving access to the AI engine checking the config that is used when logging bugs without the need to log the bug. Ideally if its a real bug, then it can point me to the bug report ready written for me

Clearly there is a cost for the AI and I would be happy to pay for this on a pay as you go basis where I think that ny config is bust  but dont think its a bug. which is where I was with one of the cases mentioned. 

logging a bug to get the analysis but that seems unfair in that it costs predbat for theuse of the AI. 

**Additional context**
I am a reasonable sysadmin but not a SW engineer. I understand what predbat does but not how it does it.  I can clearly see when its not doing the right thing. often I can see where or when it broke, but not how it broke.   Generally the logs make some sense to me but they are far from obvious in the necessary detail to identify a problem.   Having had to log bugs twice and both times had the AI find the issue and a workaround solution in minutes, after me looking for much longer, I feel that incorporating this into the UI would help more than me to save much time. 

I dont have any AI accounts so I am prepared to pay as needed for tokens to run this.

springfall2008:
This is an automated draft PR generated from issue #4768 — a maintainer should review it before merging.

Fixes #4768

## Summary

The MCP server already covered part of #4768 via `get_apps`/`get_config`, but had no way to reach `predbat.log` or Predbat's internal state — the artefacts a bug report normally has to carry. This adds both, and closes the credential leak that made pointing a cloud AI at `get_apps` unwise.

### `get_log`

Serves `predbat.log` (plus the rotated `predbat.1.log`), with optional `filter` (`all`/`info`/`warnings`/`errors`, default `warnings`), `search`, `hours` and `max_lines`. Lines come back oldest-first with severity tagged; `max_lines` defaults to 500 and is capped at 5000 so a request can't pull a 10MB log through the protocol. Continuation lines with no timestamp of their own (tracebacks) inherit the timestamp of the entry they belong to, so `hours` keeps or drops a multi-line entry as a whole.

### `get_state`

Deliberately **not** a debug-yaml download. Sizing a real dump ([`coverage/cases/predbat_debug_agile1.yaml`](coverage/cases/predbat_debug_agile1.yaml), 313 top-level keys, 5.3MB) shows 272 of those keys are under 1KB and total under 10KB between them, while 11 keys are 85% of the file and every one is a per-minute series. So the split is:

- Called with **no arguments**, `get_state` returns every variable within the per-variable budget — 371 keys / ~2.4KB (~590 tokens) on a fresh instance — and describes the rest.
- Large values go into an `omitted` section carrying type, length, sample keys and min/max/mean, so the caller can ask for one by name rather than guessing. A size guard, not a size limit.
- `keys`, `filter` (regex) and `max_bytes` narrow or widen the request. Collections over 200 entries are refused on entry count alone, so a 2,880-entry per-minute dict is never serialised just to discover it doesn't fit.
- Values `json.dumps` can't encode (datetimes, inverter objects) are coerced rather than failing the call.

### Credential handling

- **`get_apps` now redacts by default**, matching what the web UI's own apps.yaml download already did. Pass `masked: false` to opt out. Previously `_execute_get_apps` returned `self.base.args` verbatim.
- **`mask_secret_args` widened** to match `secret` and `token` alongside `_key`/`password`. `sigenergy_app_secret`, `solis_api_secret`, `deye_app_secret`, `alphaess_app_secret`, `solis_access_token`, `gateway_mqtt_token` and `mcp_secret` were all being served in the clear. `*_expires_at` is exempt — a token expiry is exactly what you want visible when debugging a dead cloud integration.
- **The debug yaml's exclusion list moved to `utils.is_debug_excluded_key`** and is now shared with `create_debug_yaml`, so `get_state` can never return anything a debug dump would not (`ha_interface`, `components`, `secrets`, the URL caches, `db*`, credentials). Verified against a live instance: the widened rule newly excludes **zero** attributes on a baseline config, so it costs debug dumps nothing and only bites on cloud-integration credentials.

### Shared filtering

Log level rules (`classify_log_line`/`log_line_included`) are now shared between `get_log` and `/api/log` so the two views of the same log can't drift. The web handler's behaviour is unchanged — its HTML escaping and search highlighting are pinned by a regression test.

### Docs

[`docs/components.md`](docs/components.md) lists every MCP tool in a table, documents `get_log`'s and `get_state`'s arguments and the redaction behaviour, and adds a short "ask an AI assistant to review your setup" section describing the workflow the issue asked for.

## Testing

- New [`apps/predbat/tests/test_web_mcp.py`](apps/predbat/tests/test_web_mcp.py) (registered as `web_mcp`), 10 groups: `mask_secret_args` key matching, `is_debug_excluded_key`, `read_predbat_log` rotation, the shared filter helpers and timestamp parsing, the `get_state` value helpers, `get_apps` redaction, `get_log`, `get_state`, `tools/list` registration, and a regression test that `/api/log` still filters, escapes and highlights as before.
- `get_state` was probed against a real `PredBat` instance, not just a fake. That surfaced a genuine bug the fake missed: `measure_state_value` used `None` as its "too large" sentinel, so the 14 state variables that are legitimately `None` were being reported as omitted. Fixed by returning an explicit `fits` flag, with a regression test.
- `tools/triage_test.sh web_mcp` — passed. `tools/triage_test.sh debug_cases` — passed (run explicitly since `create_debug_yaml`'s filter changed; it's marked slow and skipped by `--quick`).
- `cd coverage && ./run_pre_commit` — all hooks passed, and the `run_all --quick` suite it runs passed in full.
- `interrogate` fails at 89% on `main` today and is not a pre-commit hook; changed files each moved up or stayed level, and everything added here is documented.

## Notes

No debug-yaml download tool. An MCP server cannot write to the client's filesystem and everything it returns lands in the client's context, so a 5MB dump has no viable path through the protocol; `get_state` reaches the same data incrementally. The "button in the web UI" form of the request needs a billing/key-custody decision and is left for @springfall2008.

### DIFF
--- a/apps/predbat/tests/test_web_mcp.py
+++ b/apps/predbat/tests/test_web_mcp.py
@@ -1,0 +1,519 @@
+# -----------------------------------------------------------------------------
+# Predbat Home Battery System
+# Copyright Trefor Southwell 2026 - All Rights Reserved
+# This application maybe used for personal use only and not for commercial use
+# -----------------------------------------------------------------------------
+# fmt off
+# pylint: disable=consider-using-f-string
+# pylint: disable=line-too-long
+# pylint: disable=attribute-defined-outside-init
+
+"""Tests for the MCP server's log and apps.yaml tools (#4768).
+
+Covers the new get_log tool, get_apps' credential redaction, the widened
+mask_secret_args() key match, and the log-filter helpers now shared between the
+MCP tool and the web log view.
+"""
+
+import asyncio
+import json
+import os
+import tempfile
+from datetime import datetime, timedelta
+
+import web
+import web_mcp
+from web import WebInterface
+from web_mcp import MCPServerWrapper, LOG_FILTER_TYPES, MCP_LOG_DEFAULT_LINES, MCP_LOG_MAX_LINES, parse_bool_argument
+from utils import mask_secret_args, is_secret_key, read_predbat_log, classify_log_line, log_line_included, parse_log_timestamp
+
+
+class FakeRequest:
+    """A minimal aiohttp-request stand-in exposing only the query string a handler reads."""
+
+    def __init__(self, query=None):
+        """Store the query string a handler will read."""
+        self.query = query or {}
+
+
+def _stamp(minutes_ago):
+    """Return a predbat.log-style timestamp prefix for a line written N minutes ago."""
+    return "{}: ".format(datetime.now() - timedelta(minutes=minutes_ago))
+
+
+def _sample_log():
+    """Return a small synthetic predbat.log, newest line last, spanning three days."""
+    lines = [
+        _stamp(60 * 48) + "Info: Predbat starting up two days ago",
+        _stamp(60 * 48) + "Warn: An old warning from before the window",
+        _stamp(60 * 3) + "Info: Fetched Octopus rates",
+        _stamp(60 * 2) + "Warn: battery_rate_max_charge clamped to 3.0",
+        _stamp(60) + "Error: Failed to write charge window to inverter",
+        "Traceback (most recent call last):",  # a continuation line, written without its own stamp
+        _stamp(30) + "Fetched 48 slots of solar forecast",
+        _stamp(10) + "Warn: iboost target not reached",
+    ]
+    return "\n".join(lines) + "\n"
+
+
+def _make_mcp(my_predbat):
+    """Build an MCPServerWrapper bound to my_predbat without starting any server."""
+    return MCPServerWrapper(my_predbat, log_func=my_predbat.log)
+
+
+def _make_web(my_predbat):
+    """Build a minimal WebInterface bound to my_predbat, bypassing ComponentBase.__init__
+    (which would stand up the real aiohttp app) - same pattern as test_web_debug_history_routes.py.
+    """
+    w = WebInterface.__new__(WebInterface)
+    w.base = my_predbat
+    w.log = my_predbat.log
+    w.prefix = my_predbat.prefix
+    return w
+
+
+def _call_tool(mcp, name, arguments=None):
+    """Run one tool through the MCP tools/call path and return the decoded JSON result."""
+    result = asyncio.run(mcp._handle_tools_call({"name": name, "arguments": arguments or {}}))
+    return json.loads(result["content"][0]["text"]), result.get("isError", False)
+
+
+def test_mask_secret_args(my_predbat):
+    """Credential-like apps.yaml keys are redacted, including the *_secret and *_token names
+    the original "_key"/"password" match missed (#4768), while timestamps stay readable.
+    """
+    failed = False
+    print("**** Testing mask_secret_args key matching ****")
+
+    args = {
+        "ha_key": "ha-key-value",
+        "octopus_api_key": "octopus-key-value",
+        "deye_password": "deye-password-value",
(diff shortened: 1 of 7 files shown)


springfall2008:
Thanks — five of the six findings were good and are fixed in 15d4f660. One I've not actioned, with reasoning below.

**Fixed**

- **`max_lines` ordering description.** Correct catch, and my wording was the bug: the tool keeps the most recent matching lines but returns them oldest-first. Fixed in the tool schema, the tool description and `docs/components.md`.
- **Argument validation.** Added `parse_number_argument` and `compile_filter_argument`, which raise a named `MCPArgumentError` reported as `{"success": false, "error": "'max_lines' must be a whole number, got 'lots'"}` rather than a raw exception string. This matters more than usual here because the consumer is an AI assistant that can only correct itself if told which argument was wrong. Applied to `get_state` and `get_log` as flagged, and also to `get_apps`/`get_config`/`get_entities`, which have carried the same unvalidated `re.search` since before this branch — leaving three tools with a worse error path than the two new ones would be a trap. Compiling the pattern once also drops ~370 `re.search` calls per `get_state`.
- **Log encoding** (suppressed comment). The best of the six. One non-UTF-8 byte in `predbat.log` would raise `UnicodeDecodeError` and take out both `/api/log` and `get_log`. Now decoded as UTF-8 with `errors="replace"`, with a test that feeds it a latin-1 degree sign. Worth noting this is pre-existing — `web.py` has always opened the log without an explicit encoding; centralising it in this PR is what makes it a one-line fix.

**Not actioned: the `is_secret_key` exemption breadth**

The claim is that `partner_token_expiry` (Axle) and `_token_birth` (Ohme) will now be excluded from debug dumps, losing useful context. Both attributes exist, but they are set on the **component objects** — `axle.py:74` and `ohme.py:786` — not on `PredBat`. `create_debug_yaml` and `get_state` walk `self.__dict__` plus `self.inverters`, and `components` is in `DEBUG_EXCLUDE_LIST`, so neither attribute has ever reached a debug dump, before or after this change.

I checked the generalisation rather than just the two examples: scanning every `PredBat` mixin for `self.*secret*` / `self.*token*` assignments turns up three hits, all on `OctopusAPI(ComponentBase)` rather than the `Octopus` mixin that `PredBat` inherits. No `PredBat` base attribute matches at all, which is why probing a live instance showed the widened rule newly excluding **zero** keys.

The underlying concern about substring matching is still fair as future-proofing, so `SECRET_KEY_EXEMPT_SUFFIXES` now covers `_expiry`, `_expires`, `_expiration` and `_birth` alongside `_expires_at` — which would handle exactly those two names if such an attribute ever moved onto the base.

**Verification:** `run_pre_commit` green (all hooks + `run_all --quick` in full), `tools/triage_test.sh debug_cases` green, and the `web_mcp` suite grew two groups covering the validation helpers across all five tools and the log-decoding fix.

PR #4806, opened 2026-08-28
fix(web): correct inverted grid arrow and icon on Power Flow diagram

### LINKED ISSUE #4797: Power Flow diagram: grid arrow points the wrong way (import/export inverted)
**Describe the bug**

On the Power Flow diagram, the grid arrow points *grid → house* while the system is exporting, and *house → grid* while importing. The `get_grid_power_icon` indicator is inverted the same way.

The battery arm of the same diagram was corrected in ef2f306 (24 Aug), but the grid arm was left on the old sign convention.

`web.py:616`:

```python
grid_importing = grid_power <= -10  # Grid is importing power (negative value)
```

This contradicts the convention documented in `predbat_metrics.py:101`:

```python
self.grid_power = _gauge("predbat_grid_power", "Current grid power in kW (positive for import, negative for export)")
```

**Expected behaviour**

Arrow points house → grid when exporting, grid → house when importing.

**Steps to reproduce**

Export to the grid (PV exceeding house load with the battery full) and open the Power Flow page.

Live reproduction on my system, mid-export:

| Reading | Value |
|---|---|
| `pv_power` | 1588 W |
| `load_power` | 279 W |
| `battery_power` | -6 W (idle) |
| `grid_power` (after `grid_power_invert`) | **-1247 W** |

Energy balance confirms export: 1588 = 279 (house) + 6 (battery) + 1247 (export), within losses. The diagram drew the grid arrow **into** the house, labelled -1131 W.

The opposite case from the same log, charging from grid at night with no PV - `battery_power -1244W`, `load power 143W`, `PV Power 0W` - reads `grid power 1469W`, i.e. **positive while importing**, matching `predbat_metrics.py` and contradicting the comment on line 616.

**Suggested fix**

```python
# web.py:616
grid_importing = grid_power >= 10
```

Second instance, `get_grid_power_icon` at `web.py:4253-4256` - the two icon names are swapped:

```python
if power >= 10:
    icon_text = "transmission-tower-export"   # should be -import
elif power <= -10:
    icon_text = "transmission-tower-import"   # should be -export
```

(The `transmission-tower-export` use at `web.py:4336` reads `binary_sensor.<prefix>_exporting` rather than `grid_power`, and is not affected.)

**Origin**

Commit ef2f306, "feat(web): draw the power flow nodes as icons in their own colours" (2026-08-24), fixed the battery sign in this function:

```diff
- battery_discharging = battery_power <= -10  # Battery is discharging (negative value)
+ battery_to_house    = battery_power >= 10   # Battery is discharging into the house
```

but left `grid_importing` unchanged, and added a comment stating the opposite convention:

```diff
+ # (gateway.py negates the firmware's sign for exactly this reason) and grid_power is
+ # negative when importing, so the reading and the arrow run opposite ways round.
```

Still present on `main` at time of writing.

**Predbat version**

v8.53.5 (also reproduced by inspection on `main`)

**Environment details**

- Inverter: FoxESS (foxess_modbus), single inverter, 9.73 kWh
- `grid_power: sensor.grid_ct` with `grid_power_invert: True`
- HAOS, Predbat add-on

**Screenshots**
Image

**Log file / predbat_debug.yaml**

Not attached, as I don't think they add anything here - happy to provide either if you disagree.

This is a pure display bug with no plan involvement: the branch is decided by a single comparison on `grid_power` at `web.py:616`, and the raw values that drive it are quoted above for both the export and import cases. The log would only restate those same `Inverter 0 SoC: ... grid power NNNNW` lines, and `predbat_debug.yaml` captures the plan and full settings, none of which affect this code path.

**Notes**

Display-only - planning and control read `grid_power` directly and are unaffected.

Possibly related to #4788 (car arm of the same diagram).

springfall2008:
This is an automated draft PR generated from issue #4797 — a maintainer should review it before merging.

Fixes #4797

## Summary

`web.py`'s `render_power_flow` computed `grid_importing = grid_power <= -10`, but `grid_power` is positive while importing and negative while exporting (`predbat_metrics.py:101`) — the opposite convention, and the opposite of the sign the battery arm's equivalent fix (ef2f306) established for `battery_power`. This drew the grid arrow the wrong way round for every import/export state, and the comment introduced alongside that commit asserted the wrong convention. `get_grid_power_icon` had the same two icon names swapped (`transmission-tower-import`/`-export`).

Fixed both spots to `grid_power >= 10` / correct icon mapping, and corrected the misleading comment.

## Testing

- `cd coverage && ./run_pre_commit` — all hooks and the full unit test suite passed (`web_power_flow` included).
- `tools/triage_test.sh web_power_flow <log>` — passed after the fix.
- Verified the new `run_grid_power_sign_tests` regression test (added in `test_web_power_flow.py`, checks both arrow direction and icon against the documented sign convention) fails with clear errors against the pre-fix code, then re-ran clean after restoring the fix.
- Also corrected the existing geometry test's fixture, which paired `grid_power = -1000` with a comment labelled "Importing" — the same inverted convention as the bug, so it only ever checked arm geometry, never the sign mapping (as flagged in the triage comment).

## Notes

Root cause and both bug locations were identified in the issue's triage comment; this PR implements the one-line fix it suggested plus the swapped icon pair, and adds the sign-check regression test the triage comment recommended.

### DIFF
--- a/apps/predbat/tests/test_web_power_flow.py
+++ b/apps/predbat/tests/test_web_power_flow.py
@@ -272,6 +272,47 @@ def run_power_flow_colour_tests(my_predbat, web):
     return failed
 
 
+def run_grid_power_sign_tests(my_predbat, web):
+    """The grid arrow and icon follow the documented grid_power sign convention.
+
+    grid_power is positive while importing and negative while exporting (predbat_metrics.py:101),
+    the same way round as battery_power (positive while discharging). Getting this backwards drew
+    the grid arrow, and picked the grid icon, the wrong way round for every import/export state (#4797).
+    """
+    failed = 0
+    print("Test: the grid arrow and icon follow the documented grid_power sign convention")
+
+    saved = (my_predbat.grid_power, my_predbat.dashboard_index)
+    my_predbat.dashboard_index = ["dummy"]
+
+    cases = [
+        (1500, "Grid", "transmission-tower-import", "importing"),
+        (-1500, "House", "transmission-tower-export", "exporting"),
+    ]
+    for grid_power, expected_tail, icon_name, description in cases:
+        my_predbat.grid_power = grid_power
+        html = web.get_power_flow_diagram()
+        grid_lines = [line for colour, line in arrow_lines(html) if colour == "#757575"]
+        if not grid_lines:
+            print(f"  ERROR: no grid arm drawn while {description} (grid_power={grid_power})")
+            failed += 1
+            continue
+        x1, y1, x2, y2 = grid_lines[0]
+        tail_at_grid = distance((x1, y1), NODE_BY_COLOUR["#757575"][1]) < distance((x1, y1), HOUSE_CENTRE)
+        actual_tail = "Grid" if tail_at_grid else "House"
+        if actual_tail != expected_tail:
+            print(f"  ERROR: while {description} (grid_power={grid_power}) the grid arm starts at {actual_tail}, expected {expected_tail}")
+            failed += 1
+
+        icon_html = web.get_grid_power_icon()
+        if f"mdi-{icon_name}" not in icon_html:
+            print(f"  ERROR: while {description} (grid_power={grid_power}) expected the {icon_name} icon, got: {icon_html}")
+            failed += 1
+
+    (my_predbat.grid_power, my_predbat.dashboard_index) = saved
+    return failed
+
+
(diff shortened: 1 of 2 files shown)


springfall2008:
Sorry I don't agree this is a bug

my car and battery are charging and its fine, I think your grid maybe the wrong way around?

image

Issue #4835, opened 2026-08-29
Weekend Happy Hours with code: null are ignored by octopus_free_session


73ttv627re77:
## Describe the bug

Booked Octopus **Weekend Happy Hours** are published correctly by HomeAssistant-OctopusEnergy v19.0.1, but Predbat silently ignores them when the event payload contains `code: null`.

This is a focused follow-up to #4548. The Power Up/Power Down migration and entity naming are not the failure here: the configured legacy free-electricity event entity exists and receives both booked sessions.

## Environment

- Predbat: v8.53.5
- Home Assistant Core: 2026.8.3
- HomeAssistant-OctopusEnergy: v19.0.1
- `octopus_free_session` points to the documented legacy event entity:

```yaml
octopus_free_session: 're:(event.octopus_energy_([0-9a-z_]+|)_octoplus_free_electricity_session_events)'
```

## Reproduction

1. Earn and book two Weekend Happy Hours in the Octopus Saving Sessions dashboard.
2. Wait for the Octopus HA integration coordinator to refresh.
3. Confirm that `event.octopus_energy_<account>_octoplus_free_electricity_session_events` contains the booked sessions.
4. Allow Predbat to complete multiple planning cycles.
5. Inspect Predbat's free-event logs and calculated import rates/plan for those periods.

The anonymised HA event payload was:

```yaml
events:
  - id: 5798
    code: null
    start: "2026-08-30T12:00:00+01:00"
    end: "2026-08-30T13:00:00+01:00"
    duration_in_minutes: 60
  - id: 5800
    code: null
    start: "2026-08-30T14:00:00+01:00"
    end: "2026-08-30T15:00:00+01:00"
    duration_in_minutes: 60
```

## Observed behaviour

- HA published both booked sessions and continued refreshing the event entity.
- Four consecutive completed Predbat cycles did not log either free event.
- Both periods remained priced at the normal 30.46p/kWh import rate in Predbat.
- Neither became a charge window; the plan instead allowed an export window to span the Happy Hour periods.

The relevant parser currently gates planning on all three values being truthy:

```python
if start and end and code:
```

Because Weekend Happy Hours arrive with `code: null`, both valid sessions are discarded without a warning.

## Expected behaviour

Predbat should accept a free/Power Up event for planning when it has valid `start` and `end` values, even if `code` is absent or null. The event ID could be used as a diagnostic identifier, but a join code is not required for an already-booked/auto-joined Happy Hour.

[...]

Issue #4531, opened 2026-08-15
Octopus saving sessions notifications repeated


raldred:
**Describe the bug**
Since recently updating I've started recieving savings sessions notifications multiple times for each even and they are repeating each time the plan calculation runs, every 5 minutes.

**Expected behaviour**
I should only receive one notification per session being joined.

**Predbat version**
v8.48.1

**Environment details**

- Luxpower
- Standard HAOS installer

**Screenshots**

<img width="1079" height="1283" alt="Image" src="https://github.com/user-attachments/assets/48ecd786-3ab0-4f47-b439-ad4edd11bfa7" />

**Log file**
Can you capture a log file from the time of the issue, debug mode is not normally required.

[predbat.log](https://github.com/user-attachments/files/31101763/predbat.log)

**Predbat debug yaml file**
This is important for any plan related issues.

[predbat_debug.bin.txt](https://github.com/user-attachments/files/31101770/predbat_debug.bin.txt)

motocodeltd:
i see this too, and I think it's related to the free electricity hour octopus is giving between 11:00 and 12:00 on sunday 16/08 (at least for me, region j) due to eclipse. You can see that the octopus saving event is at 0.0 p/kWh. I also don't see that price reflected in the predbat plan yet.

kingjpc:
Also seeing this. Looks like it thinks we have not joined the session. Probably as it’s an auto session we have been given for the eclipse event. 

I actually get 3 notifications with different times in each but only when it restarts.

gcoan:
@raldred I think there are two problems, neither are predbat issues

first one, the repeated 'joined saving session' messages, in your log file I can see errors that Octopus data isn't present so my guess is that for whatever reason the integration has lost connection to your Octopus account and so when it tries to join the saving session, it doesn't, and so next time predbat runs it tries again.  I see this repeatedly for the 2p eclipse session

<img width="1058" height="270" alt="Image" src="https://github.com/user-attachments/assets/8554eb04-16fa-4a48-a776-7f324a107891" />

the second issue is that Octopus have setup 3 incorrect saving sessions for tomorrow, from 12-1, 1-2 and 2-3pm, each for 0 octopoints,.  When predate tries to join these, Octopus errors as these are not valid sessions

<img width="1053" height="381" alt="Image" src="https://github.com/user-attachments/assets/c13071c4-5c57-4e2f-b4fb-c231c0dd48f6" />

[...]

PR #4852, opened 2026-08-30
fix(octopus): treat a joined zero-reward session as free electricity

### LINKED ISSUE #4851: Octopus: joined Power Up (free electricity) sessions reported at 0 p/kWh are dropped from the plan
**Describe the bug**

A **joined** Power Up (free electricity) session that Octopus reports at `0` octopoints/kWh never reaches the plan. Predbat has the event — correct start, correct end, in `joined_events` — and then discards it, so the import rate for that period stays at the normal tariff rate and no charge window is planned.

The only way to get the free hour into the plan today is to type it in by hand as a manual import rate override, e.g.

```
select.predbat_manual_import_rates = "+Sun 11:00=0.0,Sun 11:30=0.0"
```

which reproduces exactly what the automatic path should have done.

This is the planning-side counterpart of #4548 point 5. That point covers the *auto-join* behaviour (Predbat trying to join events the integration rejects, fixed by the rate guard in #4593/#4595). The join side is now correct — but nothing was done on the side that turns those same events into rates, so the data is recognised well enough to be skipped for joining and then thrown away rather than used.

**To Reproduce**

Seen on Octopus Intelligent Go with Octopus Energy Direct, but the consumer is shared so the BottleCapDave path behaves the same way.

1. Octopus publishes the day's Power Up sessions into the Power Down / saving-session data set at `octopoints_per_kwh: 0` (the mislabelling described in #4548 point 5). One of them is already joined:

```
joined_events:    [{"id": ..., "code": "EVENT_nn_ddmmyy",
                    "start": "...T10:00:00+00:00", "end": "...T11:00:00+00:00",
                    "octopoints_per_kwh": 0, "rewarded_octopoints": null}]
available_events: [{... 11:00-12:00 ..., "octopoints_per_kwh": 0}, ...]
```

2. `sensor.<prefix>_free_electricity` is `off` — the free-electricity/Power Up feed returns nothing for the day (also #4548 point 5, "Nothing populated in the free electricity/power up event data"). Any events it does hold are historical, so `load_free_slot()` correctly skips them.

3. The plan for the joined window keeps the standard import rate and plans no charge.

**Expected behaviour**

A joined session reported at 0 p/kWh is a free import hour. It should land in `octopus_free_slots` and set the import rate to 0 for its period, so the planner can use it — the same result the manual override produces.

**Root cause**

`fetch_octopus_sessions()` in `apps/predbat/octopus.py`, the `joined_events` loop:

```python
octopoints_kwh = event.get("octopoints_per_kwh", None)
if octopoints_kwh is not None:
    saving_rate = octopoints_kwh / octopoints_per_penny
elif default_rate_pence > 0:
    saving_rate = default_rate_pence
# Skip events with no rate info unless default is configured
if start and end and (octopoints_kwh is not None or default_rate_pence > 0) and saving_rate > 0:
```

`0 / 8 == 0.0`, so `saving_rate > 0` is false and the event is dropped entirely. The guard is right that this is not a *saving* session — a zero reward means there is nothing to earn by exporting. But zero reward on a Power Up event means the opposite of "ignore this": importing during it is free.

The auto-join loop above it already treats `octopoints_kwh <= octopus_saving_session_min_octopoints_per_kwh` (default 0) as "do not attempt to join", with a comment explaining why. So both loops identify these events correctly and both then do nothing with them.

Note the two loops disagree in a way that matters for the fix: the join side skips at `<= threshold` (configurable), the planning side at `> 0` (fixed). A user who raises `octopus_saving_session_min_octopoints_per_kwh` to skip low-value Power Down sessions should not thereby lose free Power Up hours, so the planning side should key on `== 0` specifically, not on the threshold.

**Proposed fix**

In the `joined_events` loop, route an event whose reported reward is exactly `0` into `octopus_free_slots` at rate 0, rather than discarding it. Events with no reported rate (`None`) keep their current meaning — rate not reported, `octopus_saving_session_rate` applies if configured.

This is deliberately narrow: it does not attempt to distinguish Power Up from Power Down by event type (#4548 point 7), which needs the API change described there. It only stops a joined, zero-reward, known-duration session from being silently dropped.

**Predbat version**

8.53.5

**Environment**

Octopus Energy Direct, Octopus Intelligent Go.

mgazza:
Fixes #4851.

## Problem

Octopus puts Power Up (free electricity) events into the Power Down / saving-session data set at `0` p/kWh — the mislabelling described in #4548 point 5. The **auto-join** side already handles this: #4593/#4595 added a rate guard so Predbat stops trying to join events the integration will reject.

Nothing was done on the side that turns those same events into **rates**. The `joined_events` loop drops them:

```python
if start and end and (octopoints_kwh is not None or default_rate_pence > 0) and saving_rate > 0:
```

`0 / 8 == 0.0`, so a joined, zero-reward session fails the test and is discarded. The import rate for that period stays at the standard tariff rate and no charge window is planned.

So Predbat recognises the event well enough to avoid join-spamming it, then throws it away. Observed on Octopus Intelligent Go via Octopus Energy Direct: the only way to get the free hour into the plan was to type it in by hand —

```
select.predbat_manual_import_rates = "+Sun 11:00=0.0,Sun 11:30=0.0"
```

— which is exactly what the automatic path should have produced. The plan row confirmed the source, showing `import_rate_adjust_type: "manual"` against a published tariff rate of 28.56p.

The free-electricity feed does not cover for this either: it returned nothing for the day (also #4548 point 5, *"Nothing populated in the free electricity/power up event data"*), so `load_free_slot()` had no current events to apply.

## Change

Route a joined session reporting exactly `0` into `octopus_free_slots` at rate 0 instead of discarding it.

Zero reward on a Power **Down** event means there is nothing to earn by exporting. On a Power **Up** event it means the opposite of "ignore this" — importing is free.

Two deliberate choices:

- **Keyed on `== 0`, not on `octopus_saving_session_min_octopoints_per_kwh`.** That threshold exists so a user can decline low-value Power Down sessions; raising it must not also cost them free Power Up hours. The two loops are testing different things on purpose.
- **A null rate keeps its existing meaning** — rate not reported, `octopus_saving_session_rate` applies if configured — and is not treated as free.

This does not attempt to distinguish Power Up from Power Down by event type (#4548 point 7); that needs the upstream API change described there. It only stops a joined, zero-reward, known-duration session being silently discarded.

## Tests

New `test_saving_session_zero_octopoints_joined_is_free_slot` (registered in `unit_test.py`), covering:

- a joined 0 p/kWh session becomes a free slot, while a genuine 500-octopoint session alongside it still becomes a saving slot
- `load_free_slot()` zeroes exactly one contiguous hour and leaves every other minute untouched, with the replicate reason marked
- a null reward rate still produces no free slot

Confirmed to fail against the unfixed code (`got []`, `expecting 60 minutes at rate 0, got 0`).

`unit_test.py -k saving` and `-k octopus` both pass in full.

### DIFF
--- a/apps/predbat/octopus.py
+++ b/apps/predbat/octopus.py
@@ -315,6 +315,7 @@ octoplus_saving_session_query = """query {{
 			startAt
 			endAt
       devEvent
+      eventType
       targetRegion {{
         regionId
       }}
@@ -1186,6 +1187,7 @@ class OctopusAPI(ComponentBase):
         joined_ids = {}
         event_reward = {}
         event_code = {}
+        event_type = {}
 
         # Default saving session rate in octopoints/kWh
         # octopus_saving_session_rate is in p/kWh, convert to octopoints
@@ -1220,6 +1222,7 @@ class OctopusAPI(ComponentBase):
             if event_id:
                 event_reward[event_id] = reward
                 event_code[event_id] = code
+                event_type[event_id] = event.get("eventType", None)
             target_regions = [region.get("regionId") for region in (event.get("targetRegion", None) or []) if region]
             if target_regions and account_region_id not in target_regions:
                 self.log("OctopusAPI: Skipping saving event code {} - not eligible for account region {} (event targets regions {})".format(code, account_region_id, target_regions))
@@ -1227,7 +1230,7 @@ class OctopusAPI(ComponentBase):
             if start and end and event_id not in joined_ids:
                 endDataTime = parse_date_time(end)
                 if endDataTime > self.now_utc_exact:
-                    return_available_events.append({"start": start, "end": end, "octopoints_per_kwh": reward, "code": code, "id": event_id})
+                    return_available_events.append({"start": start, "end": end, "octopoints_per_kwh": reward, "code": code, "id": event_id, "event_type": event.get("eventType", None)})
 
         for event in joined_events:
             start = event.get("startAt", None)
@@ -1237,7 +1240,9 @@ class OctopusAPI(ComponentBase):
             if reward is None:
                 reward = default_octopoints  # Inject default when API doesn't provide reward
             if start and end:
-                return_joined_events.append({"start": start, "end": end, "octopoints_per_kwh": reward, "rewarded_octopoints": event.get("rewardGivenInOctoPoints", None), "id": event_id, "code": event_code.get(event_id, None)})
+                return_joined_events.append(
+                    {"start": start, "end": end, "octopoints_per_kwh": reward, "rewarded_octopoints": event.get("rewardGivenInOctoPoints", None), "id": event_id, "code": event_code.get(event_id, None), "event_type": event_type.get(event_id, None)}
+                )
 
         saving_attributes = {"friendly_name": "Octopus Intelligent Saving Sessions", "icon": "mdi:currency-usd", "joined_events": return_joined_events, "available_events": return_available_events}
 
@@ -3427,6 +3432,35 @@ class Octopus:
                         saving_rate = octopoints_kwh / octopoints_per_penny  # Octopoints per pence
                     elif default_rate_pence > 0:
                         saving_rate = default_rate_pence  # Use configured default rate
+
+                    # Octopus publishes Power Up and Power Down events through the same savingSessions
+                    # feed and distinguishes them with eventType (#4548 point 5/7):
+                    #   TURN_DOWN           - reduce consumption, rewarded in octopoints
+                    #   WEEKEND_HAPPY_HOUR  - an earned free import hour, reward 0
+                    #   TURN_UP             - increase consumption, rewarded in octopoints
+                    # Only WEEKEND_HAPPY_HOUR is free. TURN_UP is a rewarded increase event, not free
+                    # import, so it stays on the saving path - do not be tempted to fold it in here on
+                    # the grounds that both mean "import more".
+                    #
+                    # The reward value cannot stand in for the type. A TURN_DOWN reporting 0 - an
+                    # unknown reward published as 0 rather than null - would be zeroed on IMPORT,
+                    # making Predbat grid-charge through a session where the user is meant to be
+                    # reducing import. eventType is the only sound discriminator (#4851).
+                    #
+                    # An absent eventType (older API, or a feed that does not carry it) falls through
+                    # to the existing behaviour rather than guessing.
+                    if start and end and event.get("event_type", None) == "WEEKEND_HAPPY_HOUR":
+                        try:
+                            start_time = str2time(start)
+                            end_time = str2time(end)
+                        except (ValueError, TypeError):
+                            self.log("Warn: Bad start time for joined Octopus Weekend Happy Hour: {}-{}".format(start, end))
+                            continue
+                        if abs((start_time - self.now_utc).days) <= 3:
+                            self.log("Octopus: Joined Octopus Weekend Happy Hour {}-{} - treating as a free electricity session".format(start_time.strftime("%a %d/%m %H:%M"), end_time.strftime("%H:%M")))
+                            octopus_free_slots.append({"start": start, "end": end, "rate": 0})
+                        continue
+
                     # Skip events with no rate info unless default is configured
                     if start and end and (octopoints_kwh is not None or default_rate_pence > 0) and saving_rate > 0:
                         # Save the saving slot?
--- a/apps/predbat/tests/test_saving_session.py
+++ b/apps/predbat/tests/test_saving_session.py
@@ -1352,3 +1352,134 @@ def test_saving_session_entity_regex_power_rename(my_predbat):
                 print(f"PASS: {description} resolved to {resolved}")
 
     return failed
+
+
+def test_saving_session_zero_octopoints_joined_is_free_slot(my_predbat):
+    """
+    Test that the free/saving split is driven by eventType, not by the reward value.
+    Covers GitHub issue #4851.
+
+    Octopus publishes Power Up and Power Down through the same savingSessions feed and tells them
+    apart with eventType: TURN_DOWN (reduce, rewarded), WEEKEND_HAPPY_HOUR (earned free import hour,
+    reward 0), TURN_UP (increase, rewarded). Only WEEKEND_HAPPY_HOUR is free.
+
+    The reward value cannot stand in for the type: a TURN_DOWN reporting 0 would be zeroed on IMPORT,
+    making Predbat grid-charge through a session where the user is meant to be reducing import.
+    """
+    print("Test joined session free/saving split keys on eventType (issue #4851)")
+    ha = my_predbat.ha_interface
+    failed = False
+    date_today = datetime.now().strftime("%Y-%m-%d")
+    tz_offset = int(my_predbat.midnight_utc.tzinfo.utcoffset(my_predbat.midnight_utc).total_seconds() / 3600)
+    tz_offset = f"{tz_offset:02d}"
+
+    session_binary = """
+state: off
+current_joined_event_start: null
+current_joined_event_end: null
+current_joined_event_duration_in_minutes: null
+next_joined_event_start: null
+next_joined_event_end: null
+next_joined_event_duration_in_minutes: null
+icon: mdi:leaf
+friendly_name: Octoplus Saving Session
+"""
+
+    def run(joined_yaml):
+        session_sensor = f"""
+state: '2025-01-23T12:10:11.108+{tz_offset}:00'
+event_types: octopus_energy_all_octoplus_saving_sessions
+event_type: octopus_energy_all_octoplus_saving_sessions
+account_id: A-4DD6C5EE
+available_events: []
+joined_events:
+{joined_yaml}
+friendly_name: Octoplus Saving Session Events
+"""
+        ha.dummy_items.clear()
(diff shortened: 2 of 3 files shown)


mgazza:
**Not a duplicate of #4835 / #4837** — same user-visible symptom, different feed, and #4837 does not cover this path.

**What the events are.** Weekday Power Down participation earns a **Weekend Happy Hour**: Octopus offers several candidate hours at the weekend and the customer books one. All the offered hours appear in the saving-session `available_events` at `octopoints_per_kwh: 0`; the booked one moves to `joined_events`, also at 0. The lists are disjoint, so `joined_events` is a reliable signal of what the customer actually has — and it carries genuine weekday Power Down sessions (68 octopoints/kWh observed) in the same list, which is why the reward value is the right discriminator and this change keys on it.

**Why #4837 does not cover it:**

| | feed | guard that dropped it |
|---|---|---|
| #4835 / #4837 (fixed) | BottleCapDave `octopus_free_session` → `event.*_octoplus_free_electricity_session_events` | `if start and end and code:` — booked Happy Hours publish `code: null` |
| this PR | **Octopus Direct** saving-session `joined_events` | `saving_rate > 0` — booked Happy Hours report `octopoints_per_kwh: 0` |

On Octopus Direct the free-electricity feed carries nothing for the day (`sensor.*_free_electricity` is `off`, its events historical), so the booked hour *only* ever appears in `joined_events`. The two reports are of the same weekend: #4835's reporter booked event ids 5798 and 5800, two of the hours left unbooked in the case behind this PR.

Observed dropped on at least two consecutive weekends, with a hand-typed manual import rate override as the workaround both times.

springfall2008:
How does this code know its a power up and not a power down with zero reward?

mgazza:
It doesn't, and you're right to push on it — the change infers it from the reward being zero, and that inference isn't sound.

**What I actually have is correlation, not a discriminator.** On the account this came from, `joined_events` carried weekday Power Down sessions at 68 octopoints/kWh and weekend Happy Hours at 0, consistently, across two weekends. Combined with #4548 point 5 ("all national power up event times are populated in the power down/saving session 'available events' data set ... all with 0p/kWh") I took zero to mean Power Up. But that's a description of a known upstream mislabelling, not a guarantee, and I can't rule out a genuine Power Down arriving at 0 — an unknown-yet reward published as `0` rather than `null` would do it.

**And the failure mode is bad, not neutral.** If a real Power Down did report 0, this change would zero the *import* rate for it, so Predbat would grid-charge through a session where the user is supposed to be reducing import — paying full price believing it's free, and losing the participation. That's worse than the bug it fixes. It shouldn't merge on an inference.

**There is an authoritative source already wired up.** `async_get_flexibility_campaign_events()` queries `customerFlexibilityCampaignEvents` *per campaign slug*, which separates the two properly:

```python
campaign_slug="octoplus-saving-sessions"   # Power Down
campaign_slug="free_electricity"           # Power Up
```

No inference needed — the API says which campaign an event belongs to.

The reason it isn't already solving this is that `free_electricity` returns nothing current. On the account here it returns 15 events, `last: 50`, newest **2025-10-25** — i.e. the old free-electricity campaign, which appears to have ended. Weekend Happy Hours are evidently a *different campaign*, so they fall through to the legacy `savingSessions` query, which is exactly where the two kinds get mixed at 0 p/kWh.

So the correct fix looks like: find the campaign slug Happy Hours are published under, query it alongside the other two, and feed the results into `free_electricity_events` — which already flows into `octopus_free_slots` through the existing path. That needs no guessing about reward values at all, and would also cover the case where a Power Down legitimately reports 0.

I don't know the slug — I only have the two above, and the inconsistent naming (`octoplus-saving-sessions` hyphenated vs `free_electricity` underscored) suggests they came from different places, so I'd rather not guess. Do you know it, or is there someone with a booked Happy Hour who could dump `customerFlexibilityCampaignEvents` for a few candidate slugs?

Happy to rework this PR that way, or close it if you'd rather fold it into #4548 point 7. Either way I don't think the current version should go in as-is.

springfall2008:
Ran pr-cleanup on this PR: CI is green (kernel-binaries, pre-commit) and there are no unaddressed inline review comments — the only outstanding item is the top-level exchange between @springfall2008 and @mgazza above, which already reaches a conclusion.

I checked whether the open question in that exchange (is there a way to identify the correct campaign slug for Happy Hour events without guessing) could be resolved from the codebase. `async_get_flexibility_events()` (octopus.py:1371-1420) queries `customerFlexibilityCampaignEvents` with two hardcoded slug literals, `octoplus-saving-sessions` and `free_electricity` — there's no query anywhere that enumerates a customer's actual campaigns, so the codebase itself has no way to surface the slug Happy Hour events are published under. I don't have a live account to dump `customerFlexibilityCampaignEvents` against candidate slugs either.

That leaves mgazza's assessment as the correct read: the `== 0` inference this PR currently relies on is unverified, and its failure mode — silently zeroing the import rate for a genuine Power Down that happens to report 0 — is worse than the bug it fixes (#4851). There's no safe code change to make without the campaign slug or account data mgazza asked for, so I'm not implementing anything further here rather than guessing. Leaving this open for @springfall2008 to weigh in on the slug, or for a user with a booked Happy Hour to supply the data.

mgazza:
Answered — and thank you, my original approach was wrong.

`eventType` is in the API and Predbat simply wasn't asking for it. From @sammort's payload dumps on #4531:

```json
{ "id": 4809, "code": "EVENT_41_050826", "rewardPerKwhInOctoPoints": 93, "eventType": "TURN_DOWN",          "status": "DONE" }
{ "id": 5144, "code": "EVENT_45_160826", "rewardPerKwhInOctoPoints": 0,  "eventType": "WEEKEND_HAPPY_HOUR", "status": "UPCOMING" }
```

with the schema description for the field being *"Whether the event is a reduction event (TURN_DOWN) or an increase event (TURN_UP)."*

Our `savingSessions` query requests `id, code, rewardPerKwhInOctoPoints, startAt, endAt, devEvent, targetRegion` — no `eventType`, so both kinds arrived indistinguishable and the reward value was the only thing left to look at. That was the whole problem.

**Reworked** (force-pushed, single commit):

- request `eventType` in the query
- carry it through to the joined and available event lists
- route a joined `WEEKEND_HAPPY_HOUR` into `octopus_free_slots`

`TURN_UP` deliberately stays on the saving path — it means import more, but it is rewarded in octopoints rather than free, so folding it in here would misprice it. An absent `eventType` falls through to existing behaviour, so the BottleCapDave feed and older API responses are unaffected.

Your specific case is now a test: a `TURN_DOWN` reporting 0 must not become free. Worth spelling out why it matters, since it is the failure mode I had built in — a zero-reward `TURN_DOWN` treated as free zeroes the **import** rate, so Predbat would grid-charge through a session where the user is meant to be *reducing* import: full price, believing it free, and the participation lost. Strictly worse than the bug being fixed.

Four cases covered, and reverting to the reward-value test fails two of them:

- joined `WEEKEND_HAPPY_HOUR` becomes a free slot, while a rewarded `TURN_DOWN` in the same `joined_events` list stays a saving slot
- `TURN_DOWN` at 0 reward does not become free
- `TURN_UP` does not become free
- missing `eventType` does not become free

One thing I could use a second opinion on: I've left `available_events` alone, so an *offered but unbooked* Happy Hour is not treated as free. That matches what I can observe — Octopus offers several candidate hours and the customer books one, and only the booked one moves to `joined_events`. But if Octopus ever auto-enrols rather than requiring a booking, the joined list would be the wrong place to look. Do you know if that varies?

`unit_test.py -k saving` and `-k octopus` both pass in full.

gcoan:
> One thing I could use a second opinion on: I've left available_events alone, so an offered but unbooked Happy Hour is not treated as free. That matches what I can observe — Octopus offers several candidate hours and the customer books one, and only the booked one moves to joined_events. But if Octopus ever auto-enrols rather than requiring a booking, the joined list would be the wrong place to look. Do you know if that varies?

@mgazza The changes made look to match what I understand of what the BCD integration does

Answering your last point though, Octopus can and DID auto-join people to a free saving session, the one after the Eclipse power down, people were told what hour they were allocated.  The events that were available that they were not allocated to still appeared in available events, the Octopus allocated event appeared in joined events.

I'm not sure what the issue is you are suggesting , the Octopus selected joined events appears the same as the user selected power up free hour

mgazza:
@gcoan thank you — that settles it, and it means my worry was unfounded. If Octopus allocates rather than the user booking, the allocated event still lands in `joined_events` and the ones they weren't allocated stay in `available_events`. So `joined_events` is the right place to look either way, and there's no case to handle differently. Nothing to change there.

I've taken your other point from #4851 and pushed it as a second commit: **Weekend Happy Hours are no longer offered in `available_events`**, matching BCD v19.0.1. They can't be joined through the API, so listing them only produces the rejected join attempts from #4593/#4595 and fills the join selector with unselectable options. Now that `eventType` is being requested they're identifiable, so this can be done precisely rather than by reward value.

One ordering trap worth flagging for review, because it's invisible in the diff: the reward/code/type lookup maps are built from the same `events` list, and a *joined* Happy Hour reads its type out of those maps. Skipping before they're populated leaves the joined event with `event_type: None` **and** the injected default reward — so a free hour gets priced as an 80p/kWh saving session, which is worse than the original bug. The skip therefore sits after the maps are filled, and there's a test that fails if it's moved above them.

Also fixed the Copilot nits on #4849 (test list numbering, docstring phrasing) and #4850 (log prefix — the new warnings now use `Warn: Kraken: …` like the rest of the component).

PR #4862, opened 2026-08-30
fix(debug): keep the 5-minute step arrays in the debug yaml so dumps replay

### LINKED ISSUE #4853: Debug yaml downloaded with debug disabled has empty step arrays and cannot be replayed
**Describe the bug**

A debug yaml downloaded from the web interface while `switch.predbat_debug_enable` is **off** ships the 5-minute step arrays as empty dicts, and cannot be replayed.

Since that is how most reporters produce the file - they hit Download on the Debug panel without turning the switch on first - the primary triage artefact is broken for exactly the users who most need hand-holding.

**Reproduce**

1. Run Predbat with `switch.predbat_debug_enable` off (the default).
2. Download the debug yaml from the web interface Debug panel.
3. `cd coverage && ./run_all --debug_file <that file>`

```
File "apps/predbat/prediction.py", line 769, in run_prediction
    pv_now = pv_forecast_minute_step_flat[minute]
KeyError: 0
```

The file itself shows why:

```yaml
load_minutes_step: {}
load_minutes_step10: {}
pv_forecast_minute_step: {}
pv_forecast_minute10_step: {}
```

**Cause**

`update_pred()` frees them at the end of every cycle:

```python
# Free memory
if not self.debug_enable:
    self.load_minutes_step = {}
    self.load_minutes_step10 = {}
    self.pv_forecast_minute_step = {}
    self.pv_forecast_minute10_step = {}
```

and they are only ever repopulated inside `calculate_plan()`'s `recompute` branch. `create_debug_yaml()` just serialises whatever is in memory, so with the switch off it always captures the freed state.

The asymmetry confirms it - the two step arrays that are *not* in the free list survive into the dump:

| array | in the free list | in a real dump taken with debug off |
|---|---|---|
| `load_minutes_step` | yes | `{}` |
| `load_minutes_step10` | yes | `{}` |
| `pv_forecast_minute_step` | yes | `{}` |
| `pv_forecast_minute10_step` | yes | `{}` |
| `load_minutes_step90` | no | populated |
| `pv_forecast_minute90_step` | no | populated |

**Why `--redo` is not a fix**

`./run_all --debug_file X --redo` does replay the file, but it recomputes rates, the load model and Octopus slots, so the inputs are no longer the ones the reporter had. A plan bug replayed that way may simply not reproduce, and any conclusion drawn from it is unsafe.

**Note on the debug-history snapshots**

`debug_history_snapshot_*` files are unaffected - that feature requires the debug switch, so their step arrays are always populated. Only the on-demand web download is affected.

**Version**

Reproduced on 8.54.1 with a dump from a live install, and the code path is unchanged on current main.

chalfontchubby:
Fixes #4853

A debug yaml downloaded from the web interface while `switch.predbat_debug_enable` is off ships the 5-minute step arrays as `{}`, and cannot be replayed — `run_prediction()` indexes `pv_forecast_minute_step` directly and raises `KeyError: 0`.

`update_pred()` frees `load_minutes_step`, `load_minutes_step10`, `pv_forecast_minute_step` and `pv_forecast_minute10_step` at the end of every cycle unless debug is on, and only `calculate_plan()`'s recompute branch ever refills them. `create_debug_yaml()` serialises whatever is in memory, so with the switch off it always captures the freed state.

This matters more than it first looks because `DEBUG_ENABLE_MAX_HOURS = 2` auto-disables the switch. Even a reporter who follows the instruction to turn debug on ships a freed dump if they grab it more than two hours later, so the primary triage artefact is broken for most bug reports. `--redo` replays such a file but recomputes rates, load model and Octopus slots, so the inputs are no longer the reporter's and a plan bug may simply not reproduce.

## What this does

**`create_debug_yaml()` rebuilds the arrays if they have been freed**, from the same inputs `calculate_plan()` uses, so a dump is self-contained however it was taken. Already-populated arrays are left alone, so a dump taken with debug on still reflects the plan that actually ran rather than a fresh rebuild. A failure to rebuild is logged and swallowed — the dump is the whole point of the exercise and is still useful without step data, so it must never be lost to this.

**The replay harness rebuilds them too when a dump arrives without them**, since every dump already attached to a bug report predates this change. It resets only the load model, not rates or Octopus slots, so those replays stay faithful in a way `--redo` does not. Checked against a real dump from a live install: it now replays without `--redo` and reproduces `--redo`'s result exactly (orig `-111.7811`, final `18.9157`).

## Notes

The two `*90` step arrays are not in the free list and so were never affected — that asymmetry is what confirmed the diagnosis. They are left as they are here rather than widened, since freeing them would need the same rebuild treatment.

## Testing

New `debug_yaml_step_data` test module covering the rebuild, that the dump itself carries the arrays, that a failing rebuild does not cost the user their dump, and that populated arrays are left untouched. `./run_all --quick` and `./run_all --test debug_cases` both pass.

### DIFF
--- a/apps/predbat/tests/test_debug_yaml_step_data.py
+++ b/apps/predbat/tests/test_debug_yaml_step_data.py
@@ -1,0 +1,95 @@
+# -----------------------------------------------------------------------------
+# Predbat Home Battery System
+# Copyright Trefor Southwell 2026 - All Rights Reserved
+# This application maybe used for personal use only and not for commercial use
+# -----------------------------------------------------------------------------
+# fmt: off
+# pylint: disable=consider-using-f-string
+# pylint: disable=line-too-long
+# pylint: disable=attribute-defined-outside-init
+
+"""Tests that a debug yaml always carries the 5-minute step arrays.
+
+update_pred() frees load_minutes_step, load_minutes_step10, pv_forecast_minute_step and
+pv_forecast_minute10_step at the end of every cycle unless switch.predbat_debug_enable is on, and
+only calculate_plan()'s recompute branch ever refills them. A debug yaml downloaded from the web
+interface with the switch off therefore shipped them as {}, and replaying such a dump raised
+KeyError on minute 0 inside run_prediction() - so the artefact most bug reports carry could not be
+replayed at all.
+"""
+
+STEP_KEYS = ["load_minutes_step", "load_minutes_step10", "pv_forecast_minute_step", "pv_forecast_minute10_step"]
+
+
+def test_debug_yaml_step_data(my_predbat):
+    """Verify create_debug_yaml() repopulates the step arrays when update_pred() has freed them."""
+    failed = False
+    print("**** Testing debug yaml step data ****")
+
+    saved = {key: getattr(my_predbat, key, {}) for key in STEP_KEYS}
+    # Set by fetch_config_options() on a running install; the test instance has not run a cycle.
+    for name, value in (("metric_load_divergence", None), ("load_scaling", 1.0), ("load_scaling10", 1.1)):
+        if not hasattr(my_predbat, name):
+            setattr(my_predbat, name, value)
+    try:
+        print("Test: step arrays freed by update_pred are rebuilt for the dump")
+        for key in STEP_KEYS:
+            setattr(my_predbat, key, {})
+
+        my_predbat.rebuild_debug_step_data()
+
+        for key in STEP_KEYS:
+            rebuilt = getattr(my_predbat, key)
+            if not rebuilt:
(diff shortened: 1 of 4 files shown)


springfall2008:
Not sure about this, didn't I drop them and use --redo instead?

chalfontchubby:
Does --redo get exactly the same result as would have been the case at the time the debug was captured?
I just noticed I wasn't able to run yamls the way I have been - so long as we can exactly replicate the users experience without the arrays then fine - I'll figure out what I'm doing wrong.

springfall2008:
Technically not, in that it re-creates the step data but in theory the source data is the same so it should be the same (normally it is for me).

Issue #4725, opened 2026-08-25
ALPHAESS: EV hold (Probably) won't work.


dragon2611:
I've not yet tested this (I'm happy to do so but I don't actually need to charge the EV at the moment) but looking at the code (with LLM assistance) and based on my observations of how my inverter behaves when configuring Charge/Discharge I don't believe the current EV hold function (That is to prevent the EV discharging the battery) will actually work.

This is because with no valid slots configured the default action is to disable scheduled discharge, I'm also not convinced 0W is an acceptable power limit.

**Expected behaviour**
Predbat holds the current SOC and prevents discharging to load 

**Predbat version**

8.53.1

**Environment details**

- Alphaess Smile G3 5KW
- HAOS

I'd suggest either (Based on my obsversations on how the schedule affects the inverter behavour.

1) Set a charge window covering the EV charge session but set the target SOC to be lower than the current SOC, the effect is this will basically act like a hold.  -  This is basically how I was doing it manually before Predbat was able to control the battery

OR

2) set a scheduled discharge window that lands outside the current time period.

I actually think option 1 may be cleaner.

springfall2008:
🤖 **Automated first-pass triage** — this is a bot's initial analysis against current `main`. A maintainer will review before any action is taken, so treat the below as a starting point rather than a verdict.

**Classification:** `bug` (an unverified hardware assumption in the experimental AlphaESS Cloud integration) · **Priority:** `priority_low` — confined to AlphaESS, but there is no in-Predbat workaround if the assumption is wrong.

Investigated against `main` @ `e3ada1a4` (`v8.53.0-64-ge3ada1a4`). The hold code you're questioning landed in `v8.52.0`, so it is definitely in the 8.53.1 you're running.

### The good news: the code already does roughly your option 2

Your reading of "with no valid slots configured the default action is to disable scheduled discharge" is a real branch, but it is **not** the one the EV hold takes. The zero-rate case is decided *first*, and deliberately:

- **Legacy two-window path** (`alphaess.py:1115-1124`): `export_rate <= 0` returns `ctrDis: 1` with `timeDisf1/e1/f2/e2` all `00:00` — discharge time control **on**, with no permitted period. `ctrDis: 0` (demand mode, what you were worried about) is only reached at `alphaess.py:1136-1138`, i.e. when the rate is non-zero but no window is programmed.
[...]

PR #4867, opened 2026-08-30
ALPHAESS: Use charge to 10% @ 100w to implement hold.

dragon2611:
Since there's only 2 ways to stop the AlphaESS to discharge to load which is 

1) Set a discharge schedule OUTSIDE of the current time.

2) Set a charge schedule with a SOC below the current SOC, on my Smile G3 this will just hold the battery at the current charge.   

I went with option 2.

Note since 10% is the minimum the API accepts I set a 100w charge limit as well, if the battery is below 10% it will still charge from the grid and there's nothing we can really do about that.

https://github.com/springfall2008/batpred/issues/4725 was annoying me enough to take a stab at it.

Codex/ GPT 5.6 sol (medium) was used for this.

I'll need more time to do full H/W testing but it seems to be working with my Smile G3 stack.

### DIFF
--- a/apps/predbat/alphaess.py
+++ b/apps/predbat/alphaess.py
@@ -131,6 +131,12 @@ class AlphaESSAPI(ComponentBase):
         # allowance every time.
         self.write_burst_start = {}
         self.write_burst_writes = {}
+        # Target latched for the lifetime of one generic discharge hold. AlphaESS ignores
+        # an enabled discharge schedule with no periods, so a hold is expressed as a charge
+        # profile whose target is already below the battery SOC. Latching stops rising solar
+        # SOC from changing the target and consuming another cloud write for every 1% gained.
+        self.hold_target_soc = {}
+        self._hold_power_warned = set()
         # Serials Predbat has actually been asked to drive, i.e. ones whose write button has
         # been pressed at least once. The reconcile loop only re-applies for these, so a
         # startup cycle can never clobber an inverter before there is a plan to apply.
@@ -1020,6 +1026,102 @@ class AlphaESSAPI(ComponentBase):
         except (TypeError, ValueError):
             return low
 
+    def _window_active_now(self, window):
+        """Return whether a stored window covers Predbat's current minute.
+
+        The generic hold-charge path disables the charge-enable switch after deciding the
+        target has been reached, but deliberately leaves the start/end entities intact. The
+        times therefore retain the missing intent needed by this adapter. Hours above 24 are
+        accepted because Predbat represents a window crossing midnight as (for example)
+        23:00-25:00 before the AlphaESS boundary splits it.
+        """
+        start = hm_to_minutes(window.get("start"))
+        end = hm_to_minutes(window.get("end"))
+        if start == end:
+            return False
+        now = int(getattr(self.base, "minutes_now", 0)) % (24 * 60)
+        if end > 24 * 60 and now < start:
+            now += 24 * 60
+        if start < end:
+            return start <= now < end
+        return now >= start or now < end
+
+    def _hold_charge_window(self, sn, schedule):
+        """Translate Predbat's generic discharge hold into an AlphaESS charge profile.
+
+        Predbat represents every no-discharge hold (freeze charge, EV/iBoost hold and other
+        generic callers) by writing discharge power zero. AlphaESS ignores the previous
+        translation, discharge scheduling enabled with no periods. Live SMILE G3 testing in
+        GH#4725 established that an enabled CHARGE profile whose target is below current SOC
+        is the working device primitive instead.
+
+        A planned hold charge can arrive with normal discharge power: execute_plan has
+        already disabled its charge-enable switch and raised the reserve, while the original
+        target and times remain. A currently-active window at/below SOC recovers that intent.
+        A real charge target above SOC is kept verbatim even if another feature also asks for
+        a discharge hold; charging already prevents the battery feeding the load.
+
+        Returns a copied, enabled charge window while a hold is requested, otherwise None.
+        """
+        charge = schedule.get("charge", {}) or {}
+        export = schedule.get("export", {}) or {}
+        current_soc = self._as_float(self.device_values.get(sn, {}).get("soc"), 0.0)
+        charge_target = self._as_float(charge.get("soc"), 0.0)
+        charge_power = int(self._as_float(charge.get("power"), 0.0))
+        export_power = self._as_float(export.get("power"), 0.0)
+
+        discharge_hold = export_power <= 0
+        # A disabled window is only a generic hold while its computed charge rate is zero.
+        # Once execute_plan releases the hold it restores the rate before the old target/time
+        # entities are necessarily replaced; requiring zero prevents those stale fields from
+        # extending the hold. An explicitly enabled target-at/below-SOC profile is itself a
+        # hold regardless of the valid non-zero power it carries.
+        charge_window_hold = charge_target > 0 and charge_target <= current_soc and self._window_active_now(charge) and (bool(charge.get("enable")) or charge_power <= 0)
+        if not discharge_hold and not charge_window_hold:
+            self.hold_target_soc.pop(sn, None)
+            self._hold_power_warned.discard(sn)
+            return None
+
+        # Do not replace a genuine charge just because an EV/iBoost hold is present too.
+        # The active grid charge already prevents discharge and its requested target/rate
+        # must survive unchanged.
+        if bool(charge.get("enable")) and charge_target > current_soc and charge_power > 0:
+            self.hold_target_soc.pop(sn, None)
+            self._hold_power_warned.discard(sn)
+            return dict(charge)
+
+        power = charge_power if charge_power > 0 else int(self.battery_rate_max(sn))
+        if power <= 0:
+            if sn not in self._hold_power_warned:
+                self._hold_power_warned.add(sn)
+                self.log("Warn: AlphaESS {} cannot create its hold charge profile because neither the schedule nor the inverter rating supplies a positive power; keeping the legacy hold payload".format(sn))
+            return None
+        self._hold_power_warned.discard(sn)
+
+        if sn not in self.hold_target_soc:
+            below_current = self._clamp_percent(current_soc - 1, low=10, high=100)
+            # A genuine planned hold target is already below current SOC and should remain
+            # the target. Freeze charge and demand-mode holds carry an equal/empty target,
+            # so they use current SOC minus one instead.
+            if 10 <= charge_target < current_soc and charge_window_hold:
+                target = self._clamp_percent(charge_target, low=10, high=100)
+            else:
+                target = below_current
+            self.hold_target_soc[sn] = target
+
+        if charge_window_hold:
+            start = charge.get("start", "00:00:00")
+            end = charge.get("end", "23:45:00")
+        else:
+            # Generic demand-mode holds carry no end time. A stable daily profile avoids a
+            # rolling boundary that would change the payload and spend an AlphaESS cloud
+            # write every 15 minutes; reconciliation removes it as soon as the hold signal
+            # clears on the next Predbat cycle.
+            start = "00:00:00"
+            end = "23:45:00"
+
+        return {"enable": True, "soc": self.hold_target_soc[sn], "power": power, "start": start, "end": end}
+
     def split_window(self, start, end):
         """Split a window at midnight, returning ((start1, end1), (start2, end2)) in HH:mm.
 
@@ -1073,7 +1175,7 @@ class AlphaESSAPI(ComponentBase):
         A FULL REPLACEMENT, not a patch: all seven fields are always present, because the
         endpoint silently resets anything omitted.
         """
-        window = schedule.get("charge", {}) or {}
+        window = self._hold_charge_window(sn, schedule) or (schedule.get("charge", {}) or {})
         enabled = bool(window.get("enable"))
         # charge_rate zero is Predbat signalling freeze charge / no cross-charging. There
         # is no pause endpoint, so a zeroed rate is the only signal available and it
@@ -1113,15 +1215,23 @@ class AlphaESSAPI(ComponentBase):
         reserve = self._clamp_percent(schedule.get("reserve", 0))
         export_rate = self._as_float(window.get("power"), 0.0)
 
-        # discharge_rate zero means hold SOC (freeze export): discharge time control ON
-        # with no permitted period, so the battery cannot discharge at all. This does NOT
-        # depend on the charge rate - Predbat reaches discharge_rate == 0 together with
-        # charge_rate == 0 through ordinary, reachable combinations (set_freeze_export_during_demand
-        # zeroing the charge rate while a car-charging-from-battery hold or
-        # iboost_prevent_discharge zeroes the discharge rate in the same pass), and that
-        # combination is an explicit hold, not an absence of a plan. Treating it as "no plan"
-        # would let the battery discharge into the EV or iBoost load against the hold Predbat
-        # asked for.
+        # The working AlphaESS no-discharge primitive is the charge profile built by
+        # build_charge_payload, not an empty discharge programme. Leave discharge time
+        # control off so the ignored empty schedule cannot compete with that profile.
+        if self._hold_charge_window(sn, schedule):
+            return {
+                "sysSn": sn,
+                "ctrDis": 0,
+                "timeDisf1": ALPHAESS_TIME_DISABLED,
+                "timeDise1": ALPHAESS_TIME_DISABLED,
+                "timeDisf2": ALPHAESS_TIME_DISABLED,
+                "timeDise2": ALPHAESS_TIME_DISABLED,
+                "batUseCap": reserve,
+            }
+
+        # Fallback only: reached when a hold was requested but no positive charge-profile
+        # power could be derived. Hardware testing says AlphaESS ignores this empty schedule,
+        # but retaining it is safer than silently returning to demand when no rating is available.
         if export_rate <= 0:
             return {
                 "sysSn": sn,
@@ -1519,7 +1629,8 @@ class AlphaESSAPI(ComponentBase):
         holds schedule["reserve"] rather than an arbitrary constant - see
         build_discharge_payload for the same one-field-two-purposes rule on the legacy pair.
         """
-        charge = schedule.get("charge", {}) or {}
+        hold_charge = self._hold_charge_window(sn, schedule)
+        charge = hold_charge or (schedule.get("charge", {}) or {})
         export = schedule.get("export", {}) or {}
         reserve = schedule.get("reserve", 10)
         charge_rate = self._as_float(charge.get("power"), 0.0)
@@ -1531,14 +1642,10 @@ class AlphaESSAPI(ComponentBase):
             charge_on = False
         charge_list = [self._periodic_entry(charge_start, charge_end, charge.get("soc", 100), charge_rate)] if charge_on else [self._periodic_entry(ALPHAESS_TIME_DISABLED, ALPHAESS_TIME_DISABLED, 10, 0)]
 
-        # export_rate zero means hold SOC (freeze export), exactly as build_discharge_payload:
-        # discharge time control ON with no permitted period, so the battery cannot discharge
-        # at all. Decided before, and independently of, the window/enable check - Predbat
-        # reaches export_rate == 0 through ordinary, reachable combinations (see
-        # build_discharge_payload's comment), and that is an explicit hold, not the absence of
-        # a plan. Getting this wrong lets the battery discharge into the EV or iBoost load
-        # against the hold Predbat asked for; ctrDisCycle 0 here is DEMAND MODE, not a hold.
-        export_hold = export_rate <= 0
+        # While the hold charge profile is active it is the whole control action. Suppress
+        # any export schedule and leave discharge time control in demand mode; AlphaESS
+        # ignores the former ctrDisCycle=1 plus empty-period representation.
+        export_hold = hold_charge is not None
         if export_hold:
             export_on = False
         else:
@@ -1572,11 +1679,9 @@ class AlphaESSAPI(ComponentBase):
             "chargeTimeList": charge_list,
             "dischargeTimeList": discharge_list,
             "gridChargeCycle": 1 if charge_on else 0,
-            # export_hold sets this ON (1) with no permitted period - the hard hold.
-            # export_on sets it ON for a real scheduled window. Otherwise it is OFF (0),
-            # demand mode, where the battery covers the house down to the reserve floor
-            # carried by the idle discharge entry above.
-            "ctrDisCycle": 1 if (export_on or export_hold) else 0,
+            # A real export uses discharge time control. Hold charge deliberately leaves it
+            # off: the enabled charge profile is what prevents discharge on AlphaESS.
+            "ctrDisCycle": 1 if export_on else 0,
         }
 
     async def apply_settings(self, sn, schedule, force=False):
--- a/apps/predbat/alphaess.py
+++ b/apps/predbat/alphaess.py
@@ -131,12 +131,6 @@ class AlphaESSAPI(ComponentBase):
         # allowance every time.
         self.write_burst_start = {}
         self.write_burst_writes = {}
-        # Target latched for the lifetime of one generic discharge hold. AlphaESS ignores
-        # an enabled discharge schedule with no periods, so a hold is expressed as a charge
-        # profile whose target is already below the battery SOC. Latching stops rising solar
-        # SOC from changing the target and consuming another cloud write for every 1% gained.
-        self.hold_target_soc = {}
-        self._hold_power_warned = set()
         # Serials Predbat has actually been asked to drive, i.e. ones whose write button has
         # been pressed at least once. The reconcile loop only re-applies for these, so a
         # startup cycle can never clobber an inverter before there is a plan to apply.
@@ -1052,12 +1046,13 @@ class AlphaESSAPI(ComponentBase):
         Predbat represents every no-discharge hold (freeze charge, EV/iBoost hold and other
         generic callers) by writing discharge power zero. AlphaESS ignores the previous
         translation, discharge scheduling enabled with no periods. Live SMILE G3 testing in
-        GH#4725 established that an enabled CHARGE profile whose target is below current SOC
-        is the working device primitive instead.
+        GH#4725 established that an enabled CHARGE profile with a low target is the working
+        device primitive instead. Synthetic holds use a fixed 10% target and 100 W power.
 
         A planned hold charge can arrive with normal discharge power: execute_plan has
         already disabled its charge-enable switch and raised the reserve, while the original
-        target and times remain. A currently-active window at/below SOC recovers that intent.
+        target and times remain. A currently-active window at/below SOC recovers that intent
+        and converts it to the same fixed hold profile.
         A real charge target above SOC is kept verbatim even if another feature also asks for
         a discharge hold; charging already prevents the battery feeding the load.
 
@@ -1078,36 +1073,13 @@ class AlphaESSAPI(ComponentBase):
         # hold regardless of the valid non-zero power it carries.
         charge_window_hold = charge_target > 0 and charge_target <= current_soc and self._window_active_now(charge) and (bool(charge.get("enable")) or charge_power <= 0)
         if not discharge_hold and not charge_window_hold:
-            self.hold_target_soc.pop(sn, None)
-            self._hold_power_warned.discard(sn)
             return None
 
         # Do not replace a genuine charge just because an EV/iBoost hold is present too.
         # The active grid charge already prevents discharge and its requested target/rate
         # must survive unchanged.
         if bool(charge.get("enable")) and charge_target > current_soc and charge_power > 0:
-            self.hold_target_soc.pop(sn, None)
-            self._hold_power_warned.discard(sn)
             return dict(charge)
-
-        power = charge_power if charge_power > 0 else int(self.battery_rate_max(sn))
-        if power <= 0:
-            if sn not in self._hold_power_warned:
-                self._hold_power_warned.add(sn)
-                self.log("Warn: AlphaESS {} cannot create its hold charge profile because neither the schedule nor the inverter rating supplies a positive power; keeping the legacy hold payload".format(sn))
-            return None
-        self._hold_power_warned.discard(sn)
-
-        if sn not in self.hold_target_soc:
-            below_current = self._clamp_percent(current_soc - 1, low=10, high=100)
-            # A genuine planned hold target is already below current SOC and should remain
-            # the target. Freeze charge and demand-mode holds carry an equal/empty target,
-            # so they use current SOC minus one instead.
-            if 10 <= charge_target < current_soc and charge_window_hold:
-                target = self._clamp_percent(charge_target, low=10, high=100)
-            else:
-                target = below_current
-            self.hold_target_soc[sn] = target
 
         if charge_window_hold:
             start = charge.get("start", "00:00:00")
@@ -1120,7 +1092,11 @@ class AlphaESSAPI(ComponentBase):
             start = "00:00:00"
             end = "23:45:00"
 
-        return {"enable": True, "soc": self.hold_target_soc[sn], "power": power, "start": start, "end": end}
+        # Keep the synthetic profile deterministic. Zero power disables charging in
+        # Predbat and may cause AlphaESS to discard the profile, while 100 W is a small,
+        # positive setpoint on the periodic API. The legacy endpoint has no power field,
+        # but still receives the same enabled 10% target and time window.
+        return {"enable": True, "soc": 10, "power": 100, "start": start, "end": end}
 
     def split_window(self, start, end):
         """Split a window at midnight, returning ((start1, end1), (start2, end2)) in HH:mm.
@@ -1213,7 +1189,6 @@ class AlphaESSAPI(ComponentBase):
         window = schedule.get("export", {}) or {}
         enabled = bool(window.get("enable"))
         reserve = self._clamp_percent(schedule.get("reserve", 0))
-        export_rate = self._as_float(window.get("power"), 0.0)
 
         # The working AlphaESS no-discharge primitive is the charge profile built by
         # build_charge_payload, not an empty discharge programme. Leave discharge time
@@ -1222,20 +1197,6 @@ class AlphaESSAPI(ComponentBase):
             return {
                 "sysSn": sn,
                 "ctrDis": 0,
-                "timeDisf1": ALPHAESS_TIME_DISABLED,
-                "timeDise1": ALPHAESS_TIME_DISABLED,
-                "timeDisf2": ALPHAESS_TIME_DISABLED,
-                "timeDise2": ALPHAESS_TIME_DISABLED,
-                "batUseCap": reserve,
-            }
-
-        # Fallback only: reached when a hold was requested but no positive charge-profile
-        # power could be derived. Hardware testing says AlphaESS ignores this empty schedule,
-        # but retaining it is safer than silently returning to demand when no rating is available.
-        if export_rate <= 0:
-            return {
-                "sysSn": sn,
-                "ctrDis": 1,
                 "timeDisf1": ALPHAESS_TIME_DISABLED,
                 "timeDise1": ALPHAESS_TIME_DISABLED,
                 "timeDisf2": ALPHAESS_TIME_DISABLED,
--- a/apps/predbat/alphaess.py
+++ b/apps/predbat/alphaess.py
@@ -11,8 +11,13 @@
 Registers each discovered AlphaESS system as an ``AlphaESSCloud`` Predbat inverter,
 publishing monitoring sensors and schedule control entities. Predbat's controls map
 straight onto the AlphaESS schedule fields - gridCharge/timeChaf/batHighCap for charging
-and ctrDis/timeDisf/batUseCap for export - and the inverter does the timing, so there is
-no per-instant work mode to derive.
+and ctrDis/timeDisf/batUseCap for scheduled discharge permission - and the inverter does
+the timing, so there is no per-instant work mode to derive.
+
+Hardware testing confirms that the periodic discharge schedule only controls when the
+battery may serve local load; it does not force grid export. The legacy endpoint has no
+power field at all, so forced-export behaviour there remains unverified and cannot be
+requested at a particular wattage. The private VPP control used by providers is unavailable.
 
 Auth is stateless: every request carries appId, timeStamp and
 sign = sha512(appId + appSecret + timeStamp). There is no token and no refresh, so the
@@ -61,6 +66,8 @@ from alphaess_const import (
     window_is_empty,
     ALPHAESS_TIME_DISABLED,
     ALPHAESS_TIME_MAX,
+    ALPHAESS_HOLD_SOC,
+    ALPHAESS_HOLD_POWER,
     ALPHAESS_SETTLE_POLLS,
     ALPHAESS_WRITE_SETTLE_SECONDS,
     ALPHAESS_WRITE_BURST_MAX,
@@ -73,6 +80,9 @@ from alphaess_const import (
     ALPHAESS_TTL_CONFIG,
     ALPHAESS_TTL_ENERGY,
 )
+
+
+_HOLD_NOT_EVALUATED = object()
 
 
 class AlphaESSAPI(ComponentBase):
@@ -1026,14 +1036,14 @@ class AlphaESSAPI(ComponentBase):
         The generic hold-charge path disables the charge-enable switch after deciding the
         target has been reached, but deliberately leaves the start/end entities intact. The
         times therefore retain the missing intent needed by this adapter. Hours above 24 are
-        accepted because Predbat represents a window crossing midnight as (for example)
+        accepted because Predbat can represent a window crossing midnight as (for example)
         23:00-25:00 before the AlphaESS boundary splits it.
         """
         start = hm_to_minutes(window.get("start"))
         end = hm_to_minutes(window.get("end"))
         if start == end:
             return False
-        now = int(getattr(self.base, "minutes_now", 0)) % (24 * 60)
+        now = int(self.minutes_now) % (24 * 60)
         if end > 24 * 60 and now < start:
             now += 24 * 60
         if start < end:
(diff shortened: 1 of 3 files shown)


apps/predbat/alphaess.py
```diff
@@ -1020,6 +1020,84 @@ def _clamp_percent(value, low=0, high=100):
...
+        # extending the hold. An explicitly enabled target-at/below-SOC profile is itself a
+        # hold regardless of the valid non-zero power it carries.
+        charge_window_hold = charge_target > 0 and charge_target <= current_soc and self._window_active_now(charge) and (bool(charge.get("enable")) or charge_power <= 0)
+        if not discharge_hold and not charge_window_hold:
+            return None
+
+        # Do not replace a genuine charge just because an EV/iBoost hold is present too.
+        # The active grid charge already prevents discharge and its requested target/rate
+        # must survive unchanged.
+        if bool(charge.get("enable")) and charge_target > current_soc and charge_power > 0:
```
springfall2008: **Bug (verified): the real-charge escape fires for a FUTURE window and defeats an active hold.** `if bool(charge.get("enable")) and charge_target > current_soc and charge_power > 0: return dict(charge)` has no `_window_active_now` guard. With a car/iBoost hold active now (export power 0) and an enabled real charge window that has not started yet, this branch returns the (future) charge dict - so `build_discharge_payload` sees a non-None hold and writes ctrDis=0 demand mode, while nothing enforces the hold now: the battery discharges into the EV/iBoost load. Pre-PR, export_rate<=0 produced the ctrDis=1 hard hold at any hour. Suggest adding `self._window_active_now(charge)` to the escape condition.

apps/predbat/alphaess.py
```diff
@@ -1020,6 +1020,84 @@ def _clamp_percent(value, low=0, high=100):
...
+        charge_power = int(self._as_float(charge.get("power"), 0.0))
+        export_power = self._as_float(export.get("power"), 0.0)
+
+        discharge_hold = export_power <= 0
+        # A disabled window is only a generic hold while its computed charge rate is zero.
+        # Once execute_plan releases the hold it restores the rate before the old target/time
+        # entities are necessarily replaced; requiring zero prevents those stale fields from
+        # extending the hold. An explicitly enabled target-at/below-SOC profile is itself a
+        # hold regardless of the valid non-zero power it carries.
+        charge_window_hold = charge_target > 0 and charge_target <= current_soc and self._window_active_now(charge) and (bool(charge.get("enable")) or charge_power <= 0)
```
springfall2008: **Bug (verified): a real, active export window is silently dropped.** `charge_window_hold` has no `export_power` term, and the early return passes when it is true even while `discharge_hold` is False (real export). The escape below cannot rescue this case because it requires `charge_target > current_soc`, contradicting `charge_target <= current_soc` here. Result: `build_discharge_payload` short-circuits to ctrDis=0 with all periods disabled and `build_periodic_payload` sets export_hold=True / export_on=False - the planned export (ctrDis=1, real times, batUseCap) is replaced by the 10%/100W profile for as long as reported SOC sits at/above the charge target. This state is anticipated by the codebase itself: the periodic overlap-trim (export 12:00-13:30 vs charge 13:00-14:00) exists precisely because charge_on and export_on co-occur, and execute.py runs the charge and export blocks independently in the same pass. Consider gating `charge_window_hold` on `export_power <= 0`, or having `_hold_charge_window` return the hold kind so a charge-only hold cannot veto a planned export.

apps/predbat/alphaess.py
```diff
@@ -1111,21 +1189,14 @@ def build_discharge_payload(self, sn, schedule):
...
-        # iboost_prevent_discharge zeroes the discharge rate in the same pass), and that
-        # combination is an explicit hold, not an absence of a plan. Treating it as "no plan"
-        # would let the battery discharge into the EV or iBoost load against the hold Predbat
-        # asked for.
-        if export_rate <= 0:
+
+        # The working AlphaESS no-discharge primitive is the charge profile built by
+        # build_charge_payload, not an empty discharge programme. Leave discharge time
+        # control off so the ignored empty schedule cannot compete with that profile.
+        if self._hold_charge_window(sn, schedule):
```
springfall2008: **Bug (verified): the hold decision is evaluated twice per legacy apply and the two evaluations can disagree.** `build_discharge_payload` re-runs the full `_hold_charge_window` computation (live SOC read + window-active check) independently of the identical call inside `build_charge_payload`. `apply_settings` awaits `asyncio.sleep(self.api_delay)` between building the charge payload and the discharge payload, and `_write_payload` pacing (alphaess_min_write_interval, default 300s) can delay the charge write, so the second evaluation can see a different `device_values[sn][soc]` or a crossed window boundary: the charge side is then written as the 10% hold profile while the discharge side decides not-a-hold and writes ctrDis=1 with a real export period - exactly the competing-representation state this comment says it is eliminating (and the inverse: no hold profile plus ctrDis=0). Compute the hold once in `apply_settings` and pass it to both builders.

apps/predbat/alphaess.py
```diff
@@ -1020,6 +1020,84 @@ def _clamp_percent(value, low=0, high=100):
...
+        export_power = self._as_float(export.get("power"), 0.0)
+
+        discharge_hold = export_power <= 0
+        # A disabled window is only a generic hold while its computed charge rate is zero.
+        # Once execute_plan releases the hold it restores the rate before the old target/time
+        # entities are necessarily replaced; requiring zero prevents those stale fields from
+        # extending the hold. An explicitly enabled target-at/below-SOC profile is itself a
+        # hold regardless of the valid non-zero power it carries.
+        charge_window_hold = charge_target > 0 and charge_target <= current_soc and self._window_active_now(charge) and (bool(charge.get("enable")) or charge_power <= 0)
+        if not discharge_hold and not charge_window_hold:
```
springfall2008, line 1090: **Bug (verified): the planned-hold recovery path looks unreachable in production.** `charge_window_hold` requires `charge_power <= 0` when the enable switch is off, but during freeze-charge / Hold charging execute.py sets or restores the charge rate to max (execute.py:292 and 634-640; `find_charge_rate` returns max_rate once soc >= target), so by the time the adapter reads the entities charge_power is the full rate, not 0, and the recovered windowed profile is never built. The new test only passes because its fixture hand-writes power=0 - a state Predbat does not produce. Real planned holds therefore fall through to the all-day 00:00-23:45 synthetic profile (or, on periodic, the same) rather than the recovered windowed hold this docstring describes. Worth verifying what entity state execute.py actually leaves behind before relying on power<=0 as the hold signature.

apps/predbat/alphaess.py
```diff
@@ -1020,6 +1020,84 @@ def _clamp_percent(value, low=0, high=100):
...
+        times therefore retain the missing intent needed by this adapter. Hours above 24 are
+        accepted because Predbat represents a window crossing midnight as (for example)
+        23:00-25:00 before the AlphaESS boundary splits it.
+        """
+        start = hm_to_minutes(window.get("start"))
+        end = hm_to_minutes(window.get("end"))
+        if start == end:
+            return False
+        now = int(getattr(self.base, "minutes_now", 0)) % (24 * 60)
+        if end > 24 * 60 and now < start:
```
springfall2008, line 1047: **Bug (plausible, narrower than the docstring claims): a start>end hold window collapses to NO hold.** This `end > 24 * 60` branch is dead in production and the docstring premise is false: Predbat writes window times mod-24 (inverter.py:3142, strftime %H:%M:%S) and, for AlphaESSCloud (can_span_midnight False), execute.py clamps crossing plan windows to end 23:59 before adjust_charge_window, so a >24:00 end never reaches the schedule entities. A start>end pair CAN still occur via a large `inverter_clock_skew_end` pushing a clamped 23:59 end past midnight (skew is applied after the clamp, inverter.py:3138-3139), or manual edits of the schedule selects - and then the recovered hold dict carries start>end, `split_window` treats it as a same-day window (end_min <= 24*60), `window_is_empty` collapses it to disabled, and the charge payload becomes gridCharge=0 / gridChargeCycle=0 while `build_discharge_payload` writes ctrDis=0: the hold vanishes entirely, a regression vs the old ctrDis=1 hard hold. The existing midnight tests mask this by feeding "26:00" ends that production never writes - a wrap-branch test with realistic times would document the real behaviour.

apps/predbat/alphaess.py
springfall2008: **Review finding 1 (CONFIRMED, severity: high) — a real enabled export window is silently destroyed.**

`charge_window_hold` has no `export_power` term, so whenever an active charge window is at/below SOC (`enable` True, or power<=0), `_hold_charge_window` returns non-None and:
- `build_discharge_payload` returns `ctrDis: 0` with all discharge periods disabled (line ~1212), and
- `build_periodic_payload` sets `export_on = False` and `ctrDisCycle: 0` (line ~1645).

That drops a genuine export window (`export.enable` True, power>0) entirely. Reachability is real: execute.pys "Hold charging" fallback (execute.py:350/363) leaves the charge window **enabled at its original target** once SoC reaches it, and a concurrent export window is exactly the co-occurrence this adapters own overlap-trim code anticipates.

Suggested fix: gate `charge_window_hold` (or the export suppression that follows from it) on `export_power <= 0`, i.e. only treat the charge profile as the whole control action when there is no real export schedule to preserve.

apps/predbat/alphaess.py
```diff
@@ -1020,6 +1020,84 @@ def _clamp_percent(value, low=0, high=100):
...
+            # write every 15 minutes; reconciliation removes it as soon as the hold signal
+            # clears on the next Predbat cycle.
+            start = "00:00:00"
+            end = "23:45:00"
+
+        # Keep the synthetic profile deterministic. Zero power disables charging in
+        # Predbat and may cause AlphaESS to discard the profile, while 100 W is a small,
+        # positive setpoint on the periodic API. The legacy endpoint has no power field,
+        # but still receives the same enabled 10% target and time window.
+        return {"enable": True, "soc": 10, "power": 100, "start": start, "end": end}
```
springfall2008: **Bug (plausible): the synthetic hold actively charges the battery.** This is an ENABLED 10%/100W charge profile, so when reported SOC is below 10% it grid-charges the battery to 10% around the clock - a control action the old zero-rate path never took (it moved no energy). cbat is percent-quantised so SOC < 10 is a normal, reachable state, and the 10% floor is above some users reserves. Consider clamping the synthetic target to max(10, schedule reserve) or at least documenting that a hold can consume grid energy to reach 10%.

apps/predbat/alphaess.py
```diff
@@ -1020,6 +1020,84 @@ def _clamp_percent(value, low=0, high=100):
...
+        target and times remain. A currently-active window at/below SOC recovers that intent
+        and converts it to the same fixed hold profile.
+        A real charge target above SOC is kept verbatim even if another feature also asks for
+        a discharge hold; charging already prevents the battery feeding the load.
+
+        Returns a copied, enabled charge window while a hold is requested, otherwise None.
+        """
+        charge = schedule.get("charge", {}) or {}
+        export = schedule.get("export", {}) or {}
+        current_soc = self._as_float(self.device_values.get(sn, {}).get("soc"), 0.0)
```
springfall2008, line 1077: **Risk (plausible): payload now depends on live telemetry that is quantised, 1-minute, and absent at startup.** Two consequences: (a) a real charge whose reported SOC crosses the charge target flips the payload between the real window and the 10%/100W hold on each crossing - the guard at line 1081 fails on equality, so target == reported SOC converts a genuine charge to a hold; each flip is a byte-different payload, paced only by min_write_interval (300s) against an API documented as writable once per 24h. (b) Before the first poll (and after a restart, since device_values is not persisted) soc defaults to 0.0, so `charge_target <= current_soc` is unsatisfiable and a planned hold silently degrades to the all-day 00:00-23:45 profile instead of the recovered window. Mitigations: exclude the equality case (>= is Predbat own hold semantics anyway, but a hysteresis band would stop the flap), and consider a last-known-SOC fallback.

apps/predbat/alphaess.py
springfall2008: **Review finding 2 (CONFIRMED, severity: high) — the real-charge escape ignores whether the window is active.**

`if bool(charge.get("enable")) and charge_target > current_soc and charge_power > 0: return dict(charge)` fires for a **future / not-yet-active** charge window too: `_window_active_now` is only consulted in `charge_window_hold`, not here.

Failure: during an EV/iBoost discharge hold (`export_power <= 0`) with a charge window that has not started yet, this branch returns non-None, so `build_discharge_payload` writes `ctrDis: 0` and the periodic path writes `ctrDisCycle: 0` (demand mode) with only the future charge window posted. Outside the charge window there is no hold at all and the battery discharges into the EV / iBoost load — the exact failure mode the old code guarded against (`export_rate <= 0` produced a hard `ctrDis=1` at all hours on main, and `export_hold = export_rate <= 0` made `ctrDisCycle=1` unconditionally).

The reserve raise in execute.py:596-598 is only a partial backstop (and is itself conditional). Suggested fix: require `self._window_active_now(charge)` (or an imminent start) before trusting the verbatim charge profile as the hold.

apps/predbat/alphaess.py
```diff
@@ -1519,7 +1590,8 @@ def build_periodic_payload(self, sn, schedule):
         holds schedule["reserve"] rather than an arbitrary constant - see
         build_discharge_payload for the same one-field-two-purposes rule on the legacy pair.
         """
-        charge = schedule.get("charge", {}) or {}
+        hold_charge = self._hold_charge_window(sn, schedule)
+        charge = hold_charge or (schedule.get("charge", {}) or {})
```
springfall2008, line 1621: **Bug (plausible): the periodic path drops the second period of a split hold window.** `charge = hold_charge or ...` funnels the recovered hold back through the generic machinery, and the legacy path splits a crossing window into two periods - but here `(charge_start, charge_end), _ = self._snapped_periods(...)` keeps only periods[0]. For a hold window crossing midnight the periodic payload ends at 23:45 while ctrDisCycle=0 leaves demand mode, so the post-midnight remainder is a discharge hole exactly during the held window (the legacy pair covers both periods). Either build charge_list directly from the fixed hold profile (skipping the re-validation of what `_hold_charge_window` just guaranteed) or handle the split here too.

apps/predbat/alphaess.py
```diff
@@ -1572,11 +1640,9 @@ def build_periodic_payload(self, sn, schedule):
...
             "dischargeTimeList": discharge_list,
             "gridChargeCycle": 1 if charge_on else 0,
-            # export_hold sets this ON (1) with no permitted period - the hard hold.
-            # export_on sets it ON for a real scheduled window. Otherwise it is OFF (0),
-            # demand mode, where the battery covers the house down to the reserve floor
-            # carried by the idle discharge entry above.
-            "ctrDisCycle": 1 if (export_on or export_hold) else 0,
+            # A real export uses discharge time control. Hold charge deliberately leaves it
+            # off: the enabled charge profile is what prevents discharge on AlphaESS.
+            "ctrDisCycle": 1 if export_on else 0,
```
springfall2008, line 1673: **Review finding 3 (CONFIRMED mechanically, severity: high) — a rejected charge write leaves the device in demand mode with no hold.**

`apply_settings` (immediately below this hunk, ~line 1660) writes the charge profile and then the discharge payload with no early return on charge-write failure. If the charge write is rejected or pacing-held while the discharge write succeeds, the device gets `ctrDis 1 -> 0` / `ctrDisCycle 1 -> 0` (demand mode) **without** the hold charge profile that is now the only thing preventing discharge — the old code held via the discharge schedule itself, so a failed companion write could not un-hold the battery.

Suggested fix: skip / defer the discharge write (or fall back to the legacy `ctrDis=1` empty-period representation) when the hold charge write did not succeed.

apps/predbat/alphaess.py
springfall2008, line 1621: **Review finding 4 (PLAUSIBLE, severity: medium) — the periodic path drops period 2 of a midnight-split hold.**

`hold_charge` carries raw start/end that may cross midnight (e.g. 23:00-25:00 from clock-skew offsets, a clamped write, or stale/manual entity values). `build_charge_payload` splits it via `split_window`/`_snapped_periods` and writes **both** periods, but the periodic path here does `(charge_start, charge_end), _ = ...` and keeps only period 1, then feeds `_snapped_periods` again inside the periodic builder — for a crossing window both snapped periods come back empty, so `charge_on` flips False and the chargeTimeList becomes the disabled filler.

Result for crossing windows: `gridChargeCycle: 0` + `ctrDisCycle: 0` (this line) — the entire hold is lost on the periodic API while the legacy pair still reports it, i.e. a regression vs the old unconditional `ctrDisCycle=1` hold, and the two payload builders diverge from each other.

Suggested fix: reuse the already-split periods from `build_charge_payload` (or have `_hold_charge_window` return pre-split periods) so both paths agree.

apps/predbat/tests/test_alphaess_control.py
```diff
@@ -1733,6 +1806,10 @@ def run_alphaess_control_tests(my_predbat):
         ("batusecap_is_reserve", test_alphaess_batusecap_is_the_reserve_outside_an_export_window),
         ("rate_zero_is_freeze", test_alphaess_rate_zero_is_freeze),
         ("both_rates_zero_still_holds", test_alphaess_both_rates_zero_still_holds_the_battery),
+        ("disabled_planned_hold_recovers_window", test_alphaess_disabled_planned_hold_recovers_the_charge_window),
+        ("restored_rate_releases_hold", test_alphaess_restored_rate_releases_a_disabled_planned_hold),
+        ("hold_profile_constant", test_alphaess_hold_profile_is_constant_while_soc_changes),
```
springfall2008, line 1869: **Test-coverage gaps (CLAUDE.md: unit tests must be added for all new code).** Missing cases for the new logic: (1) the real-charge escape firing OUTSIDE an active window (it currently has no `_window_active_now` guard, so a future charge window defeats an active car/iBoost hold); (2) a start>end wrap window reaching `_window_active_now` - the existing midnight tests feed "26:00" ends that production never writes (Predbat writes mod-24 and AlphaESSCloud clamps to 23:59), so the wrap branch is untested against realistic input; (3) a real enabled export surviving a `charge_window_hold` (currently the hold suppresses it); (4) the periodic second-period remainder for a split hold window; (5) the rewrite drops the old assertions that the discharge filler carries the reserve floor (batUseCap / dischargeTimeList chargeLimit = reserve), so that invariant is no longer pinned.

apps/predbat/alphaess.py
```diff
@@ -1020,6 +1020,84 @@ def _clamp_percent(value, low=0, high=100):
...
+        target has been reached, but deliberately leaves the start/end entities intact. The
+        times therefore retain the missing intent needed by this adapter. Hours above 24 are
+        accepted because Predbat represents a window crossing midnight as (for example)
+        23:00-25:00 before the AlphaESS boundary splits it.
+        """
+        start = hm_to_minutes(window.get("start"))
+        end = hm_to_minutes(window.get("end"))
+        if start == end:
+            return False
+        now = int(getattr(self.base, "minutes_now", 0)) % (24 * 60)
```
springfall2008: **Cleanup (reuse): bypasses the ComponentBase.minutes_now property.** `int(getattr(self.base, "minutes_now", 0))` is the only getattr-with-default access in the module; `AlphaESSAPI` subclasses `ComponentBase`, whose `minutes_now` property (component_base.py:150) is the documented accessor every other base delegation uses, and it already guarantees a 0-1439 value (making the `% (24 * 60)` here redundant). The silent 0 default means a wiring mistake or rename degrades into evaluating every window at midnight instead of failing loudly - a 01:00-05:00 planned hold would be missed while a 00:00-start window latches as a hold forever. Use `self.base.minutes_now` directly.

apps/predbat/alphaess.py
```diff
@@ -1020,6 +1020,84 @@ def _clamp_percent(value, low=0, high=100):
...
+        if not discharge_hold and not charge_window_hold:
+            return None
+
+        # Do not replace a genuine charge just because an EV/iBoost hold is present too.
+        # The active grid charge already prevents discharge and its requested target/rate
+        # must survive unchanged.
+        if bool(charge.get("enable")) and charge_target > current_soc and charge_power > 0:
+            return dict(charge)
+
+        if charge_window_hold:
```
springfall2008, line 1099: **Review finding 5 (PLAUSIBLE, severity: medium) — a detected hold whose times collapse writes NO hold at all.**

The `charge_window_hold` branch copies `start`/`end` verbatim, but the builders then snap them inward via `_snapped_periods`. Two ways this collapses:

1. **Inverted pair** (start > end): `window_is_empty(end) <= window_is_empty(start)` semantics make `_snapped_periods` return two empty periods -> `gridCharge: 0` on the legacy pair while `ctrDis: 0` — neither a hold profile nor a discharge hold; the battery discharges freely. Trigger: non-default `inverter_clock_skew_start/end`, a clamped write, or stale/manual start-end entities (Predbat itself will not write inverted times here — can_span_midnight is False and execute.py clamps to 23:59 — but the entities are user/stale-editable).
2. **Sub-15-minute window** (e.g. 01:05-01:10): active per `_window_active_now` on raw times, but snapping collapses it to a single point -> both periods disabled, same net result.

Either way this is a regression vs the old `ctrDis=1` hold, which did not depend on the times being parseable. Suggested fix: if the snapped periods come back empty, fall back to the full-day 00:00-23:45 synthetic profile (the same shape used for generic demand-mode holds).

apps/predbat/alphaess.py
```diff
@@ -1020,6 +1020,84 @@ def _clamp_percent(value, low=0, high=100):
...
+            start = charge.get("start", "00:00:00")
```
springfall2008: **Cleanup (reuse): day-boundary and hold-profile literals re-derived inline.** "00:00:00"/"23:45:00" appear here, at 1092-1093 and in the returned dict, while alphaess_const.py already owns ALPHAESS_TIME_DISABLED ("00:00") and ALPHAESS_TIME_MAX ("23:45") - and hhmmss_to_hhmm normalises HH:MM:SS down to HH:mm anyway, so the ":SS" suffix buys nothing. If the documented 23:45 device maximum ever changes, split windows and snapped periods follow the constant but the synthetic hold keeps the hard-coded end, written every cycle as a full replacement payload. The 10%/100W constants are likewise inline and duplicated in the periodic builder; every other AlphaESS tuning value lives in alphaess_const.py or the alphaess_* config block (config.py), where a per-system hold target/power (e.g. alphaess_hold_soc_percent) would fit.

apps/predbat/alphaess.py
```diff
@@ -1020,6 +1020,84 @@ def _clamp_percent(value, low=0, high=100):
...
+            end = charge.get("end", "23:45:00")
+        else:
+            # Generic demand-mode holds carry no end time. A stable daily profile avoids a
+            # rolling boundary that would change the payload and spend an AlphaESS cloud
+            # write every 15 minutes; reconciliation removes it as soon as the hold signal
+            # clears on the next Predbat cycle.
+            start = "00:00:00"
```
springfall2008: **Review finding 6 (CONFIRMED mechanically, severity: medium) — the hold representation flips at the window boundary and burns the 1-write/24h budget.**

While the stale charge window is active the profile is (window times, 10, 100); once it ends (`charge_window_hold` goes False while `discharge_hold` persists) it becomes (00:00-23:45, 10, 100). Each boundary changes `timeChaf1/timeChae1` and the periodic chargeTimeList, so `applied_payload` differs and both endpoints get rewritten — twice per boundary, twice a day, per held serial — against endpoints this file documents as writable once per 24 hours.

Failure: a multi-day EV/iBoost hold consumes ~4 writes/day; when the follow-up write is pacing-held the inverter is stranded on whichever profile lost the race, and the 00:00-23:45@10% profile also destroys any user-programmed charge window until the next eligible tick.

Suggested fix: prefer one stable shape for the duration of the hold (e.g. always the full-day profile once a hold is detected, keeping the windowed shape only for the genuine-charge escape).

apps/predbat/alphaess.py
```diff
@@ -1020,6 +1020,84 @@ def _clamp_percent(value, low=0, high=100):
         except (TypeError, ValueError):
             return low
 
+    def _window_active_now(self, window):
```
springfall2008, line 1033: **Cleanup (reuse): third hand-rolled window-overlap predicate.** `_window_active_now` re-implements the same containment logic as `TeslemetryAPI.in_window` (teslemetry.py:512) and `solis.time_in_window` (solis.py:217), with subtly different edge semantics (solis is end-inclusive, teslemetry and this one end-exclusive, and only this one handles an end > 24:00). There is no shared util for this anywhere, which is the real reuse gap: promoting one helper (Teslemetry.in_window is the closest match) into a shared util would stop a boundary-condition fix (half-open vs closed end, wrap handling) landing in one adapter only and the three disagreeing about whether now == end is inside the window.

apps/predbat/alphaess.py
springfall2008, line 1077: **Review finding 7 (PLAUSIBLE, severity: medium) — hold-vs-real-charge decisions ride on live, quantised percent SOC.**

`current_soc` comes from `device_values[sn]["soc"]`, a live **percent** (cbat) refreshed per poll and never persisted, and `charge_target` is also a percent. Two consequences:

- A stale-high SOC reading (or percent quantisation) makes an explicitly **enabled** charge window with a target above the true SOC satisfy `charge_target <= current_soc`, so a genuine charge (e.g. target just raised 40%->50%, battery really at 45%, last poll said 55%) is converted into the fixed 10%/100W hold and the charge window is missed until telemetry catches up and the payload flips back — two extra writes on a 1-write/24h budget.
- Near the equality boundary the same comparison flaps the payload between hold and real-charge shapes on every poll (mitigated by the default 300s `min_write_interval`, but the flapping still reaches the wire).

Suggested fix: hysteresis on the comparison (e.g. require `current_soc >= charge_target + margin` before treating an enabled window as a hold), and/or use the target Predbat just computed rather than device telemetry for the enable-side decision.

apps/predbat/alphaess.py
```diff
@@ -1073,7 +1151,7 @@ def build_charge_payload(self, sn, schedule):
         A FULL REPLACEMENT, not a patch: all seven fields are always present, because the
         endpoint silently resets anything omitted.
         """
-        window = schedule.get("charge", {}) or {}
+        window = self._hold_charge_window(sn, schedule) or (schedule.get("charge", {}) or {})
```
springfall2008: **Cleanup: stale comment contradicts the new logic.** The retained block just below - "charge_rate zero is Predbat signalling freeze charge / no cross-charging ... a zeroed rate is the only signal available and it overrides the window outright" - describes the OLD translation. Under this PR a zero charge rate no longer overrides the window when a hold is detected: `_hold_charge_window` converts that same zero into an ENABLED 100W profile. The two comments in this function now assert opposite semantics for the same input; a maintainer reading top-down would conclude zero rate => gridCharge 0 and distrust or "fix" the hold path. Rewrite to state the new precedence (zero rate disables charging only when no hold profile was derived).

apps/predbat/alphaess.py
```diff
@@ -1020,6 +1020,84 @@ def _clamp_percent(value, low=0, high=100):
...
+        # hold regardless of the valid non-zero power it carries.
+        charge_window_hold = charge_target > 0 and charge_target <= current_soc and self._window_active_now(charge) and (bool(charge.get("enable")) or charge_power <= 0)
+        if not discharge_hold and not charge_window_hold:
+            return None
+
+        # Do not replace a genuine charge just because an EV/iBoost hold is present too.
+        # The active grid charge already prevents discharge and its requested target/rate
+        # must survive unchanged.
+        if bool(charge.get("enable")) and charge_target > current_soc and charge_power > 0:
+            return dict(charge)
```
springfall2008, line 1097: **Review finding 8 (PLAUSIBLE, severity: low-medium) — after a restart the recovery branch is dead until telemetry lands, causing an un-hold then re-hold.**

`device_values` is repopulated per poll and never persisted, so immediately after an app restart `current_soc` defaults to 0.0 via `_as_float(..., 0.0)`. That makes `charge_target <= current_soc` false and the escape here false too — a serial that was held pre-restart briefly gets un-held payloads (`ctrDis 0` / `ctrDisCycle 0` with export suppression off) until the first SOC poll lands, then flips back to the hold profile. Narrow (alphaess.py:575 refuses to drive serials with no SOC at all, so most affected serials are already excluded), but on a 1-write/24h budget each flip is expensive.

Suggested fix: persist the last SOC per serial through the Storage component, or treat a missing SOC as "keep the previous payload" rather than 0.0.

apps/predbat/alphaess.py
```diff
@@ -1531,14 +1603,10 @@ def build_periodic_payload(self, sn, schedule):
...
-        # at all. Decided before, and independently of, the window/enable check - Predbat
-        # reaches export_rate == 0 through ordinary, reachable combinations (see
-        # build_discharge_payload's comment), and that is an explicit hold, not the absence of
-        # a plan. Getting this wrong lets the battery discharge into the EV or iBoost load
-        # against the hold Predbat asked for; ctrDisCycle 0 here is DEMAND MODE, not a hold.
-        export_hold = export_rate <= 0
+        # While the hold charge profile is active it is the whole control action. Suppress
+        # any export schedule and leave discharge time control in demand mode; AlphaESS
+        # ignores the former ctrDisCycle=1 plus empty-period representation.
+        export_hold = hold_charge is not None
```
springfall2008: **Cleanup (naming): `export_hold` no longer means an export hold.** It is now `hold_charge is not None` - the charge profile doing the holding, with discharge explicitly NOT held - whereas the name (and the removed comment above it) meant the old export_rate <= 0 freeze. A future reader, or a merge that re-adds an export-hold concept, could reinstate ctrDisCycle=1 under this variable, reintroducing the empty-period write AlphaESS ignores (the exact GH#4725 bug this PR fixes). Rename to `hold_profile_active` or inline the check so the name cannot re-seed the discarded mental model.

apps/predbat/alphaess.py
```diff
@@ -1020,6 +1020,84 @@ def _clamp_percent(value, low=0, high=100):
...
+        """Return whether a stored window covers Predbat's current minute.
+
+        The generic hold-charge path disables the charge-enable switch after deciding the
+        target has been reached, but deliberately leaves the start/end entities intact. The
+        times therefore retain the missing intent needed by this adapter. Hours above 24 are
+        accepted because Predbat represents a window crossing midnight as (for example)
+        23:00-25:00 before the AlphaESS boundary splits it.
+        """
+        start = hm_to_minutes(window.get("start"))
+        end = hm_to_minutes(window.get("end"))
```
springfall2008, line 1043: **Review finding 9 (PLAUSIBLE, severity: low) — an unparseable end time makes the hold look active all day while the written payload collapses.**

`hm_to_minutes` returns 0 for unusable input, so a missing/garbled `end` yields `end=0 < start`, which falls into the wrap branch below and makes `_window_active_now` return True for essentially every minute of the day. The hold is then detected (and discharge suppressed via `ctrDis: 0`), but the copied `start/end` collapse to empty periods in `_snapped_periods`, so no charge profile actually gets enabled — the battery is left in demand mode with no hold, the worst combination of the failure modes above.

Suggested fix: treat an unparseable/zero end as "window unknown" and fall back to the full-day synthetic profile rather than the wrap branch.

apps/predbat/alphaess.py
```diff
@@ -1020,6 +1020,84 @@ def _clamp_percent(value, low=0, high=100):
...
+        A real charge target above SOC is kept verbatim even if another feature also asks for
+        a discharge hold; charging already prevents the battery feeding the load.
+
+        Returns a copied, enabled charge window while a hold is requested, otherwise None.
+        """
+        charge = schedule.get("charge", {}) or {}
+        export = schedule.get("export", {}) or {}
+        current_soc = self._as_float(self.device_values.get(sn, {}).get("soc"), 0.0)
+        charge_target = self._as_float(charge.get("soc"), 0.0)
+        charge_power = int(self._as_float(charge.get("power"), 0.0))
```
springfall2008: **Review finding 10 (minor, CONFIRMED mechanically) — int() truncation makes the power discriminator disagree with the builders.**

`charge_power = int(self._as_float(charge.get("power"), 0.0))` truncates: a computed charge rate of 0<p<1 W becomes 0 here, so `(bool(enable) or charge_power <= 0)` treats the window as hold-eligible, while `build_charge_payload`/`build_periodic_payload` compare the raw float `rate <= 0` and treat the same value as a real (enabled, non-zero) charge. The two paths can therefore disagree about whether a window is a hold.

Suggested fix: drop the `int()` and compare the float directly (matching the builders), or round consistently in both places.

apps/predbat/alphaess.py
```diff
@@ -1020,6 +1020,84 @@ def _clamp_percent(value, low=0, high=100):
...
+        if start == end:
+            return False
+        now = int(getattr(self.base, "minutes_now", 0)) % (24 * 60)
+        if end > 24 * 60 and now < start:
+            now += 24 * 60
+        if start < end:
+            return start <= now < end
+        return now >= start or now < end
+
+    def _hold_charge_window(self, sn, schedule):
```
springfall2008, line 1053: **Altitude: hold intent is reconstructed from leftover entity state instead of an explicit signal.** This helper infers "hold" from (a) which fields execute_plan happened to leave untouched when it called `disable_charge_window()`, (b) live device SOC and (c) the wall clock. Core already makes exactly this decision (execute.py:305-342) and already has a sentinel convention for the export direction (EXPORT_LIMIT_FREEZE=99 in export soc, which deye/sunsynk read as freeze). Writing an explicit hold marker into the shared schedule/entities would let this adapter - and future ones - test a value instead of reverse-engineering which fields a foreign function skips; as written, any refactor of disable_charge_window (e.g. clearing times via adjust_idle_time like other inverter types, or zeroing the soc entity) silently breaks or permanently latches the inference, with the failure mode being either the GH#4725 bug returning or a 10%/100W forced charge all day.

apps/predbat/alphaess.py
springfall2008: **Review finding 11 (cleanup) — `_hold_charge_window` is evaluated twice per apply, straddling `api_delay`.**

On the legacy path `apply_settings` calls `build_charge_payload` (which evaluates `_hold_charge_window`), sleeps `api_delay`, then calls `build_discharge_payload` (which evaluates it again here). The second evaluation re-reads `device_values` SOC and `self.base.minutes_now` after the delay, so the two payloads can be built from **different** hold decisions — e.g. a window boundary crossing inside the delay yields an enabled charge profile paired with an un-suppressed discharge write (or vice versa). It is also wasted work on a hot path.

Suggested fix: evaluate `_hold_charge_window` once in `apply_settings` and pass the result down to both builders (an optional `hold_charge=None` parameter keeps the existing signatures usable).

apps/predbat/alphaess.py
springfall2008: **Review finding 12 (cleanup) — `getattr(self.base, "minutes_now", 0)` bypasses the existing property and defaults silently to midnight.**

ComponentBase already exposes a `minutes_now` property (component_base.py:150) that resolves the base time; this getattr shadows that path with a raw attribute lookup and silently falls back to 0 (midnight) if the attribute is missing. A silent midnight default makes every window look "active" early in the day rather than surfacing the wiring problem.

Suggested fix: use the `minutes_now` property (it raises loudly if the base is not wired) or log once when falling back.

apps/predbat/alphaess.py
springfall2008: **Review finding 13 (cleanup) — hardcoded time literals instead of the existing constants.**

`"00:00:00"` / `"23:45:00"` here (and in the fallback return below) restate `ALPHAESS_TIME_DISABLED` and `ALPHAESS_TIME_MAX`, which the rest of this module uses for exactly these boundary values (note the constants are HH:mm while these literals are HH:mm:ss — the mixed formats are exactly why they drifted apart). If `ALPHAESS_TIME_MAX` ever changes (e.g. to 23:59), the synthetic hold profile silently disagrees with the rest of the adapter.

Suggested fix: use the constants, converting format once via `hhmmss_to_hhmm` if the seconds form is required.

apps/predbat/alphaess.py
```diff
@@ -1073,7 +1151,7 @@ def build_charge_payload(self, sn, schedule):
         A FULL REPLACEMENT, not a patch: all seven fields are always present, because the
         endpoint silently resets anything omitted.
         """
-        window = schedule.get("charge", {}) or {}
+        window = self._hold_charge_window(sn, schedule) or (schedule.get("charge", {}) or {})
         enabled = bool(window.get("enable"))
         # charge_rate zero is Predbat signalling freeze charge / no cross-charging. There
         # is no pause endpoint, so a zeroed rate is the only signal available and it
```
springfall2008: **Review finding 14 (cleanup) — stale comment plus a now-inverted variable name.**

Two naming/doc issues introduced by the indirection:

1. The comment two lines below (`charge_rate zero is Predbat signalling freeze charge / no cross-charging ... a zeroed rate is the only signal available`) is no longer true for holds: the hold is now realised by `_hold_charge_window` overriding the window, not by the zeroed rate being passed through. It should say the zero-rate path only applies when no hold was detected.
2. In `build_periodic_payload` (~line 1608) `export_hold = hold_charge is not None` now means "a hold charge profile exists", which can be true with `export_power > 0` (finding 1) — the name asserts the opposite of what it tests and `ctrDisCycle` no longer follows it. Something like `hold_profile_present` would describe what it actually checks.

Suggested fix: reword both comments and rename the flag together with finding 1, since the semantics are what is changing.

apps/predbat/tests/test_alphaess_control.py
```diff
@@ -343,26 +351,102 @@ def test_alphaess_both_rates_zero_still_holds_the_battery():
...
+
+    Recover that active window and use the fixed hold target instead of turning a
+    target-reached charge back into demand mode.
+    """
+    failed = False
+    client = _client()
+    client.base.minutes_now = 2 * 60
+    schedule = _schedule(reserve=41, charge={"enable": False, "soc": 40, "power": 0, "start": "01:00:00", "end": "05:00:00"})
+    charge = client.build_charge_payload("AL70", schedule)
+    discharge = client.build_discharge_payload("AL70", schedule)
```
springfall2008, line 385: **Review finding 15 (cleanup / coverage) — `_window_active_now` and the periodic hold path have no test coverage.**

The new helper is only ever exercised incidentally (the new tests set `minutes_now = 2*60` inside a window), and CLAUDE.md says unit tests must be added for all new code. Gaps:

- `_window_active_now` itself: no test pins its wrap branches (`start < end` vs the `now >= start or now < end` arm, the `end > 24*60` shift, `start == end` short-circuit, and the `hm_to_minutes` returns 0 fallback that finding 9 exploits).
- `build_periodic_payload` with a recovered/bounded hold: the new tests assert the legacy pair; nothing checks the periodic chargeTimeList / gridChargeCycle / ctrDisCycle for a hold with **bounded** (non-full-day) times, which is where finding 4 (period 2 of a midnight split) hides.
- The rewritten tests also dropped the old assertions on `timeDisf1/timeDise1/timeDisf2/timeDise2` and the `batUseCap` reserve behaviour, so the disabled-periods shape of the discharge payload is now unasserted.

Suggested fix: add a small direct test table for `_window_active_now` plus one periodic-payload test for a bounded hold with a crossing window.

dragon2611:
I'm having a look, I think some of these are inherent to the way the alphaESS API works so cannot be fixed but there's probably some edge cases and behavour I can tidy up

dragon2611:
Thanks for the detailed review. I’ve gone through the findings and pushed changes for the cases that can be improved within the limitations of the AlphaESS API.

## Changes made

- A configured future charge window is no longer treated as satisfying a hold now. The genuine-charge escape now requires the charge window to be active.
- Planned hold recovery now matches production behaviour. Predbat restores the normal/max charge power after disabling a target-reached window, so zero charge power was not a valid signature.
- A recovered planned hold is released when its original charge window ends, preventing stale target/time entities from extending it.
- The hold decision is calculated once per apply and passed to both payload builders.
- On the legacy two-endpoint path, `ctrDis=0` is not written unless the companion charge hold profile was successfully applied. If the charge write is rejected or pacing-held, the discharge update is deferred.
- The hold target, power and time boundaries now use named constants, and the related comments and variable names have been clarified.
- Tests were added for:
  - future charge windows during a current EV/iBoost hold;
  - planned holds with production’s normal/max charge power;
  - release at the end of the planned window;
  - rejected legacy hold transitions.
- AlphaESS schedule times continue to be snapped inward to the required 15-minute grid (`:00`, `:15`, `:30`, `:45`). A window that collapses after snapping is disabled rather than sent as an undocumented wrap-around.

## Why the hold uses 10% and 100 W

The choice of 100 W is **not** based on evidence that AlphaESS explicitly rejects a zero value. I do not know whether an explicit zero would be rejected or interpreted as the inverter’s full/default power.

Predbat’s periodic payload builder currently omits `chargePower` when it is non-positive. That would leave the resulting charge power entirely to undocumented AlphaESS behaviour. A fixed 100 W is therefore a deliberately small, positive and deterministic setpoint, rather than relying on an omitted or zero value with unknown semantics.

The 10% target is used because 10% is the minimum charge target accepted by the API.

If the battery SOC is below 10%, the hold may consequently charge it towards 10% at 100 W. There is no public pause/hold endpoint that avoids this trade-off: hardware testing established that enabling discharge scheduling with no periods is ignored.

## Discharge scheduling versus forced export

On the periodic scheduling path, AlphaESS’s “discharge” control does not force export to the grid. Hardware testing shows that it only permits the battery to discharge to local load during the configured period.

Actual forced export appears to require the private VPP API used by providers.

The legacy endpoint has no discharge-power field at all. I cannot test that path on my system, so I am not claiming that legacy mode can never export; its forced-export behaviour remains unverified and it cannot request a particular export wattage.

Consequently, findings based on preserving a genuine periodic “forced export” assume a capability that this public scheduling API does not provide.

## Write frequency and atomicity

The hold workaround consumes the same charge schedule used for genuine grid charging, and the AlphaESS endpoints use full-replacement payloads.

Some writes at transitions are therefore unavoidable: Predbat cannot simultaneously install an all-day hold and preserve a different future charge profile in the same schedule.

The legacy API also separates charge and discharge into two endpoints, so the transition cannot be genuinely atomic. The new sequencing makes entry into a hold fail safe, but it cannot turn two independent AlphaESS requests into one atomic operation.

## SOC and unusual window states

SOC is only needed to recover a target-reached planned charge hold. Explicit EV/iBoost/freeze holds still arrive through Predbat’s zero-discharge-power signal and do not depend on that SOC comparison.

Equality is intentional because AlphaESS holds once the charge target has been reached.

Predbat configures AlphaESS with `can_span_midnight=False`, clamps normal crossing plan windows, and snaps them to the API’s 15-minute grid. The `26:00` examples are therefore not representative of normal production entity state. Manually edited, malformed or heavily clock-skewed windows remain defensive-input cases rather than normal plan output.

The targeted AlphaESS tests, full quick regression suite and all pre-commit checks now pass.

## Alternative hold approach considered

The only other approach I can see working with the available public APIs would be to configure a discharge-permission window for a time that is not now. Because AlphaESS prevents battery discharge outside the configured window, that would indirectly hold the battery at the current time.

I do not favour this approach because the discharge schedule is persistent. Predbat would need to reliably clear it afterwards and continually move the window so that the current time never crosses into it. If cleanup or rescheduling were delayed or rejected by the API, the inverter could enter the discharge-permitted window and the hold would silently stop working.

The low-target charge profile is therefore the more deterministic workaround, despite its documented limitations.

dragon2611:
Yes, that comment was mostly me guiding the AI towards what I wanted to say. I’m not claiming this is the ideal implementation, because it absolutely isn’t, but I think it’s better than what we had before, where a hold seemingly did nothing.

I’d welcome suggestions for a cleaner approach, particularly if there’s another way to achieve a reliable hold using the public AlphaESS APIs.

I'll be running my fork for a while to do some testing against real hardware, it's also possible different regions behave differently i know some have a "self consumption plus" but I can only test with the device I have.

dragon2611:
Also should probably document if we haven't already that it's not currently possible to prevent solar PV charging (or at least I haven't seen a way to do so) the exposed controls seem to only stop discharge to load and control charging from the grid.

As an aside 100w may need to change it's accepted but when I forced a higher soc it seemed to be flipping between charging and discharge it might actually be too low a number to get sensible behaviour or that might just be the sensor/app updates

PR #4871, opened 2026-08-30
fix(fetch): make the cloud/load divergence model actually modulate

### LINKED ISSUE #4870: Cloud/load divergence model is inert: operator precedence makes it a no-op on every array except PV10
_Written by Claude, posted on behalf of @chalfontchubby._

**Describe the bug**

The cloud/load divergence model in `step_data_history()` does nothing on every array except one. `switch.predbat_metric_cloud_enable` only ever perturbs the PV10 scenario and never the central PV forecast, and `metric_load_divergence` — a documented, user-facing setting — has no effect at all.

**Cause**

`apps/predbat/fetch.py`:

```python
# Simple divergence model keeps the same total but brings PV/Load up and down every 5 minutes
if cloud_factor and cloud_factor > 0:
    for minute in range(0, self.forecast_minutes, step):
        cloud_on = (int((minute + self.minutes_now) / 5) + 1 if flip else 0) % 2
        if cloud_on > 0:
            cloud_diff += min(values[minute] * cloud_factor, values.get(minute + 5, 0) * cloud_factor)
            values[minute] += cloud_diff
        else:
            subtract = min(cloud_diff, values[minute])
            values[minute] -= subtract
            cloud_diff = 0
```

A conditional expression binds looser than `+`, so this parses as `((int(...) + 1) if flip else 0) % 2`. With `flip=False` that is a constant `0`:

```
flip=False  [0, 0, 0, 0, 0, 0]
flip=True   [1, 0, 1, 0, 1, 0]
```

Every iteration then takes the `else` branch and subtracts a `cloud_diff` that never becomes non-zero, so the series is returned unchanged.

`flip=True` is passed by exactly one caller:

| array | caller | modulated? |
|---|---|---|
| `pv_forecast_minute_step` | `plan.py:1475` | **no** |
| `pv_forecast_minute10_step` | `plan.py:1476`, `flip=True` | yes |
| `pv_forecast_minute90_step` | `plan.py:1478` | **no** |
| `load_minutes_step` / `_step10` / `_step90` | `plan.py:1437/1450/1471` | **no** |

The intent looks clear from `flip` itself: it only means anything as an *antiphase* to a modulated base, so it is the offset that should be conditional — `+ (1 if flip else 0)` — not the whole expression. The comment above the loop says "brings PV/Load up and down every 5 minutes", which is not what happens.

**Why it has gone unnoticed**

It still shifts plans, because PV10 feeds the weighted metric via `pv_metric10_weight` — so the setting appears to do *something*. And nothing in the test suite asserted that either knob had an effect; every reference to `cloud_factor` under `apps/predbat/tests/` is a caller replicating the plan setup.

No lint rule catches it either. I checked all 918 rules in ruff 0.11.4 with `--select ALL` and nothing flags ternary/arithmetic precedence — the closest, `RUF021`, covers only `and`/`or` chaining.

Unchanged since a2789970 (#1655), which was a file split, so it likely predates that.

**Impact**

Both settings default on, so fixing this changes plans for everyone on defaults. In the two golden debug cases, `agile1` moves only trivially (five charge targets by ~4Wh) but `pre_saving1` moves genuinely — a charge window shifts two hours later, its limit drops 5.02 → 3.02kWh, and an export target moves 10 → 15%. All twenty random benchmark scenarios move.

That makes this a behaviour change wanting a maintainer's judgement rather than a quiet fix, which is why it is raised separately rather than folded into anything else.

**Related**

Came out of #4840. The triage bot there suggested @gcoan enable `metric_cloud_enable`, and he replied that he had turned it off deliberately because it made Predbat "chop and change the plan across a half hour slot". Worth knowing he was judging a feature that only ever perturbed the 10% scenario.

**Secondary observation**

The model does not preserve totals exactly, contrary to its comment: what it adds to a step it can only take back off the next one if that step is large enough, so a decaying series drifts. Harmless in itself, but it means the nominal and PV10 load totals no longer sit in the pure `load_scaling` ratio once divergence is live.

**Predbat version**

Present on current main (v8.54.2).

chalfontchubby:
_Written by Claude, posted on behalf of @chalfontchubby._

Fixes #4870

`step_data_history()`'s divergence model picks its phase with

```python
cloud_on = (int((minute + self.minutes_now) / 5) + 1 if flip else 0) % 2
```

A conditional expression binds looser than `+`, so that is `((int(...) + 1) if flip else 0) % 2` — a constant `0` whenever `flip` is False. Every such iteration takes the `else` branch and subtracts a `cloud_diff` that never becomes non-zero, so the series comes back untouched.

`flip=True` is passed by exactly one caller, `pv_forecast_minute10_step`. So `switch.predbat_metric_cloud_enable` only ever perturbed the PV10 scenario, never the central PV forecast, and `metric_load_divergence` did nothing whatsoever. `flip` only means anything as an antiphase to a modulated base, which gives away the intent: it is the offset that should be conditional, not the whole expression.

## This changes plans on the defaults

Both settings default on, so this is a behaviour change for everyone rather than a quiet fix — flagging it prominently because it wants your judgement, not just a review of the diff.

| baseline | movement |
|---|---|
| `predbat_debug_agile1` | trivial — five charge targets by ~4Wh |
| `predbat_debug_pre_saving1` | **real** — a charge window shifts two hours later, its limit drops 5.02 → 3.02kWh, an export target moves 10 → 15% |
| random benchmark scenarios | all 20 move |

Baselines regenerated. `runtime_s` in `random_results.json` is left at the previous values so the diff shows only genuine changes rather than 20 lines of timing noise.

## Testing

New `cloud_modulation` test module asserting the model actually modulates without `flip`, that `flip` produces antiphase, that raises and lowers alternate step by step, and that the total is roughly preserved. Confirmed it fails without the fix:

```
ERROR: cloud_factor made no difference without flip - the divergence model is inert
       for the central PV and load forecasts
ERROR: expected both series to raise some steps, got 0 plain and 144 flipped
```

There was no such coverage before — every reference to `cloud_factor` under `apps/predbat/tests/` was a caller replicating the plan setup, so nothing asserted either knob had an effect and the suite passed either way.

`test_pv90`'s clamp check compared the nominal and PV10 load totals against the pure `load_scaling` ratio. Those arrays take different divergence factors (PV10 is +0.5) and the model does not redistribute exactly — what it adds to a step it can only take back off the next one if that step is large enough — so the ratio no longer holds once divergence is live. Disabled there via `metric_load_divergence_enable`, since that test is about the clamp, not divergence.

`./run_all --quick` and `./run_all --test debug_cases` both pass.

## Notes

Unchanged since a2789970 (#1655), a file split, so it likely predates that. No lint rule catches this class — all 918 rules in ruff 0.11.4 under `--select ALL` are silent on it, and the nearest, `RUF021`, covers only `and`/`or` chaining.

The total-drift behaviour noted above contradicts the loop's own "keeps the same total" comment. Left alone here as pre-existing and separable, but worth knowing it is now live for the load arrays for the first time.


springfall2008 (commented):
## Review: the precedence fix itself is correct — but the activation it ships has substantive defects

The one-line fix is right (`(int(...) + 1) if flip else 0` vs `(int(...) + (1 if flip else 0))` — confirmed against Python precedence), and the new unit test pins it well (non-inertness, antiphase, strict alternation). Conventions, docstrings, cspell and test registration are all clean.

The real blast radius is not the fix, though — it is that the previously-dead divergence block (`fetch.py:200-210`) now runs on **all five central/downside arrays** (load p50/p10/p90, PV central and p90) for every user with `metric_cloud_enable` / `metric_load_divergence_enable`, both default-True (`config.py:571-583`). Findings from the review, most severe first:

**Correctness of the activated model**

1. **Non-Solcast users get a ±50% sawtooth on their central PV forecast.** Only Solcast publishes a PV10 series; with `pv_forecast_minute10` empty, `pv_total10 = 0`, so `get_cloud_factor` (`fetch.py:62-65`) computes `pv_diff/pv_total = 1.0` → `metric_cloud_coverage` pinned at exactly **0.5 regardless of weather**. Pre-PR that factor was inert for non-flip callers; post-PR `plan.py:1475/1478` modulate the central PV array with it — buckets alternate 1.5×/0.5×.

2. **`cloud_factor` 1.0 zeroes every off-bucket, discarding the baselines the same function just applied.** `load_minutes_step10` uses `min(divergence + 0.5, 1.0)` (`plan.py:1450`) and `get_load_divergence` caps at 1.0 — a household with divergence ≥ 0.5 (EV/shower on the history days) gets cf10 = 1.0: each on-bucket borrows the full neighbouring value and the off-bucket subtract `min(cloud_diff, values[minute])` drives it to exactly **0** — after `base_load` and the `load_baseline` floor (`fetch.py:196-197`) were enforced. The planner sees a house drawing nothing for 5 minutes at a time and can schedule discharge/export into those fake troughs. Off-buckets also put load10 *below* nominal (central 1.3v/0.7v vs load10 1.98v/0.22v at divergence 0.3), inverting the pointwise ordering the `plan.py:1408-1417` clamp comment claims to guarantee (totals still order correctly; only pointwise breaks).

3. **The model is not total-preserving.** (a) The loop runs `range(0, forecast_minutes, step)` while `values` is built to `forecast_minutes + plan_interval_minutes` (`fetch.py:157`): when the last in-loop slot lands on an "on" parity its borrow is never repaid — wall-clock-parity-dependent net energy creation, up to one bucket per plan. (b) A clamped repay (`min(cloud_diff, values[minute])`) destroys the remainder via the unconditional `cloud_diff = 0`. The new test's 1% tolerance cannot see (a) by construction (one bucket of ~288 ≈ 0.35%), so the suite passes while the invariant it names is false.

4. **Phase inverts every cycle → plan flapping.** The phase is `int((minute + self.minutes_now)/5) % 2` and `minutes_now` advances 5 per `RUN_EVERY`, so every bucket flips raised/lowered between consecutive plan calcs; the same forecast scores differently cycle-to-cycle, interacting with the 2.0p `metric_min_improvement_plan` retention gate (the GH#4840 pattern) and the known `calculate_plan_every`(10) vs `RUN_EVERY`(5) double-recompute.

5. **The phase reads `self.minutes_now`, shadowing the function's own `minutes_now` parameter** — every other minute computation in `step_data_history` keys off the parameter. Latent today only because every modulated caller happens to pass `self.minutes_now`; `output.py:3038-3040` already calls it with `minutes_now=0` (safe only because it omits `cloud_factor`).

6. **Hardcoded 5-minute grid vs the `step` parameter.** `int(.../5)` and `values.get(minute + 5, 0)` assume step == 5. A future `step=10` caller gets constant parity (all-on: unbounded cumulative inflation; all-off: silent no-op) and a `.get` that always misses. The signature advertises a parameter the block ignores.

7. **Central PV and load modulate in phase**, so the adverse combination the model exists to probe (PV below forecast while load above) is structurally unreachable — only pv10's `flip=True` inverts, and no load array is paired antiphase with anything.

8. **Downstream mirrors were not updated:** `annual.py:1087-1092` passes `cloud_factor=None` while its docstring claims to mirror `calculate_plan` (annual-vs-live savings diverge); `load_today_comparison` (`output.py:2663-2700`) replicates the scalings but not the modulation, so `today_remaining` disagrees with the plan's load by the modulation delta (a #4496-class report waiting to recur); `predbat.load_power`/`grid_power` are single-step samples of the now-sawtoothing arrays (`prediction.py:1366-1368`) so dashboards go square-wave; `test_single_debug.py` aliases p90 slots to the central arrays and omits the 90-array divergence factors, so replay metric diffs include artificial p90 differences.

**Tests and goldens**

9. The pv90 clamp change removes the suite's only tight-total assertion *and* its only default-on coverage: with the fix applied and divergence left at its defaults, the test **fails** (verified empirically on this tree: pv10 load total 32.3741 vs expected 32.3190). Disabling the knob makes it pass, but a future modulation regression then has no failing test anywhere in the suite.

10. The regenerated goldens enshrine real plan changes (pre_saving1 shifts two charge windows and a limit 5.02 → 3.02 kWh; agile1 drifts all five charge targets). That is the honest consequence of the activation, but it means `debug_cases` can no longer catch modulation regressions in those windows.

None of this argues against the precedence fix — it argues for either (a) fixing the activated model's conservation/parity/floor interactions before the default-on activation lands, or (b) explicitly gating the activation and following with the model fixes. Findings 1 and 2 are the ones I would block on.

apps/predbat/fetch.py
```diff
@@ -199,7 +199,12 @@ def step_data_history(
         # Simple divergence model keeps the same total but brings PV/Load up and down every 5 minutes
```
springfall2008: **Stale invariant comment.** "keeps the same total" is not true of the code below it: the repay on an off-slot is clamped to `min(cloud_diff, values[minute])` and the bank is then zeroed, so anything that could not be repaid is created net; and when the last slot of line 201's `range` (which stops at `forecast_minutes`) lands on an "on" parity, its borrow is never repaid at all. With this block now live for all five arrays, that is a wall-clock-parity-dependent net energy creation of up to one 5-minute bucket per plan. The PR's own tests acknowledge it (the 1% tolerance, "does not redistribute exactly" in test_pv90.py). Please either reword the comment or make the model actually conserve: keep a pristine copy of the pre-modulation values, repay exactly, and zero the residual at the horizon.

apps/predbat/fetch.py
```diff
@@ -199,7 +199,12 @@ def step_data_history(
...
         if cloud_factor and cloud_factor > 0:
             for minute in range(0, self.forecast_minutes, step):
-                cloud_on = (int((minute + self.minutes_now) / 5) + 1 if flip else 0) % 2
+                # The inner parentheses matter: a conditional expression binds looser than "+", so
+                # "int(...) + 1 if flip else 0" is "(int(...) + 1) if flip else 0" - which is a
+                # constant 0 whenever flip is False, leaving every non-flipped caller unmodulated.
+                # flip only means anything as an antiphase to a modulated base, so it is the offset
+                # that is conditional, not the whole expression.
+                cloud_on = (int((minute + self.minutes_now) / 5) + (1 if flip else 0)) % 2
```
springfall2008: `self.minutes_now` shadows the function's own `minutes_now` parameter — every other minute computation in this function (`minute_absolute`, the forward reads, the `load_adjust`/`load_baseline` lookups) keys off the parameter. It is latent today only because every modulated caller happens to pass `self.minutes_now`; `output.py:3038-3040` already calls this with `minutes_now=0` (safe only because it omits `cloud_factor`). Should be `int((minute + minutes_now) / 5)` — that also makes the output a pure function of the inputs, which is what the replay/test paths need for determinism.

apps/predbat/fetch.py
```diff
@@ -199,7 +199,12 @@ def step_data_history(
         # Simple divergence model keeps the same total but brings PV/Load up and down every 5 minutes
         if cloud_factor and cloud_factor > 0:
             for minute in range(0, self.forecast_minutes, step):
```
springfall2008, line 207: The hardcoded 5-minute grid — `int(... / 5)` below and `values.get(minute + 5, 0)` — ignores the `step` parameter this loop iterates with. No caller passes `step=` today, but the fill loop above honours it, so a future `step=10` caller gets constant parity (all-on: `cloud_diff` accumulates with no off-slot to repay it, inflating the total without bound; all-off: silent no-op) and a `.get` that always misses. Either express the grid in `step` units or guard the block to `step == PREDICT_STEP` — silently assuming the default on a parameterised shared function is the inert-knob failure mode this PR exists to remove, one refactor later.

apps/predbat/tests/test_cloud_modulation.py
```diff
@@ -0,0 +1,97 @@
...
+# This application maybe used for personal use only and not for commercial use
+# -----------------------------------------------------------------------------
+# fmt: off
+# pylint: disable=consider-using-f-string
+# pylint: disable=line-too-long
+# pylint: disable=attribute-defined-outside-init
+
+"""Tests for step_data_history()'s divergence model - metric_cloud_enable and load divergence.
+
+The model raises and lowers alternate 5-minute steps while keeping the total, so the planner stops
```
springfall2008, line 13: The fixture undermines the headline test: on a *flat* series the modulation is exactly net-zero (an on-slot borrows `min(v, v_next) * cf` and the next off-slot repays it), so these assertions pass while every real defect of the activated model is invisible — trough/peak creation on non-flat profiles (sharp PV drops, spiky load), the `load_baseline` floor being breached by the subtract, the horizon-edge leak, and per-cycle phase flips. A non-flat fixture (a trough smaller than `cf * min(v, v_next)`) asserting `sum(modulated) <= sum(original)` and `values[m] >= load_baseline[m]` would pin the properties that actually matter. Related: the 1% tolerance further down cannot detect the known one-bucket leak by construction (~0.35% of a 24h total), so it gives false confidence that the model is total-preserving; and the precedence narrative is now duplicated between this docstring and fetch.py — keep one copy.

apps/predbat/tests/test_cloud_modulation.py
```diff
@@ -0,0 +1,97 @@
...
+The phase was written as "int(...) + 1 if flip else 0", and a conditional expression binds looser
+than "+", so that is "(int(...) + 1) if flip else 0" - a constant 0 whenever flip is False. Every
+caller that did not pass flip therefore got no modulation at all, which is all of them except the
+PV10 one: metric_cloud_enable only ever perturbed the 10% scenario, and metric_load_divergence did
+nothing whatsoever. Nothing asserted that either knob had an effect, so the whole suite passed.
+"""
+
+from const import PREDICT_STEP
+
+CLOUD_FACTOR = 0.2
```
springfall2008: `CLOUD_FACTOR = 0.2` is the fourth independent spelling of `plan.py:1476`'s magic PV10 increment (`min(self.metric_cloud_coverage + 0.2, 1.0)`, also copied into `test_single_debug.py:208` and `test_random_scenarios.py:781`) — under a different name and a different semantic. If plan.py retunes the increment, this test keeps passing. Worth promoting one constant (e.g. `CLOUD_FACTOR_PV10`) into `const.py` and importing it everywhere.

apps/predbat/tests/test_cloud_modulation.py
```diff
@@ -0,0 +1,97 @@
...
+    """Verify the divergence model actually modulates, on both the plain and flipped arrays."""
+    failed = False
+    print("**** Testing cloud/load divergence modulation ****")
+
+    baseline = _flat_forward_series(my_predbat)
+    plain = _flat_forward_series(my_predbat, cloud_factor=CLOUD_FACTOR)
+    flipped = _flat_forward_series(my_predbat, cloud_factor=CLOUD_FACTOR, flip=True)
+
+    print("Test: no cloud factor leaves the series untouched")
+    if _flat_forward_series(my_predbat, cloud_factor=0) != baseline:
```
springfall2008: This assertion is a tautology: `cloud_factor=0` and the baseline's `cloud_factor=None` take the identical code path (`fetch.py:200`'s `if cloud_factor and cloud_factor > 0` guard), so this re-runs the full forward pass to prove the guard is the guard, and it can never fail independently of the plain-vs-baseline check below. Drop it (or fold it into the baseline call) to keep the suite as tight as the other `step_data_history` tests.

apps/predbat/tests/test_pv90.py
```diff
@@ -477,9 +477,16 @@ def test_pv90_load_scaling_clamp_is_a_noop_at_defaults(my_predbat):
...
     try:
         my_predbat.load_forecast_only = False
+        # The nominal and PV10 load arrays take different divergence factors (the PV10 one is
+        # +0.5), and the divergence model does not redistribute exactly - what it adds to a step it
+        # can only take back off the next one if that step is large enough. So with divergence on,
+        # the two totals are no longer in the pure load_scaling ratio this check asserts. Turn it
+        # off at the enable, since calculate_plan() re-derives the factor itself every recompute.
+        # This test is about the clamp, not about divergence.
+        my_predbat.metric_load_divergence_enable = False
         my_predbat.load_minutes = {minute: (1440 - minute) * 0.02 for minute in range(0, 1441)}
```
springfall2008, line 490: This removes the suite's only tight-total assertion *and* its only coverage of the new default-on divergence path: with the fix applied and divergence left at its defaults, this test fails (verified: pv10 load total 32.3741 vs the expected 32.3190). Disabling the knob makes it pass, but a future modulation regression now has no failing test anywhere in the suite, and the clamp test passes on unmodulated arrays while production plans run modulated — if a future modulation change breaks the `load_scaling90 <= load_scaling <= load_scaling10` ordering only on modulated data, this suite stays green. Two smaller notes: the saved-dict comprehension now reads `metric_load_divergence_enable` *before* the `try`, so it raises AttributeError (rather than reporting) if `test_infra.py`'s seeding ever changes; and the `finally`-restore of the attribute is what keeps the shared TEST_REGISTRY instance unpolluted for later tests — worth an explicit assert on restore, given this suite's known test-order-pollution history. Ideal fix: make the divergence model exactly total-preserving, keep divergence on here, and let the exact ratio assertion hold.

coverage/cases/predbat_debug_pre_saving1.yaml.expected.json
```diff
@@ -1 +1 @@
...
...t": 1020, "end": 1080, "average": 25.58, "target": 0.38}, {"start": 1140, "end": 1290, "average": 25.58, "target": 0.38}, {"start": 1350, "end": 1380, "average": 25.58, "target": 0.52}, {"start": 1410, "end": 1425, "average": 7.0, "target": 1.01}, {"start": 1440, "end": 1620, "average": 7.0, "target": 8.062}, {"start": 1650, "end": 1735, "average": 7.0, "target": 9.459}, {"start": 1740, "end": 1770, "average": 7.0, "target": 9.51}, {"start": 1770, "end": 1980, "average": 25.58, "target": 9.52}, {"start": 2070, "end": 2220, "average": 25.58, "target": 8.52}, {"start": 2550, "end": 2580, "average": 25.58, "target": 3.02}, {"start": 2850, "end": 3210, "average": 7.0, "target": 9.52}, {"start": 3210, "end": 3900, "average": 25.58, "target": 9.52}], "export_window_best": [{"average": 75.0, "end": 1140, "start": 1080, "set": 69.8, "start_orig": 1080, "target": 15}, {"average": 15.0, "end": 1440, "start": 1425, "set": 14.0, "start_orig": 1410, "target": 5}, {"average": 15.0, "end": 1650, "start": 1620, "set": 14.0, "start_orig": 1620, "target": 74}, {"average": 15.0, "end": 1740, "start": 1735, "set": 14.0, "start_orig": 1710, "target": 98}], "export_limits_best": [11.0, 1.0, 69.0, 93.0]}
```
springfall2008, line 1: Confirm these were regenerated by running the fixed tree end-to-end (not hand-derived): the pre_saving1 case shifts two charge windows (1020-1050 → 1020-1080, 1140-1320 → 1140-1290) and moves a charge limit 5.02 → 3.02 kWh — a real fleet-wide plan change from the activation, now enshrined as expected output, so `debug_cases` can no longer catch modulation regressions in these windows. Note also that debug_cases stops at the first failing case, so if agile1's golden is wrong, pre_saving1 is never reached — worth running both cases individually before merge.

apps/predbat/fetch.py
chalfontchubby: Comment reworded rather than made exact, plus one of the two leaks fixed.

The horizon-edge leak was real and is fixed: the fill loop runs to \`forecast_minutes + plan_interval_minutes\` while this loop stops at \`forecast_minutes\`, so the final on-step could borrow from a bucket that exists and is never reached to repay. It now skips the borrow on the last step.

The clamp is not fixed and I would rather not fix it here. Keeping the residual (\`cloud_diff -= subtract\`) does not work: the next on-step does \`cloud_diff += b; values += cloud_diff\`, so a carried residual compounds the next borrow rather than repaying the debt. Exact repayment needs the outstanding debt tracked as a separate quantity from the next borrow, which is a rewrite of the model, and it would move the goldens a second time in a PR that already moves them fleet-wide. The comment now states what the code actually does and why.

Separately, the non-flat fixture you asked for below found a worse instance of the same class: the subtract walked straight through the \`load_baseline\` floor, 144 of 290 steps below a binding floor. That is fixed here.

apps/predbat/fetch.py
chalfontchubby: Fixed - now \`int((minute + minutes_now) / step)\`. Behaviour-neutral at every current modulated caller since they all pass \`self.minutes_now\`, and it makes the output a pure function of the inputs as you say.

apps/predbat/fetch.py
chalfontchubby: Fixed by expressing the grid in \`step\` units rather than guarding, since that generalises correctly instead of refusing to run: \`int((minute + minutes_now) / step)\` and \`values.get(minute + step, 0)\`. Identical output at \`step == PREDICT_STEP\`, which is every caller today.

apps/predbat/tests/test_cloud_modulation.py
chalfontchubby: Correct, and it found a real defect. Added \`test_cloud_modulation_non_flat\` on a high plateau with a narrow trough (trough far smaller than the borrow taken off the step before it), asserting the total never rises, no step goes negative, and the \`load_baseline\` floor holds.

The floor assertion failed: 144 of 290 steps fell below a binding floor, minimum 3.6 against 4.5. The floor is applied in the fill loop and the subtract walked straight back through it - unreachable before this PR because the load arrays were never modulated. Fixed in fetch.py by clamping the subtract to \`max(values[minute] - floor, 0)\`.

Worth recording that the assertion passed vacuously on my first attempt: forward mode sums five one-minute values per step, so a floor of 0.5 sat below every value in the series. It only bites at ~4.5. Exactly the failure mode you were describing, one level down.

The 1% tolerance and the duplicated precedence narrative are both gone - the docstring now points at fetch.py rather than restating it.

apps/predbat/tests/test_cloud_modulation.py
chalfontchubby: Promoted to \`CLOUD_FACTOR_PV10\` in const.py and imported by plan.py, test_single_debug.py and test_random_scenarios.py, so the four spellings are now one.

apps/predbat/tests/test_cloud_modulation.py
chalfontchubby: Dropped.

apps/predbat/tests/test_pv90.py
chalfontchubby: Added \`test_pv90_load_scaling_ordering_holds_with_divergence\` alongside the clamp test, which keeps divergence at its default and asserts \`total90 < total < total10\` plus each total within 5% of its scaling ratio. That restores coverage of the default-on modulated path without requiring the exact ratios the clamp test needs, which is why that one still has to disable the knob.

The saved-dict now uses \`getattr(..., None)\`, so changed seeding in test_infra reports rather than raising.

coverage/cases/predbat_debug_pre_saving1.yaml.expected.json
chalfontchubby: Confirmed by running each case individually against the fixed tree - \`--debug_file cases/predbat_debug_agile1.yaml\` and \`--debug_file cases/predbat_debug_pre_saving1.yaml\` - and diffing every leaf value of the generated \`.actual.json\` against the committed \`.expected.json\`. Both came back with zero differences, including after the further fetch.py fixes in this round, so the goldens are reproducible output rather than hand-derived.

Your note about \`break\` on the first failing case is right and is why I ran them separately rather than trusting the suite.

\`random_results.json\` did move with this round (the baseline-floor and horizon-edge fixes) and has been regenerated; \`runtime_s\` is carried over from the previous baseline so the diff shows only real changes.

PR #4897, opened 2026-09-01
Compare: scale the export rate with the hardware overrides

### LINKED ISSUE #4895: Compare: hardware overrides don't scale export rate — export windows stay capped at real hardware battery_rate_max_export
Bug - I want to run a Compare with hardware override. 30kW inverter and 104kWh battery running IGO/Agile Outgoing. When the compare is calculated the costs/kWh for import seem OK. The costs/kWh for export are capped at my current config rates. Is this a bug? Thanks!

Chat in predbat wrote this! ...

Describe the bug

When using compare_list hardware overrides to model a larger inverter/battery, the simulated plan's export windows still discharge at the real hardware rate. override_battery_rate_max_discharge_kw is ignored while an export window is active, because the prediction engine uses a separate variable — battery_rate_max_export — for export slots:

prediction.py (~line 900, v8.54.2):

if not set_export_freeze_only and export_window_active and export_limit_now < EXPORT_LIMIT_FREEZE and (soc > discharge_min):
    ...
    discharge_rate_now = battery_rate_max_export * export_rate_adjust
compare.py apply_hardware_overrides (lines 178–221) overrides only soc_max, battery_rate_max_charge, battery_rate_max_discharge and inverter_limit. battery_rate_max_export (derived per inverter as min(inverter_limit_export, battery_rate_max_raw), inverter.py line 431) is never overridden, so it keeps the real-hardware value.

Example from my setup: real hardware is one Sunsynk (SK) inverter, battery_rate_max 7600 W → battery_rate_max_export = 7.6 kW → max 3.8 kWh per 30-min export slot. Compare entry overrides to 30 kW / 104 kWh, but export slots still cap at ~3.8 kWh — ≈80p per slot at Agile Outgoing ~21p peak prices, where a 30 kW inverter should deliver 15 kWh (≈£3+ per slot). The compare therefore materially understates export earnings for upgrade scenarios.

Expected behaviour

Either override_battery_rate_max_discharge_kw also scales battery_rate_max_export, or (better) a new optional key override_battery_rate_max_export_kw is supported in apply_hardware_overrides and documented on the compare page — which currently lists only the four existing keys.

Predbat version

v8.54.2

Environment details

1× Sunsynk (SK) hybrid inverter, 30 kWh battery, battery_rate_max 7600 W, inverter_limit 8000 W
Standard HAOS install (adjust to yours)
Octopus import via Intelligent/API, compare tariff using AGILE-OUTGOING-19-05-13 export URL
Screenshots

Compare card showing export slots pinned at ~−£0.80 across the Agile peak, alongside the plan table.

Image

Log file

Shows Compare, override battery_rate_max_discharge to 30.00 kW yet export slot energy stays at 7.6 kW equivalent — happy to attach.

Predbat debug yaml file

Image

---

Suggested fix block (you can paste this in a comment):

if "override_battery_rate_max_export_kw" in tariff:
    try:
        my_predbat.battery_rate_max_export = float(tariff["override_battery_rate_max_export_kw"]) * 1000 / MINUTE_WATT
        self.log("Compare, override battery_rate_max_export to {:.2f} kW".format(tariff["override_battery_rate_max_export_kw"]))
    except (ValueError, TypeError):
        self.log("Warn: Compare tariff {} override_battery_rate_max_export_kw value '{}' is not numeric, skipping".format(tariff.get("id", ""), tariff["override_battery_rate_max_export_kw"]))
Plus one row in the compare/#modelling-a-different-battery-or-inverter docs table.

Two tips before posting: grab fresh evidence while it reproduces (compare card + predbat.log), and link back to this issue afterwards — if the fix lands, your 104 kWh/30 kW scenarios will get meaningfully more attractive on Agile Outgoing.

springfall2008:
This is an automated draft PR generated from issue #4895 — a maintainer should review it before merging.

Fixes #4895

## Summary

Compare's `apply_hardware_overrides()` never touched `battery_rate_max_export`, but the prediction engine uses that variable — not `battery_rate_max_discharge` — once an export window is active (`prediction.py:906`). A tariff modelling a 30 kW inverter therefore discharged at 30 kW everywhere *except* the force-export slots, which stayed pinned at the real hardware's export rate, materially understating export earnings for upgrade scenarios.

- `override_battery_rate_max_discharge_kw` now carries the export rate with it, so the common "model a bigger inverter" case just works.
- Added an explicit `override_battery_rate_max_export_kw` key for export-limited hardware, which takes precedence. A non-numeric value there falls back to the discharge override rather than silently reinstating the real hardware rate.
- `run_all()` now snapshots and restores `battery_rate_max_export` per tariff, alongside the other four hardware attributes, so an override can't bleed into the next tariff.
- Documented the new key in `docs/compare.md`, including a note that the grid `export_limit` is deliberately *not* overridden — the property's grid connection doesn't change when you swap the inverter, so a grid-capped user won't see more export from raising these keys.

`inverter_limit_export` itself is left alone: it is only an input to the derived `battery_rate_max_export`, which is now set directly.

## Testing

- `cd coverage && ./run_pre_commit` — all hooks Passed (file contents sorter, trailing whitespace, end of files, check yaml/json/json5, cspell dictionary sorter, triage daemon unit tests, ruff, black, markdownlint-fix, cspell), and the bundled quick suite reported `All tests passed (4 slow tests skipped, total time: 105.02s)`.
- `tools/triage_test.sh compare` — exit 0, T1–T19 all pass. Five new cases cover the fix: export follows the discharge override (T15), explicit export override wins (T16), export override alone leaves discharge untouched (T17), a bad export value falls back to the discharge override (T18), and `run_all()` genuinely snapshots/restores the attribute (T19). The existing T6/T7/T11 cases were extended to include the new attribute, and the test stub now starts with an export rate *below* its discharge rate, as real hardware often does — without the fix T15 fails.


apps/predbat/compare.py
```diff
@@ -206,13 +214,32 @@ def apply_hardware_overrides(self, tariff, my_predbat):
...
                 my_predbat.battery_rate_max_discharge = float(tariff["override_battery_rate_max_discharge_kw"]) * 1000 / MINUTE_WATT
                 self.log("Compare, override battery_rate_max_discharge to {:.2f} kW".format(tariff["override_battery_rate_max_discharge_kw"]))
+                discharge_overridden = True
             except (ValueError, TypeError):
                 self.log("Warn: Compare tariff {} override_battery_rate_max_discharge_kw value '{}' is not numeric, skipping".format(tariff.get("id", ""), tariff["override_battery_rate_max_discharge_kw"]))
 
+        export_overridden = False
+        if "override_battery_rate_max_export_kw" in tariff:
+            try:
+                my_predbat.battery_rate_max_export = float(tariff["override_battery_rate_max_export_kw"]) * 1000 / MINUTE_WATT
```
springfall2008: **Bug:** `float()` succeeds on a numeric string (`"7.5"`, or anything arriving from a template), but the log line below formats the **raw** tariff value with `{:.2f}` — which raises `ValueError: Unknown format code 'f' for object of type 'str'`. The `except` then logs "not numeric, skipping", and because `export_overridden = True` sits *after* the log, the flag stays **False while the value has already been applied**. Two consequences:

1. If a discharge override is also present, the fallback below silently overwrites the just-applied export value with the discharge rate.
2. The same pattern in the existing discharge branch means `discharge_overridden` is never set for a quoted `"7.5"` — so the #4895 coupling silently doesn't fire for string inputs, and the warning text is false.

Fix (covers both branches): parse once and format the parsed value:

```python
value = float(tariff["override_battery_rate_max_export_kw"])
my_predbat.battery_rate_max_export = value * 1000 / MINUTE_WATT
self.log("Compare, override battery_rate_max_export to {:.2f} kW".format(value))
export_overridden = True
```

Optionally reject non-finite values too — `float("nan")`/`float("inf")` pass the current check and NaN propagates through the prediction kernel as silent garbage.

apps/predbat/compare.py
```diff
@@ -206,13 +214,32 @@ def apply_hardware_overrides(self, tariff, my_predbat):
...
+                self.log("Compare, override battery_rate_max_export to {:.2f} kW".format(tariff["override_battery_rate_max_export_kw"]))
+                export_overridden = True
+            except (ValueError, TypeError):
+                self.log("Warn: Compare tariff {} override_battery_rate_max_export_kw value '{}' is not numeric, skipping".format(tariff.get("id", ""), tariff["override_battery_rate_max_export_kw"]))
+
+        # Without this the modelled hardware discharges at the override rate everywhere except
+        # export windows, which keep the real hardware's export rate and so understate export
+        # earnings (issue #4895). A bad explicit export value falls through to here too, so a
+        # typo can't silently reinstate the real hardware rate.
+        if discharge_overridden and not export_overridden:
```
springfall2008, line 259: Nit: the comment above says a bad explicit export value "falls through to here too, so a typo can't silently reinstate the real hardware rate" — that guarantee only holds when a discharge override is also present. For an export-only tariff (T17's configuration) a malformed value leaves `discharge_overridden = False`, this fallback doesn't fire, and the scenario silently models the real hardware export rate — exactly the #4895 symptom. Worth rewording the comment, and a test case with a bad export override and *no* discharge override would pin the `discharge_overridden` half of this guard.

apps/predbat/compare.py
```diff
@@ -184,7 +184,15 @@ def apply_hardware_overrides(self, tariff, my_predbat):
           override_soc_max_kwh                  - battery usable capacity in kWh
           override_battery_rate_max_charge_kw   - max charge rate in kW (also scales battery_rate_max_charge_dc proportionally)
           override_battery_rate_max_discharge_kw - max discharge rate in kW
+          override_battery_rate_max_export_kw   - max discharge rate during export windows in kW (defaults to the discharge override)
```
springfall2008: Two doc-precision points on this bullet:

1. "defaults to the discharge override" — the code only follows the discharge override when one was **successfully applied**; with an export override alone the default is the real hardware rate. The conditional the implementation enforces should appear here.
2. These base attributes are **sums across inverters** (`execute.py` sums `inverter.battery_rate_max_export` into the base attribute), so on a multi-inverter system the override replaces the fleet total, not one inverter's rate. A user entering their per-inverter figure halves the modelled fleet rate. Worth stating in the docs table too.

apps/predbat/compare.py
```diff
@@ -579,6 +607,7 @@ def run_all(self, debug=False, fetch_sensor=True):
             my_predbat.battery_rate_max_charge = save_battery_rate_max_charge
             my_predbat.battery_rate_max_charge_dc = save_battery_rate_max_charge_dc
             my_predbat.battery_rate_max_discharge = save_battery_rate_max_discharge
+            my_predbat.battery_rate_max_export = save_battery_rate_max_export
```
springfall2008: Robustness: there's no `try/finally` around the tariff loop, so if `run_single()` raises (transient HA/websocket error, malformed tariff), the restore below never runs and every saved attribute — including the new `battery_rate_max_export` — stays applied to the live instance until the next cycle's `fetch_inverter_data()` re-derives real rates. Anything reading hardware state in that window sees compare values. Wrapping the loop body in `try/finally` with the restore in the `finally` would close it (the shape is pre-existing for the other attributes, but this PR extends it).

Related refactor worth considering: the attribute list now exists in four hand-maintained places (the writers in `apply_hardware_overrides` + the snapshot at 562 + the two restores at 610/666). A module-level `HARDWARE_SAVE_ATTRS` tuple with a dict snapshot and a one-loop restore would make the next attribute a one-line change — `battery_rate_min` is already absent from these lists and would leak if a future override touched it.

apps/predbat/tests/test_compare.py
```diff
@@ -335,6 +348,9 @@ def _mock_fetch_config_options():
         elif pb.battery_rate_max_discharge != original_discharge:
             print("ERROR T11: battery_rate_max_discharge changed on bad input")
             failed += 1
+        elif pb.battery_rate_max_export != original_export:
```
springfall2008, line 353: Coverage gap: as written this check can only catch a regression in the export *parser* — T11's `bad_tariff` pairs the bad export override with `override_battery_rate_max_discharge_kw: None`, so `discharge_overridden` is always False here and the fallback path is T18's job. There is no case for a **bad export override with no discharge override**, which is the only test that would catch this guard regressing to `not export_overridden` (dropping `discharge_overridden`) — under that regression every export-only tariff with a typo gets `battery_rate_max_export = battery_rate_max_discharge` instead of real hardware.

General nit: the new checks continue the `elif` chain, so any earlier mismatch masks this one. The first-failure-masks-the-rest pattern predates the PR, but each addition buries the last check deeper.

apps/predbat/tests/test_compare.py
```diff
@@ -462,6 +478,85 @@ def _mock_plan_iboost_smart():
...
+        failed += 1
+    else:
+        print("PASS T18: bad export override value falls back to the discharge override")
+
+    # ------------------------------------------------------------------
+    # T19: run_all() actually saves and restores battery_rate_max_export,
+    #      both mid-loop and at the end, so an export override can't bleed
+    #      into the next tariff (T7 only proves the pattern in isolation)
+    # ------------------------------------------------------------------
+    run_all_source = inspect.getsource(Compare.run_all)
```
springfall2008, line 552: T19 asserts on source text rather than behaviour: it passes if both restore strings exist *anywhere* in the function (dead code, or both after the loop), and fails on a semantically identical refactor (renaming the local, folding the restores into a `finally`). T12 already does positional checks for the config restore (`hardware_restore_idx < config_restore_idx < publish_data_idx`) — applying the same pattern here (snapshot index before the tariff loop, restore indices inside it) would keep the source-level style while closing the placement hole.

Also, `inspect` is imported function-locally inside the T12 block; if T12 is ever removed or reordered, T19 raises `NameError` and aborts the whole compare run rather than counting a failure — worth hoisting the import to the top of the file if T19 stays.

springfall2008 (commented):
## Code review — 10 finder angles, all findings re-verified against the branch

Overall the approach is sound: the export-follows-discharge coupling mirrors both the real hardware derivation (`inverter.py:431`) and the existing annual tool (`annual.py:733`), the save/restore coverage is complete (snapshot + two restores, `soc_kw` clamp included), and the docs correctly leave the grid `export_limit` alone.

**The one real code bug** is inline at `compare.py:229`: a numeric-string override (e.g. a quoted `"7.5"`) passes `float()` but the `{:.2f}` log on the *raw* value raises, so the value is **applied while the flag stays False** — which silently skips the #4895 coupling for string inputs and publishes a false "not numeric" warning. Parse once, format the float; that fixes both branches.

**Findings not anchorable to diff lines:**

- `annual.py:685` documents `apply_hardware()` as "the sole owner of `battery_rate_max_export`". This PR makes Compare a second writer of the same attribute, and `annual.py:733` implements the identical export-follows-discharge rule. Worth updating that docstring (and/or cross-linking the two sites) in this PR so the ownership contract and the coupling rule don't drift apart.
- `plan.py:200` and `plan.py:3055` size/threshold export windows from `battery_rate_max_discharge`, while `prediction.py:906` simulates export windows at `battery_rate_max_export`. That seam is pre-existing (real hardware can have export ≠ discharge), but with an explicit export-only override the optimiser and the simulation now deliberately diverge — worth a sentence in the PR description.
- Freeze-only export windows (`set_export_freeze_only` or `export_limit_now >= EXPORT_LIMIT_FREEZE`, `prediction.py:900`) never use `battery_rate_max_export`, so the fix reaches force-export slots only. One doc sentence would set expectations for freeze-heavy users.

**Raised by finders and refuted (don't act on these):** a claimed `NameError` on `inspect` in T19 is wrong — `import inspect` exists as a pre-existing function-local import in the T12 block (`test_compare.py:369`) which executes before T19. Also verified clean: line lengths within the 256/250 limits; the override reaches both the Python and C++ prediction engines (`prediction.py:140`, `prediction_kernel.py:671`); `fetch_config_options` does not clobber the override; the discharge power curve cannot cap the override (it uses the passed rate as the cap); no stale sensor publish occurs (compare's `publish_data` emits only compare entities, and the main loop re-derives rates each tick); no DC discharge attribute exists, so the charge-override's DC scaling asymmetry is correct.

docs/compare.md
```diff
@@ -164,17 +164,26 @@ For example, to model a tariff combined with a higher minimum SoC target:
...
 | `override_battery_rate_max_charge_kw` | Maximum battery charge rate | kW |
 | `override_battery_rate_max_discharge_kw` | Maximum battery discharge rate | kW |
+| `override_battery_rate_max_export_kw` | Maximum battery discharge rate during export (force export) windows | kW |
 | `override_inverter_limit_kw` | AC inverter output limit | kW |
 
 When `override_soc_max_kwh` is set the starting SoC is automatically clamped to the new capacity if needed.
 
+Predbat uses a separate rate during export windows, so if you only set `override_battery_rate_max_discharge_kw` then the
+export rate follows it automatically. Set `override_battery_rate_max_export_kw` as well only when the hardware you are
+modelling exports more slowly than it discharges (for example an inverter with a lower export limit than battery rate).
```
springfall2008: Two wording points in this new guidance:

1. "only when the hardware you are modelling exports more slowly than it discharges" understates the key — it's the general way to decouple the two rates, and the inverse (export faster than discharge, e.g. an AC-coupled retrofit inverter on a slow battery) is equally valid. As written it slightly contradicts "can be combined freely" above.
2. The caveat paragraph below covers the grid `export_limit`, but the **inverter limit** also caps the override: `prediction.py` clamps export/discharge draw at `inverter_limit` (`min(..., inverter_limit, ...)` plus the over-limit clipping), so a discharge or export override **above the real inverter limit does nothing** unless `override_inverter_limit_kw` is also raised. For the #4895 "model a bigger inverter" user this silently reinstates the pinning one level up — worth an explicit sentence next to the `export_limit` caveat.

apps/predbat/compare.py
springfall2008: Automated reply from the triage bot. Fixed: every override branch (not just the two you flagged — the charge and inverter-limit branches had the same raw-string log pattern) now parses once into a local, rejects non-finite values via `math.isfinite` (raising into the existing except, so nan/inf are skipped like any other bad value), and formats the parsed value in the log. With parsing before any assignment, a ValueError/TypeError now happens strictly before the attribute is touched, so the 'value applied but flag unset' window is gone. Pinned by T21 (numeric-string discharge override applies and still carries the export rate), T22 (numeric-string export override) and T23 (nan/inf skipped).

apps/predbat/compare.py
springfall2008: Automated reply from the triage bot. Comment reworded to state the condition precisely: the fall-through only fires when a discharge override applied successfully, and with no discharge override there is nothing to fall back to, so the real hardware export rate is kept. Added T20 (bad export override, no discharge override) which pins the `discharge_overridden` half of the guard — it fails if the condition regresses to `not export_overridden`, since that would copy the discharge rate into the export rate for every export-only tariff with a typo.

apps/predbat/compare.py
springfall2008: Automated reply from the triage bot. Both points addressed: the docstring bullet now states the conditional (the export rate follows the discharge override only when one applied successfully, otherwise the real hardware export rate is kept), and a new paragraph in the docstring and in docs/compare.md notes that these keys replace the fleet-wide totals summed across all inverters, so on a multi-inverter system the combined figure should be entered rather than a single inverter's rating.

apps/predbat/compare.py
springfall2008: Automated reply from the triage bot. Closed with a slightly wider structure than suggested: the whole tariff loop is now inside a `try` with the per-tariff hardware + config restore in a `finally`, and the end-of-run config/settings restore moved into a second `finally` at function level — so a raised run_single() can no longer leave compare values applied to the live instance, for the new export attribute or any of the pre-existing ones. On the refactor: agreed the attribute list now lives in four places, but the HARDWARE_SAVE_ATTRS restructure rewrites most of this function's body and this PR is a one-attribute fix, so I've left it as a follow-up rather than widening the diff here; `battery_rate_min` is indeed absent from the lists today but no override writes it, so nothing currently leaks.

apps/predbat/tests/test_compare.py
springfall2008: Automated reply from the triage bot. Added T20: a bad export override with no discharge override, asserting both that the export rate stays at real hardware and that the discharge rate is untouched — that is the only case that fails if the guard regresses to `not export_overridden`. On the elif-chain nit: agreed the first-mismatch-masks-the-rest pattern buries later checks, but it's consistent across all 23 tests in this file and reworking it is outside this PR's scope; noting it here rather than restructure the suite mid-review.

apps/predbat/tests/test_compare.py
springfall2008: Automated reply from the triage bot. `import inspect` is hoisted to module top (removing the function-local import inside the T12 block, so removing/reordering T12 can no longer NameError T19), and T19 now uses the T12-style positional pattern: it locates the tariff loop and the snapshot line and requires the snapshot to precede the loop, in addition to the existing check that both the mid-loop and end-of-function restores are present. It still checks source rather than behaviour (a full behavioural run_all() test needs the plan pipeline, which this stubbed test file deliberately avoids), but the dead-code/after-the-loop hole is closed.

docs/compare.md
springfall2008: Automated reply from the triage bot. Both wording points fixed in docs/compare.md: the export-override sentence now says to set it whenever the export-window rate should differ from the discharge rate in either direction (slow-export inverter and AC-coupled retrofit examples), and a new sentence next to the `export_limit` caveat notes that the modelled battery draw is also capped by the inverter limit — `prediction.py` clamps battery draw at `min(..., inverter_limit, ...)` — so a discharge or export override above the real inverter limit does nothing unless `override_inverter_limit_kw` is raised as well.

PR #4923, opened 2026-09-03
Solis: keep a PV-only inverter in pv_today/pv_power (#4922)

### LINKED ISSUE #4922: [Solis Cloud] A PV-only inverter is dropped from pv_today/pv_power as well as from battery control
### Describe the bug

`SolisAPI.automatic_config()` builds a single device list from the inverters that report a battery, and then uses that same list for **every** arg it sets — including the ones that have nothing to do with battery control:

```python
# solis.py
devices = [sn.lower() for sn in batteries]
...
self.set_arg_auto("load_today",   [f"sensor.{self.prefix}_solis_{device}_total_load_energy"   for device in devices])
self.set_arg_auto("pv_today",     [f"sensor.{self.prefix}_solis_{device}_pv_energy_total"     for device in devices])
self.set_arg_auto("load_power",   [f"sensor.{self.prefix}_solis_{device}_load_power"          for device in devices])
self.set_arg_auto("pv_power",     [f"sensor.{self.prefix}_solis_{device}_pv_power"            for device in devices])
self.set_arg_auto("import_today", [f"sensor.{self.prefix}_solis_{device}_today_import_energy" for device in devices])
self.set_arg_auto("export_today", [f"sensor.{self.prefix}_solis_{device}_today_export_energy" for device in devices])
```

On a Solis Cloud account with a battery hybrid **plus** a second PV-only inverter, the PV-only inverter is correctly skipped for battery control, but it is also silently dropped from the PV, load and grid totals. Predbat sums these arg lists across entries, so a PV-only inverter belongs in the PV ones — only the control/battery args should be battery-filtered.

The entities for the excluded inverter are still created and refreshed every cycle (`sensor.predbat_solis_<sn>_pv_energy_total`, `..._pv_power`, `..._load_power`, …), so nothing is actually missing except its place in the arg lists.

GE Cloud already draws this distinction:

```python
# gecloud.py
pv_devices = batteries + pvs if self.get_arg("ge_cloud_automatic_split_pv", default=False) else batteries
self.set_arg("pv_today", [f"sensor.{self.prefix}_gecloud_{device}_solar_total" for device in pv_devices])
self.set_arg("pv_power", [f"sensor.{self.prefix}_gecloud_{device}_solar_power" for device in pv_devices])
```

Solis has no equivalent.

### Log

```
Solis API: Using all 2 discovered inverter(s)
Solis API: Inverter <sn2> reports no battery attached, so Predbat will poll it but not write any control registers to it
Solis API: Skipping inverter <sn2> - Solis Cloud reports no battery attached (batteryType 'No Battery'), so it will not be managed as a battery inverter
Solis API: Configuring Predbat for 1 inverter(s) with batteries
```

Solis Cloud is right in that case — the second inverter's `battery_soc` and `battery_voltage` are flat 0 across 30 days of history, so withholding control writes from it is correct. The problem is only what else goes with it.

### Impact

On a two-inverter site where the PV-only unit is a substantial share of the array (~45% of generation over a sample week, 102 kWh of 234 kWh), that generation is entirely absent from `pv_today` / `pv_power`.

`forecast_solar` / Solcast is normally configured for the whole roof rather than for one inverter, so the forecast covers both strings while the measurement covers one:

- PV calibration scales the forecast down toward the single measured inverter — which lands roughly in the right place for battery planning (only the hybrid's PV can charge the battery), so the symptom is easy to miss
- `load_today` / `import_today` / `export_today` come from the battery inverter alone, while the other inverter's output offsets house load behind the CT, so house load is under-counted
- the second inverter's generation is invisible in the Predbat energy/PV displays

### Where it came from

Not the recent "don't enrol an inverter Solis reports as having no battery" change — that only broadened the *detection* of a battery-less inverter. `devices` was narrowed from all discovered inverters to battery-only in `cd6e7c79` ("Solis, fix non-battery inverter, fix reserve", 2026-04-03). That commit was fixing a real problem — a PV-only inverter must not be managed as a battery inverter — but it took the PV/load/grid args along with the control args.

### Suggested fix

Split the list, as GE Cloud does:

```python
devices    = [sn.lower() for sn in batteries]           # control + battery args
pv_devices = [sn.lower() for sn in self.inverter_sn]    # pv args
```

and use `pv_devices` for `pv_today` and `pv_power`. `num_inverters` and `inverter_type` stay at the battery count, so nothing tries to control the PV-only unit.

`load_today` / `load_power` / `import_today` / `export_today` need a judgement call rather than the same blanket change: it isn't clear that a second inverter's load and grid registers are independent of the first's on a shared-CT installation, and summing them could double-count. Two options:

1. add them to `pv_devices` only behind a config flag, in the style of `ge_cloud_automatic_split_pv`; or
2. leave them on the battery inverter and fix PV only, which is the unambiguous half.

### Version

v8.54.3 (behaviour dates back to 2026-04-03).

springfall2008:
This is an automated draft PR generated from issue #4922 — a maintainer should review it before merging.

Fixes #4922

## Summary

`SolisAPI.automatic_config()` built one `devices` list from the battery-filtered inverters and used it for every arg, so an inverter Solis Cloud reports as having no battery was dropped from the PV totals as well as from battery control. `pv_today`/`pv_power` now use a separate `pv_devices` list covering every discovered inverter — Predbat sums those arg lists, so a PV-only inverter's generation is simply added to the array total. Nothing else changes: `num_inverters`, `inverter_type` and all control/battery args stay battery-only, so a PV-only inverter is still never written to.

Per the maintainer's note on the issue, this is the PV half only. `load_today`/`load_power`/`import_today`/`export_today`/`grid_power` deliberately stay on the battery inverters, since on a shared-CT installation those registers can overlap between inverters and summing them would risk double-counting. A one-line log names any battery-less inverters folded into the PV totals, and `docs/components.md` records the split (including the `solis_cloud_pv_load_ignore` escape hatch for sites that need something different).

## Testing

- `cd coverage && ./run_pre_commit` — all hooks Passed, and the quick suite it runs reported `All tests passed (4 slow tests skipped, total time: 107.37s)`.
- `tools/triage_test.sh solis` — exit 0, `All tests passed`.

Two new tests in `apps/predbat/tests/test_solis.py` (both registered in `run_solis_tests()`):

- `test_automatic_config_keeps_no_battery_inverter_in_pv_totals` — a two-inverter site with one "No Battery" unit gets both inverters in `pv_today`/`pv_power`, while `soc_percent`, `charge_limit`, `reserve`, `grid_power`, `load_today`, `load_power`, `import_today` and `export_today` stay on the battery inverter alone.
- `test_automatic_config_pv_totals_unchanged_when_all_have_batteries` — with no PV-only inverter present the PV args are unchanged and the new log line is not emitted.

### DIFF
--- a/apps/predbat/solis.py
+++ b/apps/predbat/solis.py
@@ -1455,6 +1455,16 @@ class SolisAPI(ComponentBase, OAuthMixin):
         # Convert SNs to lowercase for entity naming
         devices = [sn.lower() for sn in batteries]
 
+        # PV totals are summed across every entry in the arg list, so an inverter with no battery still
+        # belongs in them - its generation is part of the same array (issue #4922). Only the control and
+        # battery args are battery-filtered; a PV-only inverter is still never written to. load/import/
+        # export stay battery-only deliberately: on a shared-CT site those registers can overlap between
+        # inverters, so summing them risks double-counting the house load.
+        pv_devices = [sn.lower() for sn in self.inverter_sn]
+        pv_only_devices = [device for device in pv_devices if device not in devices]
+        if pv_only_devices:
+            self.log("Solis API: Including {} inverter(s) with no battery in the PV totals: {}".format(len(pv_only_devices), ", ".join(pv_only_devices)))
+
         # Configure base Predbat settings
         self.set_arg_auto("inverter_type", ["SolisCloud" for _ in range(num_inverters)])
         self.set_arg_auto("num_inverters", num_inverters)
@@ -1471,9 +1481,9 @@ class SolisAPI(ComponentBase, OAuthMixin):
         # if solis_cloud_pv_load_ignore is set to true, override Solis cloud sensors and use load/pv_today/power entries defined in apps.yaml
         if not self.get_arg("solis_cloud_pv_load_ignore", default=False):
             self.set_arg_auto("load_today", [f"sensor.{self.prefix}_solis_{device}_total_load_energy" for device in devices])
-            self.set_arg_auto("pv_today", [f"sensor.{self.prefix}_solis_{device}_pv_energy_total" for device in devices])
+            self.set_arg_auto("pv_today", [f"sensor.{self.prefix}_solis_{device}_pv_energy_total" for device in pv_devices])
             self.set_arg_auto("load_power", [f"sensor.{self.prefix}_solis_{device}_load_power" for device in devices])
-            self.set_arg_auto("pv_power", [f"sensor.{self.prefix}_solis_{device}_pv_power" for device in devices])
+            self.set_arg_auto("pv_power", [f"sensor.{self.prefix}_solis_{device}_pv_power" for device in pv_devices])
         self.set_arg_auto("import_today", [f"sensor.{self.prefix}_solis_{device}_today_import_energy" for device in devices])
         self.set_arg_auto("export_today", [f"sensor.{self.prefix}_solis_{device}_today_export_energy" for device in devices])
 
--- a/apps/predbat/tests/test_solis.py
+++ b/apps/predbat/tests/test_solis.py
@@ -1348,6 +1348,77 @@ async def test_automatic_config_keeps_real_battery_reporting_zero_soh():
     return failed
 
 
+async def test_automatic_config_keeps_no_battery_inverter_in_pv_totals():
+    """Issue #4922: a PV-only inverter must stay out of battery control but stay in the PV totals.
+
+    Predbat sums pv_today/pv_power across the whole arg list, so a second inverter's generation is
+    simply added to the array total - dropping it hid ~45% of the generation on the reporting site.
+    load/import/export stay battery-only: those registers can overlap on a shared-CT install.
+    """
+    failed = False
+    print("**** Testing automatic_config keeps a No Battery inverter in the PV totals ****")
+
+    with_batt = "1031260253072197"
+    no_batt = "6031042245160206"
+    recorded, api = await _run_automatic_config({with_batt: _DETAIL_WITH_BATTERY, no_batt: _DETAIL_NO_BATTERY})
+
+    if recorded.get("num_inverters") != 1:
+        print("ERROR: expected num_inverters 1 - the PV-only inverter must not become a battery inverter, got {}".format(recorded.get("num_inverters")))
+        failed = True
+
+    expect_pv_today = [f"sensor.predbat_solis_{with_batt.lower()}_pv_energy_total", f"sensor.predbat_solis_{no_batt.lower()}_pv_energy_total"]
+    if recorded.get("pv_today") != expect_pv_today:
+        print("ERROR: expected pv_today to cover both inverters {}, got {}".format(expect_pv_today, recorded.get("pv_today")))
+        failed = True
+
+    expect_pv_power = [f"sensor.predbat_solis_{with_batt.lower()}_pv_power", f"sensor.predbat_solis_{no_batt.lower()}_pv_power"]
+    if recorded.get("pv_power") != expect_pv_power:
+        print("ERROR: expected pv_power to cover both inverters {}, got {}".format(expect_pv_power, recorded.get("pv_power")))
+        failed = True
+
+    # Control and battery args must not have widened along with the PV ones
+    for arg in ("soc_percent", "charge_limit", "reserve", "grid_power", "load_today", "load_power", "import_today", "export_today"):
+        entities = recorded.get(arg) or []
+        if len(entities) != 1 or no_batt.lower() in " ".join(entities):
+            print("ERROR: {} must stay on the battery inverter alone, got {}".format(arg, entities))
+            failed = True
+
+    if not any("in the PV totals" in message for message in api.log_messages):
+        print("ERROR: expected a log line naming the battery-less inverter(s) added to the PV totals")
+        failed = True
+
+    if not failed:
+        print("PASSED: automatic_config keeps a No Battery inverter in the PV totals")
+    return failed
+
+
+async def test_automatic_config_pv_totals_unchanged_when_all_have_batteries():
+    """With no PV-only inverter present the PV args must still be exactly the battery inverters."""
+    failed = False
+    print("**** Testing automatic_config leaves the PV totals alone when every inverter has a battery ****")
+
+    first = "1031260253072197"
+    second = "6031052256280133"
+    recorded, api = await _run_automatic_config({first: _DETAIL_WITH_BATTERY, second: _DETAIL_WITH_BATTERY_ALT_FIRMWARE})
+
+    if recorded.get("num_inverters") != 2:
+        print("ERROR: expected num_inverters 2, got {}".format(recorded.get("num_inverters")))
+        failed = True
+
+    expect_pv_today = [f"sensor.predbat_solis_{first.lower()}_pv_energy_total", f"sensor.predbat_solis_{second.lower()}_pv_energy_total"]
+    if recorded.get("pv_today") != expect_pv_today:
+        print("ERROR: expected pv_today {}, got {}".format(expect_pv_today, recorded.get("pv_today")))
+        failed = True
+
+    if any("in the PV totals" in message for message in api.log_messages):
+        print("ERROR: the PV-only inclusion must not be logged when there is no PV-only inverter")
+        failed = True
+
+    if not failed:
+        print("PASSED: automatic_config leaves the PV totals alone when every inverter has a battery")
+    return failed
+
+
 def run_solis_tests(my_predbat):
     """
     Run all Solis API tests
@@ -1372,6 +1443,8 @@ def run_solis_tests(my_predbat):
         failed |= asyncio.run(test_automatic_config_skips_no_battery_named_only_in_battery_list())
         failed |= asyncio.run(test_automatic_config_keeps_battery_when_battery_list_contradicts_battery_type())
         failed |= asyncio.run(test_automatic_config_keeps_real_battery_reporting_zero_soh())
+        failed |= asyncio.run(test_automatic_config_keeps_no_battery_inverter_in_pv_totals())
+        failed |= asyncio.run(test_automatic_config_pv_totals_unchanged_when_all_have_batteries())
         failed |= asyncio.run(test_read_cid())
         failed |= asyncio.run(test_read_batch())
         failed |= asyncio.run(test_read_and_write_cid())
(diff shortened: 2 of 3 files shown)


apps/predbat/solis.py
```diff
@@ -1455,6 +1455,16 @@ async def automatic_config(self):
         # Convert SNs to lowercase for entity naming
         devices = [sn.lower() for sn in batteries]
 
+        # PV totals are summed across every entry in the arg list, so an inverter with no battery still
+        # belongs in them - its generation is part of the same array (issue #4922). Only the control and
+        # battery args are battery-filtered; a PV-only inverter is still never written to. load/import/
+        # export stay battery-only deliberately: on a shared-CT site those registers can overlap between
+        # inverters, so summing them risks double-counting the house load.
+        pv_devices = [sn.lower() for sn in self.inverter_sn]
```
springfall2008, line 1463: **Finding (verified, plan-affecting):** `pv_devices` order interacts badly with the pad=False history consumers. `minute_data_import_export`/`minute_data_load` accumulate entities sequentially and the final merge loop only covers `range(60*24*days)` for the *last* entity's own history depth (fetch.py:674-676 sizes each `minute_data` call with that entity's `age_days`, utils.py accumulate block merges `accumulate` only within that window), so minutes older than the last entity's history are silently dropped from the merged dict.

On a Solis site the merged `pv_today` history is therefore truncated to the **last-listed** entity's recorder age. Two consumers hit this:
- solcast.py:871 (`pad=False`): if the PV-only inverter's `pv_energy_total` has <3 days of recorder history and comes after the battery inverter in `inverter_sn`, `pv_today_hist_days < 3` and PV calibration is silently disabled (solcast.py:899-902) - a site-wide plan behaviour change with only an info log.
- load_ml_component.py:385 (`pad=False`): `days_to_fetch = min(days_to_fetch, pv_minutes_age)` collapses the ML PV training window to the new entity's age.

Pre-PR, `pv_today` held only long-established battery-inverter sensors, so this path was stable. Suggest either fetching/merging each entity over its own full window (max, not last), or at least documenting the recorder-history dependency.

springfall2008, line 1463: **Finding (verified):** `pv_devices` preserves `self.inverter_sn` order, but every per-inverter arg in this function uses `devices` (battery-inverter) order, and inverter ids are indices into that order (execute.py:917-920 `for id in range(self.num_inverters): Inverter(self, id)`, read at inverter.py:1456 `get_arg("pv_power", index=self.id)`). If the PV-only inverter precedes a battery inverter in `inverter_sn`, the battery inverter reads the PV-only unit's sensor at its index and the last battery inverter's PV falls into the "Additional PVs without inverters" bucket (execute.py:1031-1037).

Fleet totals stay correct (every entry is summed exactly once), so the impact is misattribution in the per-inverter status/BALANCE debug logs - but the cheap fix removes the trap entirely and matches the existing convention: `pv_devices = devices + pv_only_devices` (gecloud.py:1242 does exactly `batteries + pvs`, battery inverters first).

apps/predbat/solis.py
```diff
@@ -1455,6 +1455,16 @@ async def automatic_config(self):
...
+        pv_only_devices = [device for device in pv_devices if device not in devices]
+        if pv_only_devices:
+            self.log("Solis API: Including {} inverter(s) with no battery in the PV totals: {}".format(len(pv_only_devices), ", ".join(pv_only_devices)))
```
springfall2008, line 1466: **Finding (verified, log accuracy):** this line is emitted unconditionally, but it can be wrong in two configs:

1. `solis_cloud_pv_load_ignore: true` - the `set_arg_auto("pv_today"/"pv_power", ...pv_devices...)` calls sit inside the `if not self.get_arg("solis_cloud_pv_load_ignore")` block below, so with the flag set the PV args come from apps.yaml and the named inverters are *not* included - yet the log has already told the user they were.
2. An inverter excluded from `batteries` because `batteryHealthSoh` is missing/unparseable (`battery_soh = None`, lines 1441-1447) still lands in `pv_only_devices` even when `batteryType` names a real pack - it gets labelled "with no battery" here, and this is the *only* trace of that exclusion (no "Skipping" log is emitted on that path, unlike the `_reports_no_battery` path at line 1437).

Suggest moving the log inside the ignore gate and rewording to cover both exclusion reasons.

apps/predbat/solis.py
springfall2008, line 1463: **Finding (verified, log noise):** widening the list adds the PV-only inverter's `pv_energy_total` to a `required=True` history fetch (fetch.py:897 -> minute_data_import_export defaults `required=True`). If that entity has no recorder history (HA recorder exclude rules, or the entity was only just created and `get_history_wrapper` returns None), fetch.py:657-659/693-698 emits `Warn: No history data found` / `Warn: Failure to fetch history` for it on **every 5-minute cycle**, indefinitely, and it counts as a fetch failure in `record_status`. Worth either noting in the docs that the entity must be recorder-stored, or fetching it with `required=False` since it only improves the PV total.

apps/predbat/solis.py
```diff
@@ -1471,9 +1481,9 @@ async def automatic_config(self):
         # if solis_cloud_pv_load_ignore is set to true, override Solis cloud sensors and use load/pv_today/power entries defined in apps.yaml
         if not self.get_arg("solis_cloud_pv_load_ignore", default=False):
             self.set_arg_auto("load_today", [f"sensor.{self.prefix}_solis_{device}_total_load_energy" for device in devices])
-            self.set_arg_auto("pv_today", [f"sensor.{self.prefix}_solis_{device}_pv_energy_total" for device in devices])
+            self.set_arg_auto("pv_today", [f"sensor.{self.prefix}_solis_{device}_pv_energy_total" for device in pv_devices])
             self.set_arg_auto("load_power", [f"sensor.{self.prefix}_solis_{device}_load_power" for device in devices])
-            self.set_arg_auto("pv_power", [f"sensor.{self.prefix}_solis_{device}_pv_power" for device in devices])
+            self.set_arg_auto("pv_power", [f"sensor.{self.prefix}_solis_{device}_pv_power" for device in pv_devices])
```
springfall2008, line 1486: **Finding (plausible, display-only):** this widening feeds the PV-only inverter's `pac` into the fleet `pv_power` total (execute.py:1031-1035) while `load_power`/`grid_power` deliberately stay battery-CT-only. On exactly the shared-CT topology the new comment cites, the PV-only inverter's output is already netted inside the battery inverter's load/grid CT readings, so the web power-flow diagram (web.py:600-615, which reconciles those four values) now counts that generation once in the PV circle *and* implicitly as reduced load/grid - the unexplained residual grows by the PV-only inverter's full output whenever it generates. The PR comment applies the double-counting argument to load but not to the PV side it widens; worth a sentence in the docs, or accepting it as cosmetic.

docs/components.md
```diff
@@ -1177,6 +1177,7 @@ Integrates with Solis inverters for monitoring and controlling Solis battery sys
     - Leave `soc_max` unset and allow Predbat to automatically detect battery size from historical charging data (requires several days of data)
 - Supports both V1 (older firmware) and V2 (newer firmware) time window formats
 - Automatic configuration available - sets up all required Predbat sensors automatically
+- **PV-only inverters**: an inverter Solis Cloud reports as having no battery is never managed as a battery inverter and is never written to, but its generation is still included in `pv_today` and `pv_power` so the array total covers the whole roof. Its load and grid readings are left out, as those registers can overlap with the battery inverter's on a shared-CT installation - set `pv_today`/`pv_power`/`load_today` manually with `solis_cloud_pv_load_ignore: true` if you need something different
```
springfall2008, line 1180: **Finding (docs precision, verified):** two overclaims in this bullet:

1. Following the bullet's own advice - `solis_cloud_pv_load_ignore: true` - gates the entire widened block (solis.py:1482-1486), so it silently reverts `pv_today`/`pv_power` to whatever apps.yaml says, i.e. it *disables* the feature the bullet describes. It also gates `load_power`/`pv_power`, which the bullet omits.
2. "so the array total covers the whole roof" holds only for mixed sites: when *every* discovered inverter is battery-less, `automatic_config` returns at the `num_inverters == 0` guard before any `set_arg_auto`, so a pure-PV Solis site still gets no PV args at all (pre-existing behaviour, but the new sentence overclaims).

Suggest scoping the sentence to sites with at least one battery inverter and noting the all-PV case needs manual `pv_today`/`pv_power`.

apps/predbat/tests/test_solis.py
```diff
@@ -1348,6 +1348,77 @@ async def test_automatic_config_keeps_real_battery_reporting_zero_soh():
...
+    Predbat sums pv_today/pv_power across the whole arg list, so a second inverter's generation is
+    simply added to the array total - dropping it hid ~45% of the generation on the reporting site.
+    load/import/export stay battery-only: those registers can overlap on a shared-CT install.
+    """
+    failed = False
+    print("**** Testing automatic_config keeps a No Battery inverter in the PV totals ****")
+
+    with_batt = "1031260253072197"
+    no_batt = "6031042245160206"
+    recorded, api = await _run_automatic_config({with_batt: _DETAIL_WITH_BATTERY, no_batt: _DETAIL_NO_BATTERY})
```
springfall2008, line 1363: **Finding (test gaps, verified):** the new tests green-light the two cases where the widening behaves least obviously:

1. `_run_automatic_config` sets `api.inverter_sn = list(details.keys())`, and both tests put the battery inverter **first** (`{with_batt: ..., no_batt: ...}`, line 1363), so `pv_devices` order coincides with the battery-inverter index order. A test with the PV-only inverter first (`{no_batt: ..., with_batt: ...}`) is the one case where per-inverter index reads (inverter.py:1456) diverge from the arg order - currently untested.
2. No test exercises `solis_cloud_pv_load_ignore: true` (zero hits in the file; `MockSolisAPI.get_arg` returns the default False for every test), so the path where pv args come from apps.yaml while the new log still fires is uncovered.
3. `test_automatic_config_pv_totals_unchanged_when_all_have_batteries` asserts `pv_today` but never `pv_power`, so a regression that widens only the adjacent `pv_power` line would pass the whole suite.

Also worth asserting per-inverter alignment at the inverter.py read level so the ordering question is pinned, not inferred.

apps/predbat/solis.py
springfall2008, line 1463: **Nit (schema drift):** `APPS_SCHEMA` still declares `"pv_power": {"type": "sensor_list", "entries": "num_inverters"}` (config.py:2580), but Solis auto-config now intentionally produces more entries than `num_inverters` (validation only errors when the list is *shorter*, predbat.py:1453, so nothing breaks today). The schema no longer describes reality for Solis auto-config sites; worth a comment on the schema entry (or `optional_entries`/docs note) so a future tightening of that check doesn't silently break this feature.

PR #4926, opened 2026-09-04
Sigenergy: redact accessToken from the MQTT publish log line

### LINKED ISSUE #4920: Sigenergy: MQTT accessToken logged in plaintext
## Description

`SigenergyComponent._publish_mqtt()` logs the full MQTT payload dict, including the plaintext `accessToken` used as the MQTT password. This token is written to the app log on every mode-switch and battery command.

https://github.com/springfall2008/batpred/blob/main/apps/predbat/sigenergy.py#L1037

```python
await client.publish(topic, payload=json.dumps(payload_dict), qos=1)
self.log("SigenergyAPI: MQTT published to {} - {}".format(topic, payload_dict))
```

Both callers put a live `accessToken` directly into `payload_dict` before it reaches this log line:

- `set_operating_mode()` — https://github.com/springfall2008/batpred/blob/main/apps/predbat/sigenergy.py#L884-L888
- `send_battery_command()` — https://github.com/springfall2008/batpred/blob/main/apps/predbat/sigenergy.py#L1102-L1110

Example log output (token redacted here, but not in the real log):

```
SigenergyAPI: MQTT published to openapi/instruction/command - {'accessToken': '<REDACTED>', 'commands': [{'systemId': '...', 'activeMode': 'selfConsumption', 'startTime': 1788373857, 'duration': 720}]}
```

Since this fires on every export/charge mode change, the same live token ends up in the log repeatedly, in plaintext, for as long as it's valid.

## Suggested fix

Redact known-sensitive keys before logging the payload, e.g. mirroring the existing pattern in `deye.py`/`deye_const.py` (`DEYE_DEBUG_REDACT_KEYS` + `DeyeAPI.redact()`), which already masks `accessToken`/`refreshToken`/`password`/etc. for the Deye integration:

https://github.com/springfall2008/batpred/blob/main/apps/predbat/deye_const.py#L110
https://github.com/springfall2008/batpred/blob/main/apps/predbat/deye.py#L185

## Impact

Anyone with read access to a Predbat instance's logs (log files, log aggregation, support bundles, etc.) can extract a live Sigenergy `accessToken` and use it to issue commands against the account's Sigenergy system(s) until the token expires/rotates.

springfall2008:
This is an automated draft PR generated from issue #4920 — a maintainer should review it before merging.

Fixes #4920

## Summary

`SigenergyAPI._publish_mqtt()` logged the full command payload dict, which both callers (`set_operating_mode()` and `send_battery_command()`) populate with the live `accessToken` — the same token used as the MQTT broker password. Since MQTT commands fire on every mode switch and battery command, the token landed in the app log in plaintext, repeatedly, for as long as it was valid.

This adds a `SigenergyAPI.redact()` static method and a `SIGENERGY_LOG_REDACT_KEYS` tuple, mirroring the existing `DeyeAPI.redact()` / `SunsynkAPI.redact()` pattern the report pointed at, and applies it to the one leaking log line. Redaction is log-only — the real token still goes to the broker on the wire. `redact()` recurses into dicts and lists because the battery command payload nests its per-system commands one level down inside a list.

Scope is deliberately limited to that log line: the other token paths were checked and are clean today — the subscription publishes (`sigenergy.py:1465`) log only a count, `get_access_token()` logs only `code`/`msg`/`expiresIn`, and `_request()` sends the token in the `Authorization` header rather than in the `params`/`json_data` it logs.

## Testing

- `tools/triage_test.sh sigenergy` — all Sigenergy tests pass, including two new ones:
  - `test_sigenergy_redact` — masking at top level and nested inside a list, every key in the redact list, and scalar/list/None passthrough.
  - `test_sigenergy_publish_mqtt_redacts_token` — the broker still receives the real token while the log line contains `<redacted>` and not the token, non-credential payload content (`systemId`) is still logged, and the caller's payload dict is not mutated.
- Confirmed the new test is not vacuous: with the redaction call reverted it fails with `Token must not appear in the log`, and passes again once restored.
- `coverage/run_pre_commit` — all hooks pass (ruff, black, cspell, markdownlint) and the quick test suite is green.

### DIFF
--- a/apps/predbat/sigenergy.py
+++ b/apps/predbat/sigenergy.py
@@ -146,6 +146,11 @@ SIGENERGY_MQTT_TOPIC_ALARM = "openapi/alarm/{app_key}/{system_id}"      # alarm 
 SIGENERGY_MQTT_TOPIC_COMMAND = "openapi/instruction/command"            # battery command publish
 SIGENERGY_MQTT_TOPIC_MODE = "openapi/instruction/mode"                  # V1 operating mode switch (MQTT)
 
+# Payload keys masked before a payload is written to the log. The MQTT command payloads
+# carry the live accessToken (it doubles as the MQTT broker password), and Predbat logs are
+# routinely pasted into GitHub issues, so anything credential-bearing has to be masked first.
+SIGENERGY_LOG_REDACT_KEYS = ("accessToken", "refreshToken", "appKey", "appSecret", "password", "token", "key")
+
 # Operating mode enums (REST mode switch endpoint — MSC and FFG only; NBI is not used)
 SIGENERGY_MODE_MSC = 0   # Maximum Self-Consumption (eco)
 SIGENERGY_MODE_FFG = 5   # Fully Feed-in to Grid
@@ -1020,6 +1025,20 @@ class SigenergyAPI(ComponentBase):
         self._tls_context = tls_context
         return tls_context
 
+    @staticmethod
+    def redact(payload):
+        """Return payload with credential-bearing keys masked, for safe logging.
+
+        Recursive over dicts and lists, as deye.py's and sunsynk.py's redact are: the MQTT
+        command payload nests its per-system commands one level down inside a list, so a
+        top-level-only rewrite would still leak anything a future payload carries there.
+        """
+        if isinstance(payload, dict):
+            return {key: ("<redacted>" if key in SIGENERGY_LOG_REDACT_KEYS else SigenergyAPI.redact(value)) for key, value in payload.items()}
+        if isinstance(payload, list):
+            return [SigenergyAPI.redact(value) for value in payload]
+        return payload
+
     async def _publish_mqtt(self, topic, payload_dict):
         """Publish a JSON payload to the Sigenergy MQTT broker.
 
@@ -1047,7 +1066,7 @@ class SigenergyAPI(ComponentBase):
                 keepalive=30,
             ) as client:
                 await client.publish(topic, payload=json.dumps(payload_dict), qos=1)
-            self.log("SigenergyAPI: MQTT published to {} - {}".format(topic, payload_dict))
+            self.log("SigenergyAPI: MQTT published to {} - {}".format(topic, self.redact(payload_dict)))
             return True
         except Exception as e:
             self.log("Warn: SigenergyAPI: MQTT publish to {} failed: {}".format(topic, e))
--- a/apps/predbat/sigenergy.py
+++ b/apps/predbat/sigenergy.py
@@ -1033,7 +1033,7 @@ class SigenergyAPI(ComponentBase):
         Sigenergy MQTT command payloads nest per-system commands one level down inside a list,
         so a top-level-only rewrite would still leak anything a future payload carries there.
         Tuples are included because json.dumps() serialises tuples as JSON arrays.
-        Sets/frozensets are handled for log safety, even though json.dumps() does not
+        Sets/frozen sets are handled for log safety, even though json.dumps() does not
         serialise them by default.
         """
         if isinstance(payload, dict):
--- a/apps/predbat/sigenergy.py
+++ b/apps/predbat/sigenergy.py
@@ -1029,12 +1029,12 @@ class SigenergyAPI(ComponentBase):
     def redact(payload):
         """Return payload with credential-bearing keys masked, for safe logging.
 
-        Recursive over dicts, lists, tuples and sets, as deye.py's and sunsynk.py's redact are
-        over dicts and lists: the MQTT command payload nests its per-system commands one level
-        down inside a list, so a top-level-only rewrite would still leak anything a future
-        payload carries there. Sequences beyond list are covered because json.dumps() serialises
-        tuples and sets as JSON arrays too — redacting only lists would let a tuple-shaped
-        payload be published as an array yet logged unmasked.
+        Recursive over dicts and common sequences. DeyeAPI/SunsynkAPI recurse over dicts + lists;
+        Sigenergy MQTT command payloads nest per-system commands one level down inside a list,
+        so a top-level-only rewrite would still leak anything a future payload carries there.
+        Tuples are included because json.dumps() serialises tuples as JSON arrays.
+        Sets/frozensets are handled for log safety, even though json.dumps() does not
+        serialise them by default.
         """
         if isinstance(payload, dict):
             return {key: ("<redacted>" if key in SIGENERGY_LOG_REDACT_KEYS else SigenergyAPI.redact(value)) for key, value in payload.items()}
--- a/apps/predbat/sigenergy.py
+++ b/apps/predbat/sigenergy.py
@@ -471,7 +471,7 @@ class SigenergyAPI(ComponentBase):
             "Content-Type": "application/json",
         }
 
-        self.log("Requesting {} {} with params={} json={}".format(method, path, params, json_data))
+        self.log("Requesting {} {} with params={} json={}".format(method, path, self.redact(params), self.redact(json_data)))
 
         for attempt in range(retries):
             await self._enforce_rate_limit()
@@ -513,7 +513,7 @@ class SigenergyAPI(ComponentBase):
                             self.log("Warn: SigenergyAPI: Failed to decode response from {}: {}".format(path, e))
                             return None
 
-                        self.log("SigenergyAPI: Response from {} {}: {}".format(method, path, body))
+                        self.log("SigenergyAPI: Response from {} {}: {}".format(method, path, self.redact(body)))
 
                         code = body.get("code", -1)
                         if code != 0:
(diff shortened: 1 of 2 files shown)


apps/predbat/sigenergy.py
```diff
@@ -146,6 +146,11 @@
 SIGENERGY_MQTT_TOPIC_COMMAND = "openapi/instruction/command"            # battery command publish
 SIGENERGY_MQTT_TOPIC_MODE = "openapi/instruction/mode"                  # V1 operating mode switch (MQTT)
 
+# Payload keys masked before a payload is written to the log. The MQTT command payloads
+# carry the live accessToken (it doubles as the MQTT broker password), and Predbat logs are
+# routinely pasted into GitHub issues, so anything credential-bearing has to be masked first.
+SIGENERGY_LOG_REDACT_KEYS = ("accessToken", "refreshToken", "appKey", "appSecret", "password", "token", "key")
```
springfall2008, line 152: **Same bug class on the inbound path — the topic embeds `app_key` and the listener logs it raw.**

The listener logs the full `topic_str` on every received message (`sigenergy.py:1521`) and in the non-JSON warning (`sigenergy.py:1507`), and those topics embed `app_key` (`SIGENERGY_MQTT_TOPIC_CHANGE/PERIOD/ALARM`, `sigenergy.py:148-150`) — a key this very list classifies as credential-bearing (it is the MQTT broker username and half of the base64 login key). So a user pasting a debug log still leaks it on every inbound line; the fix covers only the outbound publish line.

Worth either `redact()`-ing the topic (or the format args) in those two listener lines as part of this PR, or an explicit follow-up. The CLI tools (`test_sigenergy_api`, `test_mqtt_connection`) deliberately print only a 10-char `app_key` prefix, but the listener lines print the full topic.

Separately: the `key`/`token` catch-alls here will mask benign diagnostic fields in what is often the only log line for debugging a broker-side rejection (an alarm/instruction entry carrying a literal `key` id renders as `<redacted>`), while the match is case-exact, so a snake_case `access_token` variant would slip through. Sibling lists cover both directions (Deye adds `tokenHash`; Sunsynk covers snake_case plus `Authorization`/`sign`). Worth a conscious decision on both edges.

apps/predbat/sigenergy.py
```diff
@@ -1047,7 +1066,7 @@ async def _publish_mqtt(self, topic, payload_dict):
                 keepalive=30,
             ) as client:
                 await client.publish(topic, payload=json.dumps(payload_dict), qos=1)
-            self.log("SigenergyAPI: MQTT published to {} - {}".format(topic, payload_dict))
+            self.log("SigenergyAPI: MQTT published to {} - {}".format(topic, self.redact(payload_dict)))
```
springfall2008, line 1072: `redact()` now guards exactly one log line. `_request()` still logs `params`/`json_data` verbatim (`sigenergy.py:474`) and the full parsed response body (`sigenergy.py:530`) — both clean today (no current caller passes a credential, no current body carries one, and the token goes in the `Authorization` header which is never logged), but any future endpoint that accepts or echoes a token-bearing body re-opens the leak through a site the new invariant doesn't cover.

Applying `redact()` at those two lines too (or routing all payload/body logging through one helper) would make the guarantee hold by construction rather than by call site — this is the same bandaid pattern that regressed once already in a sibling integration.

apps/predbat/sigenergy.py
```diff
@@ -1020,6 +1025,20 @@ def _build_tls_context(self):
...
+
+        Recursive over dicts and lists, as deye.py's and sunsynk.py's redact are: the MQTT
+        command payload nests its per-system commands one level down inside a list, so a
+        top-level-only rewrite would still leak anything a future payload carries there.
+        """
+        if isinstance(payload, dict):
+            return {key: ("<redacted>" if key in SIGENERGY_LOG_REDACT_KEYS else SigenergyAPI.redact(value)) for key, value in payload.items()}
+        if isinstance(payload, list):
+            return [SigenergyAPI.redact(value) for value in payload]
+        return payload
```
springfall2008, line 1043: Minor latent edge: this branch recurses `list` only, but `json.dumps()` serialises tuples as JSON arrays too — so a future payload with `{"commands": ({...},)}` would be *published* as an array yet *logged* unmasked, silently breaking the "redaction is log-only, content-equal" assumption. Cheap fix: accept any non-str/bytes sequence.

Also worth noting this is now the fourth hand-rolled `redact` copy (deye, sunsynk, alphaess — the last non-recursive) with four independently-drifting key tuples; a single parameterised helper in `utils.py` next to `mask_secret_args` would make the next integration correct by default. Not a blocker for this PR, which deliberately mirrors the existing pattern.

apps/predbat/tests/test_sigenergy.py
```diff
@@ -1343,6 +1343,66 @@ def test_sigenergy_publish_mqtt_success(my_predbat):
...
+    assert redacted["accessToken"] == "<redacted>", "accessToken masked"
+    assert redacted["systemId"] == "SIG1", "Non-credential key untouched"
+
+    # Nested inside a list, as the battery command payload nests its commands
+    nested = SigenergyAPI.redact({"commands": [{"systemId": "SIG1", "password": "hunter2"}]})
+    assert nested["commands"][0]["password"] == "<redacted>", "Credential masked inside a nested list"
+    assert nested["commands"][0]["systemId"] == "SIG1", "Nested non-credential key untouched"
+
+    # Every documented credential key is covered
+    for key in ("accessToken", "refreshToken", "appKey", "appSecret", "password", "token", "key"):
```
springfall2008: Two test-strength gaps:

1. This loop re-hardcodes the 7-key tuple instead of importing `SIGENERGY_LOG_REDACT_KEYS`, so a key *added* to the constant later is silently untested — the "every documented credential key is covered" guarantee only holds for today's snapshot. Iterating the constant closes that.

2. The nesting case covered is dict-in-list; dict-in-dict (`{"outer": {"accessToken": "x"}}`) never exercises the dict-value recursion branch, so deleting that recursion would leave the suite green. One added case covers it.

apps/predbat/tests/test_sigenergy.py
```diff
@@ -1343,6 +1343,66 @@ def test_sigenergy_publish_mqtt_success(my_predbat):
...
+def test_sigenergy_publish_mqtt_redacts_token(my_predbat):
+    """Test _publish_mqtt keeps the live token on the wire but masks it in the log (#4920)."""
+    failed = False
+    api = MockSigenergyAPI()
+    api.access_token = "tok123"
+    api.mqtt_host = "openapi-eu.sigencloud.com" # cspell:disable-line
+    api.mqtt_port = 8883
+
+    mock_client = _make_mock_aiomqtt_client()
+    payload = {"accessToken": "live-secret-token", "commands": [{"systemId": "SIG1", "activeMode": "charge"}]}
```
springfall2008: The nested `commands[0]` here carries no credential key, so both tests would stay green if `_publish_mqtt` regressed to a *top-level-only* redaction — yet the exact shape of the #4920 bug (a credential nested inside `commands`) is never asserted through the publish path; `test_sigenergy_redact` covers nesting but tests the function directly, not the wiring. Adding e.g. a `"password"` to `commands[0]` and asserting it is masked in the published log line closes that gap.

Minor: `topic` is unpacked from `mock_client.publishes[0]` but never asserted (the sibling `publish_mqtt_success` test does assert it), so a topic swap would also pass this test.

apps/predbat/sigenergy.py
springfall2008: Automated reply from the triage bot (pr-cleanup).

Fixed in 0b92ad00: both listener lines now log a `safe_topic` with `app_key` masked (`topic_str.replace(self.app_key, "<redacted>")`, computed once after topic parsing and used for the non-JSON warning and the message line). I redacted only the topic, not the `value_dict` format arg, on purpose: inbound `value` payloads are device telemetry, and running them through `redact()` would amplify the over-masking problem flagged below onto every inbound state log line for no credential risk.

On the two edges of the key list — conscious decision, unchanged: the catch-alls stay, because the failure direction is asymmetric (a benign field rendering as `<redacted>` costs one diagnostic detail; a missed credential leaks it permanently), and case-exact matching stays to mirror the sibling integrations (deye/sunsynk are also exact-match). Verified the Sigenergy API is camelCase throughout (`accessToken`/`refreshToken` in the login/refresh responses, camelCase in the value dicts), so the snake_case variant has no live path here. Worth revisiting only if the shared-helper refactor in the third thread ever lands.

apps/predbat/sigenergy.py
springfall2008: Automated reply from the triage bot (pr-cleanup).

Done in 0b92ad00: both `_request()` sites now route through `redact()` — the `Requesting ... params=... json=...` line redacts `params` and `json_data`, and the `Response from ...` line redacts the parsed body. That covers the future-endpoint case: if any caller ever passes or echoes a token-bearing body, the log stays safe without a new call site having to remember.

Note the "clean today" assessment still holds after re-checking: `get_access_token()` builds its own raw `aiohttp` session and never logs the `{"key": base64(AppKey:AppSecret)}` payload, so no current `_request()` caller passes a credential — this change makes the invariant hold by construction rather than fixing an active leak.

Covered by a new registered test `test_sigenergy_request_log_redacts_credentials`, which calls `_request()` with a credential in `params`, in `json_data`, and in the mocked response body, and asserts each is `<redacted>` in the log while non-credential content survives. (The test patches `SIGENERGY_MIN_REQUEST_INTERVAL` to 0 so it doesn't sleep through the 10 req/min limiter.)

The shared-helper idea is noted but not part of this PR, same as the fourth thread.

apps/predbat/sigenergy.py
springfall2008: Automated reply from the triage bot (pr-cleanup).

Fixed in 0b92ad00: the recursion branch now accepts `(list, tuple, set, frozenset)` and returns a list, so any sequence `json.dumps()` would serialise as a JSON array is redacted before logging. Docstring updated to state why (publish-as-array/log-unmasked divergence), and `test_sigenergy_redact` gained a tuple case (`{"commands": ({"password": ...},)}` asserted masked).

Agreed the shared parameterised helper in `utils.py` is the right eventual shape — that's now four drifting copies across integrations — but as noted it's out of scope for this PR, which deliberately follows the existing per-integration pattern.

apps/predbat/tests/test_sigenergy.py
springfall2008: Automated reply from the triage bot (pr-cleanup).

Both closed in 0b92ad00:

1. The loop now iterates the imported `SIGENERGY_LOG_REDACT_KEYS` constant (with a comment stating that keys added later are automatically covered) instead of a re-hardcoded tuple.

2. Added a dict-in-dict case asserting `{"outer": {"accessToken": "x"}}` masks the nested value — so deleting the dict-value recursion branch now fails the suite.

apps/predbat/tests/test_sigenergy.py
springfall2008: Automated reply from the triage bot (pr-cleanup).

Both closed in 0b92ad00:

1. `commands[0]` in the publish-path test now carries a `password` (with a comment saying it mirrors the shape of the #4920 leak), and the test asserts both that the broker still receives the nested value and that it is absent from the published log line — a regression to top-level-only redaction now fails the test, not just the direct-function one.

2. `topic` is now asserted to equal `openapi/instruction/command`.

PR #4929, opened 2026-09-04
fix(gecloud): honour battery_rate_max over charge_rate's max attribute

### LINKED ISSUE #4908: GEC (GivEnergy Cloud) with ge_cloud_automatic: true uses hardcoded 2600W default instead of correctly-fetched 20000W inverter capability (3-phase HV inverter)
[apps.yaml.txt](https://github.com/user-attachments/files/31745251/apps.yaml.txt)
[predbat.log](https://github.com/user-attachments/files/31745252/predbat.log)
[predbat_debug.yaml.txt](https://github.com/user-attachments/files/31745253/predbat_debug.yaml.txt)

**Describe the bug**
With **ge_cloud_automatic: true**, Predbat charges/discharges/plans as though the inverter is limited to 2600W, even though GivEnergy's own Cloud API correctly reports this inverter's real maximum rate as 20000W, and Predbat's own derived sensors correctly reflect that value too.

Log evidence from the same startup sequence:

```text
2026-09-02 06:13:26.541216: GECloud: Updated data for device tj2515g023: battery capacity 47.92kWh, max charge rate 20000W, max inverter rate 20000W
...
2026-09-02 06:15:00.762327: Inverter 0 with soc_max 40.73kWh, nominal_capacity 47.92kWh, battery rate raw 2600.0W, charge rate 2.6kW, charge rate DC 30.0kW, discharge rate 2.6kW, battery_rate_min 0.0W, AC limit 20.0kW, export limit 11.0kW, reserve 4%, current_reserve 5%, temperature 38°C
```

Under two minutes apart, in the same log, for the same device: GECloud correctly fetches and logs 20000W for both max charge rate and max inverter rate, but the inverter initialisation immediately afterwards shows battery rate raw 2600.0W / charge rate 2.6kW / discharge rate 2.6kW, while AC limit 20.0kW on the same line is correct.

This is confirmed independently by the debug config snapshot, which shows Predbat's own derived sensors already holding the correct value at the time:

```yaml
sensor.predbat_gecloud_tj2515g023_max_charge_rate:
  state: 20000
sensor.predbat_gecloud_tj2515g023_max_inverter_rate:
  state: 20000
sensor.predbat_gecloud_tj2515g023_model:
  state: GIV-3HY-20
  attributes:
    details:
      info:
        max_charge_rate: 20000
        max_discharge_rate: 20000
        model: GIV-3HY-20
```

So the correct 20000W value is present and available via GECloud's own device data and Predbat's own sensors, but something in the **ge_cloud_automatic: true** path substitutes what looks like the default 2600W value instead.

**Expected behaviour**
With **ge_cloud_automatic: true**, Predbat should use the inverter's actual reported max charge/discharge rate (20000W here) rather than defaulting to 2600W.

**Workaround found:**
Setting **ge_cloud_automatic: false** and manually specifying inverter_limit, battery_rate_max, etc. as literal values in apps.yaml resolves the charge/discharge rate issue, but introduces a separate bug (reported separately) where Predbat attempts an invalid select.turn_on service call against a legacy select.givtcp_{serial}_mode entity.

**Predbat version**
v8.54.2

**Environment details**
GivEnergy 3-phase HV inverter, model GIV-3HY-20.
Controlled via GivEnergy Cloud Direct (GEC) 
ge_cloud_automatic: true 
ge_cloud_data: false (local GivTCP used for load/import/export data only).

**Screenshots**
None

**Log file**
Attached together with to the best of my knowledge the apps.yaml from the same period.

kennedy-nic:
- GEC/GEE (GivEnergy Cloud) inverters always derived their max charge/discharge rate from the `charge_rate` entity's `max` attribute, which GivEnergy's cloud API can report as a stale/low value (2600W on a 20kW 3-phase inverter) instead of the real capability.
- `gecloud.py` already computes and publishes an accurate `battery_rate_max` sensor from the device's model/capability data, but it was being silently ignored for GEC/GEE.
- `battery_rate_max` now takes priority over `charge_rate`'s `max` attribute when configured, fixing the incorrect 2600W cap.

**Precedent:** this brings GE-family inverters in line with how `battery_rate_max` already works for AlphaESS, which is documented as an override, in watts, for the battery's max charge/discharge rate specifically because its API doesn't reliably report one. Same situation here – an untrustworthy API-reported value with a manual correction – so this reuses the existing key rather than introducing a new, differently-named one for the same job.

**Who this affects:** I checked all three default GivEnergy templates (`givenergy_givtcp.yaml`/GE, `givenergy_cloud.yaml`/GEC, `givenergy_ems.yaml`/GEE) and none of them set `battery_rate_max`, commented or otherwise – the GEE template in fact states that apps.yaml entries in that section are ignored entirely in favour of Cloud auto-configuration. So only a user who has deliberately added `battery_rate_max` themselves (as this fix requires) would see a behaviour change; anyone on a stock template is unaffected regardless of inverter type.

Fixes #4908

## Test plan
- [x] Added `test_gecloud_battery_rate_max_priority` regression test (GEC uses the `battery_rate_max` sensor; GE still falls back to `charge_rate`'s max attribute)
- [x] `./run_all --test inverter` passes
- [x] `./run_pre_commit` – passes except two pre-existing failures confirmed to reproduce identically on a clean `main` tree, unrelated to this change (a Windows-only `triage-daemon-tests` POSIX-path bug, and a `file-contents-sorter` CRLF no-op on `requirements.txt`)
- [x] `./run_all --quick` – no regressions from this change; two further pre-existing failures confirmed environment-specific to this fresh Windows setup (a missing `encoding="utf-8"` in an unrelated test helper, and a `debug_cases` golden-plan mismatch caused by the compiled prediction kernel not being built here)

https://claude.ai/code/session_01Djbv5fZh1Nr7MEYF7AqznT

### DIFF
--- a/apps/predbat/inverter.py
+++ b/apps/predbat/inverter.py
@@ -406,10 +406,14 @@ class Inverter:
             self.nominal_capacity = self.base.get_arg("soc_max", default=0.0, index=self.id)
             self.soc_max = self.nominal_capacity * self.battery_scaling
 
-            if self.inverter_type in ["GE", "GEC", "GEE"]:
-                self.battery_rate_max_raw = self.base.get_arg("charge_rate", attribute="max", index=self.id, default=2600.0, required_unit="W")
-            elif "battery_rate_max" in self.base.args:
+            if "battery_rate_max" in self.base.args:
+                # battery_rate_max, when configured, is a more authoritative source than the charge_rate
+                # entity's max attribute - e.g. GECloud (GEC/GEE) always populates it from the device's
+                # reported model/capability data, whereas the charge_rate entity's max attribute there
+                # reflects a live register limit that can read back as a stale/wrong low value (#4908).
                 self.battery_rate_max_raw = self.base.get_arg("battery_rate_max", index=self.id, default=2600.0, required_unit="W")
+            elif self.inverter_type in ["GE", "GEC", "GEE"]:
+                self.battery_rate_max_raw = self.base.get_arg("charge_rate", attribute="max", index=self.id, default=2600.0, required_unit="W")
             else:
                 self.battery_rate_max_raw = 2600.0
 
--- a/apps/predbat/tests/test_inverter.py
+++ b/apps/predbat/tests/test_inverter.py
@@ -2760,6 +2760,54 @@ def test_inverter_type_default_warning(test_name, my_predbat):
     return failed
 
 
+def test_gecloud_battery_rate_max_priority(test_name, my_predbat):
+    """
+    GECloud (GEC/GEE) always publishes an accurate battery_rate_max sensor derived from the
+    device's reported model/capability data. The charge_rate entity's max attribute for GEC
+    instead reflects a live register limit that can read back low/stale (e.g. 2600W on a 20kW
+    3-phase inverter, despite GECloud's own logs correctly showing 20000W) - issue #4908.
+    battery_rate_max must take priority over charge_rate's max attribute when both are configured,
+    while GE (local GivTCP), which never sets battery_rate_max, keeps using charge_rate's max
+    attribute as before.
+    """
+    failed = False
+    print("**** Running Test: {} ****".format(test_name))
+
+    ha = my_predbat.ha_interface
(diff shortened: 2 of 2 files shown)


springfall2008:
Fixed a slightly different way here: https://github.com/springfall2008/batpred/pull/4954

PR #4934, opened 2026-09-04
fix(chargers): isolate car plans and preserve composed measurements

mgazza:
A charger moved to a different slot by #4928 can consume the previous occupant's published plan. Track the registry allocation version from configuration fetch through plan publication, and make all four charger controls wait for a plan built from the current mapping. Also keep each car's published window list separate: the previous list was shared across loop iterations, so later cars inherited earlier cars' windows.

Fix two measurement regressions: Ohme retains its energy and power contribution when another component registers first or discovery repeats; legacy and discovered measurements of the same charger are counted once. Matching uses the existing exact-entity and bounded manufacturer-serial rules, and legacy measurements are restored when discovery releases them.

Validation:
- The full `unit_test.py --quick` suite passed (four slow tests skipped).
- All seven affected component/registry suites passed, including 62 registry regressions and 97 Ohme tests.
- All pre-commit hooks for changed files passed.
- Both energy regressions were reproduced against the original PR. New tests cover each control loop waiting after reassignment and rejecting a plan published for an older allocation.

Slot numbering still follows the ordering documented in #4928; adding a charger may require adjusting per-car settings. This follow-up targets `feat/charger-registry`, keeping the review fixes separate from the original feature.

### DIFF
--- a/.cspell/custom-dictionary-workspace.txt
+++ b/.cspell/custom-dictionary-workspace.txt
@@ -408,6 +408,7 @@ nord
 nordpool
 noreferrer
 nosc
+nosuch
 nosuchrun
 notgithub
 notsn
--- a/apps/predbat/charger_registry.py
+++ b/apps/predbat/charger_registry.py
@@ -1,0 +1,553 @@
+# -----------------------------------------------------------------------------
+# Predbat Home Battery System
+# Copyright Trefor Southwell 2026 - All Rights Reserved
+# This application maybe used for personal use only and not for commercial use
+# -----------------------------------------------------------------------------
+# fmt off
+# pylint: disable=consider-using-f-string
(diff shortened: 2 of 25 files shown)


mgazza:
Superseded by #4928 — please close.

These fixes are now folded into #4928 (`569f3255`) rather than stacked on it. #4928 hadn't been reviewed yet, so keeping them separate would have meant offering a two-deep stack whose base had four known defects — a worse review surface, not a better one.

Everything here is in #4928: the registry allocation version carried from configuration fetch through to plan publication so no control loop acts on a plan built for an older mapping, the per-car published window list isolation, Ohme's retained energy and power contribution, and the legacy/discovered measurement dedupe — with their regressions. The branch is also rebased onto current main.

PR #4938, opened 2026-09-05
fix(config): trust APPS_SCHEMA float type over an int default in get_arg() (#4925)

### LINKED ISSUE #4925: solcast_poll_hours 4.8 is silently truncated to 4h, exhausting the Solcast daily quota
Found and drafted with help of GLM 5.3 Flash in the new AI chat. 

[predbat (6).log](https://github.com/user-attachments/files/31807237/predbat.6.log)

**Describe the bug**

`UserInterface.get_arg()` coerces the resolved value to the *Python type of the default it was given*:
elif isinstance(default, int) and not isinstance(default, bool):
  Convert to int?
value = int(float(value)) # userinterface.py — truncates 4.8 -> 4

Component args are resolved through this path with the defaults declared in `COMPONENT_LIST` (components.py). `solcast_poll_hours` is declared `"default": 8` (a bare int), so `get_arg("solcast_poll_hours", 8)` returns `int(float(4.8)) = 4`, and `solcast.py` computes `max_age = 4 * 60 = 240` minutes.

With two Solcast sites this causes 6 refreshes/day (12 API calls) instead of the documented 5/day (10 calls), which exhausts the hobbyist 10 poll/day limit and produces nightly HTTP 429 failures. Validation doesn't catch this because it validates the raw apps.yaml value (a valid float, per APPS_SCHEMA `"type": "float"`) before get_arg runs — and there is no warning on the lossy coercion, so apps.yaml still says 4.8 while the runtime behaves as 4.

**Related: #4296 / #4441 — same mechanism, different code path**

#4441 fixed this exact truncation class for CONFIG_ITEMS-driven reads (metric_battery_cycle 0.5 → 0 via the same `int(float(value))`, because CONFIG_ITEMS declared `"default": 0` as an int) by normalising int defaults to float in `get_ha_config()`. `solcast_poll_hours` is not a CONFIG_ITEMS entry: it is resolved through the component framework, where `Components.initialize()` passes `COMPONENT_LIST`'s spec default directly to `get_arg()`. The `get_ha_config()` normalisation cannot reach this path because there is no `config_index` entry to match. v8.54.2 includes #4441 (merged 2026-08-08) and still truncates.

**Scope — other keys on the same path**

Same int-default/float-schema pattern, all confirmed in the installed source and on `main`:

| Key | COMPONENT_LIST default | APPS_SCHEMA type | Consumer | Impact if user configures a fraction |
|---|---|---|---|---|
| `solcast_poll_hours` | `8` | float | solcast.py:555 `max_age = x * 60` | poll TTL 4h instead of 4.8h → quota overuse → 429s |
| `forecast_solar_max_age` | `8` | float | solcast.py:458 `max_age = x * 60` | Forecast.Solar cache TTL silently shortened |
| `open_meteo_forecast_max_age` | `4` | float | solcast.py:236/310 `max_age = x * 60` | Open-Meteo cache TTL silently shortened |
| `axle_pence_per_kwh` | `100` | float | Axle pricing | **money**: configured 12.5p/kWh becomes 12p |

(Plausible further candidates with int defaults and fractional semantics: `octopus_saving_session_rate`, `car_charging_battery_size`.)

**Expected behaviour**

Either:
1. `solcast_poll_hours: 4.8` works as documented (the docs at apps-yaml/#solcast-solar-forecast recommend 4.8 for two-array hobbyist accounts to hit exactly 10 polls/day), or
2. at minimum, a warning is logged whenever get_arg's coercion is lossy — the failure is otherwise invisible: the YAML still says 4.8 and validation passes.

**Suggested fix**

Mirror #4441's "fix at the source" approach on the component path: when resolving arg specs in `Components.initialize()`, normalise an int default to float where APPS_SCHEMA declares the key `"type": "float"` — so a future component author writing `8` vs `8.0` can't change behaviour, rather than hand-editing the four literals above. If a narrower change is preferred, the one-line alternative is float literals in `COMPONENT_LIST` (`8.0`, `8.0`, `4.0`, `100.0`). A warning on lossy coercion in `get_arg()` would also be worth having regardless.

Regression test: extend `apps/predbat/tests/test_integer_config.py` (the #4296 home) — assert that a component arg spec with an int default and an APPS_SCHEMA float type resolves a configured 4.8 to `max_age == 288` minutes.

**Workaround in the meantime:** set an integer — `solcast_poll_hours: 5` gives 5 refreshes/day = 10 calls, the same outcome the 4.8 documentation recommendation was aiming for.

**Version:** v8.54.2. Still present on `main` as of 2026-09-03 — `apps/predbat/components.py` still declares `"default": 8` for solcast_poll_hours.

chalfontchubby:
Fixes #4925.

## Problem

`get_arg()` coerces its return value on the *type* of the default it was handed, not on the key's declared `APPS_SCHEMA` type, and applies that coercion to whatever value was resolved — real configured value or not. So `get_arg("solcast_poll_hours", 8)` ran a genuinely configured `4.8` through `int(float(value))` and returned `4`, shortening the Solcast poll TTL from 4.8h to 4h and pushing a two-site hobbyist account past its 10 poll/day quota into nightly HTTP 429s. Nothing warned: `validate_config()` checks the raw apps.yaml value against the float schema and passes it, so apps.yaml still reads `4.8` while the runtime behaves as `4`.

## Why not the issue's suggested fix

The issue proposes normalising int defaults inside `Components.initialize()`. That only covers the component-framework path. Auditing every float-declared `APPS_SCHEMA` key against its `get_arg` call sites found three more truncating reads that fix would miss — `octopus_saving_session_rate` and `octopus_saving_session_min_octopoints_per_kwh` are read directly from `octopus.py`, not through a component arg spec — plus a fifth affected component arg (`alphaess_api_delay`) the issue's table didn't list.

## Fix

Fixed at the source in `get_arg()` (`userinterface.py`): promote an int default to float whenever `APPS_SCHEMA` declares the key float, before the coercion at the end of the function ever sees it. One normalisation point, mirroring where #4441 fixed the equivalent `CONFIG_ITEMS`-route bug in `get_ha_config()`, so no call site — present or future — can change a float-declared key's runtime type by writing `8` instead of `8.0`.

Not included: the issue's second ask (a warning on lossy coercion). With the root cause fixed at the source, a remaining truncation means the schema genuinely declares the key an integer, so the warning would fire on intended behaviour every ~5 minutes on a routine sensor read.

## Testing

- `test_get_arg_float_schema_int_default_not_truncated` — the reported 4.8 → 288min case, an `axle_pence_per_kwh` money case, a direct (non-component) call site, unset-fallback typing, and negative controls for integer-declared and boolean keys.
- `test_component_arg_specs_resolve_float_declared_keys_as_float` — sweeps every `COMPONENT_LIST` arg spec whose `APPS_SCHEMA` type is float, resolving each the way `Components.initialize()` does (currently 9 keys), so it stays correct for specs added later rather than asserting on today's literals.
- Both verified to fail against unpatched `userinterface.py` and pass with the fix.
- `./run_all --quick` green, `./run_pre_commit` green.

---
Written by Claude on behalf of @chalfontchubby.

### DIFF
--- a/apps/predbat/tests/test_integer_config.py
+++ b/apps/predbat/tests/test_integer_config.py
@@ -478,3 +478,106 @@ def test_metric_battery_value_scaling_step_resolves_export_margin(my_predbat):
 
     print("✓ Test passed: metric_battery_value_scaling step {} keeps the first nudge ({:.4f}) clear of the {:.4f} flip point and keeps old 0.1-step values valid".format(step, first_nudge, flip_point))
     return False
+
+
+def test_get_arg_float_schema_int_default_not_truncated(my_predbat):
+    """
+    Regression test for #4925: a fractional value for a key APPS_SCHEMA declares "float" must
+    survive get_arg() even when the call site passed a bare int as its default.
+
+    get_arg() coerces its return value on the *type* of the default it was handed, not on the
+    key's declared type, and applies that coercion to whatever value was resolved - real
+    configured value or not. So `get_arg("solcast_poll_hours", 8)` ran a genuinely configured 4.8
+    through `int(float(value))` and returned 4, shortening the Solcast poll TTL from 4.8h to 4h
+    and pushing a two-site hobbyist account past its 10 poll/day quota into nightly HTTP 429s.
+    Nothing warned: validate_config() checks the raw apps.yaml value against the float schema and
+    passes it, so apps.yaml still reads 4.8 while the runtime behaves as 4.
+
+    Fixed at the source in get_arg() by trusting APPS_SCHEMA over the literal the caller happened
+    to write, so this cannot depend on which call site does the reading. #4296 was the same
+    mechanism on the CONFIG_ITEMS route (see the get_ha_config tests above); this is the
+    apps.yaml/APPS_SCHEMA route, which get_ha_config's normalisation cannot reach because there
+    is no config_index entry to match.
+    """
+    print("**** test_get_arg_float_schema_int_default_not_truncated ****")
+
+    original_args = my_predbat.args.copy()
+    try:
+        # The reported case: solcast_poll_hours read the way components.py's arg spec reads it.
+        my_predbat.args["solcast_poll_hours"] = 4.8
+        value = my_predbat.get_arg("solcast_poll_hours", 8)
+        assert value == 4.8, "get_arg('solcast_poll_hours', 8) should return 4.8, got {} ({})".format(value, type(value))
+        assert value * 60 == 288, "Solcast poll TTL should be 288 minutes, got {}".format(value * 60)
+
+        # A money value on the same path - a configured 12.5p/kWh must not be charged as 12p.
+        my_predbat.args["axle_pence_per_kwh"] = 12.5
+        value = my_predbat.get_arg("axle_pence_per_kwh", 100)
+        assert value == 12.5, "get_arg('axle_pence_per_kwh', 100) should return 12.5, got {} ({})".format(value, type(value))
+
+        # Not only the component path: a direct get_arg() call site with an int default is equally
+        # affected, which is why the fix lives in get_arg() rather than in Components.initialize().
+        my_predbat.args["octopus_saving_session_rate"] = 12.5
+        value = my_predbat.get_arg("octopus_saving_session_rate", 100)
+        assert value == 12.5, "get_arg('octopus_saving_session_rate', 100) should return 12.5, got {} ({})".format(value, type(value))
+
+        # An unset key still falls back to the caller's default, just typed to match the schema.
+        my_predbat.args.pop("solcast_poll_hours", None)
+        value = my_predbat.get_arg("solcast_poll_hours", 8)
+        assert value == 8 and isinstance(value, float), "Unset float-declared key should fall back to a float 8.0, got {} ({})".format(value, type(value))
+
+        # An integer-declared key must be unaffected - truncation there is the declared behaviour.
+        my_predbat.args["num_cars"] = 1.8
+        value = my_predbat.get_arg("num_cars", 1)
+        assert value == 1 and isinstance(value, int), "Integer-declared key should still coerce to int, got {} ({})".format(value, type(value))
+
+        # A boolean default must not be dragged into the float branch by isinstance(True, int).
+        my_predbat.args["forecast_solar_open_meteo_first"] = True
+        value = my_predbat.get_arg("forecast_solar_open_meteo_first", False)
+        assert value is True, "Boolean-declared key should stay boolean, got {} ({})".format(value, type(value))
+    finally:
+        my_predbat.args = original_args
+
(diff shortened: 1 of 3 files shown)


springfall2008:
Not sure about this, shouldn't we fix the caller for get_arg()?

chalfontchubby:
(Posted by Rik; written by Claude.)

Fair challenge, and normally I would agree — fix the caller, not the accessor. Two things pushed me the other way here.

**"The caller" is not one place.** Auditing main: 15 float-declared `APPS_SCHEMA` keys, affected in two unrelated mechanisms.

| Where | Keys |
| --- | --- |
| `components.py` arg specs | `solcast_poll_hours`, `forecast_solar_max_age`, `open_meteo_forecast_max_age` |
| `octopus.py` direct `get_arg()` calls | `octopus_saving_session_rate` (x2), `octopus_saving_session_min_octopoints_per_kwh` |

The original issue proposed normalising in `Components.initialize()`, which fixes the first three and cannot reach the `octopus.py` ones — those never go through the component framework. So the caller-side fix is six edits across two mechanisms rather than one.

**More to the point, it only holds while every future caller keeps getting it right.** The trap resets every time someone writes `8` instead of `8.0` against a float-declared key, and it fails silently in the worst way: `validate_config()` reads the raw apps.yaml value against the float schema and passes it, so apps.yaml still says `4.8` while the runtime behaves as `4`. Nothing warns, and the symptom surfaces somewhere unrelated — in #4925 as nightly Solcast HTTP 429s. A correctness property that depends on every call site remembering an invisible convention will drift back.

The framing that convinced me: the type is already declared authoritatively in `APPS_SCHEMA`, and `get_arg()` currently prefers an incidental literal over it. That reads as the accessor being wrong, not the callers.

There is also direct precedent — #4441 fixed this exact mechanism on the `CONFIG_ITEMS` route inside `get_ha_config()`, normalising an int default to float at the source for the same reason. This is the same bug on the apps.yaml/`APPS_SCHEMA` route, which `get_ha_config()`s normalisation cannot reach because there is no `config_index` entry to match. Fixing it in `get_arg()` keeps both routes consistent.

Two honest costs, so they are on the record rather than buried:

- it adds an `APPS_SCHEMA` lookup to a hot accessor (dict `.get` on a resolved-once global, but it is not free); and
- it makes `get_arg()`s return type depend on that global rather than purely on its arguments.

If you would rather keep the accessor dumb, the alternative I would want is all six call sites fixed **plus** a test that fails when a new int default appears against a float-declared key — the sweep in `test_component_arg_specs_resolve_float_declared_keys_as_float` already does roughly that for the component specs and could be widened. Without the guard I think it regresses quietly. Happy to switch it round if you prefer that shape — just say which.

chalfontchubby:
(Posted by Rik; written by Claude.)

@springfall2008 following up — my reply above (2026-09-10) laid out the case for fixing this in `get_arg()` rather than at each caller, plus an alternative if you'd rather keep the accessor dumb (fix all six call sites + a regression test). Just flagging so it doesn't fall off your radar; happy to switch approach if you tell me which way you'd prefer.

Issue #4720, opened 2026-08-24
Rolling debug logs not accessible with HA Companion app


gcoan:
**Describe the bug**
The rolling debug log feature introduced with #4417 is not accessible if you use the HA companion app.  Only works with a web browser.

Clicking on the 'download all (.tgz) results in the TGZ being rendered in the companion app, not downloaded anywhere:
![image](https://github.com/user-attachments/assets/aa0006b1-3835-438c-a9e8-5cb15aef07e0)

Similar behaviour if clicking on a snapshot in the debug plan view:
![image](https://github.com/user-attachments/assets/4da21f96-08bd-4082-b458-bb6fda6cebbf)

**Expected behaviour**
To be able to use the HA companion app to download debug log files.

I think the problem is that the debug download capability assumes that when a file is clicked, it will be downloaded as a separate file and stored on the users file system (where it can then be attached to a github issue).

This isn't the behaviour of the HA companion app.  Same problem exists with 'create predbat_debug.yaml', for which the workaround has been to turn on debug mode and then download the file from the debug directory.

The idea behind the rolling debug is to regularly capture debug logs so that you have one available to attach to a github issue.  The way it's implemented means it doesn't work with the Companion app.

I'd suggest that the rolling debug log feature should instead write the rolling logs directly to the debug directory so they can easily be attached to an issue.  Currently they're only stored in binary format in the cache directory.

See also #674 about the file extension used for the debug files as well.

**Predbat version**
8.53.1

**Environment details**
HAOS 2026.8.3
HA Companion app on iPad

springfall2008:
> **Automated first-pass triage.** This was produced by an automated pass over the issue and the current `main` tree — a maintainer will review before any action is taken.

**Classification:** `enhancement` (no priority set — priorities are skipped for feature requests). The underlying capability works as designed in a desktop browser; what is being asked for is a new delivery route for the snapshots. Happy to be re-labelled `bug` if the maintainers see "feature unreachable for Companion-app users" as a defect rather than a gap.

**Investigated against** `main` at `v8.53.0-57-gcc8c494e` (i.e. current main, which is 8.53.1 plus a few commits — the same code you are running). No log or `predbat_debug.yaml` was needed for this one, so nothing is being requested from you.

**What I found**

[...]

PR #4941, opened 2026-09-05
Name debug-history snapshots .yaml.txt so they attach to GitHub without renaming

### LINKED ISSUE #4932: Automatic debug files are being written with .yaml not .txt extension
**Describe the bug**
Change was made recently to write the automatic periodic predbat debug files in the debug directory not the cache directory.

That change has happened but the files are being written with a .yaml extension not .txt so can't be readily attached to github without having to SSH in and rename the files - not ideal for novices.

Also notice that the github issue template still refers to this renaming as being required, perhaps update the template to reflect the correct process and where to find the debug files

**Expected behaviour**
Debug files written with ,txt extension (as they used to be when written to the cache directory)

**Predbat version**
8.54.3

**Screenshots**
image

**Log file**
[predbat.log](https://github.com/user-attachments/files/31849207/predbat.log)

springfall2008:
This is an automated draft PR generated from issue #4932 — a maintainer should review it before merging.

Fixes #4932

## Summary

Since #4720 the rolling debug-history snapshots are written into `debug/` (so HA Companion-app users can pick them up over Samba/File editor, the download links being unusable in that webview) — but under `snapshot_filename()`'s `predbat_debug_<id>.yaml` name. GitHub refuses a bare `.yaml` as an issue attachment, so getting one onto a bug report meant SSHing in to rename it first.

`snapshot_filename()` now returns `predbat_debug_<id>.yaml.txt`, matching what the interactive web download has always served (`predbat_debug.yaml.txt`). The content is unchanged — still plain YAML — and keeping `.yaml` in the middle means `unit_test.py --debug_file` and editors still recognise it.

Because renaming the on-disk file orphans everything an upgrading install already has in `debug/`, a `_legacy_snapshot_filename()` fallback was added alongside the existing pre-#4720 cache-key one:

- `resolve_and_load_snapshot()` falls back to the old `.yaml` name, so snapshots captured before the upgrade stay downloadable for the rest of their normal time in the ring.
- `_discard_snapshot()` deletes both names, so eviction still prunes them. Without this an upgrade would drop them from the index but leave them on disk forever — up to `debug_history_count` files at several MB each.

Also, as the issue asks, the stale instruction in the bug-report template: the Dash-tab download already arrives as `predbat_debug.yaml.txt`, so the "rename the download file" step is no longer needed. The template now says to upload it as-is, and adds where to find the files in `debug/` for anyone the download link doesn't work for. `docs/customisation.md`'s debug-history section is updated to the new filename.

## Testing

From `coverage/`:

- `./run_pre_commit` — all 12 hooks Passed, and the quick suite it runs finished `All tests passed (4 slow tests skipped, total time: 109.20s)`.

From the repo root:

- `tools/triage_test.sh debug_history` — PASSED (exit 0), including the six new cases: the `.yaml.txt` name, that a fresh capture writes only the new name, that a pre-rename `.yaml` file still loads through the fallback (both against `FakeStorage` and a real `StorageLocalFiles` backend on disk), and that eviction deletes the old name off disk.
- `tools/triage_test.sh web_debug_history_routes` — PASSED (exit 0), with a new assertion that the single-snapshot download route's `Content-Disposition` filename ends in `.yaml.txt`.

## Notes

GitNexus MCP tools were not available in this session, so the blast radius for `snapshot_filename()` was established by search instead: five call sites — `debug_history.py:87,121,160,228` and `web.py:2946` — all of which are covered above. The archive member names inside `predbat_debug_history.tgz.dmp` change to `.yaml.txt` as a consequence, which seems desirable for the same reason.


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -36,4 +36,6 @@ This is important for any plan related issues.
 
 Once you have captured the issue go to the Web UI, in the Dash tab click on **predbat_debug.yaml** and wait 30 seconds for it to download. Select 'keep' if your Web Browser thinks the file is dangerous.
 
-Rename the download file to predbat_debug.yaml.txt and upload it to GitHub. This will allow your plan to be reproduced and also stores all your settings for review.
+The file arrives already named predbat_debug.yaml.txt, so upload it to GitHub as-is - no renaming needed. This will allow your plan to be reproduced and also stores all your settings for review.
+
+If the download link does nothing (the HA Companion app's built-in browser cannot save files, it just shows the content on screen), browse to the `debug/` folder under your Predbat config directory instead - with the File editor or Samba add-on - and attach a `predbat_debug_<timestamp>.yaml.txt` file from there. Predbat keeps a rolling history of these automatically, so there is usually one from around the time of the problem even if you had not turned debug mode on.
```
springfall2008: This advice doesn't hold on an install that has *just* upgraded across the rename: every snapshot captured before the upgrade is still on disk as `predbat_debug_<timestamp>.yaml` - the new fallback only re-serves the old bytes under the new name on download, the on-disk file keeps its old name until the ring rotates it out (up to ~45h at the default 15 snapshots x 3h). A Companion-app user following this paragraph finds only `.yaml` files for the incident window, and GitHub refuses the attachment - the exact problem this PR exists to remove. Worth a one-line caveat.

It also doesn't distinguish the `switch.predbat_debug_enable` per-cycle dumps sitting in the same folder (`predbat_debug_HH_MM_SS.yaml`, userinterface.py:759), which keep their bare `.yaml` name permanently - "attach a predbat_debug_<timestamp>.yaml.txt from there" can send the user to the wrong, still-unattachable file.

docs/customisation.md
```diff
@@ -871,7 +871,7 @@ Turning on `switch.predbat_debug_enable` only captures debug information from th
...
+Each snapshot is written as a plain `predbat_debug_<timestamp>.yaml.txt` file into the same `debug/` folder as the `switch.predbat_debug_enable` output described above - useful in its own right (a full rolling history sitting on disk, not just what the switch happened to catch), and it's the way to get a snapshot to attach to a GitHub issue from the HA Companion app, where the archive/single-file download links above don't work (see the note above). The contents are ordinary YAML; the `.txt` on the end is there because GitHub refuses to accept a `.yaml` file as an issue attachment, so a snapshot picked up from `debug/` can be attached to a bug report as-is with no renaming. These files are pruned in step with the ring buffer itself, so there are never more of them on disk than `debug_history_count` snapshots. Snapshots written by a version before this naming change (`predbat_debug_<timestamp>.yaml`) still download and are still pruned normally.
```
springfall2008: Same gap as the bug-report template: on an install upgraded across this change, the rolling files in debug/ are still named `predbat_debug_<timestamp>.yaml` for up to a ring's worth of time (default ~45h at 15 x 3h) - and those are exactly the files GitHub refuses as attachments. The closing sentence says they "still download and are still pruned normally" but never tells the reader they are still named `.yaml` on disk until evicted, so a user following this paragraph picks one up and the upload is rejected. One sentence would close it, e.g. "(files written before the upgrade keep the old `.yaml` name on disk until they are pruned; those still need renaming before attaching)".

apps/predbat/debug_history.py
```diff
@@ -202,12 +212,31 @@ async def capture_snapshot(storage, yaml_text, now_utc, max_count, max_age=None)
 
 
 def snapshot_filename(snapshot_id):
-    """Return the filename used both for a single-snapshot download and inside the bulk archive.
+    """Return the filename used on disk, for a single-snapshot download and inside the bulk archive.
 
     Deliberately keyed on snapshot_id alone (a capture timestamp, stable and unique) and
     not on steps_back (the snapshot's position in the current ring, which shifts as newer
     captures push it back) - two archives downloaded hours apart must name the same real
```
springfall2008, line 224: This docstring's own invariant - "two archives downloaded hours apart must name the same real capture identically, or they can't be merged/deduplicated by filename" - is broken by the rename across the upgrade boundary: the same capture id is `predbat_debug_<id>.yaml` in an archive downloaded before the upgrade and `...yaml.txt` in one downloaded after, so filename-keyed merging/dedup of the two archives duplicates it. One-time and minor, but either the sentence or the transition population needs a caveat (or a one-off rename of existing files on first capture/load would make the invariant hold unconditionally).

apps/predbat/debug_history.py
```diff
@@ -85,6 +87,8 @@ async def resolve_and_load_snapshot(storage, snapshot_id):
             return None, None
         snapshot_id = snapshots[0]["id"]
     data = await storage.load_debug_copy(snapshot_filename(snapshot_id))
+    if data is None:
```
springfall2008: The `if data is None` chain treats a zero-byte current-name file as a successful load: `load_debug_copy()` returns "" (not None) for an empty file, so both the pre-#4932 `.yaml` fallback and the legacy cache fallback are skipped, and the download route serves a 200 with empty content even when an intact legacy copy for the same id exists. Narrow but real trigger: `save_debug_copy()` opens with mode "w" (truncating immediately) then writes, so an interrupted re-capture of an id that already existed pre-rename leaves a 0-byte `predbat_debug_<id>.yaml.txt` alongside the good `predbat_debug_<id>.yaml`. `if not data:` would make the chain fall through instead.

apps/predbat/debug_history.py
```diff
@@ -202,12 +212,31 @@ async def capture_snapshot(storage, yaml_text, now_utc, max_count, max_age=None)
...
+
+    The trailing ".txt" is load-bearing, not decoration (see #4932). GitHub refuses a bare
+    ".yaml" as an issue attachment, so a snapshot picked straight out of debug/ - the only
+    route a HA Companion-app user has, since that webview cannot save the download links
+    (see #4720) - could not be attached to a bug report without an SSH session to rename it
+    first. The content is unchanged, still plain YAML: this matches what the interactive
+    web download has always served (web.py's "predbat_debug.yaml.txt"), and keeping the
+    ".yaml" in the middle means unit_test.py --debug_file and any editor still recognise it.
+    """
+    return "predbat_debug_{}.yaml.txt".format(snapshot_id)
```
springfall2008, line 241: Cosmetic follow-on worth a line in the PR body at least: the web file browser and its download route classify files by final extension, and both are unchanged here. In web.py the viewer's syntax-class check only matches `.yaml`/`.yml` (web.py:5416), and the browse download's content-type table maps `.txt` to `text/plain` (web.py:5573-5580) - so snapshot files picked from debug/ via the built-in file browser lose YAML highlighting and download as text/plain where pre-rename they were classified as YAML. No functional harm (the bytes are unchanged, and the /debug_history_download route is unaffected), but it is an untested behaviour change the rename causes.

apps/predbat/tests/test_debug_history.py
```diff
@@ -331,10 +331,72 @@ def test_debug_history(my_predbat):
...
+        print("  ERROR: test would be vacuous - the legacy and current filenames are the same")
+        failed = True
+
+    print("Test: a pre-rename snapshot survives a real Storage backend round trip (the file genuinely on disk under the old name)")
+    tmpdir = tempfile.mkdtemp()
+    try:
+        real_storage = StorageLocalFiles(tmpdir, lambda msg: None)
+        pre_rename_text = sample_yaml_text("pre-rename-on-disk")
+        pre_rename_real_id = now.strftime("%Y%m%d-%H%M%S")
+        asyncio.run(real_storage.save(STORAGE_MODULE, INDEX_NAME, [{"id": pre_rename_real_id, "timestamp": now.isoformat()}], format="json"))
```
springfall2008, line 423: Minor test gap: this index save is dead setup - `load_snapshot()` here is called with an explicit non-"latest" id, and `resolve_and_load_snapshot()` only consults the index for a falsy/"latest" id, so nothing in this block reads it. If index persistence or `list_snapshots()`'s json parse regressed, this test would stay green while every `?id=latest` download and the History-view snapshot links 404 (the "latest" route is covered, but only in the other test file / other FakeStorage tests, not against a real backend here). Either drive this assertion through the "latest" path or drop the line.

apps/predbat/tests/test_debug_history.py
```diff
@@ -331,10 +331,72 @@ def test_debug_history(my_predbat):
...
+    if asyncio.run(load_snapshot(pre_rename_storage, pre_rename_id)) is not None:
+        print("  ERROR: a discarded pre-rename snapshot should no longer load")
+        failed = True
+
+    print("Test: the two on-disk names are distinct, so the fallback is actually reaching a different file")
+    if _legacy_snapshot_filename("20260804-140000") == snapshot_filename("20260804-140000"):
+        print("  ERROR: test would be vacuous - the legacy and current filenames are the same")
+        failed = True
+
+    print("Test: a pre-rename snapshot survives a real Storage backend round trip (the file genuinely on disk under the old name)")
```
springfall2008, line 417: Two upgrade-boundary cases this new block doesn't cover:

1. Both names present for one id - a fresh capture after the upgrade writes `.yaml.txt` while the pre-rename `.yaml` for the same id (same-second re-capture) is still on disk. The current-name-wins preference in `resolve_and_load_snapshot()` is untested, and these tests always simulate exactly one file.
2. The cross-rename archive case the `snapshot_filename()` docstring's merge/dedup invariant turns on: the existing "archive filename is stable across downloads" test downloads both archives inside one version, so it structurally cannot catch the `.yaml` -> `.yaml.txt` split of the same capture across an upgrade.

apps/predbat/debug_history.py
```diff
@@ -202,12 +212,31 @@ async def capture_snapshot(storage, yaml_text, now_utc, max_count, max_age=None)
...
+
+
+def _legacy_snapshot_filename(snapshot_id):
```
springfall2008, line 244: Nit: this re-hardcodes the `predbat_debug_<id>.yaml` stem in a second format string instead of deriving it from `snapshot_filename()` (e.g. `snapshot_filename(snapshot_id)[:-4]`) or a shared stem constant. The two names now encode the same knowledge twice - if the on-disk name ever changes again, someone has to remember to update both or the fallback/delete reaches a filename nothing was ever written under, which is exactly the class of silent failure this fallback exists to paper over.

apps/predbat/debug_history.py
```diff
@@ -85,6 +87,8 @@ async def resolve_and_load_snapshot(storage, snapshot_id):
             return None, None
         snapshot_id = snapshots[0]["id"]
     data = await storage.load_debug_copy(snapshot_filename(snapshot_id))
+    if data is None:
+        data = await storage.load_debug_copy(_legacy_snapshot_filename(snapshot_id))
```
springfall2008, line 96: Efficiency note, low priority: this fallback now costs a failed `load_debug_copy()` open (plus the pre-existing legacy-cache `load()`) on *every* snapshot load, forever - including `load_all_snapshots()`'s loop, i.e. up to ~15 extra debug/ opens + 15 cache loads per archive download, for the life of the install even after every legacy file has been reaped. Honest cost is small (archive downloads are rare and user-triggered), but a rename-on-first-load - when the fallback hits, re-save the text under `snapshot_filename()` and delete the legacy file - would bound the cost to once per legacy snapshot and let this whole branch die out naturally.

docs/customisation.md
springfall2008: **[Review xhigh] Upgrade-window guidance in the docs is wrong for pre-rename snapshots (CONFIRMED)**

`snapshot_filename()` now returns `predbat_debug_<id>.yaml.txt` and nothing renames pre-existing files — the fallback in `resolve_and_load_snapshot()` only *reads* the old `.yaml` name. For up to the full ring turnover (15 snapshots x ~3h ≈ 45h) after an install upgrades, retained captures are still on disk as bare `.yaml`, so a user told to "attach a `predbat_debug_<timestamp>.yaml.txt` file from there" (bug_report.md:41) will not find that name for pre-upgrade captures — and the file they *do* find is the exact bare `.yaml` that GitHub refuses, the failure mode #4932 was meant to remove.

Worth one sentence in both docs: files written before the upgrade keep the `.yaml` name on disk and need renaming (or re-capturing) before GitHub will take them.

apps/predbat/debug_history.py
```diff
@@ -202,12 +212,31 @@ async def capture_snapshot(storage, yaml_text, now_utc, max_count, max_age=None)
...
+    """Return the filename used on disk, for a single-snapshot download and inside the bulk archive.
 
     Deliberately keyed on snapshot_id alone (a capture timestamp, stable and unique) and
     not on steps_back (the snapshot's position in the current ring, which shifts as newer
     captures push it back) - two archives downloaded hours apart must name the same real
     capture identically, or they can't be merged/deduplicated by filename.
+
+    The trailing ".txt" is load-bearing, not decoration (see #4932). GitHub refuses a bare
+    ".yaml" as an issue attachment, so a snapshot picked straight out of debug/ - the only
+    route a HA Companion-app user has, since that webview cannot save the download links
```
springfall2008, line 235: **[Review xhigh] The rename breaks this docstring's own cross-archive dedup invariant across the upgrade boundary (CONFIRMED, one-time and self-healing)**

The docstring still promises "two archives downloaded hours apart must name the same real capture identically, or they can't be merged/deduplicated by filename" — but for captures already in the ring when an install upgrades, an archive downloaded pre-upgrade has member `predbat_debug_<id>.yaml` while a post-upgrade archive for the *same* still-retained capture (served via the legacy fallback) has `predbat_debug_<id>.yaml.txt`. Extracting both and deduping by filename keeps two copies of the same capture.

Bounded to one ring lifetime and gone after each pre-rename file is naturally evicted, so acceptable — but the docstring should acknowledge the exception rather than state an invariant the very next section of the same PR deliberately breaks.

apps/predbat/debug_history.py
springfall2008, line 96: **[Review xhigh] A truncated (0-byte) current-name file short-circuits both legacy fallbacks (PLAUSIBLE — pattern pre-existing, extended here)**

`load_debug_copy()` returns `""` for an existing 0-byte file and `None` only when the file is missing, so `if data is None` lets an empty-but-present file pass and skips both the legacy-`.yaml` and legacy-cache fallbacks: the single-snapshot route serves a 200 with an empty body even though a good legacy copy exists, and `load_all_snapshots()` includes a zero-length tar member. `save_debug_copy()` opens with "w" (truncate-then-write), so a same-minute recapture whose rewrite fails mid-write, or external truncation, can leave a 0-byte file behind a live index entry.

Reachability is narrow and the single-fallback version of this hole predates the PR (main has the same `is None` check); the PR adds the second fallback level that would have rescued it. Using truthiness (`if not data`) or an empty-string special-case would close both levels.

apps/predbat/tests/test_debug_history.py
```diff
@@ -331,10 +331,72 @@ def test_debug_history(my_predbat):
...
+    if list(rename_storage.debug_copies.keys()) != [snapshot_filename(renamed_id)]:
+        print("  ERROR: expected exactly one debug/ file under the new name, got {}".format(list(rename_storage.debug_copies.keys())))
+        failed = True
+
+    print("Test: a snapshot written before the #4932 rename (debug/ file still named .yaml) still loads via the filename fallback")
+    pre_rename_storage = FakeStorage()
+    pre_rename_id = asyncio.run(capture_snapshot(pre_rename_storage, sample_yaml_text("pre-rename"), now, max_count=15))
+    # Simulate an install upgrading across the rename: the index entry is unchanged, but the
+    # file on disk is still under the old name.
+    pre_rename_storage.debug_copies.pop(snapshot_filename(pre_rename_id), None)
```
springfall2008, line 367: **[Review xhigh] No test pins the fallback ORDER when both names exist for one id (CONFIRMED, test gap)**

Every fallback test first pops the current-name entry before installing the legacy one, so no test covers the both-files-present case (`predbat_debug_<id>.yaml` *and* `predbat_debug_<id>.yaml.txt` in `debug_copies` for the same id) that would assert the current name wins. The documented chain order in `resolve_and_load_snapshot()` — current name → legacy `.yaml` → legacy cache key — is entirely unpinned; a swapped or reordered chain would pass the suite. A same-id/both-names test (and ideally one for `_discard_snapshot` deleting both) would close this cheaply.

apps/predbat/tests/test_debug_history.py
```diff
@@ -331,10 +331,72 @@ def test_debug_history(my_predbat):
         failed = True
```
springfall2008, line 340: **[Review xhigh] FakeStorage.save() discards `expiry`, so the fake-level reap tests never exercise the mechanism the reap relies on (PLAUSIBLE, partial)**

`FakeStorage.save()` (line 41) accepts and throws away `expiry`, so this test's `load_snapshot(...) is None` comes from the data having been overwritten to `None`, not from expiry — the test never inspects the expiry argument, and no fake-based test makes an expired entry unloadable *by expiry*. The real-backend tests do cover write-side expiry (real `cleanup()` only reaps past-expiry entries, so they would fail if `_discard_snapshot` wrote `expiry=None`), which keeps this from being tautological — but the real backend's *load-time* expiry gate (storage.py load path) remains untested. Either honour `expiry` in FakeStorage or add a real-backend load-after-expiry assertion.

apps/predbat/tests/test_debug_history.py
springfall2008, line 423: **[Review xhigh] Dead setup: this index save is never read (CONFIRMED, cleanup)**

The test loads and discards by explicit id, and `resolve_and_load_snapshot()` only consults `list_snapshots()` (the index) when the id is falsy or "latest" — so this index entry, like the one in the earlier real-backend legacy-key test, is never read. Either drop the index save (leaving only the content file, which is what the load path actually needs) or extend the test to resolve via the index (e.g. also check `list_snapshots`/the "latest" path through the real backend).

apps/predbat/tests/test_debug_history.py
```diff
@@ -331,10 +331,72 @@ def test_debug_history(my_predbat):
...
 
+    print("Test: snapshots are named .yaml.txt so they can be attached to a GitHub issue straight out of debug/ without renaming (#4932)")
+    if not snapshot_filename("20260804-140000").endswith(".txt"):
+        print("  ERROR: a snapshot GitHub will not accept as an attachment defeats the point of writing it into debug/ at all, got {!r}".format(snapshot_filename("20260804-140000")))
+        failed = True
+    if ".yaml" not in snapshot_filename("20260804-140000"):
+        print("  ERROR: the .yaml should be kept in the middle so the content type stays obvious, got {!r}".format(snapshot_filename("20260804-140000")))
+        failed = True
+
+    print("Test: a fresh capture writes only the new .yaml.txt name into debug/, not the pre-#4932 .yaml one")
```
springfall2008, line 355: **[Review xhigh] Structural duplication with the two existing real-backend tests (PLAUSIBLE, minor)**

This reuses the exact template of the pre-existing real-backend tests (index entry + content written to a real backend, load via fallback, `_discard_snapshot`, verify the file gone) — the boilerplate shape is duplicated. The *behaviour* it adds (loading the pre-#4932 `.yaml` debug/ filename and deleting that legacy-named file on eviction through the real backend) is genuinely new coverage, so no action strictly needed; a small shared helper for the tempdir-backend setup would remove the copy-paste. Noting for the record only.

apps/predbat/tests/test_web_debug_history_routes.py
```diff
@@ -123,6 +123,11 @@ def test_web_debug_history_routes(my_predbat):
         if first_id not in resp.headers.get("Content-Disposition", ""):
             print("  ERROR: expected first_id in the Content-Disposition filename, got {!r}".format(resp.headers.get("Content-Disposition")))
             failed = True
```
springfall2008, line 126: **[Review xhigh] The web layer never exercises a snapshot that only exists under the legacy .yaml name (PLAUSIBLE, test gap)**

Every route test's snapshot bytes live under the current `.yaml.txt` name, so `html_debug_history_download()` serving a pre-rename capture — `resolve_and_load_snapshot()` falling back to `_legacy_snapshot_filename()` while the Content-Disposition is built from `snapshot_filename(resolved_id)` — is only covered at the unit level, never through the route. A regression that makes the route 404 on (or mis-name) legacy-only snapshots would pass this suite. Adding one case that installs only the legacy `.yaml` file and asserts the route returns 200 with the `.yaml.txt` disposition name would pin the route/fallback interaction the PR is built on.

Separately, docs/web-interface.md:56 still tells users to "rename the file extension ... to '.txt'" before attaching — contradicting the new bug_report.md text this PR edits; it is one paragraph away from the same instruction and is now doubly stale since the download already arrives as `.yaml.txt`.

.github/ISSUE_TEMPLATE/bug_report.md
springfall2008: Automated reply from the triage bot. Fixed in e2096ac6: the paragraph now says files captured before the upgrade to the .txt-suffix version keep the old plain-.yaml name until pruned and need renaming, and it now also names the per-cycle predbat_debug_HH_MM_SS.yaml files that debug mode writes (userinterface.py:761) as always being plain .yaml too, so neither failure mode sends the user to an unattachable file.

docs/customisation.md
springfall2008: Automated reply from the triage bot. Fixed in e2096ac6: the closing sentence now states the pre-rename files keep the old plain-.yaml name on disk until they are pruned, and so need renaming to .yaml.txt before GitHub will accept them as an attachment.

apps/predbat/debug_history.py
springfall2008: Automated reply from the triage bot. Added the caveat to the docstring in e2096ac6: it now names the cross-upgrade exception explicitly - a capture still in the ring at upgrade time is .yaml in a pre-upgrade archive and .yaml.txt (via the fallback) in a post-upgrade one, so filename-keyed dedup keeps two copies of that one capture, bounded to a single ring lifetime. Not adding a one-off rename on load/capture: resolve_and_load_snapshot() is deliberately read-only (the base-class save_debug_copy() is a documented no-op returning False for filesystem-less backends, so a write-based migration there is a silent-failure path), and an eager rename on capture would not touch files that are never captured again. The natural eviction reap already bounds the inconsistency.

apps/predbat/debug_history.py
springfall2008: Automated reply from the triage bot. Fixed in e2096ac6: the chain now uses 'if not data', so a zero-byte current-name file falls through to the legacy .yaml fallback and the legacy cache key instead of short-circuiting them; the resolve_and_load_snapshot() docstring documents the empty-means-missing semantics. Covered by a new test that truncates the current-name file to 0 bytes and asserts the intact legacy copy loads.

apps/predbat/debug_history.py
springfall2008: Automated reply from the triage bot. Noted rather than changed: the browse viewer/download route are generic, extension-keyed UI shared with every other file in debug/, and keying on the final extension is their existing design - a .txt mapping to text/plain is the standard classification, so I have left web.py alone here rather than adding snapshot-specific special cases. Added a line to the PR body recording the behaviour change (snapshot files in the built-in file browser lose YAML highlighting and download as text/plain; /debug_history_download is unaffected).

apps/predbat/tests/test_debug_history.py
springfall2008: Automated reply from the triage bot. Extended in e2096ac6 rather than dropped: the test now also loads via 'latest' against the real backend, which is what makes the index entry live setup and additionally pins the ?id=latest resolve path working through a real StorageLocalFiles backend with only a legacy-named file on disk.

apps/predbat/tests/test_debug_history.py
springfall2008: Automated reply from the triage bot. Both closed with new tests in e2096ac6: (1) a both-names-for-one-id case asserting the current .yaml.txt name wins (pins the documented fallback order) and that _discard_snapshot then deletes both; (2) a load_all_snapshots case installing only a legacy .yaml file and asserting the archive member is named under the current .yaml.txt name - which is the post-upgrade half of the cross-rename split the snapshot_filename() docstring now caveats (the pre-upgrade half is an archive built by the old code, which this version's tests cannot run against).

apps/predbat/debug_history.py
springfall2008: Automated reply from the triage bot. Fixed in e2096ac6: _legacy_snapshot_filename() now derives from snapshot_filename() by stripping the trailing '.txt', so the current name is the only place the stem is written and the two cannot drift apart. The existing 'the two on-disk names are distinct' test still guards against the derivation collapsing to the same name.

apps/predbat/debug_history.py
springfall2008: Automated reply from the triage bot. Leaving this as-is: a rename-on-first-load makes resolve_and_load_snapshot() a write path, and the Storage base class' save_debug_copy() is a documented no-op returning False for backends with no real filesystem, so the migration write would be a silent-failure path on exactly the backends this code must stay agnostic about - for a cost the comment itself concedes is negligible (rare, user-triggered downloads, a couple of failed opens each). The dead branch also dies out naturally anyway, since every legacy file is reaped by _discard_snapshot within one ring lifetime.

docs/customisation.md
springfall2008: Automated reply from the triage bot. Fixed in e2096ac6 - both docs, in fact: customisation.md's closing sentence now states pre-rename files keep the plain-.yaml name on disk until pruned and need renaming to .yaml.txt before GitHub will take them, and bug_report.md's debug/ paragraph carries the same caveat (plus the per-cycle predbat_debug_HH_MM_SS.yaml files debug mode writes, which are always plain .yaml).

apps/predbat/debug_history.py
springfall2008: Automated reply from the triage bot. The docstring now caveats the exception instead of asserting the unconditional invariant: a capture still in the ring across the #4932 upgrade is .yaml in a pre-upgrade archive and .yaml.txt (via the fallback) in a post-upgrade one, so filename-keyed dedup keeps two copies of that one capture, bounded to a single ring lifetime. Not adding a one-off rename to make it hold unconditionally: resolve_and_load_snapshot() is deliberately read-only (the base-class save_debug_copy() is a documented no-op returning False for filesystem-less backends), so the fallback stays read-only too.

apps/predbat/debug_history.py
springfall2008: Automated reply from the triage bot. Fixed in e2096ac6: the chain uses 'if not data' now, so an empty-but-present current-name file falls through to the legacy .yaml fallback and the legacy cache key at both levels; the docstring records the empty-means-missing semantics (save_debug_copy() truncates on open, so an interrupted re-capture can leave the 0-byte file behind a live index entry). New test covers it: current-name file truncated to 0 bytes, intact legacy copy loads.

apps/predbat/tests/test_debug_history.py
springfall2008: Automated reply from the triage bot. Covered in e2096ac6: a new test installs both names for one id (with distinguishable content) and asserts the current .yaml.txt name wins, then that _discard_snapshot deletes both - that pins the documented current-name > legacy-.yaml > legacy-cache order for the two debug/ levels. (The legacy cache level's position after the .yaml fallback was already pinned by the pre-#4720 test, where no debug/ file exists at either name.)

apps/predbat/tests/test_debug_history.py
springfall2008: Automated reply from the triage bot. Fixed on the fake side in e2096ac6: FakeStorage.save() now records expiry and load() returns None once it has passed, so the fake-level reap test genuinely exercises the mechanism - it now asserts _discard_snapshot's legacy overwrite carries an expiry in the past, and the follow-up load returning None comes from the expiry gate rather than the overwritten-to-None data. (Deliberately not adding a real-backend load-after-expiry assertion: the real backend's load path is generic storage.py machinery outside this PR's scope, and its write-side expiry is already covered by the cleanup()-based test.)

apps/predbat/tests/test_debug_history.py
springfall2008: Automated reply from the triage bot. Extended in e2096ac6 rather than dropped: the test now also loads via 'latest' against the real backend, making the index entry live setup and pinning the ?id=latest resolve path through a real StorageLocalFiles backend with only a legacy-named file on disk.

apps/predbat/tests/test_debug_history.py
springfall2008: Automated reply from the triage bot. Left as-is: the behaviour is genuinely new coverage, and the two pre-existing real-backend tests are the only other users of that shape - extracting a shared helper for three call sites would cost more readability than the duplication does while this is the last test likely to use it.

apps/predbat/tests/test_web_debug_history_routes.py
springfall2008: Automated reply from the triage bot. Both fixed in e2096ac6: the route test now captures a snapshot, renames its on-disk file to the legacy .yaml name only, and asserts /debug_history_download still serves the data 200 with a .yaml.txt Content-Disposition - pinning the route/fallback interaction the PR is built on. web-interface.md's now-doubly-stale 'rename the file extension' paragraph is rewritten: the debug download arrives as predbat_debug.yaml.txt and attaches as-is; only apps.yaml still needs renaming.

PR #4955, opened 2026-09-06
perf: reduce Predbat's idle memory footprint

springfall2008:
## Summary

Cuts steady-state idle RSS on Linux/glibc by roughly a third and removes the transient spike from the debug dump. Measured with a live config, Load ML off, Ubuntu 24.04 / Python 3.12 (peak / idle RSS, MB):

| Build | Peak | Idle | Threads |
|---|---|---|---|
| `main` | 266 | 162 | 42 |
| + per-key debug yaml, `malloc_trim`, arena cap | 138 | 126 | 43 |
| + lazy component imports | 118 | 107 | 41 |

Plan calculation timings are unchanged (1.75-2.75 s per cycle) and all 13 configured components still initialise.

Three independent commits, each measured on its own:

- **`userinterface.py`** - `create_debug_yaml()` now streams the dump one top-level key at a time through `dump_debug_yaml()`. PyYAML builds a Node tree ~30x the text size, so dumping the whole dict at once was the largest peak in the cycle; per-key bounds it by the biggest key. A shared anchor counter keeps aliases unique across the per-key dumps so the file still loads as a single document.
- **`utils.py` / `predbat.py`** - `malloc_trim()` after each cycle's `gc.collect()` hands freed heap pages back to the OS, and `limit_malloc_arenas()` at startup caps glibc at 2 malloc arenas (the ~40 threads were each growing their own). Both are no-ops that return `False` on macOS/musl.
- **`components.py`** - `COMPONENT_LIST` entries name their class as a `"module.ClassName"` string and `load_component_class()` imports it only when the component is enabled, so startup no longer compiles all 30 component modules. An `ImportError` (e.g. `gateway` without protobuf) logs a warning and skips the component instead of failing the registry.

Also adds a "Memory usage high" row to the triage debug journal recording what idle memory is made of and which levers moved it.

## Tests

- `debug_yaml_scope`: proves the plain-Dumper duplicate-anchor hazard, round-trips the per-key dump, and checks with `tracemalloc` that per-key dumping bounds the node tree.
- `memory_release` (new): the glibc helpers call `malloc_trim(0)` / `mallopt(M_ARENA_MAX, n)` correctly and report `False` without glibc.
- `components`: every registry entry imports and subclasses `ComponentBase` (so a compile error in any component fails the quick suite), a bare `import predbat` in a subprocess leaves the registry modules unloaded, and a missing dependency skips the component with a warning.

Full quick suite passes (307 tests); pre-commit clean.


springfall2008:
Addressed the review in f8f22f08 and went a step further: a component whose module will not import *or* whose constructor raises is now contained in `Components.initialize()` - logged (with traceback unless it is a plain `ImportError`), recorded in `component_errors`, and the rest of the phase still initialises. It is reported as an **error** rather than silently listed as disabled: `binary_sensor.predbat_components_healthy` goes off, the run status names it, and the `/components` page shows the card in red with the reason and a Restart button. The "Initialising" line is now logged only after the import succeeds, and the test docstring is rewritten (the test itself is now three: import failure, constructor failure, status reporting).

1e6247c7 also makes the web interface test listen on a free port instead of 5052 so a second test run or a Predbat in another session cannot answer its requests.

PR #4957, opened 2026-09-06
fix(octopus): price completed IOG dispatches off-peak regardless of location

### LINKED ISSUE #4946: Octopus IOG: completed dispatches with location AWAY should still be priced as off-peak
## Summary

`rate_add_io_slots()` only applies the Intelligent Octopus Go off-peak override to dispatches whose `location` is `AT_HOME` (or unset):

```python
if source != "bump-charge" and source != "BOOST" and (not location or location == "AT_HOME"):
```

`apps/predbat/octopus.py` — the same test appears three times, around lines 2760, 2790 and 2861 as of v8.53.0.

Checking this against the supplier's half-hourly **metered billing data** shows that `location` does not determine the rate actually charged. Octopus exposes, per half hour, which rate bucket the energy was billed in — `CONSUMPTION_CHARGE_EV_DEVICE_OFF_PEAK_*` for smart-charge energy, alongside `..._ECO7_NIGHT_*` and `..._ECO7_DAY_*`. (The trailing suffix differs per account, so it needs prefix matching, not an exact string compare.)

`location` is also not stable. The same completed dispatches have been observed reported as `AT_HOME` for several hours and then retroactively relabelled `AWAY` long after they had finished. Because `today_cost()` re-prices the whole day from `rate_import` on every cycle, such a relabel makes the day's cost step up in one go — the increase landing on an hour with no import at all, for energy the supplier had in fact billed at the off-peak rate.

## Suggested change

For **completed** dispatches, drop the `location` term so `AWAY` and `UNABLE_TO_IDENTIFY` are priced the same as `AT_HOME`.

Suggest **not** extending this to planned/flex dispatches. A genuinely-away car produces exactly the low-consumption pattern above, and for a future slot there is no way to tell the two apart yet — marking it off-peak would have the planner import against a cheap window that will not materialise.

The `bump-charge` / `BOOST` exclusions are unaffected

springfall2008:
This is an automated draft PR generated from issue #4946 — a maintainer should review it before merging.

Fixes #4946

## Summary

`rate_add_io_slots()` applied its `location` test to *every* dispatch, including ones that had already finished. Octopus bills a completed smart-charge dispatch at the off-peak rate whatever location it ends up reported at, and that label is not stable — the same completed dispatch can be relabelled `AT_HOME` to `AWAY` hours after it ran. Since `rate_import` is rebuilt from the tariff every cycle, such a relabel silently un-stamped the cheap rate on minutes that had already been metered, and `today_cost()` re-prices the whole day, so the reported cost stepped up in one go on an hour with no import at all.

The location test is now scoped to dispatches that haven't finished yet (`end_minutes <= minutes_now` marks a dispatch as completed). Planned and in-progress slots keep the existing behaviour, as the issue suggested, so a genuinely-away car still can't have the planner import against a cheap window that will not materialise. The `bump-charge` / `BOOST` exclusions are unchanged.

## Testing

Run from `coverage/`: `./run_pre_commit` — all 12 pre-commit hooks passed and the full `./run_all --quick` suite passed (4 slow tests skipped, 123s). Targeted module: `tools/triage_test.sh rate_add_io_slots` passed.

Existing Test 10 asserted the behaviour being fixed (a past `AWAY` slot staying at the default rate), so it has been rescoped to a *future* `AWAY` slot — still ignored, which is the part of the behaviour that should not change. Four tests were added:

- Test 20 — a completed `AWAY` dispatch is priced off-peak (the #4946 regression).
- Test 21 — same for `UNABLE_TO_IDENTIFY`.
- Test 22 — an in-progress `AWAY` dispatch is still ignored, fixing the completed/not-completed boundary.
- Test 23 — completed `bump-charge` and `BOOST` slots stay excluded now that `location` no longer blocks them.

Test 22 is anchored on `minutes_now` rather than a wall-clock hour: rate minutes are indexed from local midnight, so a fixed UTC hour lands either side of `minutes_now` depending on the DST offset.

## Notes

Approach informed by the Octopus entry in `.claude/skills/issue-triage/references/debug-journal.md`, which records this mechanism and flags that Test 10 asserts the current behaviour. One caveat worth a maintainer's eye: the supplier-billing premise — that completed smart-charge energy is billed off-peak regardless of `location` — comes from the reporter's half-hourly metered billing data and could not be verified independently here. The code-side mechanism and the cost-jump symptom are confirmed.

Note that completed non-`AT_HOME` dispatches now consume the daily low-rate slot budget, which is intended: they were real charges billed off-peak.


apps/predbat/octopus.py
```diff
@@ -3169,8 +3169,20 @@ def rate_add_io_slots(self, car_n, rates, octopus_slots):
...
+                # completed smart-charge dispatch at the off-peak rate regardless of the location it ends
+                # up reported at, and that label is not stable - a completed dispatch can be retroactively
+                # relabelled AT_HOME to AWAY hours after it ran. Since rate_import is rebuilt from the
+                # tariff every cycle, such a relabel would silently un-stamp the cheap rate on minutes that
+                # have already been metered, and today_cost() re-prices the whole day, so the day's cost
+                # steps up on an hour with no import at all (issue #4946). A slot still to come is a
+                # different matter and keeps the location test: a genuinely-away car would otherwise have
+                # the planner import against a cheap window that never materialises.
+                dispatch_completed = end_minutes <= self.minutes_now
+                location_at_home = dispatch_completed or not location or location == "AT_HOME"
```
springfall2008: **Review (recall sweep, high effort): budget accounting diverges from `load_octopus_slots`**

With this change, a completed non-AT_HOME dispatch now enters the stamping block and consumes the midday-to-midday low-rate budget here, but `load_octopus_slots`' independent counter (octopus.py:3064-3071, unchanged) still only counts slots passing the old location test — and it drops completed slots at :3094, so only its *counter* diverges.

Failure scenario: IOG-SMB tariff (12-block cap); `fetch.py:1423-1427` merges `completed` ahead of `planned` and `rate_add_io_slots` does not sort, so completed-AWAY blocks claim budget first. A later planned AT_HOME slot is then left at `rate_max_base` in `rate_import` (what `today_cost()` and the planner read) while `load_octopus_slots` still sees blocks free and prices the same slot's chunk at `rate_min_base` — `car_charging_slots`' published `average`/`cost` contradicts the stamped rates. Exposure is near-cap only and lands on the display sensors, and this PR's side is arguably the more faithful model of Octopus's real cap — but the two counters should agree, either by giving `load_octopus_slots` the same completed exemption for its counter or by commenting the asymmetry as deliberate.

apps/predbat/octopus.py
```diff
@@ -3169,8 +3169,20 @@ def rate_add_io_slots(self, car_n, rates, octopus_slots):
...
+                # The location test only applies to dispatches that haven't finished yet. Octopus bills a
+                # completed smart-charge dispatch at the off-peak rate regardless of the location it ends
+                # up reported at, and that label is not stable - a completed dispatch can be retroactively
+                # relabelled AT_HOME to AWAY hours after it ran. Since rate_import is rebuilt from the
+                # tariff every cycle, such a relabel would silently un-stamp the cheap rate on minutes that
+                # have already been metered, and today_cost() re-prices the whole day, so the day's cost
+                # steps up on an hour with no import at all (issue #4946). A slot still to come is a
+                # different matter and keeps the location test: a genuinely-away car would otherwise have
+                # the planner import against a cheap window that never materialises.
+                dispatch_completed = end_minutes <= self.minutes_now
```
springfall2008: **Review: binary completion boundary leaves already-metered minutes of an in-progress away dispatch mispriced**

`dispatch_completed` is binary on `end_minutes <= self.minutes_now`, so a multi-hour dispatch that straddles `minutes_now` with `location: AWAY` keeps the location test and its *elapsed, already-metered* minutes stay un-stamped. `today_cost()` (output.py:2046+) re-prices the whole day from `rate_import` every cycle, so those minutes are priced at the tariff rate during the dispatch and then flip to `rate_min_base` when it completes — a downward cost step after the fact, mirroring the #4946 symptom this PR fixes.

Trigger is data-dependent: the HA-sensor path (`fetch.py:1417-1427`) passes `planned_dispatches` through untrimmed, so a straddling slot with a non-home location reaches here intact; the Octopus-API path trims a started slot to `now` first. If in-progress slots never carry a non-home location this is unreachable — worth confirming, or scoping the exemption to `start_minutes <= self.minutes_now` (the metered portion) rather than completion.

apps/predbat/octopus.py
```diff
@@ -3169,8 +3169,20 @@ def rate_add_io_slots(self, car_n, rates, octopus_slots):
                 if end_minutes <= self.minutes_now and kwh <= 0:
```
springfall2008, line 3210: **Review: the zero-kWh skip still bypasses the relaxation — residual #4946-shaped hole via the kWh field**

This `continue` runs *before* the new completed-dispatch exemption, so a completed dispatch carrying `kwh <= 0` is dropped wholesale and its minutes never get stamped off-peak regardless of location. The same file documents at `load_octopus_slots` (~3005-3011) that Octopus zeroes `chargeKwh` once the car hits its target SoC while the charger may still draw — if that zeroed value persists into the completed record, a dispatch that delivered metered energy keeps the tariff price in `rate_import` and `today_cost()` prices it at peak.

Data-dependent: on the Octopus-API path completed dispatches read the separate `delta` field (with duration-synthesis fallback), so only the HA-integration path can plausibly deliver a zeroed completed record. Test 18 asserts the skip is intended for withdrawn slots, so this may be a deliberate trade-off — but it's the same symptom family the PR claims to fix and is neither covered nor acknowledged.

apps/predbat/octopus.py
```diff
@@ -3169,8 +3169,20 @@ def rate_add_io_slots(self, car_n, rates, octopus_slots):
...
+                # have already been metered, and today_cost() re-prices the whole day, so the day's cost
+                # steps up on an hour with no import at all (issue #4946). A slot still to come is a
+                # different matter and keeps the location test: a genuinely-away car would otherwise have
+                # the planner import against a cheap window that never materialises.
+                dispatch_completed = end_minutes <= self.minutes_now
+                location_at_home = dispatch_completed or not location or location == "AT_HOME"
+
                 # Ignore bump-charge slots as their cost won't change
-                if source != "bump-charge" and source != "BOOST" and (not location or location == "AT_HOME"):
+                if source != "bump-charge" and source != "BOOST" and location_at_home:
```
springfall2008: **Review: the completed => off-peak guarantee is budget-gated**

This predicate only widens the *entry* test. Inside the loop each 30-min block is still gated on `slots_per_day[day_offset] < octopus_slot_max`, and the over-cap branch (~:3228 on this branch) assigns `assumed_price = self.rate_max_base` without writing `rates[minute]` — the block is silently left unstamped.

Failure scenario: overnight smart-charge fills a period's 12 blocks; a completed dispatch later relabelled AWAY lands after the counter saturates, its already-metered minutes keep the tariff rate, and `today_cost()` steps the day's cost up exactly as #4946 reports. If Octopus genuinely bills completed dispatches off-peak regardless of the cap, this understates the fix's reach; if the real 6-hour cap applies to away dispatches, this is correct modelling. Either way the premise deserves an explicit decision — it's the same unverified billing assumption the PR body flags.

springfall2008: **Review (minor): with `octopus_slot_low_rate: false`, the relaxation extends a start-price flattening quirk to completed non-home dispatches** (anchoring here; the offending line is `assumed_price = rates.get(start_minutes, self.rate_min)` at ~:3202 on this branch)

In the `octopus_slot_low_rate=False` branch, `assumed_price = rates.get(start_minutes, self.rate_min)` uses the *floored slot start*'s price and stamps it on every budget-added block the slot touches. A completed dispatch crossing an off-peak→peak boundary therefore writes the cheap start price into peak minutes.

Scenario: `octopus_slot_low_rate: false`, completed 04:15-05:15 away dispatch where off-peak ends 04:30 → minutes 04:30-05:15 get stamped with 04:00's price, underpricing the peak tail in both the planner and `today_cost`. Pre-existing for AT_HOME slots (Tests 6/7 can't see it — uniform rates dict / single aligned block), but this PR newly admits completed AWAY/UNABLE_TO_IDENTIFY dispatches, the ones most likely to run past the off-peak window end.

springfall2008: **Review (maintainability): the stamp predicate now exists in three unshared copies with different semantics**

- octopus.py:3064 (`load_octopus_slots`) — location test only, no completion bypass, and it also gates that function's cap counter;
- octopus.py:3094 — location test ANDed with `end_minutes > minutes_now` (completed slots dropped entirely, so it is already consistent);
- here (:3185) — location test with the completed-dispatch bypass.

Nothing at the two untouched sites tells a maintainer the completed-dispatch exemption is deliberately absent there; a well-intentioned "consistency" edit would re-expose the planned-AWAY concern the comment says future slots must keep (the GH#4516 family). Suggest either a shared helper (e.g. `dispatch_billed_off_peak(source, location, end_minutes)`, with :3094's future-only emission test kept separate) or at minimum a short cross-referencing comment (#4946 / #4516) at :3064 and :3094. This also gives the 10-line inline rationale a more durable home than a per-slot loop.

apps/predbat/octopus.py
springfall2008: **Review (nit): `location_at_home` is a misleading name**

It evaluates True for a completed dispatch labelled `AWAY`/`UNABLE_TO_IDENTIFY`, so it no longer means "the location is home" — it means "this dispatch may be stamped off-peak". A reader at `if ... and location_at_home:` will conclude location still gates the branch, and a future "inconsistency fix" that removes the completed bypass would re-expose #4946.

One-word rename fixes it: `stamp_off_peak` or `treat_as_home_dispatch`.

apps/predbat/tests/test_rate_add_io_slots.py
```diff
@@ -384,6 +386,71 @@ def run_rate_add_io_slots_tests(my_predbat):
...
+    slots = [{"start": slot_start.strftime(TIME_FORMAT), "end": slot_end.strftime(TIME_FORMAT), "charge_in_kwh": 2.5, "source": "smart-charge", "location": "UNABLE_TO_IDENTIFY"}]
+
+    expected_rates = {}
+    for minute in range(180, 210):
+        expected_rates[minute] = 4.0
+
+    failed |= run_rate_add_io_slots_test("test21_completed_unknown_location_is_off_peak", my_predbat, slots, True, 12, expected_rates)
+
+    # Test 22: An in-progress AWAY dispatch has not completed yet, so it keeps the location test -
+    # the relaxation is scoped to slots that have finished, not ones that merely started.
```
springfall2008, line 438: **Review (nit): the DST rationale in this comment misdescribes the harness**

"rate minutes are indexed from local midnight, so a fixed UTC hour lands either side of minutes_now depending on the DST offset" — in this harness both `minutes_now` (line 55) and the slot times are built from the same aware `midnight_utc` (predbat.py:654, which despite its name is *local* midnight), and `datetime.strptime(..., TIME_FORMAT)` with `%z` normalises the offsets. So a fixed-hour slot like Tests 10/20/21/23 is already boundary-safe: `minutes_now` is 600 in every timezone/DST state.

Anchoring Test 22 on `minutes_now` is still the more robust habit, but the stated reason will send the next reader hunting for a frame mismatch that doesn't exist. Also worth noting: Tests 20-23 exercise the new budget behaviour only indirectly — no test covers a completed non-home dispatch consuming cap budget ahead of a planned AT_HOME dispatch under the 12-block limit (the divergence flagged at octopus.py:3064).

apps/predbat/octopus.py
springfall2008: _Automated reply from the triage bot._

Confirmed and fixed by making the two counters share one predicate rather than commenting the asymmetry. New `Octopus.dispatch_billed_off_peak(source, location, end_minutes)` now gates both `rate_add_io_slots`' stamping/cap counter and `load_octopus_slots`' cap counter (octopus.py:3101), so a completed non-AT_HOME dispatch spends a low-rate block in both places.

Verified the emission side is unchanged: the completed slot passing the new gate in `load_octopus_slots` still gets dropped by the future-only test at :3131 (`end_minutes > self.minutes_now`), so the only behaviour change there is the counter. Your fetch.py ordering point checks out — fetch.py:1421-1427 appends `completed` then `planned` and neither function re-sorts, so the completed blocks do claim budget first.

Covered by two mutation-verified tests: `test25_completed_away_consumes_cap` in test_rate_add_io_slots.py (completed AWAY yesterday 13:00 spends both blocks of a cap-2 period, planned AT_HOME at 11:00 today is left at the day rate) and "completed AWAY dispatch consumes the low-rate cap" in test_octopus_slots.py (same shape, asserting the emitted slot's `average` is `rate_max_base`). Reverting either call site individually makes the corresponding test fail.

apps/predbat/octopus.py
springfall2008: _Automated reply from the triage bot._

Not implementing the suggested rescope, but documenting the trade-off. Reasoning:

Your reachability point is right — fetch.py:1421-1427 passes `planned_dispatches` through untrimmed, so a straddling slot with a non-home location does reach `rate_add_io_slots` intact. I can't confirm whether in-progress dispatches carry non-home locations in practice.

Scoping the exemption to `start_minutes <= self.minutes_now` would stamp the **entire remaining tail** of the window cheap, not just the metered part, because the gate is per-slot and the stamping loop runs `start_minutes`→`end_minutes`. These windows are not 30 minutes: the HA integration merges contiguous dispatches, and main's own Test 20 models a 21:00→15:00 window. So that rescope trades a bounded mispricing for exactly the #4516 failure the issue asked to preserve — the planner importing against a cheap window that never materialises. Splitting the slot at `minutes_now` avoids that but means carving a partial 30-min block out of a cap-accounting loop main has just changed (#4951), which is more risk than the residual justifies.

The residual is also strictly smaller than what this PR replaces: before, **every** AWAY minute stayed mispriced indefinitely; after, only the elapsed portion of a single in-progress AWAY dispatch, until it finishes. So the change never widens the symptom family.

Recorded in the `dispatch_billed_off_peak` docstring so the next reader sees the boundary is a decision rather than an oversight.

apps/predbat/octopus.py
springfall2008: _Automated reply from the triage bot._

Confirmed the ordering: the `continue` at octopus.py:3204 does run before the eligibility test, so a completed dispatch with `kwh <= 0` is dropped whatever its location.

Leaving the behaviour as-is and acknowledging it in the code. It is the deliberate trade-off Test 18 pins: a zero-kWh past dispatch is treated as never having happened (a withdrawn planned slot Octopus hasn't dropped), and relaxing it would let withdrawn slots eat the day's cap. The target-SoC zeroing you cite is handled in `load_octopus_slots` (:3007-3012) only for dispatches straddling `minutes_now`, which are not completed and so never hit this skip; and on the Octopus-API path completed dispatches read `delta` rather than `chargeKwh`, so a zeroed completed record would have to come from the HA integration specifically. I have no evidence it does.

Added a comment at the skip stating the ordering is intentional. If a zeroed completed record is ever observed from the HA integration, the fix belongs at the skip (distinguishing withdrawn from zeroed), not in the location exemption.

apps/predbat/octopus.py
springfall2008: _Automated reply from the triage bot._

Explicit decision made: the cap **does** apply to completed dispatches, so the budget gating stays and this is intentional modelling, not an understated fix.

Octopus's 6-hour guarantee is a daily allowance; a dispatch that actually ran and was billed off-peak consumed part of it regardless of the location it was later labelled with. Treating completed dispatches as exempt from the cap would let a day report more than 6 hours of off-peak import.

That is now stated in the `dispatch_billed_off_peak` docstring — the helper is off-peak *eligibility* only, with the cap applied on top by both callers — and it is the same decision that made the counter fix in the sibling thread necessary: completed non-home dispatches spend budget in `load_octopus_slots` too now, covered by two new tests.

springfall2008: _Automated reply from the triage bot._

The flattening is real — `assumed_price = rates.get(start_minutes, self.rate_min)` is computed from the floored slot start and stamped across every block the slot touches — but I'm not changing it here.

It is pre-existing and independent of location: a completed **AT_HOME** dispatch crossing the off-peak→peak boundary is mispriced identically today. The claim that AWAY/UNABLE_TO_IDENTIFY dispatches are "the ones most likely to run past the off-peak window end" doesn't hold up — nothing about the location label correlates with window length or alignment, and this PR admits no *longer* dispatches than before.

It also isn't a one-line fix. Using `rates.get(minute)` per minute would make the whole `octopus_slot_low_rate=False` branch a no-op (each minute stamped with its own existing value), so the flattening is load-bearing for whatever that option is meant to express; changing it needs a decision about what the option should do, on a path the majority of users (`octopus_slot_low_rate` defaults True) never take.

Worth its own issue with that question stated, rather than folding an unrelated semantic change into a #4946 regression fix.

springfall2008: _Automated reply from the triage bot._

Taken the shared-helper option. `Octopus.dispatch_billed_off_peak(source, location, end_minutes)` (octopus.py:2969) now covers source exclusion, location and the completed-dispatch exemption, and is called from both octopus.py:3101 (`load_octopus_slots`' cap counter) and octopus.py:3210 (`rate_add_io_slots`).

The third site, the future-only emission test at :3131, is left as a plain location test with a cross-reference comment rather than folded in: it has no source exclusion, and because it already requires `end_minutes > self.minutes_now` the completed exemption can never fire there — so the helper would be equivalent but would silently add the bump-charge/BOOST exclusion to a test that doesn't currently have it. The comment says why it's equivalent so a "consistency" edit has something to read.

The 10-line #4946 rationale now lives in the helper docstring, along with the in-progress boundary and cap decisions from the sibling threads.

apps/predbat/octopus.py
springfall2008: _Automated reply from the triage bot._

`location_at_home` is gone. The predicate is now `Octopus.dispatch_billed_off_peak(source, location, end_minutes)`, which names what it actually decides, and the call site reads `if self.dispatch_billed_off_peak(source, location, end_minutes):` — no local variable implying the location still gates the branch.

apps/predbat/tests/test_rate_add_io_slots.py
springfall2008: _Automated reply from the triage bot._

Both points confirmed and addressed.

The DST rationale was wrong — verified against predbat.py:645-654: `now_utc = datetime.now(self.local_tz)` and `midnight_utc` is that value with `hour=0`, i.e. local midnight as an aware datetime, and the harness derives `minutes_now` from the same value, so it is 600 in every timezone/DST state and a fixed-hour slot is already boundary-safe. Comment rewritten to say the anchor is about the slot straddling `minutes_now` by construction, noting `minutes_now` is in fact fixed here.

On the coverage gap: added `test25_completed_away_consumes_cap` — a completed AWAY dispatch at 13:00 yesterday spends both blocks of a cap-2 period, leaving a planned AT_HOME slot at 11:00 today (same midday-to-midday window, since the cap keys on the slot start) at the day rate. Its mirror in test_octopus_slots.py asserts `load_octopus_slots` now agrees. Both fail if the corresponding call site is reverted to the old location-only test.

Issue #4149, opened 2026-06-29
`find_battery_size` raises `AttributeError: 'PredBat' object has no attribute 'max_days_previous'` on every tick - reproducible on v8.42.4 and v8.42.5


chibikibou:
**Describe the bug**
Every restart, `timer_tick` throws the following, repeating on every subsequent tick (roughly every 2 minutes in my log):

```
2026-06-29 20:35:44.040422: Error: Exception raised 'PredBat' object has no attribute 'max_days_previous'
2026-06-29 20:35:44.099535: Error: Traceback (most recent call last):
  File "/config/predbat.py", line 1689, in update_time_loop
    self.quick_inverter_data_update()
  File "/config/execute.py", line 883, in quick_inverter_data_update
    self.fetch_inverter_data(create=False)
  File "/config/execute.py", line 740, in fetch_inverter_data
    inverter = Inverter(self, id)
  File "/config/inverter.py", line 411, in __init__
    self.battery_size_tracking()
  File "/config/inverter.py", line 598, in battery_size_tracking
    found_size = self.find_battery_size(self.nominal_capacity)
  File "/config/inverter.py", line 722, in find_battery_size
    soc_percent_data = self.base.get_history_wrapper(entity_id=soc_percent_sensor, days=self.base.max_days_previous, required=False)
AttributeError: 'PredBat' object has no attribute 'max_days_previous'. Did you mean: 'days_previous'?
2026-06-29 20:35:44.205307: Warn: record_status Error: Exception raised 'PredBat' object has no attribute 'max_days_previous'
2026-06-29 20:35:44.205506: Error: timer_tick caught exception: 'PredBat' object has no attribute 'max_days_previous'
```

`inverter.py` line 722 (`find_battery_size`) reads `self.base.max_days_previous`, but the `PredBat` object only ever sets `self.days_previous` (the raw apps.yaml config list) — there's no derived `max_days_previous` attribute set anywhere on `self.base`, so this fails unconditionally, every time.

First seen on v8.42.4. Updated to v8.42.5 hoping it would resolve it; still present, identical traceback. Comparing the two versions, the changes between them (GE REST charge-target-enable fix, EV/EVC gateway status fix) don't touch `inverter.py`'s battery sizing code, so this seems to be a preexisting bug rather than something either version introduced.

**Expected behaviour**
`battery_size_tracking()` / `find_battery_size()` should not crash `timer_tick`.

[...]

PR #4968, opened 2026-09-06
fix: skip the 120s quick inverter update while template mode is set

### LINKED ISSUE #4965: AttributeError: 'PredBat' object has no attribute 'inverter_clock_skew_discharge_start' — same underlying defect as #4149 (initialisation-order gap), different attribute
Related to #4149, which reported an identical failure pattern for `max_days_previous` and was closed as fixed. That fix appears to have addressed only the specific attribute reported, not the underlying initialisation-order gap — I'm hitting the same class of bug at a different attribute.

**Environment:**
- Predbat app (add-on) v8.55.0 — confirmed "Up-to-date" via Home Assistant's own update mechanism at time of report
- Also reproduced identically on v8.54.3 and v9.0.0 (beta)
- GivEnergy GIV-HY5.0 Gen3 hybrid inverter + GIV-BAT9.5 battery, single inverter, controlled via GivTCP
- Home Assistant Core 2026.9.1, HAOS 18.2, Raspberry Pi 5

**Description:**
Every `update_time_loop` cycle throws, starting on the very first cycle after startup:

AttributeError: 'PredBat' object has no attribute 'inverter_clock_skew_discharge_start'

at `inverter.py`, in `update_status()`:
```python
discharge_start -= timedelta(seconds=self.base.inverter_clock_skew_discharge_start * 60)
```
`predbat.status` shows a persistent Error state with a climbing error_count — Predbat never produces a working plan.

**Troubleshooting already tried, all reproduce the identical crash:**
- `inverter_clock_skew_discharge_start: 0` set explicitly
- Commented out entirely
- General `inverter_clock_skew_start`/`inverter_clock_skew_end` set (0) with discharge-specific ones unset — no fallback occurs
- `givtcp_rest` enabled and disabled
- `discharge_target_soc` present and commented out
- Confirmed across three versions: v8.55.0 → v9.0.0 (beta) → v8.54.3 → fresh reinstall of v8.55.0 (same crash every time)

**Fix that worked:** removing `template: true` from `apps.yaml` resolved it instantly, exactly as reported by @chibikibou in #4149.

**Conclusion:**
This is the second independently-reported attribute failing the same way (`max_days_previous` in #4149, `inverter_clock_skew_discharge_start` here), with the same underlying fix. Suggests the original patch addressed the specific attribute reported rather than the initialisation-order gap itself — flagging so this can be tracked as a systemic issue rather than a one-off.

Log file and apps.yaml attached (API key redacted).

[apps_redacted.yaml](https://github.com/user-attachments/files/31882628/apps_redacted.yaml)
[predbat_crash_log_redacted2.txt](https://github.com/user-attachments/files/31882629/predbat_crash_log_redacted2.txt)

Thanks for looking into this — happy to provide any further logs or testing if useful.

Dean

springfall2008:
This is an automated draft PR generated from issue #4965 — a maintainer should review it before merging.

Fixes #4965

## Summary

With `Template: True` still set, `update_pred()` early-returns before `fetch_config_options()` runs, but the 120-second quick inverter update in `update_time_loop()` still fires, builds the inverters and calls `update_status()` — which reads attributes only `fetch_config_options()` creates (`inverter_clock_skew_discharge_start` here, `max_days_previous` in #4149), raising an AttributeError every cycle. The #4150 fix added only the one reported attribute to `__init__`, leaving the class of failure open.

This closes the gap at the choke point instead of enumerating attributes: `quick_inverter_data_update()` now returns `False` while template mode is set, mirroring the existing guard in `run_time_loop_balance()`. Guarding inside the function (rather than at the `update_time_loop()` call site suggested as the primary option in the triage comment) covers every present and future caller with a single check, and also the edge where a full update errors out before `fetch_config_options()` runs.

## Testing

- `./run_pre_commit` from `coverage/`: passed (all checks, quick suite green in 81.68s).
- `./run_all --test execute`: passed. Added a regression case to the existing `quick_inverter_data_update` block in `tests/test_execute.py` that sets `template: True` (and removes `inverter_clock_skew_discharge_start` to reproduce the unfetched-config state) and asserts the quick update is skipped; verified it fails on unpatched `main` before the fix.
- `tests/test_control_ledger.py`'s bare `Execute` stub gained a default-returning `get_arg` so the new config read works there (the stub already fakes `fetch_inverter_data` the same way).

## Notes

Root cause per the automated triage on #4965: clock-skew attributes set only in `fetch_config_options()` (apps/predbat/fetch.py:2936), read in `update_status()` (apps/predbat/inverter.py:1624), and template mode blocks the only caller of `fetch_config_options()`.


apps/predbat/execute.py
```diff
@@ -1068,6 +1068,11 @@ def quick_inverter_data_update(self):
         """
         Quick update of inverter data for dashboard
         """
+        # While template mode is set update_pred() early-returns before fetch_config_options(),
+        # so the attributes update_status() reads (e.g. inverter_clock_skew_discharge_start)
+        # were never created - running the quick update would AttributeError every cycle (#4965)
+        if self.get_arg("template", False):
```
springfall2008: Guard confirmed and consistent with predbat.py:2118, but note a side effect: this return happens before inverter_data_last_fetch is ever stamped (only set at execute.py:881 in fetch_inverter_data and predbat.py:1348 after a plan run, both unreachable in template mode). In template mode the tdiff >= INVERTER_QUICK_UPDATE_SECONDS gate at predbat.py:2040 therefore passes on every 15s tick of update_time_loop, not every 120s - pre-fix the crashing path still stamped last_fetch first, so cadence was 120s. Each tick is now a silent no-op (plus a prediction_started True/False toggle), so nothing breaks, but the 120s throttle is dead while template mode is set. Stamping inverter_data_last_fetch here (or guarding at the call site like run_time_loop_balance does) would keep the intended cadence.

apps/predbat/execute.py
```diff
@@ -1068,6 +1068,11 @@ def quick_inverter_data_update(self):
         """
         Quick update of inverter data for dashboard
         """
+        # While template mode is set update_pred() early-returns before fetch_config_options(),
```
springfall2008: Nit on the comment's accuracy: for inverter defs with has_charge_enable_time: False (e.g. Solis), the pre-fix path didn't only AttributeError - fetch_inverter_data(create=False) force-creates Inverter objects (execute.py:908, since self.inverters == [] at predbat.py:317) and update_status() reaches write_and_poll_switch("scheduled_charge_enable", ...) (inverter.py:1487) *before* the unset skew read at inverter.py:1498. So on that hardware the old code could write the real device with no plan. Worth mentioning since the guard also removes that risk.

apps/predbat/tests/test_execute.py
```diff
@@ -3145,6 +3145,26 @@ def run_execute_tests(my_predbat):
     if failed:
         return failed
 
+    # Template mode (#4965): update_pred() early-returns before fetch_config_options(), so the
+    # attributes update_status() reads were never created - the quick update must skip rather
+    # than AttributeError every 120 seconds. Deleting the attribute reproduces that state.
+    my_predbat.args["template"] = True
+    saved_skew = my_predbat.__dict__.pop("inverter_clock_skew_discharge_start", None)
+    try:
```
springfall2008, line 3155: Cleanup isn't exception-safe: the skew restore (3161-3162) and the del below only run if nothing but AttributeError escapes. Any other exception leaves args['template'] = True set on the shared my_predbat instance (args is shared with all components via component_base.py:67). Today an escaping exception also aborts the whole runner (unit_test.py:877 has no per-test try/except) so the leak is latent, but wrapping the call + cleanup in try/finally would make this self-contained.

apps/predbat/tests/test_execute.py
```diff
@@ -3145,6 +3145,26 @@ def run_execute_tests(my_predbat):
...
+    except AttributeError as e:
+        print("ERROR: quick_inverter_data_update raised AttributeError in template mode: {}".format(e))
+        failed = True
+        result = False
+    if result is not False:
+        print("ERROR: quick_inverter_data_update should return False in template mode")
+        failed = True
+    if saved_skew is not None:
+        my_predbat.inverter_clock_skew_discharge_start = saved_skew
+    del my_predbat.args["template"]
```
springfall2008: del assumes the key didn't pre-exist. coverage/apps.yaml:10 currently ships it commented out, so this is latent - but if anyone enables template: True in the test config to exercise template mode end-to-end, this deletes their key mid-run for all later tests. Saving the prior value and restoring it (as done for the skew attr) would be order-safe.

apps/predbat/tests/test_control_ledger.py
```diff
@@ -1845,6 +1845,8 @@ def test_every_entry_point_opens_its_own_cycle():
     stub = Execute.__new__(Execute)
     ledger = ControlLedger()
     stub.control_ledger = ledger
+    # The template-mode guard reads config before anything else; answer with the default (off)
```
springfall2008, line 1848: The stub is needed and correct (the lambda binds default positionally, matching the suite's existing broad-lambda pattern). One coverage gap: this test pins cycle-per-entry-point ordering but was not extended with a template=True sub-test, so nothing pins the new ordering invariant 'template guard returns before control_ledger.begin_cycle()'. A refactor that moves the guard below begin_cycle would burn a ledger cycle every 15s in template mode with no test failing - the exact must-not-burn-a-cycle invariant the inverters-None sub-test two blocks down already pins.

apps/predbat/execute.py
springfall2008: This is now the third inline copy of the template-mode gate (predbat.py:977, predbat.py:2118, here) with no shared helper, and the copies already diverge (predbat.py:977 logs + record_status; the other two return silently). A tiny is_template_mode() helper would keep detection in one place - missing one copy on a future change is exactly how the #4965 AttributeError survived.

apps/predbat/execute.py
```diff
@@ -1068,6 +1068,11 @@ def quick_inverter_data_update(self):
...
+        # so the attributes update_status() reads (e.g. inverter_clock_skew_discharge_start)
+        # were never created - running the quick update would AttributeError every cycle (#4965)
```
springfall2008: Altitude thought (optional deeper fix): the attributes this guard protects against are the four inverter_clock_skew_* attrs set only at fetch.py:2936-2939 - and they are the only fetch_config_options-only attributes read by the quick-update chain's Inverter.update_status(); every other self.base.* read there is already pre-initialised in PredBat.__init__ (e.g. forecast_minutes at predbat.py:351, computed_charge_curve at 528). Initialising the four to 0 in __init__ would close this whole AttributeError class for any future entry point, rather than adding a third entry-point-scoped guard.

apps/predbat/tests/test_execute.py
```diff
@@ -3145,6 +3145,26 @@ def run_execute_tests(my_predbat):
     if failed:
         return failed
 
+    # Template mode (#4965): update_pred() early-returns before fetch_config_options(), so the
```
springfall2008, line 3148: Simplification: with the guard in place the try/except AttributeError can never fire and the pop/restore of the skew attr is never read - the block reduces to the 4-line pattern used at 3135-3143 (args['template'] = True -> assert False return -> del). The pop does document the production state if the guard regresses, so trimming is a judgment call, but 20 lines is a lot of scaffolding for a two-line guard.

apps/predbat/tests/test_execute.py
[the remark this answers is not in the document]
springfall2008: _Automated reply from the triage bot._ Addressed in e04604ce: the call is now wrapped in try/finally, and the prior value of args["template"] is saved and restored (the key is removed on restore if it did not pre-exist), so an unexpected exception can no longer leave the shared my_predbat instance contaminated for later tests.

apps/predbat/tests/test_execute.py
springfall2008: _Automated reply from the triage bot._ Same fix as the Copilot thread above: try/finally around the call, with an order-safe save/restore of args["template"] (and the skew attribute), in e04604ce.

apps/predbat/tests/test_execute.py
springfall2008: _Automated reply from the triage bot._ The bare del is gone in e04604ce: the prior value of args["template"] is saved before the test sets it and restored in the finally block, with the key popped on restore if it did not pre-exist.

apps/predbat/tests/test_execute.py
springfall2008: _Automated reply from the triage bot._ Trimmed in e04604ce: the dead except AttributeError branch is gone and cleanup is a try/finally with an order-safe restore. I kept the pop of the skew attribute since it reproduces the production state (pre-fetch_config_options) that the guard exists for, which felt worth the two lines.

apps/predbat/execute.py
springfall2008: _Automated reply from the triage bot._ Implemented the stamp in e04604ce: the template guard now sets inverter_data_last_fetch = datetime.now() before returning False, so the quick update keeps its intended 120s cadence in template mode instead of firing on every tick of update_time_loop. I went with the stamp rather than the call-site guard since the call site branch also serves non-template paths that rely on the in-function guard ordering.

apps/predbat/execute.py
springfall2008: _Automated reply from the triage bot._ Comment updated in e04604ce to name the second pre-fix failure mode: for inverters without has_charge_enable_time the pre-fix path reached write_and_poll_switch("scheduled_charge_enable", ...) and could write the real device with no plan in place, before the AttributeError - so the guard removes that risk as well.

apps/predbat/tests/test_control_ledger.py
springfall2008: _Automated reply from the triage bot._ Added in e04604ce: a template=True sub-test in test_every_entry_point_opens_its_own_cycle where the stubbed get_arg answers True for the template arg, pinning that quick_inverter_data_update returns False with the ledger cycle unchanged - a refactor that moves the guard below begin_cycle() now fails the test.

apps/predbat/execute.py
springfall2008: _Automated reply from the triage bot._ Added Execute.is_template_mode() in e04604ce; all three gates (update_pred, run_time_loop_balance, and quick_inverter_data_update) now call it. The helper centralises detection only - the predbat.py:977 copy keeps its log + record_status side effects at the call site since those belong to that path specifically.

apps/predbat/execute.py
springfall2008: _Automated reply from the triage bot._ Implemented in e04604ce: the four inverter_clock_skew_* attributes are pre-initialised to 0 in PredBat.__init__ (fetch_config_options() overwrites them each cycle), so any future entry point that runs before the first config fetch reads 0 rather than AttributeError. The entry-point guard stays too, since attribute initialisation cannot prevent the other template-mode hazard - fetch_inverter_data(create=False) force-creating Inverter objects and writing scheduled_charge_enable to the device - which is what the guard is really for.

Issue #4927, opened 2026-09-04
Short periods of grid import preceeding and succeding charge/discharge events.


Jake-Net:
Expensive periods of grid usage just before and just after charge/discharge slots.

**Expected behaviour**
Charging and discharging events should sart from zero grid usage and end with zero grid usage during periods of high import costs.

**Predbat version**
8.54.2

**Environment details**
Givenergy AIO + GW
HAOS 2026.8.3/18.2

<img width="1187" height="472" alt="Image" src="https://github.com/user-attachments/assets/b7ed1015-5825-42ff-8028-bf90818210f5" />

<img width="742" height="435" alt="Image" src="https://github.com/user-attachments/assets/0e254a8a-9e42-4130-b1b3-b25f9d67b41c" />

<img width="692" height="463" alt="Image" src="https://github.com/user-attachments/assets/f55dfc3d-e819-4845-869e-d758369c4c1f" />

[predbat.log](https://github.com/user-attachments/files/31839869/predbat.log)

[predbat_debug.yaml.txt](https://github.com/user-attachments/files/31839923/predbat_debug.yaml.txt)

springfall2008:
Automated first-pass triage — a maintainer will review before any action is taken.

**Classification:** configuration (with a case for a small enhancement, see end)
**Priority:** low — easy workaround, and no Predbat code is misbehaving

**What I investigated**

I replayed the log (reporter is on **v8.54.3**, not 8.54.2) against the write timeline and Predbat's own per-cycle inverter snapshots. The grid-import spikes before and after the charge/discharge slots are fully explained by the GivEnergy inverter's clock running ~25 minutes slow, with skew compensation switched off.

The log measures the skew on every cycle, consistently all night:

```
2026-09-03 20:25: Inverter time 2026-09-03 18:59:49+00:00, Predbat computer time 20:25:01+01:00, difference -25.22 minutes
2026-09-03 22:55: Inverter time 2026-09-03 21:29:34+00:00, Predbat computer time 22:55:01+01:00, difference -25.47 minutes
```

The debug yaml has `clock_skew: 0` and `inverter_clock_skew_start/end`, `inverter_clock_skew_discharge_start/end` all set to `0`, so Predbat writes slot times uncompensated and the inverter executes every window ~25.5 minutes late in real terms. Two measured examples from the night in question:

[...]

PR #4991, opened 2026-09-07
Warn on a moderate inverter clock skew of 5-29 minutes

### LINKED ISSUE #4989: Warn on moderate inverter clock skew (5-29 min) instead of only >=30
Claude (via Rik) here. Spun out from #4927, where a steady ~25-minute GivEnergy inverter clock skew (with skew compensation off) produced short grid-import spikes bracketing every charge/discharge slot, and nothing in the log hinted anything was wrong.

## Problem

`apps/predbat/inverter.py:418` only warns (and calls `auto_restart`) when `abs(tdiff) >= 30` minutes. Below that, the per-cycle line

```
Inverter time ..., Predbat computer time ..., difference -25.22 minutes
```

is the only trace, and it is not a warning — it scrolls past at info level every cycle. A drift of 5–29 minutes silently shifts every slot Predbat writes, because skew compensation (`inverter_clock_skew_start/end`, `inverter_clock_skew_discharge_start/end`) is manual-only with no auto-adjust. The user sees expensive grid import at the edges of every window and has no reason to suspect the clock.

## Suggested enhancement (low priority)

1. **Lower-threshold info/warn.** When `5 <= abs(tdiff) < 30`, emit a single `Warn:` (rate-limited, not every cycle) naming the measured skew and pointing at the `inverter_clock_skew_*` settings as the fix. Keep the `>= 30` behaviour (warn + `auto_restart`) as-is.
2. **Optional:** surface the current measured skew as a sensor/attribute so it is visible without log-diving.
3. **Drive-by:** the `>= 30` guard logs `"...this is {} minutes skewed..."` and `auto_restart("Clock skew >=10 minutes")` — the reason string says 10, the guard says 30. Pick one.

Not proposing automatic skew compensation here — just making a moderate steady drift self-diagnosing.

## Scope

Small: a threshold branch plus rate-limiting in `check_inverter_clock` (`inverter.py:~410-435`), a unit test covering the 5–29 min band, and a docs line under the clock-skew settings. ~half a day.

springfall2008:
This is an automated draft PR generated from issue #4989 — a maintainer should review it before merging.

Fixes #4989

## Summary

Compensation for inverter clock skew is manual only (`inverter_clock_skew_start/end` and `inverter_clock_skew_discharge_start/end`), and `inverter.py` only warned at a skew of 30 minutes or more. A steady drift below that threshold still shifts every charge and export slot Predbat writes, but the only trace was the per-cycle `difference N minutes` line at info level — so a reporter saw expensive grid import bracketing every window with no reason to suspect the clock (#4927).

The skew banding moves out of `update_status()` into a new `Inverter.check_clock_skew()`, and a skew of 5-29 minutes now logs a `Warn:` naming the measured skew and the `inverter_clock_skew_*` settings that compensate for it. It is rate limited to once an hour per inverter (state on the base object, so it is shared with the fresh `Inverter` objects `balance_inverters` builds) rather than firing on every 5-minute cycle, and it is cleared when the skew comes back inside tolerance so a recurrence is reported promptly. The `>= 30` band keeps its existing warning, error status and `auto_restart` unchanged.

Also fixes the drive-by noted in the issue: the restart reason string said `"Clock skew >=10 minutes"` while the guard it sits behind is 30. Both now come from the same `INVERTER_CLOCK_SKEW_RESTART_MINUTES` constant.

No automatic skew compensation is added — this only makes a moderate steady drift self-diagnosing, as scoped in the issue.

## Testing

- `tools/triage_test.sh inverter` — passes. Adds `test_inverter_clock_skew_bands` to `apps/predbat/tests/test_inverter.py` (run from the existing `inverter` module), covering the one-warning-then-rate-limited sequence, the repeat after the hour, silence below 5 minutes with the rate limit cleared, the prompt re-warn on recurrence, and the `>= 30` band still warning, recording an error status and restarting with a reason quoting 30.
- Confirmed the test is not vacuous: raising `INVERTER_CLOCK_SKEW_WARN_MINUTES` to 30 (i.e. the pre-change behaviour) makes it fail with `a 25 minute skew should log exactly one warning, got []`.
- `coverage/run_pre_commit` — all hooks pass and the full quick suite passes (4 slow tests skipped).

## Notes

`tools/debug-journal.md` already records this symptom under "Every charge/export window executes ~N minutes late", from the #4927 investigation that this issue was spun out of; the behaviour it describes is what the new warning surfaces.


apps/predbat/inverter.py
```diff
@@ -138,6 +151,36 @@ def _poll_after_write(self, entity_id, matched, refresh=True, required_unit=None
...
+
+        # Below the restart threshold nothing is restarted, but a steady moderate skew still shifts every
+        # charge and export slot Predbat writes, so say so periodically rather than every cycle (#4989)
+        self.base.restart_active = False
+        if abs(tdiff) >= INVERTER_CLOCK_SKEW_WARN_MINUTES:
+            last_warn = self.base.clock_skew_warn_time.get(self.id, None)
+            if (last_warn is None) or ((now_utc - last_warn) >= timedelta(minutes=INVERTER_CLOCK_SKEW_WARN_REPEAT_MINUTES)):
+                self.base.clock_skew_warn_time[self.id] = now_utc
+                self.base.log(
+                    "Warn: Inverter {} {}. This is below the {} minute restart threshold but will still shift every charge and export slot Predbat writes - correct the inverter clock, or compensate for it with inverter_clock_skew_start/inverter_clock_skew_end and inverter_clock_skew_discharge_start/inverter_clock_skew_discharge_end in apps.yaml".format(
```
springfall2008: Nit (cosmetic): this reads as "Warn: Inverter 0 Inverter time is ..." - `skew_message` already begins with "Inverter time is", so prefixing it with "Inverter {}" doubles the word. Suggest e.g. `"Warn: Inverter {}: {} This is below...".format(self.id, skew_message, ...)` with the sentence capitalised, or dropping the leading "Inverter time is" from the prefix.

apps/predbat/inverter.py
```diff
@@ -138,6 +151,36 @@ def _poll_after_write(self, entity_id, matched, refresh=True, required_unit=None
             waited += step
             delay = min(delay * 2, INVERTER_WRITE_POLL_MAX_INTERVAL)
 
+    def check_clock_skew(self, tdiff, now_utc):
+        """
+        Act on the measured inverter clock skew in minutes, warning (and restarting) on a large skew and warning periodically on a moderate one
+        """
+        skew_message = "Inverter time is {}, Predbat computer time {}, this is {} minutes skewed".format(self.inverter_time, now_utc, tdiff)
```
springfall2008: Nit (efficiency): skew_message is formatted on every call even when the skew is inside tolerance - the common per-cycle case for a healthy inverter (inverters are rebuilt every 5-minute cycle by fetch_inverter_data(create=True)), so this string format runs per inverter per cycle for no output. Consider building it lazily inside the two branches that log it.

apps/predbat/inverter.py
springfall2008: _Automated reply from the triage bot._

Fixed in 0768a9d1 — the message now reads "Warn: Inverter {}: {}. This is below..." so the doubled "Inverter" is gone and the sentence after the skew detail is capitalised.

apps/predbat/inverter.py
springfall2008: _Automated reply from the triage bot._

Fixed in 0768a9d1 — the string is now built inside the two branches that log it, so the common healthy-inverter path formats nothing.

apps/predbat/tests/test_inverter.py
```diff
@@ -2770,6 +2770,112 @@ def _no_sleep(self, seconds):
...
+    Below 5 minutes must stay silent, and the >=30 minute band must keep its warning, error status and
+    restart - with a reason string that matches the threshold it actually uses, not the old ">=10".
+    """
+    failed = False
+    print("**** Running Test: inverter_clock_skew_bands ****")
+
+    orig_log = my_predbat.log
+    saved_skew_times = my_predbat.clock_skew_warn_time
+    saved_status = my_predbat.current_status
+    saved_had_errors = my_predbat.had_errors
```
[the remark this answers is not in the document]
springfall2008: _Automated reply from the triage bot._

Fixed in 0768a9d1 — the test now saves `restart_active` before mutating it and restores the saved value in the `finally` block (it previously unconditionally reset it to `False`), so the shared PredBat instance keeps its pre-test state.

apps/predbat/inverter.py
```diff
@@ -138,6 +151,36 @@ def _poll_after_write(self, entity_id, matched, refresh=True, required_unit=None
...
+
+        if abs(tdiff) >= INVERTER_CLOCK_SKEW_RESTART_MINUTES:
+            message = "Warn: {}, Predbat may not function correctly, please fix this by updating your inverter time, checking HA is synchronising with your inverter, or fixing Predbat computer time zone".format(skew_message)
+            self.base.log(message)
+            self.base.record_status(message, had_errors=True)
+            # Trigger restart
+            self.auto_restart("Clock skew >={} minutes".format(INVERTER_CLOCK_SKEW_RESTART_MINUTES))
```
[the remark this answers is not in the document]
springfall2008: _Automated reply from the triage bot._

Fixed in 0768a9d1 — the reason now reads "Clock skew >= 30 minutes". One note: I checked the other user-visible log/reason strings and there is no existing ">= " spacing convention to match (this is the only ">=" in a formatted string), so this is a standalone readability tweak rather than a consistency fix.

Issue #4497, opened 2026-08-12
HA Octopus v19 overlapping completed/planned dispatches truncate future car plan


cparmar:
## Summary

When Predbat consumes Intelligent Octopus dispatch attributes from the Home Assistant Octopus Energy integration, historical completed dispatches are appended before planned dispatches. The overlap reducer is order-dependent: a completed interval inside a longer planned interval can move the planned interval's end backwards into the past, discarding its entire future remainder.

The result is that the car plan can collapse to a small current slot (or no future Car kWh), even though the ready-by time, vehicle SOC gap, and future planned dispatch remain valid.

Observed with:

- Predbat v8.47.7
- HomeAssistant-OctopusEnergy v19.0.0
- octopus_intelligent_slot pointing at the integration's intelligent-dispatch binary sensor

The relevant octopus.py blob is identical in v8.47.7, the v8.48.0 pre-release, and current main, so v8.48.0 does not appear to contain a fix.

## Minimal payload shape

At 10:37 local time, use an entity with attributes shaped like:

    completed_dispatches:
      - start: 08:00
        end: 08:30
        charge_in_kwh: 5.45
      - start: 08:30
        end: 09:00
        charge_in_kwh: 5.66

    planned_dispatches:
      - start: 07:59
        end: 16:00
        source: SMART
      - start: 17:30
        end: 18:00
        source: SMART

In fetch.py, completed entries are appended before planned entries:

https://github.com/springfall2008/batpred/blob/9dbbecd472382dd25e110ee3d2bfd05d2df1e065/apps/predbat/fetch.py#L1197-L1221

In load_octopus_slots, the first overlap between 07:59-16:00 and 08:00-08:30 takes the start-before-current branch and changes the planned end to 08:00:

https://github.com/springfall2008/batpred/blob/9dbbecd472382dd25e110ee3d2bfd05d2df1e065/apps/predbat/octopus.py#L2557-L2583

That leaves 07:59-08:00, which is already past at 10:37. The valid 09:00-16:00 future remainder is lost.

## Expected behavior

- Completed dispatches can continue to drive historical cheap-rate accounting.
- A planned dispatch that spans completed historical intervals should retain its future remainder.
- Overlap handling should not depend on completed/planned list order.
- If v19's started_dispatches is needed for the active interval, it should be handled explicitly rather than relying on a duplicated completed/planned shape.

## Verified workaround

A Predbat-only Home Assistant template adapter was used to:

[...]

PR #5008, opened 2026-09-08
Fix time-shifted car charging slots in the History view (#5004)

### LINKED ISSUE #5004: Predbat not removing EV charge from history tab and ghost EV load incorrectly showing
**Describe the bug**
As discussed here: https://github.com/springfall2008/batpred/discussions/4587
it has been requested that an issue be opened for the BOT triage 

The issue is that Predbat successfully lists expected car load into the plan when an EV is plugged into a charger thats load is visible to my Sigenergy inverter.

When the car is charged, the history tab lists the EV load into load kWh and not car kWh as should be expected. 

In addition, ghost load then appears in the history tabs car kWh in time slots where no EV was charging.

To assist, here is my apps.yaml

```
pred_bat:
  module: predbat
  class: PredBat

  # Sets the prefix for all created entities in HA - only change if you want to run more than once instance
  prefix: predbat

  # Timezone to work in
  timezone: Europe/London

  # If you are using Predbat outside of HA then set the HA URL and Key (long lived access token here)
  #ha_url: 'http://homeassistant.local:8123'
  #ha_key: 'xxx'

  # Currency, symbol for main currency second symbol for 1/100s e.g. $ c or £ p or e c
  currency_symbols:
    - '£'
    - 'p'

  # Number of threads to use in plan calculation
  # Can be auto for automatic, 0 for off or values 1-N for a fixed number
  threads: auto

  #
  # Sensors, currently more than one can be specified and they will be summed up automatically
  # however if you have two inverters only set one of them as they will both read the same.
  #
  inverter_type: "SIG"
  num_inverters: 1
  #
  # Controls/status - must by 1 per inverter
  #
  load_today:
    - sensor.sigen_plant_daily_load_consumption
  import_today:
    - sensor.sigen_plant_daily_grid_import_energy
  export_today:
    - sensor.sigen_plant_daily_grid_export_energy
  pv_today:
    - sensor.sigen_plant_daily_pv_energy

  charge_rate:
    - input_number.sigen_charge_rate
  discharge_rate:
    - input_number.sigen_discharge_rate
  battery_power:
    - sensor.sigen_plant_battery_power
  pv_power:
    - sensor.sigen_plant_pv_power
  load_power:
    - sensor.sigen_plant_consumed_power
  grid_power:
    - sensor.sigen_plant_grid_active_power

  # change the positive/negative sense of power sensors to match what Predbat expects
  battery_power_invert: true
  grid_power_invert: true

  soc_percent:
    - sensor.sigen_plant_battery_state_of_charge
  soc_kw:
    - sensor.sigen_plant_available_max_discharging_capacity
  soc_max:
    - sensor.sigen_plant_rated_energy_capacity
  battery_temperature:
    - sensor.sigen_inverter_battery_average_cell_temperature
  battery_min_soc:
    - 0
  reserve:
    - number.sigen_plant_ess_discharge_cut_off_state_of_charge
  charge_limit:
    - number.sigen_plant_ess_charge_cut_off_state_of_charge

  # Services to control the inverter
  charge_start_service:
    - service: input_select.select_option
      entity_id: input_select.predbat_requested_mode
      option: "Charging"
  charge_stop_service:
    - service: input_select.select_option
      entity_id: input_select.predbat_requested_mode
      option: "Demand"
  discharge_start_service:
    - service: input_select.select_option
      entity_id: input_select.predbat_requested_mode
      option: "Discharging"
  charge_freeze_service:
    - service: input_select.select_option
      entity_id: input_select.predbat_requested_mode
      option: "Freeze Charging"
  discharge_freeze_service:
    - service: input_select.select_option
      entity_id: input_select.predbat_requested_mode
      option: "Freeze Discharging"

  # Inverter max AC limit (one per inverter)
  # If you have a second inverter for PV only please add the two values together
  inverter_limit:
    - sensor.sigen_plant_available_max_active_power

  # Set the maximum charge/discharge rate of the battery
  battery_rate_max:
    - sensor.sigen_plant_ess_rated_discharging_power

  # Export limit is a software limit set on your inverter that prevents exporting above a given level
  # When enabled Predbat will model this limit
  #export_limit:
  #  - 3600
  #  - 3600
  #
  # The maximum rate the inverter can charge and discharge the battery can be overwritten, this will change
  # the register programming and thus cap the max rates. The default is to use the maximum supported rates (recommended)
  #
  #inverter_limit_charge:
  #  - 4000
  #inverter_limit_discharge:
  #  - 4000

  # Some inverters don't turn off when the rate is set to 0, still charge or discharge at around 200w
  # The value can be set here in watts to model this (doesn't change operation)
  #inverter_battery_rate_min:
  #  - 100

  # Some batteries tail off their charge rate at high soc%
  # enter the charging curve here as a % of the max charge rate for each soc percentage.
  # the default is 1.0 (full power)
  #battery_charge_power_curve:
  #  91 : 0.91
  #  92 : 0.81
  #  93 : 0.71
  #  94 : 0.62
  #  95 : 0.52
  #  96 : 0.43
  #  97 : 0.33
  #  98 : 0.24
  #  99 : 0.24
  #  100 : 0.24

  # Inverter clock skew in minutes, e.g. 1 means it's 1 minute fast and -1 is 1 minute slow
  # Separate start and end options are applied to the start and end time windows, mostly as you want to start late (not early) and finish early (not late)
  # Separate discharge skew for discharge windows only
  inverter_clock_skew_start: 0
  inverter_clock_skew_end: 0
  inverter_clock_skew_discharge_start: 0
  inverter_clock_skew_discharge_end: 0

  # Clock skew adjusts the Appdaemon time
  # This is the time that Predbat takes actions like starting discharge/charging
  # Only use this for workarounds if your inverter time is correct but Predbat is somehow wrong (AppDaemon issue)
  # 1 means add 1 minute to AppDaemon time, -1 takes it away
  clock_skew: 0

  # Solcast cloud interface, set this or the local interface below
  #solcast_host: 'https://api.solcast.com.au/'
  #solcast_api_key: 'xxxx'
  #solcast_poll_hours: 8

  # Set these to match solcast sensor names if not using the cloud interface
  # The regular expression (re:) makes the solcast bit optional
  # If these don't match find your own names in Home Assistant
  pv_forecast_today: re:(sensor.(solcast_|)(pv_forecast_|)forecast_today)
  pv_forecast_tomorrow: re:(sensor.(solcast_|)(pv_forecast_|)forecast_tomorrow)
  pv_forecast_d3: re:(sensor.(solcast_|)(pv_forecast_|)forecast_(day_3|d3))
  pv_forecast_d4: re:(sensor.(solcast_|)(pv_forecast_|)forecast_(day_4|d4))

  # car_charging_energy defines an incrementing sensor which measures the charge added to your car
  # is used for car_charging_hold feature to filter out car charging from the previous load data
  # Automatically set to detect Wallbox and Zappi, if it doesn't match manually enter your sensor name
  # Also adjust car_charging_energy_scale if it's not in kwH to fix the units
  car_charging_energy:
    - 'sensor.sigen_inverter_dc_charger_dc_charger_predbat'
    - 'sensor.sigen_ac_charger_ac_predbat_charger'

  num_cars: 2

  # car_charging_planned is set to a sensor which when positive indicates the car will charged in the upcoming low rate slots
  # This should not be needed if you use Octopus Intelligent Slots which will take priority if enabled
  # The list of possible values is in car_charging_planned_response
  # Auto matches Zappi and Wallbox, or change it for your own
  car_charging_planned:
    - 'sensor.sigen_inverter_dc_charger_running_state'
    - 'sensor.sigen_ac_charger_system_state'

  car_charging_planned_response:
    - 'Charging'
    - 'Scheduled'
    - 'Occupied'
    - 'Reserving'
    - 'Preparing Insulation'
    - 'Preparing Comm'

  # To make planned car charging more accurate, either using car_charging_planned or Octopus Intelligent
  # specify your battery size in kwh, charge limit % and current car battery soc % sensors/values
  # If you have intelligent the battery size and limit will be extracted from Intelligent directly
  # Set the car SoC% if you have it to give an accurate forecast of the cars battery levels
  # One entry per car if you have multiple cars
  car_charging_battery_size:
    - 60.5
    - 80.0
  car_charging_limit: [100, 100]
  car_charging_soc:
    - 'sensor.byd_atto_3_battery_level'
    - 'sensor.ioniq_5_ev_battery_level'

  # If you have Octopus Intelligent Go and are not using the Octopus Direct connection method, enable the intelligent slot information to add to pricing
  # Will automatically disable if not found, or comment out to disable fully
  # When enabled it overrides the 'car_charging_planned' feature and predict the car charging based on the intelligent plan (unless Octopus intelligent charging is False)
  # This matches the intelligent slot from the Octopus Energy integration
  # octopus_intelligent_slot: 're:(binary_sensor.octopus_energy([0-9a-z_]+|)_intelligent_dispatching)'
  # octopus_ready_time: 're:((select|time).octopus_energy_([0-9a-z_]+|)_intelligent_target_time)'
  # octopus_charge_limit: 're:(number.octopus_energy([0-9a-z_]+|)_intelligent_charge_target)'

  # Carbon Intensity data from National grid
  # carbon_postcode: 'SA4 0AB'
  # carbon_automatic: True

  # Example alternative configuration for Ohme integration release >=v0.6.1
  #octopus_intelligent_slot: 'binary_sensor.ohme_slot_active'
  #octopus_ready_time: 'time.ohme_target_time'
  #octopus_charge_limit: 'number.ohme_target_percent'

  # Set this to False if you use Octopus Intelligent slot for car planning but when on another tariff e.g. Agile
  #octopus_slot_low_rate: False

  # Octopus saving session points to the saving session Sensor in the Octopus plugin, when enabled saving sessions will be at the assumed
  # Rate is read automatically from the add-in and converted to pence using the conversion rate below (default is 8)
  # octopus_saving_session: 're:(event.octopus_energy([0-9a-z_]+|)_saving_session_event(s|))'
  # octopus_saving_session_octopoints_per_penny: 8

  # Enter your Axle VPP API key if you have signed up to the Axle service in the UK
  axle_api_key: "redacted"

  # Energy rates
  # Please set one of these three, if multiple are set then Octopus is used first, second rates_import/rates_export and latest basic metric

  # Set import and export entity to point to the Octopus Energy plugin
  # automatically matches your meter number assuming you have only one
  # Will be ignored if you don't have the sensor
  # Or manually set it to the correct sensor names e.g:
  # sensor.octopus_energy_electricity_xxxxxxxxxx_xxxxxxxxxxxxx_current_rate
  # sensor.octopus_energy_electricity_xxxxxxxxxx_xxxxxxxxxxxxx_export_current_rate
  # metric_octopus_import: 're:(sensor.(octopus_energy_|)electricity_[0-9a-z]+_[0-9a-z]+_current_rate)'
  # metric_octopus_export: 're:(sensor.(octopus_energy_|)electricity_[0-9a-z]+_[0-9a-z]+_export_current_rate)'
  # metric_octopus_import: 'sensor.edf_energy_electricity_25j0036423_2600001092501_current_rate'
  # octopus_intelligent_slot: 'binary_sensor.edf_energy_electricity_25j0036423_2600001092501_off_peak'


  # Standing charge can be set to a sensor (e.g. Octopus) or manually entered in pounds here (e.g. 0.50 is 50p)
  # metric_standing_charge: 're:(sensor.(octopus_energy_|)electricity_[0-9a-z]+_[0-9a-z]+_current_standing_charge)'
  metric_standing_charge: 'sensor.edf_energy_electricity_25j0036423_2600001092501_current_standing_charge'

  # Or set your actual rates across time for import and export
  # If start/end is missing it's assumed to be a fixed rate
  # Gaps are filled with 0
  # rates_import:
  #   - start: "00:10:00"
  #     end: "05:30:00"
  #     rate: 7.5
  #   - start: "05:30:00"
  #     end: "00:10:00"
  #     rate: 30.0

  rates_import:
    - start: "23:00:00"
      end: "06:00:00"
      rate: number.edf_off_peak_rate
    - start: "06:00:00"
      end: "23:00:00"
      rate: number.edf_peak_rate

  # rates_export:
  #   - rate: 15.0
  rates_export:
    - rate: 13.0
  # Can be used instead of the plugin to get import rates directly online
  # Overrides metric_octopus_import and rates_import
  # See the 'energy rates' part of the documentation for instructions on how to find the correct URL for your tariff and DNO region
  #
  # rates_import_octopus_url : "https://api.octopus.energy/v1/products/FLUX-IMPORT-23-02-14/electricity-tariffs/E-1R-FLUX-IMPORT-23-02-14-A/standard-unit-rates"
  # rates_import_octopus_url : "https://api.octopus.energy/v1/products/AGILE-24-10-01/electricity-tariffs/E-1R-AGILE-24-10-01-A/standard-unit-rates"

  # Overrides metric_octopus_export and rates_export
  # rates_export_octopus_url: "https://api.octopus.energy/v1/products/FLUX-EXPORT-23-02-14/electricity-tariffs/E-1R-FLUX-EXPORT-23-02-14-A/standard-unit-rates"
  # rates_export_octopus_url: "https://api.octopus.energy/v1/products/AGILE-OUTGOING-19-05-13/electricity-tariffs/E-1R-AGILE-OUTGOING-19-05-13-A/standard-unit-rates/"

  # Import rates can be overridden with rate_import_override
  # Export rates can be overridden with rate_export_override
  # Use the same format as above, but a date can be included if it just applies for a set day (e.g. Octopus power ups)
  # This will override even the Octopus plugin rates if enabled
  #
  #rates_import_override:
  #  - date: '2023-09-10'
  #    start: '14:00:00'
  #    end: '14:30:00'
  #    rate: 5

  # Days previous sets how many days of load history Predbat searches when forecasting your future house load.
  # By default this uses a weighted-bucket forecast that automatically accounts for day-of-week, holiday mode
  # and recency (days_previous_auto) - see the documentation for other options, such as picking specific
  # days/weights instead, or using Load ML.
  days_previous:
    - 7

  # Number of hours forward to forecast, best left as-is unless you have specific reason
  forecast_hours: 48

  # The number of hours ahead to count in charge planning (for cost estimates)
  # It's best to set this on your charge window repeat cycle (24) but you may want to set it higher for more variable
  # tariffs like Agile
  forecast_plan_hours: 24

  # Specify the devices that notifies are sent to, the default is 'notify' which goes to all
  #notify_devices:
  #  - mobile_app_treforsiphone12_2

  # Battery scaling makes the battery smaller (e.g. 0.9) or bigger than its reported
  # If you have an 80% DoD battery that falsely reports it's kwh then set it to 0.8 to report the real figures
  battery_scaling: 1.0

  # Can be used to scale import and export data, used for workarounds
  import_export_scaling: 1.0

  # Export triggers:
  # For each trigger give a name, the minutes of export needed and the energy required in that time
  # Multiple triggers can be set at once so in total you could use too much energy if all run
  # Creates an entity called 'binary_sensor.predbat_export_trigger_<name>' which will be turned On when the condition is valid
  # connect this to your automation to start whatever you want to trigger
  # export_triggers:
  #   - name: 'large'
  #     minutes: 60
  #     energy: 1.0
  #   - name: 'small'
  #     minutes: 15
  #     energy: 0.25

  # Nordpool market energy rates
  #futurerate_url: 'https://dataportal-api.nordpoolgroup.com/api/DayAheadPrices?date=DATE&market=N2EX_DayAhead&deliveryArea=UK&currency=GBP'
  #futurerate_adjust_import: False
  #futurerate_adjust_export: False
  #futurerate_peak_start: "16:00:00"
  #futurerate_peak_end: "19:00:00"
  #futurerate_peak_premium_import: 14
  #futurerate_peak_premium_export: 6.5

  # If you have a sensor that gives the energy consumed by your solar diverter then add it here
  # this will make the predictions more accurate. It should be an incrementing sensor, it can reset at midnight or not
  # It's assumed to be in Kwh but scaling can be applied if need be
  #iboost_energy_today: 'sensor.xxxxx'
  #iboost_energy_scaling: 1.0
```

**Expected behaviour**
It would be expected that the history tab successfully strips the EV charge data from load kWh and leaves it in car kWh so that Predbat can see a history of household load not influenced by car load. It would also be expected that ghost car load be placed into load kwh instead of car kwh as this was not car load.

**Predbat version**

v9.0.1

**Environment details**

- Sigenergy 12KW Inverter + 2 x Sig BAT 10 Batteires
- Sigenergy EVDV 12kw charger
- Sigenergy EVAC EV Charger
- Home Assistant via Proxmox VM

**Screenshots**
Example of correct plan

Image

Example of incorrect History
Image
Image
Image

**Log file**

You can download the logfile from the WebUI on the Dash tab select **predbat.log**

**Predbat debug yaml file**
#

[predbat.log](https://github.com/user-attachments/files/31969474/predbat.log)
[predbat_debug_20260907-103000.yaml.txt](https://github.com/user-attachments/files/31969475/predbat_debug_20260907-103000.yaml.txt)
[predbat_debug_20260908-070000.yaml.txt](https://github.com/user-attachments/files/31969477/predbat_debug_20260908-070000.yaml.txt)

springfall2008:
This is an automated draft PR generated from issue #5004 — a maintainer should review it before merging.

Fixes #5004

## Summary

`calculate_yesterday()` fakes `self.minutes_now` to 0 before it re-simulates yesterday, then calls `yesterday_reconstruct_car_slots()` (`output.py:3301`). That function reads the car-charging energy history with `minute_previous = self.minutes_now + 24 * 60 - minute` (`output.py:3045`), which therefore evaluated as `1440 - minute`. But `yesterday_load_step` is built *before* the fake, with `base_offset = 24 * 60 + <real minutes_now>` (`output.py:3128`), and `self.car_charging_energy` is likewise still indexed in minutes before the real now.

So car energy recorded at yesterday clock time *T* was attributed to plan-axis minute *T − minutes_now*. That produces both reported symptoms:

1. The real charging session stays in the **load kWh** band, because the subtraction is applied at the wrong minutes.
2. **Ghost car kWh** appears where nothing was charging, `minutes_now` earlier than the real session.

Because `calculate_yesterday()` re-runs hourly, the offset changed on every recalculation, so the misalignment moved around between refreshes.

The fix passes the real `minutes_now` (already saved into a local at `output.py:3253` before the fake) into `yesterday_reconstruct_car_slots()` rather than reading the faked `self.minutes_now`. The live plan is unaffected — this is the History/savings view only.

## Testing

- `tools/triage_test.sh calculate_yesterday` — PASSED.
- `coverage/run_pre_commit` — all hooks Passed, and the quick suite it runs finished `All tests passed (4 slow tests skipped, total time: 82.42s)`, exit 0.
- Negative check: with only the one-line fix reverted (tests unchanged), the suite fails with exactly the reported symptoms — `car slot should span 600..630 (GH#5004 shifted it to 240), got {'start': 240, 'end': 270, ...}`, the real session left at `0.1` kWh/step in the load band, and the shifted position zeroed. So the new assertions are not vacuous.

Test changes in `apps/predbat/tests/test_calculate_yesterday.py`:

- `_test_car_slot_from_energy_sensor` (end-to-end through `calculate_yesterday()` with `minutes_now=360`) now records the session at yesterday 10:00 using the correct sensor index, asserts the synthesised slot lands at plan-axis 600, and asserts plan-axis 240 — where the bug painted it — is left untouched. Its previous fixture encoded the buggy `1440 - minute` formula as expected behaviour, which is why the existing suite passed.
- New sub-case 5g pins the invariant directly: with `self.minutes_now` faked to 0 and `minutes_now=360` passed in, the slot must follow the passed value.
- The six existing direct call sites pass `0` explicitly, matching the `minutes_now=0` they already set up.

## Notes

The debug journal already records that `calculate_yesterday()` zeroes `minutes_now` on the shared instance (`tools/debug-journal.md:56`); this is a second consumer that read the faked value where it needed the real one.

One related gap is **left unfixed** as out of scope here: the reconstruction loop only covers plan-axis `0..end_record` (yesterday), while the History view renders `end_record + minutes_now` (through to now). Car charging that happened *today* is therefore still shown in the load band rather than the car band. Extending it needs care — the inner scan would run past `now` into negative offsets, which `get_from_incrementing()` silently wraps by +1440 — so it is better handled separately.

Also noted but not changed: reconstructed slots are always appended to `car_charging_slots[0]` (`output.py:3050`), so with `num_cars: 2` all reconstructed energy folds into car 1's graph. `car_charging_energy` is a single summed series across the configured sensors, so there is no per-car split available at that point.

### DIFF
--- a/apps/predbat/output.py
+++ b/apps/predbat/output.py
@@ -3011,7 +3011,18 @@ class Output:
         )
         self.dashboard_item("binary_sensor." + self.prefix + "_demand", state="on" if isDemand else "off", attributes={"friendly_name": "Predbat is in demand mode", "icon": "mdi:battery-arrow-up"})
 
-    def yesterday_reconstruct_car_slots(self, end_record, yesterday_load_step):
+    def yesterday_reconstruct_car_slots(self, end_record, yesterday_load_step, minutes_now):
+        """
+        Rebuild yesterday's car charging slots and subtract their energy from yesterday's historical load
+
+        :param end_record: Length of the reconstructed day in minutes, i.e. the plan axis is 0 (yesterday midnight) to end_record
+        :param yesterday_load_step: Yesterday's per-step load, keyed on the same plan axis, modified in place
+        :param minutes_now: The real minutes_now of the live plan. calculate_yesterday() fakes self.minutes_now to 0 while it
+                            re-simulates yesterday, but self.car_charging_energy is still indexed in minutes before the real
+                            now - as is yesterday_load_step, which was built with base_offset = 24 * 60 + the real minutes_now.
+                            Reading the car energy with the faked self.minutes_now would time-shift every car slot by
+                            minutes_now against the load it is meant to cancel out (GH#5004).
+        """
         # Normalize to list for multi-car support
         entity_id_config = self.get_arg("octopus_intelligent_slot", indirect=False)
         if entity_id_config and not isinstance(entity_id_config, list):
@@ -3031,7 +3042,7 @@ class Output:
             for start_minute in range(0, end_record, self.plan_interval_minutes):
                 car_energy = 0
                 for minute in range(start_minute, start_minute + self.plan_interval_minutes):
-                    minute_previous = self.minutes_now + 24 * 60 - minute  # How far back in time are we looking
+                    minute_previous = minutes_now + 24 * 60 - minute  # How far back in time are we looking
                     car_energy += self.get_from_incrementing(self.car_charging_energy, minute_previous)
                 if car_energy > 0.1:
                     # Only add the slot if there isn't already one covering this time period
@@ -3287,7 +3298,8 @@ class Output:
         self.car_charging_soc = [0] * len(self.car_charging_soc)
 
         # re-construct car charging slots from non-octopus using the sensor
-        self.yesterday_reconstruct_car_slots(end_record, yesterday_load_step)
+        # Pass the real minutes_now saved above - self.minutes_now has just been faked to 0 but the car energy history is still indexed from the real now
+        self.yesterday_reconstruct_car_slots(end_record, yesterday_load_step, minutes_now)
 
         # Simulate yesterday
         self.prediction = Prediction(self, yesterday_pv_step, yesterday_pv_step, yesterday_load_step, yesterday_load_step, soc_kw=soc_yesterday)
--- a/apps/predbat/tests/test_calculate_yesterday.py
+++ b/apps/predbat/tests/test_calculate_yesterday.py
@@ -513,13 +513,18 @@ def _test_car_slot_from_energy_sensor(my_predbat, failed):
 
     No octopus_intelligent_slot is configured and car_charging_slots starts empty.
     The car_charging_energy sensor records 1.5 kWh in the 30-minute window
-    starting at minute 60 of yesterday.  yesterday_reconstruct_car_slots should
-    synthesise a slot for that window and subtract the corresponding kWh from the
-    load step data before handing it to the Prediction.
+    starting at minute 600 (10:00) of yesterday.  yesterday_reconstruct_car_slots
+    should synthesise a slot for that window and subtract the corresponding kWh
+    from the load step data before handing it to the Prediction.
 
-    Slot: start=60, end=90, kwh=1.5 → load rate = 1.5/0.5h = 3.0 kW
-    Subtraction per 5-min step: 3.0 * 5/60 = 0.25 kWh > FLAT_LOAD_KWH (0.1)
-    Expected inside-slot value: max(0.1 - 0.25, 0) = 0.0
+    This is also the regression for GH#5004: calculate_yesterday fakes
+    self.minutes_now to 0 before reconstructing the car slots, while
+    car_charging_energy and yesterday_load_step are both still indexed from the
+    real now.  Reading the car energy off the faked minutes_now attributed the
+    energy recorded at yesterday clock time T to plan-axis minute T - minutes_now,
+    so the real session stayed in the load band and a ghost car slot appeared
+    minutes_now earlier.  With minutes_now=360 the shifted position of this
+    session is axis minute 240, which must stay untouched.
 
     Also verifies that car_charging_slots and car_charging_soc are fully
     restored to their pre-call values after calculate_yesterday returns.
@@ -543,21 +548,52 @@ def _test_car_slot_from_energy_sensor(my_predbat, failed):
     my_predbat.octopus_slots = [[], [], [], []]
     my_predbat.octopus_intelligent_consider_full = False
 
-    # Inside calculate_yesterday, minutes_now is set to 0 before calling
-    # yesterday_reconstruct_car_slots, so the lookup formula becomes:
-    #   minute_previous = 0 + 1440 - minute
-    # For start_minute=60 the inner scan covers minutes 60..89.
-    # At minute=60: minute_previous = 1380.
-    # get_from_incrementing(data, 1380) = max(data[1380] - data[1381], 0).
-    # Correct representation of an incrementing kWh sensor: the sensor reads
-    # 1.5 kWh at minute 60 (index 1380) and all more-recent times (lower indices),
-    # and 0 at earlier times (indices > 1380).  The telescoping sum over the
-    # 30-minute window yields data[1351] - data[1381] = 1.5 - 0 = 1.5 kWh.
-    my_predbat.car_charging_energy = {k: 1.5 for k in range(0, 1381)}
+    # Keep the load cap deterministic regardless of the order the sub-tests run in
+    my_predbat.car_energy_reported_load = True
+    my_predbat.car_charging_loss = 1.0
+
+    # car_charging_energy is indexed in minutes before the REAL now, so the session
+    # at yesterday 10:00 (plan-axis minute 600) lives at index
+    #   minutes_now + 1440 - 600 = 360 + 1440 - 600 = 1200.
+    # An incrementing kWh sensor is monotonically non-decreasing forward in time
+    # (= decreasing index), so reading 1.5 kWh at index 1200 and every more-recent
+    # index, and 0 at earlier times (indices > 1200), represents a 1.5 kWh jump at
+    # yesterday 10:00.  get_from_incrementing(data, 1200) = 1.5 - 0 = 1.5 kWh.
+    car_session_minute = 600
+    shifted_minute = car_session_minute - 360  # where the GH#5004 bug painted it: 240
+    my_predbat.car_charging_energy = {k: 1.5 for k in range(0, 1201)}
 
     captured_load, original_run_pred = _apply_mocks(my_predbat, now_utc, cost_value=100.0, soc_value=5.0)
 
+    # Snapshot car_charging_slots as the (mocked) prediction sees them - calculate_yesterday
+    # restores the live slots before returning, so they can't be inspected afterwards
+    captured_slots = []
+    mocked_run_prediction = my_predbat.run_prediction
+
+    def _capture_car_slots(*args, **kwargs):
+        """Record the reconstructed car slots then defer to the standard run_prediction mock."""
+        captured_slots.append(copy.deepcopy(my_predbat.car_charging_slots))
+        return mocked_run_prediction(*args, **kwargs)
+
+    my_predbat.run_prediction = _capture_car_slots
+
     my_predbat.calculate_yesterday()
+
+    plan_iv = my_predbat.plan_interval_minutes  # 30 by default
+    slot_end = car_session_minute + plan_iv
+
+    # --- The synthesised slot must sit at the minute the energy was actually recorded ---
+    if not captured_slots:
+        print("ERROR: run_prediction was not called – cannot inspect car slots")
+        failed = True
+    else:
+        slots = captured_slots[0][0]
+        if len(slots) != 1:
+            print("ERROR: expected exactly 1 synthesised car slot, got {}".format(slots))
+            failed = True
+        elif slots[0].get("start") != car_session_minute or slots[0].get("end") != slot_end:
+            print("ERROR: car slot should span {}..{} (GH#5004 shifted it to {}), got {}".format(car_session_minute, slot_end, shifted_minute, slots[0]))
+            failed = True
 
     # --- Verify load subtraction via the captured load_minutes_step ---
     if len(captured_load) < 1:
@@ -565,18 +601,24 @@ def _test_car_slot_from_energy_sensor(my_predbat, failed):
         failed = True
     else:
         load_step = captured_load[0]
-        plan_iv = my_predbat.plan_interval_minutes  # 30 by default
-        slot_end = 60 + plan_iv
 
-        # Inside-slot steps: max(0.1 - 3.0*5/60, 0) = max(0.1 - 0.25, 0) = 0.0
-        for inside_min in range(60, slot_end, PREDICT_STEP):
+        # Inside-slot steps: slot kwh is capped to load_reported (6 * 0.1 = 0.6 kWh),
+        # so the subtraction is 0.6/0.5h * 5/60 = 0.1 kWh/step → max(0.1 - 0.1, 0) = 0.0
+        for inside_min in range(car_session_minute, slot_end, PREDICT_STEP):
             val = load_step.get(inside_min, -1)
             if abs(val) > 1e-9:
                 print("ERROR: step {} (inside energy-sensor slot) should be 0.0 but got {}".format(inside_min, val))
                 failed = True
 
+        # The position the GH#5004 shift used to paint must be left alone entirely
+        for ghost_min in range(shifted_minute, shifted_minute + plan_iv, PREDICT_STEP):
+            val = load_step.get(ghost_min, -1)
+            if abs(val - FLAT_LOAD_KWH) > 1e-9:
+                print("ERROR: step {} is minutes_now before the real session and must be untouched at {}, got {}".format(ghost_min, FLAT_LOAD_KWH, val))
+                failed = True
+
         # Outside-slot steps should be unchanged at FLAT_LOAD_KWH.
-        for outside_min in [0, 5, 55, slot_end, slot_end + 5, 300]:
+        for outside_min in [0, 5, 55, car_session_minute - 5, slot_end, slot_end + 5, 300]:
             val = load_step.get(outside_min, -1)
             if abs(val - FLAT_LOAD_KWH) > 1e-9:
                 print("ERROR: step {} (outside energy-sensor slot) should be {} but got {}".format(outside_min, FLAT_LOAD_KWH, val))
@@ -676,7 +718,7 @@ def _test_reconstruct_car_slots(my_predbat, failed):
     end_record = 24 * 60  # 1440 minutes
     yesterday_load_step = {m: FLAT_LOAD_KWH for m in range(0, end_record, PREDICT_STEP)}
 
-    my_predbat.yesterday_reconstruct_car_slots(end_record, yesterday_load_step)
+    my_predbat.yesterday_reconstruct_car_slots(end_record, yesterday_load_step, 0)
 
     # Exactly one slot should have been added
     slots = my_predbat.car_charging_slots[0]
@@ -734,7 +776,7 @@ def _test_reconstruct_car_slots(my_predbat, failed):
     my_predbat.car_charging_energy = {k: 1.2 for k in range(0, 1381)}  # same energy as 5a
 
     yesterday_load_step = {m: FLAT_LOAD_KWH for m in range(0, end_record, PREDICT_STEP)}
-    my_predbat.yesterday_reconstruct_car_slots(end_record, yesterday_load_step)
+    my_predbat.yesterday_reconstruct_car_slots(end_record, yesterday_load_step, 0)
 
     if len(my_predbat.car_charging_slots[0]) != 1:
         print("ERROR 5b: expected 1 slot (no duplicate), got {}".format(len(my_predbat.car_charging_slots[0])))
@@ -771,7 +813,7 @@ def _test_reconstruct_car_slots(my_predbat, failed):
     my_predbat.load_octopus_slots = _mock_load_octopus_slots
 
     yesterday_load_step = {m: FLAT_LOAD_KWH for m in range(0, end_record, PREDICT_STEP)}
-    my_predbat.yesterday_reconstruct_car_slots(end_record, yesterday_load_step)
+    my_predbat.yesterday_reconstruct_car_slots(end_record, yesterday_load_step, 0)
 
     my_predbat.load_octopus_slots = original_load_octopus_slots
 
@@ -819,7 +861,7 @@ def _test_reconstruct_car_slots(my_predbat, failed):
     my_predbat.car_charging_slots = [[cancelled_slot], [], [], []]
 
     yesterday_load_step_5d = {m: TINY_LOAD for m in range(0, end_record, PREDICT_STEP)}
-    my_predbat.yesterday_reconstruct_car_slots(end_record, yesterday_load_step_5d)
+    my_predbat.yesterday_reconstruct_car_slots(end_record, yesterday_load_step_5d, 0)
 
     # The slot kwh should have been zeroed out.
     if cancelled_slot.get("kwh") != 0:
@@ -857,7 +899,7 @@ def _test_reconstruct_car_slots(my_predbat, failed):
     my_predbat.car_charging_slots = [[adjusted_slot], [], [], []]
 
     yesterday_load_step_5e = {m: MEDIUM_LOAD for m in range(0, end_record, PREDICT_STEP)}
-    my_predbat.yesterday_reconstruct_car_slots(end_record, yesterday_load_step_5e)
+    my_predbat.yesterday_reconstruct_car_slots(end_record, yesterday_load_step_5e, 0)
 
     expected_adj_kwh = (plan_iv // PREDICT_STEP) * MEDIUM_LOAD * my_predbat.car_charging_loss  # 6 * 0.2 * 1.0 = 1.2
     if abs(adjusted_slot.get("kwh", -1) - expected_adj_kwh) > 1e-9:
@@ -905,7 +947,7 @@ def _test_reconstruct_car_slots(my_predbat, failed):
     my_predbat.car_charging_slots = [[slot_car0], [slot_car1], [], []]
 
     yesterday_load_step_5f = {m: TWO_CAR_LOAD for m in range(0, end_record, PREDICT_STEP)}
-    my_predbat.yesterday_reconstruct_car_slots(end_record, yesterday_load_step_5f)
+    my_predbat.yesterday_reconstruct_car_slots(end_record, yesterday_load_step_5f, 0)
 
     # Car 0 should be unchanged (load was sufficient).
     if abs(slot_car0.get("kwh", -1) - 0.6) > 1e-9:
@@ -930,6 +972,52 @@ def _test_reconstruct_car_slots(my_predbat, failed):
         val = yesterday_load_step_5f.get(m, -1)
         if abs(val - TWO_CAR_LOAD) > 1e-9:
             print("ERROR 5f: step {} outside slot should be {}, got {}".format(m, TWO_CAR_LOAD, val))
+            failed = True
+
+    # -----------------------------------------------------------------------
+    # 5g - the passed minutes_now indexes the energy sensor, not self.minutes_now (GH#5004)
+    # -----------------------------------------------------------------------
+    print("calculate_yesterday: Test 5g - car energy is indexed off the passed minutes_now, not the faked one (#5004)")
+
+    # self.minutes_now is 0 here exactly as calculate_yesterday fakes it, while the
+    # real plan is at 06:00 - so the energy sensor is still indexed from 06:00 today.
+    _setup_base(my_predbat, minutes_now=0)
+    real_minutes_now = 360
+    my_predbat.num_cars = 1
+    my_predbat.car_charging_slots = [[] for _ in range(4)]
+    my_predbat.car_energy_reported_load = True
+    my_predbat.car_charging_loss = 1.0
+    my_predbat.octopus_intelligent_consider_full = False
+    my_predbat.octopus_slots = [[], [], [], []]
+    my_predbat.args["octopus_intelligent_slot"] = None
+
+    # 1.2 kWh recorded at yesterday 10:00 = plan-axis minute 600, which is
+    # real_minutes_now + 1440 - 600 = 1200 minutes before the real now.
+    session_minute_5g = 600
+    shifted_minute_5g = session_minute_5g - real_minutes_now  # 240, where the bug painted it
+    my_predbat.car_charging_energy = {k: 1.2 for k in range(0, real_minutes_now + 24 * 60 - session_minute_5g + 1)}
+
+    yesterday_load_step_5g = {m: FLAT_LOAD_KWH for m in range(0, end_record, PREDICT_STEP)}
+    my_predbat.yesterday_reconstruct_car_slots(end_record, yesterday_load_step_5g, real_minutes_now)
+
+    slots_5g = my_predbat.car_charging_slots[0]
+    if len(slots_5g) != 1:
+        print("ERROR 5g: expected 1 synthesised slot, got {}".format(slots_5g))
+        failed = True
+    elif slots_5g[0].get("start") != session_minute_5g:
+        print("ERROR 5g: slot start should be {} (the bug placed it at {}), got {}".format(session_minute_5g, shifted_minute_5g, slots_5g[0].get("start")))
+        failed = True
+
+    # Load is cancelled at the real session minutes and left alone at the shifted ones
+    for m in range(session_minute_5g, session_minute_5g + plan_iv, PREDICT_STEP):
+        val = yesterday_load_step_5g.get(m, -1)
+        if abs(val) > 1e-9:
+            print("ERROR 5g: step {} inside the real session should be 0.0, got {}".format(m, val))
+            failed = True
+    for m in range(shifted_minute_5g, shifted_minute_5g + plan_iv, PREDICT_STEP):
+        val = yesterday_load_step_5g.get(m, -1)
+        if abs(val - FLAT_LOAD_KWH) > 1e-9:
+            print("ERROR 5g: step {} is the shifted position and should be untouched at {}, got {}".format(m, FLAT_LOAD_KWH, val))
             failed = True
 
     # Restore


springfall2008 (commented):
Automated extra-high-effort review of PR 5008 (10 finder angles, per-candidate 1-vote verification, gap sweep). 15 findings inline, ranked most-severe first. The core GH#5004 fix itself verified correct: the corrected index `minutes_now + 1440 - minute` matches the axis that `get_historical_base`/`step_data_history(base_offset=24*60 + minutes_now)` use, both new tests are non-vacuous (they fail with the one-line fix reverted), and the fake/restore block is symmetric. Findings are mostly the same faked-clock bug class left in the octopus half and adjacent paths, plus test-fidelity and cleanup notes.

apps/predbat/output.py
```diff
@@ -3031,7 +3042,7 @@ def yesterday_reconstruct_car_slots(self, end_record, yesterday_load_step):
             for start_minute in range(0, end_record, self.plan_interval_minutes):
                 car_energy = 0
                 for minute in range(start_minute, start_minute + self.plan_interval_minutes):
-                    minute_previous = self.minutes_now + 24 * 60 - minute  # How far back in time are we looking
+                    minute_previous = minutes_now + 24 * 60 - minute  # How far back in time are we looking
```
springfall2008, line 3045: **[review] Yesterday subtraction still walks the live today-plan car slots for non-octopus cars** (verified CONFIRMED; pre-existing, not introduced here)

The sensor path is now indexed off the real `minutes_now`, but the octopus-only reload in `yesterday_reconstruct_car_slots` (`for car_n in range(min(len(entity_id_list), ...))`) leaves every non-octopus car's `car_charging_slots[car_n]` holding the live plan from `fetch_sensor_data_car_planning` -> `plan_car_charging(car_n, self.low_rates)` (fetch.py:1356) - slots on the **today** axis (plan.py builds them from `self.low_rates`, 0 = today midnight). The subtraction loop then caps those kWh against and subtracts them from `yesterday_load_step` at those same plan minutes (i.e. yesterday's clock band), and the synthesis dedup check (`if not any(slot["start"] <= start_minute < slot["end"] ...)`) can be satisfied by those stale slots, suppressing a genuine yesterday session found via the sensor.

Trigger: `num_cars=1`, no `octopus_intelligent_slot`, `car_charging_planned=true` -> today's 10:00 planned slot is subtracted from (and capped/zeroed against) yesterday's 10:00 load band, while yesterday's real session elsewhere is untouched. Same GH#5004 class on the slot-source side; worth either clearing non-octopus slots before the reconstruction or documenting the approximation.

apps/predbat/output.py
```diff
@@ -3287,7 +3298,8 @@ def calculate_yesterday(self):
         self.car_charging_soc = [0] * len(self.car_charging_soc)
 
         # re-construct car charging slots from non-octopus using the sensor
-        self.yesterday_reconstruct_car_slots(end_record, yesterday_load_step)
+        # Pass the real minutes_now saved above - self.minutes_now has just been faked to 0 but the car energy history is still indexed from the real now
+        self.yesterday_reconstruct_car_slots(end_record, yesterday_load_step, minutes_now)
```
springfall2008, line 3302: **[review] The octopus half still runs under the faked clock: today-morning dispatches decode past the faked cap and are dropped** (PLAUSIBLE)

The fix is asymmetric: the sensor path now uses the real `minutes_now`, but `load_octopus_slots` -> `decode_octopus_slot` still caps slot ends at `self.forecast_minutes + self.minutes_now` (octopus.py:2939), which at reconstruct time are the **faked** values (1440 + 0), while the History axis published below extends to `end_record + real minutes_now`.

An IOG dispatch that ran this morning (e.g. today 01:00-02:00 with real minutes_now=360) decodes against the faked `midnight_utc` to plan minutes 1500-1560; the cap collapses end to start (1500 == 1500), `decode_octopus_slot` returns 0/0/0 and the slot is silently dropped - so the today-so-far segment of the widened History axis gets neither a car slot nor its load subtraction on the octopus path (the sensor path cannot recover it either: it only scans `range(0, end_record, ...)`). The PR notes today-band reconstruction as out of scope, but the cap collapse (not just the loop bound) is part of why - worth recording next to that note.

apps/predbat/output.py
```diff
@@ -3287,7 +3298,8 @@ def calculate_yesterday(self):
         self.car_charging_soc = [0] * len(self.car_charging_soc)
 
         # re-construct car charging slots from non-octopus using the sensor
-        self.yesterday_reconstruct_car_slots(end_record, yesterday_load_step)
+        # Pass the real minutes_now saved above - self.minutes_now has just been faked to 0 but the car energy history is still indexed from the real now
```
springfall2008, line 3301: **[review] Active-slot kWh synthesis inside load_octopus_slots keys on the faked minutes_now=0** (PLAUSIBLE; pre-existing, ten lines above the fixed formula)

`if kwh == 0 and start_minutes <= self.minutes_now < end_minutes:` (octopus.py:3145) means "the dispatch covering real now" on the live path; called from `yesterday_reconstruct_car_slots` (output.py:3026) it means "the slot covering plan-minute 0 of the **yesterday** axis", because calculate_yesterday has faked `self.minutes_now = 0`. A completed dispatch straddling yesterday midnight whose chargeKwh Octopus zeroed (target SoC reached, per the comment at octopus.py:3142; or an overlap-trimmed segment per #4497) gets `kwh = remaining * car_charging_rate / 60` synthesised from axis minute 0 with `start` snapped to 0 - a phantom slot at the very start of yesterday's History band whose energy is then subtracted from yesterday's load. Passing the real minutes_now to the octopus half too (same argument the docstring makes for the sensor half) would close this member of the class.

springfall2008, line 3301: **[review] AWAY-labelled completed dispatches are dropped from yesterday's reconstruction because the location filter runs under the faked clock** (PLAUSIBLE; pre-existing)

The octopus reload runs while `self.minutes_now` is faked to 0, so `load_octopus_slots`' location filter (the `(not location or location == "AT_HOME")` test, octopus.py:3228) has no completed-dispatch exemption on this path, and the #4946-style exemption (completed dispatches exempt when `end_minutes <= self.minutes_now`) can never fire - every yesterday-axis slot has end > 0.

A real completed IOG dispatch yesterday 14:00-15:00 retroactively relabelled AWAY (the exact #4946 relabel case) is decoded on the yesterday axis but fails the location test and is dropped from `car_charging_slots`, so its energy is never subtracted from `yesterday_load_step` and no slot renders in the History view - the car energy stays counted as house import in the baseline, overstating savings_yesterday. The docstring's own argument (history is indexed from the real now, not the faked one) applies equally to this branch - it is the one consumer of the fake the fix chose not to parameterise.

apps/predbat/output.py
```diff
@@ -3031,7 +3042,7 @@ def yesterday_reconstruct_car_slots(self, end_record, yesterday_load_step):
             for start_minute in range(0, end_record, self.plan_interval_minutes):
```
springfall2008, line 3042: **[review] Sensor reconstruction only writes car 0; a second car's sessions are never reconstructed or subtracted** (low; acknowledged in the PR notes)

`self.car_charging_slots[0].append(...)` inside `if self.num_cars > 0:` - with `num_cars: 2` and no per-car sensor split (`car_charging_energy` is a single summed series across sensors), all reconstructed energy folds into car 1's graph and car 2's sessions are never subtracted from `yesterday_load_step`. The PR body flags this as noted-but-unchanged; filing so it is tracked alongside the fix rather than lost.

apps/predbat/output.py
```diff
@@ -3011,7 +3011,18 @@ def set_charge_export_status(self, isCharging, isExporting, isDemand):
         )
         self.dashboard_item("binary_sensor." + self.prefix + "_demand", state="on" if isDemand else "off", attributes={"friendly_name": "Predbat is in demand mode", "icon": "mdi:battery-arrow-up"})
 
-    def yesterday_reconstruct_car_slots(self, end_record, yesterday_load_step):
+    def yesterday_reconstruct_car_slots(self, end_record, yesterday_load_step, minutes_now):
```
springfall2008, line 3014: **[review] Third consumer of the faked shared clock - consider fixing the mechanism, not the instance** (altitude)

Per tools/debug-journal.md (~line 56), the calculate_yesterday rewind of `self.minutes_now`/`midnight_utc` on the shared instance is already a known bug class (GH#4900, GH#4804, now GH#5004). This fix threads the real value through as a parameter for the one consumer found today, but the same function still mixes captured locals (`minutes_now`, `minutes_back`, `minute_offset = minutes_now + end_record - minute` at :3350) with faked attribute reads (`self.minutes_now` inside `load_octopus_slots`, `publish_html_plan`'s internal slot alignment, `Prediction`'s constructor snapshot) - exposure for the next reader is unchanged. The in-repo precedent for the deeper fix is `manual_time_origin()` (PR #4906), derived from `now_utc`, which nothing fakes; passing a single yesterday-axis origin object (or retiring the shared-state fake) would retire both this parameter and the class.

apps/predbat/output.py
```diff
@@ -3011,7 +3011,18 @@ def set_charge_export_status(self, isCharging, isExporting, isDemand):
...
+        """
+        Rebuild yesterday's car charging slots and subtract their energy from yesterday's historical load
+
+        :param end_record: Length of the reconstructed day in minutes, i.e. the plan axis is 0 (yesterday midnight) to end_record
+        :param yesterday_load_step: Yesterday's per-step load, keyed on the same plan axis, modified in place
+        :param minutes_now: The real minutes_now of the live plan. calculate_yesterday() fakes self.minutes_now to 0 while it
+                            re-simulates yesterday, but self.car_charging_energy is still indexed in minutes before the real
+                            now - as is yesterday_load_step, which was built with base_offset = 24 * 60 + the real minutes_now.
```
springfall2008, line 3022: **[review] The new docstring mis-states what yesterday_load_step is indexed by** (CONFIRMED; documentation accuracy)

The docstring says "self.car_charging_energy is still indexed in minutes before the real now - **as is yesterday_load_step**, which was built with base_offset = 24 * 60 + the real minutes_now". That contradicts the :param two lines above ("keyed on the same plan axis", 0 = yesterday midnight): only the *underlying data* of yesterday_load_step was fetched with that base_offset; the dict itself is plan-axis keyed, which is exactly why the car lookup must use `minutes_now + 24*60 - minute`. The e2e test docstring (~line 521: "car_charging_energy and yesterday_load_step are both still indexed from the real now") repeats the mis-statement. A maintainer who believes the docstring may "fix" the reconstruction by re-indexing the subtraction loop off the minutes-ago axis - reintroducing a minutes_now time-shift between the slots and the band they cancel. Suggest rewording to: "car_charging_energy is still indexed in minutes before the real now; yesterday_load_step is keyed on the plan axis and its data was fetched with base_offset = 24*60 + the real minutes_now".

apps/predbat/tests/test_calculate_yesterday.py
```diff
@@ -543,40 +548,77 @@ def _test_car_slot_from_energy_sensor(my_predbat, failed):
...
+
+    # car_charging_energy is indexed in minutes before the REAL now, so the session
+    # at yesterday 10:00 (plan-axis minute 600) lives at index
+    #   minutes_now + 1440 - 600 = 360 + 1440 - 600 = 1200.
+    # An incrementing kWh sensor is monotonically non-decreasing forward in time
+    # (= decreasing index), so reading 1.5 kWh at index 1200 and every more-recent
+    # index, and 0 at earlier times (indices > 1200), represents a 1.5 kWh jump at
+    # yesterday 10:00.  get_from_incrementing(data, 1200) = 1.5 - 0 = 1.5 kWh.
+    car_session_minute = 600
+    shifted_minute = car_session_minute - 360  # where the GH#5004 bug painted it: 240
```
springfall2008, line 563: **[review] Regression mock swallows base_offset, so the load axis is never actually offset in the test** (CONFIRMED; test fidelity)

`_mock(item, minutes_now, forward, step=5, scale_today=1.0, scale_fixed=1.0, **kwargs)` (test lines 51-72 / 203-215) discards the `base_offset=24*60 + minutes_now` kwarg that production passes (output.py:3117-3119) and returns FLAT_LOAD_KWH (0.1) for every key on whichever axis is asked. Because all load values are flat, the test cannot detect a partial regression that misaligns the load axis (e.g. rebuilding yesterday_load_step with base_offset=24*60 while the car lookup keeps `minutes_now + 1440 - minute`) - every assertion stays green while production reintroduces a shifted cancellation. The exact GH#5004 bug *is* caught, but only via the car-side slot-position assertion. Making the mock honour base_offset (or returning a value that varies with the key) would close that gap.

apps/predbat/tests/test_calculate_yesterday.py
```diff
@@ -543,40 +548,77 @@ def _test_car_slot_from_energy_sensor(my_predbat, failed):
...
-    # For start_minute=60 the inner scan covers minutes 60..89.
-    # At minute=60: minute_previous = 1380.
-    # get_from_incrementing(data, 1380) = max(data[1380] - data[1381], 0).
-    # Correct representation of an incrementing kWh sensor: the sensor reads
-    # 1.5 kWh at minute 60 (index 1380) and all more-recent times (lower indices),
-    # and 0 at earlier times (indices > 1380).  The telescoping sum over the
-    # 30-minute window yields data[1351] - data[1381] = 1.5 - 0 = 1.5 kWh.
-    my_predbat.car_charging_energy = {k: 1.5 for k in range(0, 1381)}
+    # Keep the load cap deterministic regardless of the order the sub-tests run in
+    my_predbat.car_energy_reported_load = True
```
springfall2008, line 552: **[review] Coverage is one-sided: only the car_energy_reported_load=True branch is exercised** (CONFIRMED; coverage note)

Every test in this file pins `car_energy_reported_load = True` (552, 806, 843, 888, 988 - "keep the load cap deterministic"), so the capped branch of the subtraction is the only one ever run. The `False` configuration (charger outside the CT clamp - a supported switch, config.py:706-711, default True) takes no subtraction at all (the whole block sits inside `if self.car_energy_reported_load:`, output.py:3052) and instead relies on fetch.py:3197 force-disabling car_charging_hold plus the simulation re-adding car energy. That no-subtraction / hold-disabled path has zero coverage; one case with the flag off (slot stays in the load band, no subtraction, hold not re-enabled) would pin it.

apps/predbat/tests/test_calculate_yesterday.py
```diff
@@ -543,40 +548,77 @@ def _test_car_slot_from_energy_sensor(my_predbat, failed):
...
+    car_session_minute = 600
+    shifted_minute = car_session_minute - 360  # where the GH#5004 bug painted it: 240
+    my_predbat.car_charging_energy = {k: 1.5 for k in range(0, 1201)}
 
     captured_load, original_run_pred = _apply_mocks(my_predbat, now_utc, cost_value=100.0, soc_value=5.0)
 
+    # Snapshot car_charging_slots as the (mocked) prediction sees them - calculate_yesterday
+    # restores the live slots before returning, so they can't be inspected afterwards
+    captured_slots = []
+    mocked_run_prediction = my_predbat.run_prediction
```
springfall2008, line 571: **[review] New/updated tests mutate shared state and monkey-patches with no restore (style-level; verified no current victim)** (low)

5g and this test set `car_energy_reported_load`, `car_charging_loss` and `args["octopus_intelligent_slot"]` and never restore them (the cleanups restore only num_cars/slots/soc/energy); `_setup_base` and `reset_inverter` (test_infra.py:612-680) reset neither, and `_setup_base(minutes_now=0)` also leaves `self.minutes_now = 0` on the shared instance. Verified the impact is nil today: the leaked values equal the fresh-instance values (fetch.py:3068, predbat.py:476, test_infra.py:528), the args key is unconfigured in coverage/apps.yaml, and no later-registered test reads them (calculate_yesterday registers near the end of TEST_REGISTRY, unit_test.py:690). The patch chain also has no try/finally, but unit_test.py:885 has no per-test try/except, so an exception aborts the whole suite rather than poisoning later tests. Hygiene only: save/restore the three values, or wrap the body in try/finally around _restore_methods.

apps/predbat/output.py
springfall2008, line 3045: **[review] The minutes-ago conversion re-implements the existing get_historical_base helper** (cleanup)

`minute_previous = minutes_now + 24 * 60 - minute; car_energy += self.get_from_incrementing(...)` is an inline copy of `get_historical_base(data, minute, base_minutes)` (fetch.py:714-723: `minute_previous = base_minutes - minute; return self.get_from_incrementing(data, minute_previous)`), inherited via the Fetch mixin - `self.get_historical_base(self.car_charging_energy, minute, minutes_now + 24*60)` is a drop-in. calculate_yesterday alone already carries three hand-written variants of this conversion (:3045 exact, the -5-fudged cost/SoC arrays at :3205/:3227, and the base_offset-parameterised one inside step_data_history); routing this one through the helper leaves one place to change the convention - and the GH#5004 class is precisely what happens when one copy drifts.

apps/predbat/tests/test_calculate_yesterday.py
```diff
@@ -932,6 +974,52 @@ def _mock_load_octopus_slots(car_n, raw_slots, consider_full):
             print("ERROR 5f: step {} outside slot should be {}, got {}".format(m, TWO_CAR_LOAD, val))
             failed = True
 
+    # -----------------------------------------------------------------------
+    # 5g - the passed minutes_now indexes the energy sensor, not self.minutes_now (GH#5004)
+    # -----------------------------------------------------------------------
+    print("calculate_yesterday: Test 5g - car energy is indexed off the passed minutes_now, not the faked one (#5004)")
+
+    # self.minutes_now is 0 here exactly as calculate_yesterday fakes it, while the
+    # real plan is at 06:00 - so the energy sensor is still indexed from 06:00 today.
```
springfall2008, line 983: **[review] Sub-case 5g duplicates the coverage the rewritten e2e test already provides** (cleanup)

The e2e test already runs calculate_yesterday with self.minutes_now faked to 0 (exactly the GH#5004 condition) and real minutes_now=360, and asserts the slot lands at 600 and the ghost band 240..270 is untouched - if yesterday_reconstruct_car_slots read the faked value, the e2e ghost-band assertion fails. 5g re-checks the same invariant through a direct call with the same fixture shape. Keeping it is defensible (it pins the parameter contract directly), but two near-identical 40-line blocks must now be kept in lockstep; consider shrinking 5g to the direct-call slot-start assert or folding it into 5a.

apps/predbat/output.py
springfall2008, line 3042: **[review] Reconstruction scan runs even when car_charging_energy is empty** (cleanup/efficiency)

The loop is guarded only by `if self.num_cars > 0:` - with a car configured but no car_charging_energy entity (load_car_energy() leaves it {}, fetch.py:3227) or a fetch that returned nothing, every hourly calculate_yesterday performs ~1440 get_from_incrementing calls (~2880 dict lookups) that can never pass the `car_energy > 0.1` gate. `if self.num_cars > 0 and self.car_charging_energy:` is provably output-identical and skips it.

apps/predbat/tests/test_calculate_yesterday.py
```diff
@@ -543,40 +548,77 @@ def _test_car_slot_from_energy_sensor(my_predbat, failed):
...
+    my_predbat.car_charging_loss = 1.0
+
+    # car_charging_energy is indexed in minutes before the REAL now, so the session
+    # at yesterday 10:00 (plan-axis minute 600) lives at index
+    #   minutes_now + 1440 - 600 = 360 + 1440 - 600 = 1200.
+    # An incrementing kWh sensor is monotonically non-decreasing forward in time
+    # (= decreasing index), so reading 1.5 kWh at index 1200 and every more-recent
+    # index, and 0 at earlier times (indices > 1200), represents a 1.5 kWh jump at
+    # yesterday 10:00.  get_from_incrementing(data, 1200) = 1.5 - 0 = 1.5 kWh.
+    car_session_minute = 600
```
springfall2008, line 562: **[review] Fixture hardcodes 360 and 1200 instead of deriving them** (cleanup, minor)

`shifted_minute = car_session_minute - 360` and `{k: 1.5 for k in range(0, 1201)}` hardcode the minutes_now that _setup_base() supplies as its default and the derived sensor index; sub-case 5g already does this properly with a named real_minutes_now and a derived expression. Hoisting `real_minutes_now = my_predbat.minutes_now` here and deriving the index makes the test survive a change to _setup_base's default instead of failing with a misleading "expected 1 slot, got 0".

apps/predbat/output.py
```diff
@@ -3011,7 +3011,18 @@ def set_charge_export_status(self, isCharging, isExporting, isDemand):
...
         self.dashboard_item("binary_sensor." + self.prefix + "_demand", state="on" if isDemand else "off", attributes={"friendly_name": "Predbat is in demand mode", "icon": "mdi:battery-arrow-up"})
 
-    def yesterday_reconstruct_car_slots(self, end_record, yesterday_load_step):
+    def yesterday_reconstruct_car_slots(self, end_record, yesterday_load_step, minutes_now):
+        """
+        Rebuild yesterday's car charging slots and subtract their energy from yesterday's historical load
+
+        :param end_record: Length of the reconstructed day in minutes, i.e. the plan axis is 0 (yesterday midnight) to end_record
+        :param yesterday_load_step: Yesterday's per-step load, keyed on the same plan axis, modified in place
+        :param minutes_now: The real minutes_now of the live plan. calculate_yesterday() fakes self.minutes_now to 0 while it
```
springfall2008, line 3020: **[review] The 5-sentence GH#5004 mechanism explanation is duplicated between this docstring and the call-site comment** (cleanup, minor)

The docstring (3020-3024) and the inline comment at the call site (~3299: "self.minutes_now has just been faked to 0 but the car energy history is still indexed from the real now") both carry the full explanation. Keeping the load-bearing copy at the call site and a one-line :param entry here ("Real minutes_now of the live plan - the car energy history is indexed from it, not from the faked self.minutes_now (GH#5004)") leaves a single copy to keep in sync (see also the docstring-accuracy comment on line 3022).

chalfontchubby:
Closing - auto pr didn;t do a great job - will consider a clean sheet redo

PR #5019, opened 2026-09-09
docs(car-charging): document car_charging_battery_size as a list, matching the validator

### LINKED ISSUE #2172: apps.yaml validation error: car_charging_battery_size
The validator is saying that my value for car_charging_battery_size should be a sensor yet the documentation says;

> car_charging_battery_size - Set this value in apps.yaml to the car's battery size in kWh which must be entered with one decimal place, e.g. 50.0. If not set, Predbat defaults to 100.0kWh. This will be used to predict when to stop car charging.

and my value is set to 77.4 (fixed value, not a sensor, as no such sensor exists).

springfall2008:
This is an automated draft PR generated from issue #2172 — a maintainer should review it before merging.

Fixes #2172

_(Rewritten from scratch and force-pushed by Claude after the review on the first attempt — see the note at the bottom.)_

## Summary

The car-charging docs told users to write `car_charging_battery_size` as a plain value ("must be entered with one decimal place, e.g. 50.0"), but `validate_config()` only wraps a *string* scalar into a list for `sensor`-typed items, so a bare number never reaches the "fixed float values are allowed" check and is reported as `'car_charging_battery_size' is not of type 'sensor'` on every run.

Per the direction on the issue — allowing scalars for any `sensor` item would invite worse configs for keys like `pv_forecast_today` and `car_charging_soc` — this fixes the documentation rather than the validator.

The bullet now shows the list form inline:

```yaml
  car_charging_battery_size:
    - 75
```

rather than pointing at the example block further up the page, which is entirely commented out — a user who uncommented only the key line would leave `#  - 75` commented, producing YAML `null` and the same validation error. The decimal-place claim is dropped too: the templates use `- 75` and `fetch.py` reads the value through `float()`.

Six subtests pin the validator's actual behaviour: a one-entry list passes, a whole number in a list passes, a bare number fails as #2172 reported, **a string scalar still passes** (strings are wrapped, so a list is not the only accepted form), and with `num_cars: 2` a two-entry list passes while an inline number fails as the *wrong type* rather than the wrong sensor — a different error message from the one-car case.

No code path changes; car charging itself was never affected, only the `config_valid` sensor and the "found N configuration errors" log line.

## What changed from the first attempt

The review found the first version's doc prose overstated the rule, and that its 49-line `templates/` + `docs/` filesystem scan was wrong in both directions — it missed the dominant key-then-next-line YAML form (which is what actually fails) while flagging the valid `[77.4]` flow list and any key line carrying a trailing comment. Both confirmed by running the regex against `yaml.safe_load`.

Rather than patch the regex, the scan is gone. It was guarding a doc example against drift, but the invariant it claimed to enforce cannot be checked by matching line starts, and a YAML-parsing version would still have missed `apps/predbat/config/apps.yaml` — the template users actually copy — and mis-flagged the many templates that ship this key commented out. The validator subtests are the real regression guard; the doc is a doc.

Two review points are deliberately **not** addressed here, as pre-existing and out of scope for a docs fix:

- `test_infra.py` sets `num_cars: 2` with `car_charging_battery_size: 100.0` (an inline scalar), so that key sits in the shared fixture's `arg_errors` baseline on every `validate_config()` call in this module.
- `car_charging_battery_size: [0]` validates clean despite the spec's `zero: False`, and `plan.py` then divides by it (`fetch.py` guards, `plan.py` does not).

Both are worth fixing; neither belongs in a two-line documentation change. Happy to file them.

## Testing

- `./run_all --test validate_config` — passes; the log shows both distinct error messages (`not of type 'sensor'` at `num_cars: 1`, `is not a list, but requires 2 entries` at `num_cars: 2`).
- `coverage/run_pre_commit` (12 hooks + `./run_all --quick`) — all passed, 4 slow tests skipped, 78.02s.

### DIFF
--- a/apps/predbat/tests/test_validate_config.py
+++ b/apps/predbat/tests/test_validate_config.py
@@ -393,6 +393,43 @@ def test_validate_config(my_predbat):
         print(f"  [transient_ok] {name} pointing at an entity that does not exist still fails")
         _run(my_predbat, {name: "sensor.test_charger_typo"}, expect_errors=[name])
 
+    # ==========================================================================
+    # car_charging_battery_size  (#2172)
+    #
+    # A fixed battery size is written as a list with one entry per car, the form
+    # every apps.yaml template uses. A bare number on the same line as the key
+    # is not accepted: the sensor branch wraps a scalar into a list only when it
+    # is a string (predbat.py), so a numeric scalar never reaches the "fixed
+    # float values are allowed" check. The car-charging docs used to describe
+    # that failing form ("must be entered with one decimal place, e.g. 50.0"),
+    # which is what #2172 reported.
+    # ==========================================================================
+    print("  [sensor float] car_charging_battery_size as a one-entry list passes")
+    _run(my_predbat, {"car_charging_battery_size": [77.4], "num_cars": 1}, expect_clean=["car_charging_battery_size"])
+
+    print("  [sensor float] car_charging_battery_size as a whole number in a list passes (templates use '- 75')")
+    _run(my_predbat, {"car_charging_battery_size": [75], "num_cars": 1}, expect_clean=["car_charging_battery_size"])
+
+    print("  [sensor float] car_charging_battery_size written inline as a number fails (#2172)")
+    _run(my_predbat, {"car_charging_battery_size": 77.4, "num_cars": 1}, expect_errors=["car_charging_battery_size"])
+
+    # A string scalar IS wrapped into a list, so it validates - the docs must not
+    # claim a list is the only accepted form, only that an inline number fails.
+    print("  [sensor float] car_charging_battery_size as an entity name still passes (scalar strings are wrapped)")
+    _run(
+        my_predbat,
+        {"car_charging_battery_size": "sensor.test_car_battery_size", "num_cars": 1},
+        extra_states={"sensor.test_car_battery_size": 77.4},
+        expect_clean=["car_charging_battery_size"],
+    )
+
+    # entries is "num_cars", so the list must have one entry per car. With two
+    # cars an inline number fails earlier, as the wrong type rather than the
+    # wrong sensor, so the reported error differs from the one-car case above.
+    print("  [sensor float] car_charging_battery_size needs one entry per car")
+    _run(my_predbat, {"car_charging_battery_size": [77.4, 64.0], "num_cars": 2}, expect_clean=["car_charging_battery_size"])
+    _run(my_predbat, {"car_charging_battery_size": 77.4, "num_cars": 2}, expect_errors=["car_charging_battery_size"])
+
     print("**** test_validate_config PASSED ****")
     return False
 
--- a/docs/car-charging.md
+++ b/docs/car-charging.md
@@ -219,7 +219,15 @@ To make Predbat-led car charging more accurate, additionally you can configure t
   #  - 're:sensor.tsunami_battery'
 ```
 
-- **car_charging_battery_size** - Set this value in `apps.yaml` to the car's battery size in kWh which *must* be entered with one decimal place, e.g. 50.0.
+- **car_charging_battery_size** - Set this value in `apps.yaml` to the car's battery size in kWh, as a list with one entry per car:
+
+```yaml
+  car_charging_battery_size:
+    - 75
+```
+
+Writing the number on the same line as the key (`car_charging_battery_size: 75`) fails Predbat's `apps.yaml` validation
+with *"is not of type 'sensor'"*, so use the list form above. A whole number is fine - a decimal place is not required.
 If not set, Predbat defaults to 100.0kWh. This will be used to predict when Predbat will stop car charging.
 
 - **car_charging_limit** - You should configure this to point to a sensor that specifies the % limit the car is set to charge to.


apps/predbat/tests/test_validate_config.py
```diff
@@ -28,9 +28,35 @@
...
+    """Check no shipped template or doc example writes car_charging_battery_size as an inline value.
+
+    A one-entry list is the only form validate_config() accepts for this item (#2172): the
+    sensor branch wraps a bare string into a list but not a number, so a numeric scalar never
+    reaches the "fixed float values are allowed" check and is reported as "not of type 'sensor'"
+    on every run. An example that puts the value on the same line as the key therefore hands the
+    reader a config that permanently fails validation, which is how #2172 was raised.
+    """
+    repo_root = os.path.dirname(os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__)))))
+    inline_value = re.compile(r"^\s*#?\s*car_charging_battery_size:[ \t]*\S")
```
springfall2008: **Bug (guard hole, dominant YAML form):** the regex only inspects the key's own line, but the layout these templates overwhelmingly use is key-then-next-line. `car_charging_battery_size:` with a plain scalar on the next line (`77.4`, no dash) parses as a float (verified with yaml.safe_load), which fails validate_config() with num_cars=1 — the exact #2172 error — yet the scan passes it. Quoted key, space-before-colon, and double-hash spellings also evade it, so the guard does not enforce the invariant its docstring states. Suggested fix: parse each file's YAML and fail on a non-list value instead of regexing line starts.

springfall2008: **Bug (guard over-blocks valid forms, false message):** `car_charging_battery_size: [77.4]` (inline flow list) validates clean via the fixed-float branch, and `car_charging_battery_size:   # comment` is a valid key-only line — but both match the regex, failing CI with "which fails apps.yaml validation", which is false for both. Commented key lines are the dominant shipped style, so any future template that adds a value hint or trailing comment on this key breaks the build. Consider anchoring the value to a numeric pattern (e.g. `[ \t]*[0-9.]+[ \t]*$`) and excluding comment lines.

docs/car-charging.md
```diff
@@ -219,7 +219,10 @@ To make Predbat-led car charging more accurate, additionally you can configure t
   #  - 're:sensor.tsunami_battery'
 ```
 
-- **car_charging_battery_size** - Set this value in `apps.yaml` to the car's battery size in kWh which *must* be entered with one decimal place, e.g. 50.0.
+- **car_charging_battery_size** - Set this value in `apps.yaml` to the car's battery size in kWh, e.g. 77.4.
```
springfall2008: **Doc inaccuracy:** "It *must* be entered as a list" overstates the validator: a bare *string* scalar is wrapped into a list (predbat.py:1652), so `car_charging_battery_size: sensor.my_car_kwh` or `"77.4"` validates fine, and the quoted error text only appears with num_cars=1 — with num_cars>=2 the same inline number is reported as "Invalid type, expected list" (predbat.py:1499). A numeric scalar is the only inline form that fails.

apps/predbat/tests/test_validate_config.py
```diff
@@ -28,9 +28,35 @@
 """
 
 import os
+import re
 import tempfile
```
springfall2008, line 33: **Test-isolation hazard:** _run() snapshots args/dummy_items but not self.manual_api, and car_charging_battery_size is in CONFIG_API_OVERRIDE (config.py:1734), so validate_config's get_arg routes through get_manual_api (userinterface.py:218). A leftover manual_api entry from an earlier-registered test would silently decide subtests 1-3, and the override path mutates the returned list in place (userinterface.py:225/237), which a shallow args copy cannot undo. Worth snapshotting/restoring manual_api in _run.

apps/predbat/tests/test_validate_config.py
```diff
@@ -28,9 +28,35 @@
...
+    A one-entry list is the only form validate_config() accepts for this item (#2172): the
+    sensor branch wraps a bare string into a list but not a number, so a numeric scalar never
+    reaches the "fixed float values are allowed" check and is reported as "not of type 'sensor'"
+    on every run. An example that puts the value on the same line as the key therefore hands the
+    reader a config that permanently fails validation, which is how #2172 was raised.
+    """
+    repo_root = os.path.dirname(os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__)))))
+    inline_value = re.compile(r"^\s*#?\s*car_charging_battery_size:[ \t]*\S")
+    checked = 0
+    for folder, suffix in (("templates", ".yaml"), ("docs", ".md")):
```
springfall2008: **Coverage gap:** the scan walks repo-root templates/ and docs/ but skips apps/predbat/config/apps.yaml — the shipped template users actually copy (line 371 carries this key, commented). The docstring's "no shipped template or doc example" claim does not cover it.

apps/predbat/tests/test_validate_config.py
```diff
@@ -28,9 +28,35 @@
...
+        folder_path = os.path.join(repo_root, folder)
+        for name in sorted(os.listdir(folder_path)):
+            if not name.endswith(suffix):
+                continue
+            with open(os.path.join(folder_path, name), encoding="utf-8") as handle:
+                lines = handle.read().splitlines()
+            checked += 1
+            for number, line in enumerate(lines, 1):
+                assert not inline_value.match(line), "{}/{}:{} sets car_charging_battery_size inline, which fails apps.yaml validation: {}".format(folder, name, number, line.strip())
+    assert checked > 30, "Expected the templates and docs to be found and checked, only saw {}".format(checked)
```
springfall2008: **Fragile coverage proxy:** checked > 30 is a combined count (31 templates + 29 docs today), so 31 templates alone satisfy it — if docs/*.md is ever moved into subdirectories the doc half of the guard disappears silently, and the os.listdir scan is non-recursive so docs/annual/ and docs/superpowers/ pages are already outside it. Per-folder minimums (e.g. >25 each) would match the intent.

apps/predbat/tests/test_validate_config.py
```diff
@@ -393,6 +419,29 @@ def test_validate_config(my_predbat):
...
+    #
+    # A fixed battery size is a list with one entry per car, which is the form
+    # every apps.yaml template and doc example uses. A bare number on the same
+    # line as the key is not accepted - with num_cars 1 it reaches the sensor
+    # branch, which only wraps strings into a list, so it is reported as
+    # "not of type 'sensor'". The car-charging docs used to describe that
+    # failing form ("must be entered with one decimal place, e.g. 50.0"), so
+    # the doc/template scan below is what keeps the documented form working.
+    # ==========================================================================
+    print("  [sensor float] car_charging_battery_size as a one-entry list passes")
```
springfall2008, line 407: **Coverage gap:** all four subtests pin num_cars=1; the spec is entries:"num_cars", so the two-car list, the silent over-length trim (predbat.py:1655), and the "Invalid number of entries" path (predbat.py:1493) for this arg are untested. A regression there ships green.

apps/predbat/tests/test_validate_config.py
```diff
@@ -28,9 +28,35 @@
 """
 
 import os
+import re
 import tempfile
 
 
+def _assert_car_battery_size_documented_as_a_list():
+    """Check no shipped template or doc example writes car_charging_battery_size as an inline value.
```
springfall2008: **Note:** the shared fixture itself ships the banned form — test_infra.py sets num_cars: 2 with car_charging_battery_size: 100.0 (inline scalar), which lands in the arg_errors baseline ("Invalid type, expected list") on every validate_config() call in this module (part of the "29 configuration errors" in the run log). Not introduced by this PR, but the PR codifies scalar-fails-forever while the project's own harness runs that form.

docs/car-charging.md
```diff
@@ -219,7 +219,10 @@ To make Predbat-led car charging more accurate, additionally you can configure t
...
+It *must* be entered as a list with one entry per car, exactly as in the example above - writing the value on the same
```
springfall2008: **Doc clarity:** "exactly as in the example above" points at an example that is entirely commented out — a user who uncomments only the key line leaves "#  - 75" commented, giving YAML null and the exact "not of type sensor" error this bullet warns about. The prose example (77.4) and the example value (- 75) also disagree.

apps/predbat/tests/test_validate_config.py
```diff
@@ -28,9 +28,35 @@
...
+
+    A one-entry list is the only form validate_config() accepts for this item (#2172): the
+    sensor branch wraps a bare string into a list but not a number, so a numeric scalar never
+    reaches the "fixed float values are allowed" check and is reported as "not of type 'sensor'"
+    on every run. An example that puts the value on the same line as the key therefore hands the
+    reader a config that permanently fails validation, which is how #2172 was raised.
+    """
+    repo_root = os.path.dirname(os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__)))))
```
springfall2008: **Cleanup:** the four nested os.path.dirname calls re-derive REPO_ROOT, which test_faq_recorder_config.py:27 already defines a readable idiom for (os.path.normpath(os.path.join(...))); and the unguarded os.listdir(folder_path) raises a bare FileNotFoundError if a folder is missing rather than a clear assert. Sharing the constant (and/or os.path.isdir asserts) would be more robust.

apps/predbat/tests/test_validate_config.py
```diff
@@ -393,6 +419,29 @@ def test_validate_config(my_predbat):
...
+    # line as the key is not accepted - with num_cars 1 it reaches the sensor
+    # branch, which only wraps strings into a list, so it is reported as
+    # "not of type 'sensor'". The car-charging docs used to describe that
+    # failing form ("must be entered with one decimal place, e.g. 50.0"), so
+    # the doc/template scan below is what keeps the documented form working.
+    # ==========================================================================
+    print("  [sensor float] car_charging_battery_size as a one-entry list passes")
+    _run(my_predbat, {"car_charging_battery_size": [77.4], "num_cars": 1}, expect_clean=["car_charging_battery_size"])
+
+    print("  [sensor float] car_charging_battery_size as a whole number in a list passes (templates use '- 75')")
```
springfall2008, line 410: **Minor/pre-existing:** the sensor branch never enforces the spec's zero: False for this item — car_charging_battery_size: [0] validates clean, and plan.py:4581/4767 then divide by it (fetch.py:1345 guards, plan.py does not). The new subtests could have pinned this contract.

chalfontchubby:
_(Claude wrote this comment, posted under my account.)_

The review was right on the substance, so rather than patch the guard I've rewritten the PR from scratch and force-pushed. Response to each thread:

**The doc-scan guard (threads on `test_validate_config.py:44/45/47/57`) — removed entirely.**

Both bug reports check out; I verified them by running the regex against `yaml.safe_load`:

| form | regex flags it | actual YAML value |
|---|---|---|
| key, then `77.4` on the next line | **no** | `77.4` (float) — *fails validation* |
| `car_charging_battery_size: [77.4]` | **yes** | `[77.4]` (list) — *valid* |
| `car_charging_battery_size:   # comment` | **yes** | `None` — *valid key-only line* |
| `car_charging_battery_size: 77.4` | yes | `77.4` (float) — fails |

So it missed the form that actually breaks and flagged two that don't — including the commented key-only style most templates ship, which would have broken CI the moment anyone added a value hint to that line.

I didn't replace it with a YAML-parsing version. It was guarding a doc example against drift, and even done properly it would still have missed `apps/predbat/config/apps.yaml` (the template users copy, correctly noted in the coverage-gap thread) and needed special-casing for every commented-out occurrence. The `checked > 30` proxy and the `os.path.dirname` nesting go away with it. The validator subtests are the real regression guard here; the doc is a doc.

**Doc inaccuracy (`car-charging.md:222`) — fixed, and this was the most important one.** "It *must* be entered as a list" was wrong: a string scalar *is* wrapped into a list, so `car_charging_battery_size: sensor.my_car_kwh` validates fine. Only a numeric scalar fails. The prose now says an inline *number* fails, and there's a new subtest pinning that a string scalar still passes, so the doc can't drift back to the absolute claim.

**Doc clarity (`car-charging.md:223`) — fixed.** "exactly as in the example above" pointed at a fully commented-out block, where uncommenting just the key line leaves `#  - 75` commented and yields YAML `null` — the same error. The bullet now carries its own uncommented example, and the value matches (`- 75`, not `77.4`).

**Coverage gap (`:433`) — fixed.** Added `num_cars: 2` subtests. Confirms the different error path: `is not a list, but requires 2 entries based on num_cars`, not `not of type 'sensor'`.

**Not addressed, deliberately — both pre-existing and both worth their own fix:**

- **Test-isolation hazard (`:34`)** — `_run()` not snapshotting `manual_api`. Real, but it's a property of the shared helper affecting every subtest in the module, not something this change introduces. Fixing it inside a docs PR would mean touching the harness all other subtests run through.
- **Fixture ships the banned form (`:36`)** — `test_infra.py` sets `num_cars: 2` with an inline `100.0`, so this key is in the `arg_errors` baseline on every call. Worth cleaning up; it would shift that baseline for the whole module.
- **`zero: False` not enforced (`:436`)** — `[0]` validates clean and `plan.py` then divides by it (`fetch.py` guards, `plan.py` doesn't). That's a genuine latent bug and the one I'd most want followed up, but it's a validator/planner fix, not a documentation change.

Happy to file the last one as an issue if you'd like it tracked rather than left in this thread.

Both remaining `run_pre_commit` runs are green (12 hooks + `--quick`, 78s).

chalfontchubby:
_(Claude wrote this comment, posted under my account.)_

Closing in favour of **#5024**, which does the same job as a clean, non-draft PR rather than a rewritten auto-draft.

The review here was right on the substance and #5024 acts on it — the doc prose no longer overstates the rule (a string scalar *is* wrapped, so only an inline **number** fails), the bullet carries its own uncommented example instead of pointing at a commented-out block, `num_cars: 2` coverage is added, and the `templates/`+`docs/` regex scan is gone rather than patched.

On that scan, for the record: verified against `yaml.safe_load`, it missed the key-then-next-line form that actually fails validation, while flagging the valid `[77.4]` flow list and any key line with a trailing comment — wrong in both directions. A YAML-parsing rewrite would still have missed `apps/predbat/config/apps.yaml`, the template users actually copy. The validator subtests are the real regression guard.

Three pre-existing issues the review surfaced are recorded as follow-ups in #5024 rather than fixed there — the most substantive being that `car_charging_battery_size: [0]` validates clean despite `zero: False`, and `plan.py` then divides by it.

Issue #5022, opened 2026-09-09
Fox Cloud: support Feed-in First (freeze export) via the work-mode setting


chalfontchubby:
_(Claude wrote this issue, posted under my account.)_

Split out of #5015, which turned off `support_feedin_first` for FoxCloud because the Cloud path never selects the mode. This issue tracks actually adding it.

## Background

The Fox hardware does have a Feed-in First work mode, and Predbat already uses it on the **modbus** path: `templates/fox.yaml` ships a `discharge_freeze_service` stanza that selects `"Feed-in First"` on `select.foxess_work_mode`, published by the FoxESS HA integration.

The **Cloud** path has no equivalent. `templates/fox_cloud.yaml` contains no `*_service` stanzas at all, so during a freeze export slot `adjust_export_immediate(..., freeze=True)` calls `call_service_template("discharge_freeze_service", ...)`, finds nothing configured, falls back to `discharge_stop_service` (also unset), and the freeze becomes a no-op. The inverter stays in Self Use.

Same hardware, different control surface.

## What makes this look implementable

The Cloud component already publishes a **writable** work-mode select:

- `automatic_config` wires `inverter_mode` to `select.{prefix}_fox_{device}_setting_workmode` (`fox.py:2221`)
- that entity carries `enumList: OPTIONS_WORK_MODE`, which **includes `"Feedin"`** (`fox.py:44`, `fox.py:57`)
- writes route `select_event` → `write_setting_from_event` → `set_device_setting`, a real Fox API write (`fox.py:1994-2005`)

So a `discharge_freeze_service` stanza in `fox_cloud.yaml` pointing at that select is plausibly most of the work — structurally the same as the modbus template, just a different entity id.

## The catch

`adjust_inverter_mode` writes to that same `inverter_mode` entity and hardcodes SelfUse for any `has_fox_inverter_mode` inverter:

```python
if self.inv_has_fox_inverter_mode:
    # For fox we only use selfuse as the rest is done by schedule
    new_inverter_mode = "SelfUse"
```

(`inverter.py:2443-2444`, added in #2659, Sept 2025.)

A service call setting `Feedin` would be reverted to `SelfUse` on the next cycle — the two would fight. So the template stanza alone is **not** sufficient; the hardcode needs to learn that freeze export is a legitimate third state. That comment's premise ("the rest is done by schedule") holds for Cloud but not for modbus, which is why one brand-named flag spanning both control paths leaves this gap.

[...]

PR #5023, opened 2026-09-09
fix(fox): don't model Feed-in First on FoxCloud, which never selects it (#5015)

### LINKED ISSUE #5015: Fox Cloud integration - Freeze Export planned but has no effect
**Describe the bug**
Using the Fox Cloud integration Predbat plans freeze export slots but the inverter stays in 'Self Use' and PV generation first covers home load, then charges the battery and finally exports once the battery is full - so effectively the behaviour I see in 'Demand' slots.

Until a few weeks ago freeze export slots were never planned, I only had 'Demand' slots in the day.  I can see some configuration on the FoxCloud inverter was changed recently to mark these supported - and to mark feed-in first supported which is when I think I started to see these slots planned. It does not appear from observed behaviour that freeze export is actually turned into any actual change in behaviour on the inverter - it stays in SelfUse mode.

**Expected behaviour**
The battery is held at the target SoC (so no charging) and excess PV after house load flows to the grid. The "Feed-in Priority" work mode would seem to be a good fit with the caveat that at least on my inverter ~200W/h still flows into the battery even though it claims it is not charging. 

**Predbat version**
v9.0.1

**Environment details**

- KH7 inverter, EP12 battery
- Fox Cloud integration with fox_automatic: True
- Standard HAOS install

**Screenshots**
The relevant plan section:

Image

From the Fox Cloud at 9/9/2026 10:20am - inverter is in 'SelfUse' during a Predbat-scheduled FreezeExport slot:

Image

**Log file**

[predbat.log](https://github.com/user-attachments/files/31999742/predbat.log)

**Predbat debug yaml file**

[predbat_debug.yaml.txt](https://github.com/user-attachments/files/31999747/predbat_debug.yaml.txt)

chalfontchubby:
_(Claude wrote this PR, posted under my account.)_

Fixes #5015.

## The bug

FoxCloud declares `support_feedin_first: True`, justified in `config.py` by a comment saying it is the *same hardware* as FoxESS. That flag tells the planner the inverter's freeze export is a real "Feed-in First" mode — load, then export, then battery — so PV past the export limit is recaptured into the battery rather than clipped (`prediction.py:1146`).

The Cloud path never selects that mode:

- it drives the inverter through the Fox API scheduler, and `apply_battery_schedule` only ever emits `ForceCharge` / `ForceDischarge` groups — `Feedin` appears in `fox.py` only in enum lists, never in a written schedule
- `adjust_inverter_mode` hardcodes `new_inverter_mode = "SelfUse"` for any `has_fox_inverter_mode` inverter (`inverter.py:2443`)
- `fox_cloud.yaml` has no `discharge_freeze_service` (no `*_service` stanzas at all), so the freeze hook finds nothing configured and no-ops

So the inverter sits in Self Use for the whole slot: PV covers house load, then charges the battery, and exports only once the battery is full — exactly the Demand behaviour the reporter observes.

Same silicon, different control surface. FoxESS (modbus) reaches the mode through `select.foxess_work_mode`, wired up by the `discharge_freeze_service` stanza in `templates/fox.yaml`.

## Why now

Two commits that are individually fine but mutually inconsistent:

- the SelfUse hardcode is from **#2659, Sept 2025** — a year old
- `support_feedin_first: True` on FoxCloud came from **3ed9151e, 24 Aug 2026**

The old hardcode was harmless until the recent planner change started relying on the mode it blocks. That matches the reporter's "until a few weeks ago freeze export slots were never planned".

Worth noting #4864 (GivTCP) is **not** implicated despite also shipping in v9.0.1 — it touched no Fox files, and the hardcode predates it by a year.

## The change

One flag, plus the two tests that asserted the old value.

`support_discharge_freeze` stays `True` — freeze export is still attempted, it is only the Feed-in First *recapture modelling* that was wrong.

The shared opt-in test's docstring claimed all five other types genuinely select the mode, so I checked before shrinking the set: they do (`solax.py:833` calls `set_default_work_mode(mode="feedin")`, sunsynk/deye have dedicated Selling First freeze handling, solis.py has the feed-in priority bit). FoxCloud was the only false member.

## Effect for users

Freeze export slots on FoxCloud go back to being Demand slots, as they were before 24 Aug. That is the correct plan for what the inverter actually does, but it will look like a regression if unexplained — the plan stops promising something the hardware was never being told to do.

Adding the mode properly to the Cloud path is tracked in #5022, which records the implementation route I verified (the Cloud component *does* publish a writable work-mode select whose enum includes `Feedin`; the blocker is that `adjust_inverter_mode` would revert it).

## Testing

- `./run_all --test inverter` passes
- full `./run_all --quick` suite green (1015 passed, exit 0)
- all 12 pre-commit hooks pass

## Known follow-up, deliberately not in this PR

`support_feedin_first` is a *type-level* constant asserting something that is really *user config* — `call_service_template` resolves `discharge_freeze_service` from `self.base.args` with no inverter-type gate anywhere. A **modbus** user who trimmed that stanza from their `apps.yaml` has this same silent mismatch today.

Gating the flag on the service actually being configured would fix that, but it changes behaviour for all five remaining opt-in types on a path nobody has reported a problem with, and it needs a design call on when the gate is evaluated (apps.yaml changes at runtime; the kernel context carries the flag). Kept out of a targeted bug fix — happy to file it separately if you want it tracked.

### DIFF
--- a/apps/predbat/config.py
+++ b/apps/predbat/config.py
@@ -2176,8 +2176,12 @@ INVERTER_DEF = {
         "write_and_poll_sleep": 2,
         "has_time_window": False,
         "support_charge_freeze": True,
-        # See FoxESS's entry above - same hardware, correctly modelled rather than disabled (#4207).
-        "support_feedin_first": True,
+        # Unlike FoxESS above, the Cloud path never actually selects Feed-in First (#5015). It drives
+        # the inverter through the API scheduler, which only ever emits ForceCharge/ForceDischarge
+        # groups, and adjust_inverter_mode pins the work mode to SelfUse - so a freeze export slot
+        # leaves the inverter in Self Use and PV charges the battery instead of being exported.
+        # Modelling recapture here made the planner pick freeze export slots that do nothing.
+        "support_feedin_first": False,
         "support_discharge_freeze": True,
         "has_idle_time": False,
         "can_span_midnight": False,
--- a/apps/predbat/tests/test_inverter.py
+++ b/apps/predbat/tests/test_inverter.py
@@ -24,11 +24,13 @@ from const import MINUTE_WATT
 
 def test_foxess_support_discharge_freeze_matches_foxcloud():
     """
-    FoxESS (modbus) and FoxCloud are the same hardware via two different connection methods - "feed-in
-    first"/freeze export does not hold SoC flat on either, PV above the export limit still charges the
-    battery instead of being clipped (#4207). Both are now True: that spillover-charges-the-battery
-    behaviour is correctly modelled (prediction.py's freeze branch, gated on support_feedin_first)
-    rather than being a reason to disable freeze outright for this hardware - see
+    FoxESS (modbus) and FoxCloud are the same hardware via two different connection methods, so both
+    support freeze export rather than having it disabled outright for this hardware (#4207) - this
+    pins support_discharge_freeze only.
+
+    support_feedin_first deliberately does NOT match across the two: whether the freeze is a real
+    Feed-in First mode depends on the control path rather than the silicon, and only the modbus side
+    selects that mode (#5015). See test_support_feedin_first_is_opt_in, and
     test_freeze_export_recapture_beyond_limit in test_optimise_solar.py for the modelling itself.
     """
     failed = False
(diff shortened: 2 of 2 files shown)


chalfontchubby:
In theory not needed now since #5022 is merged.

springfall2008:
Yes, I fixed fox :)

chalfontchubby:
A better solution. Figured I'd put the quick fix in in case.

Issue #4415, opened 2026-08-01
Solis GS_fb00: timed_discharge_current / timed_charge_current reset to 0A on every cycle and are never re-asserted, so forced export does nothing


iansyder8:
# Solis GS_fb00: `timed_discharge_current` / `timed_charge_current` reset to 0A on every cycle and are never re-asserted, so forced export does nothing

## Summary

On a Ginlong Solis (`inverter_type: GS_fb00`) via `solax_modbus`, the inverter's **timed charge/discharge current** registers are reset to `0A` outside of Predbat's control. Predbat only writes those registers when its cached charge/discharge **rate** changes — which in normal operation is rare — so once zeroed they stay at zero indefinitely.

The practical effect: Predbat sets up a forced export window and correctly writes every slot *time* register and `scheduled_discharge_enable`, but because the timed discharge current is `0A` the inverter **neither exports nor supplies the house from the battery**. The house falls back to importing at full rate with a nearly full battery sitting idle.

Measured cost in my case: ~4.1 kW of import at 30.4p/kWh through an evening export window, battery at 85%.

Evidence below shows the reset is **deterministic — it happens on every single Predbat cycle**, roughly 120 times overnight without exception.

## Environment

| | |
|---|---|
| Predbat version | v8.47.2 (auto-updated from v8.47.1 mid-log) |
| Inverter type | `GS_fb00` — Ginlong Solis (FB00) |
| Integration | `homeassistant-solax-modbus` **(version: 2026.08.1)** |
| Solis firmware | **TODO** |
| Transport | RS485 → Waveshare RS485-to-ETH gateway, Modbus TCP↔RTU, multi-host enabled, 9600 baud |
| Battery | `soc_max: 32.0` kWh, `battery_voltage: 52`, `battery_rate_max: 8000` W |

---

## Evidence 1 — Predbat writes the current register once in 16 hours

Every successful register write over a continuous 16-hour log (2026-08-01, 04:25 → 20:46):

| Register | Writes |
|---|---:|
| `idle_start_time` | 40 |
| `discharge_end_time` | 35 |
| `discharge_end_hour` | 33 |
| `idle_end_time` | 23 |
| `discharge_start_time` | 20 |
| `discharge_end_minute` | 20 |
| `discharge_start_minute` | 19 |
| `discharge_start_hour` | 12 |
| `reserve` | 3 |
| `inverter_mode` | 3 |
| `discharge_rate` | 2 |
| `charge_end_time` | 2 |
[...]

PR #5027, opened 2026-09-09
Add optional real-time dispatch control (inverter Remote Dispatch APIs)

philicibine:
## What

Optional **real-time dispatch control** for inverters that expose a Remote Dispatch API — Predbat executes export windows by re-applying a dispatch command every cycle, instead of arming a persistent scheduled slot.

## Why

Some inverters cannot honour scheduled export depth. The concrete case (wills106/homeassistant-solax-modbus#2179, Solis S5 hybrid firmware 4F0052): the timed-discharge target SOC register (43750) is write-locked and enforced as a hard floor on every scheduled/forced export path — the battery stops dead at 50% no matter what Predbat plans. The same firmware exposes a Remote Dispatch interface (44100 block, protocol Ver3.4) which bypasses that floor, but it is RAM-only with an inverter-side failsafe: no fresh command within the failsafe interval and the inverter reverts to demand mode.

That combination — real-time-only, self-reverting — maps naturally onto Predbat's cycle loop:

- **Inside an export window** Predbat re-applies the dispatch command every cycle. This both keeps the failsafe fed and corrects any drift, exactly like the per-cycle re-assert pattern already used for the timed current register (#4415).
- **If Predbat or HA dies mid-window**, the inverter's failsafe stops the export on its own — arguably *safer* than a persistent armed slot, which exports until someone notices.
- **Outside a window** dispatch is disabled, so the inverter sits in demand mode.

## Design

- **Arg-driven and generic.** New optional per-inverter apps.yaml args: `dispatch_power`, `dispatch_apply_button`, `dispatch_disable_button`, plus optional `dispatch_control_mode`, `dispatch_soc_min`, `dispatch_soc_max`, `dispatch_failsafe_interval`, `dispatch_master`, `dispatch_status`. No INVERTER_DEF changes: any integration that publishes matching entities can drive this. Absence of the args leaves the stock scheduled-slot path bit-for-bit untouched (regression test included).
- **Single hook point.** `adjust_force_export()` gains the dispatch branch — it's the one signal that is only true inside an export window. `adjust_discharge_rate()` stashes the rate (the rate setters also run in calibration/reset paths, where arming dispatch would be wrong).
- **Pre-arm gating.** `execute()` pre-arms upcoming windows with `adjust_force_export(True, ...)` up to `set_window_minutes` ahead. Dispatch does not start until the window time is actually reached (test covered).
- **No double-drive.** When dispatch is configured, the scheduled-slot writes are skipped entirely — both paths active at once would double the export power.
- **Sign convention in one place.** Predbat computes export power as a positive discharge rate in W; the dispatch protocol wants signed power (negative = discharge). The flip lives in one function, `dispatch_power_for_rate()`.
- **SOC floor.** Dispatch soc_min defaults to the inverter's reserve floor (`max(reserve_percent, reserve_min)`), matching the depth Predbat planned for.
- **Disable is idempotent** across cycles (no button spam when every cycle calls `adjust_force_export(False)` outside windows).

## New module

`dispatch.py` — `InverterDispatch` helper class, fully docstringed per repo standards, with the entity writes routed through the existing `write_and_poll_*` helpers (control-ledger aware).

## Tests

`tests/test_dispatch_control.py`, registered in `unit_test.py` as `dispatch_control`, five tests, all passing:
1. Dispatch-enabled export applies the block (mode, signed power, soc_min, master, apply press) and writes **zero** slot entities
2. Pre-armed future window does **not** start dispatch
3. Disable is idempotent across cycles
4. Unconfigured inverter keeps the stock scheduled-slot path (regression guard)
5. Sign convention: positive rate → negative dispatch power

## Live verification

Running in production on the affected inverter (Solis S5-EH1P6K-L, fw 4F0052) driving the Solis Remote Dispatch block via the companion plugin PR wills106/homeassistant-solax-modbus#2330. The dispatch path end-to-end (stage entities → master → apply → inverter discharges at commanded power below the locked floor → disable reverts cleanly) was live-verified against the real inverter and HA instance during development; the scheduled 01:20 export window is the first full production cycle on this code.

Related: wills106/homeassistant-solax-modbus#2179 (firmware lock analysis), Pho3niX90/solis_modbus (protocol reference).


philicibine:
Live root-cause update from the first production night: the stale-on deadlock described below is confirmed exactly, and fixed in 08dd7af6.

**What happened overnight (S5-EH1P6K-L, fw 4F0052):** the first dispatch-driven export ran perfectly (battery discharged straight through the locked 50% timed-slot floor to the dispatch soc_min — the feature works). But at window end the disable button cleared the inverter-side block (44100 → 0) without updating the HA switch, which stayed optimistically `on`. Every subsequent export window that night then hit `write_and_poll_switch: No write needed for dispatch_master as True == True`, never re-armed 44100, and the master-gated inverter silently discarded each realtime Apply — predbat logged `applied -5000W` every cycle while grid export stayed at 0W.

**Live verification of the fix (raw Modbus, before deploying):**
- After the inverter cleared the master, a single FC06 `44100=1` re-arms the block (committed, status 34504 → 1) — so the master is NOT failsafe-locked, it just needs re-writing.
- `switch/turn_on` on the HA switch commits the same re-arm **from the exact stale-on state** (raw 0 / HA `on`) — proving the forced service call is the right fix.
- Scattered single-register writes (the staged failsafe/soc_min numbers) only commit while the master is enabled — so the force-`turn_on` must happen **before** staging, which the fix enforces.
- `switch/turn_off` after the disable press re-syncs the HA switch (raw 0 / HA `off`), closing the loop that poisoned the next window.

**The fix (08dd7af6):** `apply_dispatch()` now issues an unconditional `switch/turn_on` for the master before staging anything (idempotent on the inverter — re-arms 44100; never trusts the HA switch state), and `disable_dispatch()` force-offs the master alongside the disable press so the HA state can never go stale-on again. Five dispatch tests still green, including two new assertions: master forced via service call (not `write_and_poll_switch`) and forced **before** the staged power write. Deployed to the live box and boot-verified; tonight's window is the first production run on the fixed code.

PR #5028, opened 2026-09-09
fix(tests): pin the prediction_batch fixture clock instead of inheriting the wall clock

### LINKED ISSUE #5026: prediction_bat test fails when run standalone. Passes with other tests.
While running tests on 9.0.1, running prediction_bat in isolation seems to fail while running as part of the full test suite passes - implying state leakage.

springfall2008:
This is an automated draft PR generated from issue #5026 — a maintainer should review it before merging.

Fixes #5026

## Summary

`run_prediction_batch_tests` took `minutes_now` from whatever `create_predbat()` left on the fixture — the host's wall clock standalone, `reset_inverter`'s noon once another module has run. The routing test's precondition (ten trial SoCs giving ten distinct costs) only holds for part of the day: at ~20:15 BST the three highest SoCs collapse onto one cost, so the distinctness guard trips before any routing is compared. That is why the module failed standalone in the evening and passed inside the full suite, on leftover state rather than on merit.

After review the fix pins the fixture clock at its source rather than per module: `FIXTURE_MINUTES_NOW = 12 * 60` (noon) now lives in `tests/test_infra.py`, and `create_predbat()` sets `minutes_now`/`now_utc` from it immediately after `update_time()`, the same way `apply_random_scenario` pins a scenario's own clock. `reset_inverter` re-asserts the same constant instead of restating the literal, so the pins cannot silently diverge and no module inherits the wall clock into those two fields. Only those two fields are pinned — `midnight_utc`/`now_utc_real` still come from the host clock.

`run_prediction_batch_tests` keeps a pin of its own as belt and braces against the fixture ever drifting: it records the caller's clock on entry, snapshots the scenario state, pins, and restores in a `finally`. `minutes_now` was added to `SCENARIO_STATE_ATTRS` so the pin is handed back rather than leaking to the next module, and the runner then fails if the restored clock is not the caller's — which pins the snapshot-before-pin ordering as well as the round trip. The kernel-unavailable early return no longer resets `prediction_kernel_enable`; that reset now runs only when the kernel tests actually ran, so the skip path writes no shared fixture state. Nothing in the batch code itself changed; the failing check was on the test's own oracle, so no result was ever actually misrouted.

Coverage for the clock hand-back is that runner-level check rather than a separate named test. An earlier revision of this PR added a standalone `test_minutes_now_is_pinned_and_restored`; review found it deep-copied the whole ~70-attribute scenario state to assert one int and carried a dead repair line, so it was replaced by the in-runner check, which exercises the same `snapshot_scenario_state` → pin → `restore_scenario_state` path with the caller's real entry value on every invocation, including the kernel-skip path.

## Testing

- Reproduced first on main (d48cb223) at 20:28 BST: `tools/triage_test.sh prediction_batch` failed with costs `[..., 2590.8878, 2590.8878, 2590.8878]`.
- After the change, the same standalone command passes at the same time of day (exit 0, all subtests including the routing test).
- A temporary registered probe test confirmed a fresh `create_predbat()` carries `minutes_now = 720` and `now_utc = midnight + 720` (probe removed before commit).
- `coverage/run_pre_commit` passes: every hook Passed, and the quick suite it runs is green — `kernel_parity` included, since it shares `SCENARIO_STATE_ATTRS`.

## Notes

Same family as the trap the debug journal already records under "Build test dates from Predbat's clock, not the machine's" (and PR #5013 / #4998): a fixture inheriting the host's clock instead of a pinned one. This variant is about `minutes_now` rather than the calendar date.

### DIFF
--- a/apps/predbat/tests/test_kernel_parity.py
+++ b/apps/predbat/tests/test_kernel_parity.py
@@ -57,9 +57,12 @@ RESULT_NAMES = [
     "final_carbon_g",
 ]
 
-# Attributes mutated by the parity scenarios, or by a module pinning the clock before them, that
-# reset_inverter/reset_rates do not restore; snapshotted before the tests and restored afterwards
-# so later tests see a clean predbat
+# Attributes mutated by the parity scenarios that reset_inverter/reset_rates do not restore;
+# snapshotted before the tests and restored afterwards so later tests see a clean predbat.
+# One exception: the parity scenarios only ever read minutes_now - its entry exists for
+# run_prediction_batch_tests, which pins the fixture clock while it runs and hands the caller's
+# back through the snapshot mechanism this list drives (#5026). Dropping the entry would leave
+# that module passing and only its own hand-back check failing, so it is documented here.
 SCENARIO_STATE_ATTRS = [
     "minutes_now",
     "soc_max",
--- a/apps/predbat/tests/test_kernel_parity.py
+++ b/apps/predbat/tests/test_kernel_parity.py
@@ -57,9 +57,11 @@ RESULT_NAMES = [
     "final_carbon_g",
 ]
 
-# Attributes mutated by the parity scenarios that reset_inverter/reset_rates do not restore;
-# snapshotted before the tests and restored afterwards so later tests see a clean predbat
+# Attributes mutated by the parity scenarios, or by a module pinning the clock before them, that
+# reset_inverter/reset_rates do not restore; snapshotted before the tests and restored afterwards
+# so later tests see a clean predbat
 SCENARIO_STATE_ATTRS = [
+    "minutes_now",
     "soc_max",
     "soc_kw",
     "reserve",
--- a/apps/predbat/tests/test_prediction_batch.py
+++ b/apps/predbat/tests/test_prediction_batch.py
@@ -34,6 +34,14 @@ from prediction_kernel import create_kernel_context
 from tests.test_kernel_parity import apply_random_scenario, kernel_available, make_step_data, make_windows, restore_scenario_state, snapshot_scenario_state
 
 
+# create_predbat() takes minutes_now from the wall clock, and only part of the day leaves the seed-5
+# routing scenario's ten trial costs distinct - an evening run collapses the three highest onto one
+# value and trips that test's precondition. Run after another module the value is whatever residue
+# that module left (reset_inverter's noon), so standalone runs failed while the full suite passed;
+# the module pins the same noon for itself rather than inheriting either (#5026).
+BATCH_TEST_MINUTES_NOW = 12 * 60
+
+
 def make_export_windows(minutes_now):
     """Build a small deterministic export window layout for the trial-input tests"""
     return [
@@ -605,20 +613,57 @@ def test_batch_thread_count_resolution():
     return failed
 
 
+def test_minutes_now_is_pinned_and_restored(my_predbat):
+    """The module must run on its own pinned clock and hand the caller's back, returns True on failure.
+
+    These tests used to inherit whatever minutes_now the fixture arrived with, which is the wall clock
+    standalone and reset_inverter's noon after another module has run. The routing test's precondition
+    - ten trial SoCs with ten distinct costs - only holds for part of the day, so the module failed
+    when run on its own in the evening and passed in the full suite purely on leftover state (#5026).
+
+    Pinning is only half of it: the pin must not follow the module out, or it would hand the next test
+    a clock the caller never set. That is what the scenario snapshot is for, so this checks the round
+    trip carries minutes_now rather than trusting the attribute list by eye.
+    """
+    print("**** Running pinned clock tests ****")
+    failed = False
+
+    if my_predbat.minutes_now != BATCH_TEST_MINUTES_NOW:
+        print("ERROR: the batch tests are running at minutes_now {}, expected the pinned {}".format(my_predbat.minutes_now, BATCH_TEST_MINUTES_NOW))
+        failed = True
+
+    # A value no fixture would set by chance, so a restore that misses it cannot pass by luck
+    state = snapshot_scenario_state(my_predbat)
+    my_predbat.minutes_now = BATCH_TEST_MINUTES_NOW + 137
+    restore_scenario_state(my_predbat, state)
+    if my_predbat.minutes_now != BATCH_TEST_MINUTES_NOW:
+        print("ERROR: the scenario snapshot does not restore minutes_now, left {} not {}".format(my_predbat.minutes_now, BATCH_TEST_MINUTES_NOW))
+        my_predbat.minutes_now = BATCH_TEST_MINUTES_NOW
+        failed = True
+
+    if not failed:
+        print("Pinned clock tests passed")
+    return failed
+
+
 def run_prediction_batch_tests(my_predbat):
     """Run every batched prediction test, returns True on failure"""
-    failed = test_export_trial_does_not_mutate_caller_window(my_predbat)
-    failed |= test_batch_thread_count_resolution()
-    failed |= test_available_cpu_count_respects_a_cgroup_quota()
-    failed |= test_batch_state_exists_without_a_base(my_predbat)
-
-    available, required_failure = kernel_available()
-    if not available:
-        print("WARNING: kernel not available - batch tests that need it are SKIPPED")
-        return failed or required_failure
-
+    # Snapshotted before anything is pinned so the caller's clock and scenario both go back untouched
     state = snapshot_scenario_state(my_predbat)
     try:
+        my_predbat.minutes_now = BATCH_TEST_MINUTES_NOW
+
+        failed = test_minutes_now_is_pinned_and_restored(my_predbat)
+        failed |= test_export_trial_does_not_mutate_caller_window(my_predbat)
+        failed |= test_batch_thread_count_resolution()
+        failed |= test_available_cpu_count_respects_a_cgroup_quota()
+        failed |= test_batch_state_exists_without_a_base(my_predbat)
+
+        available, required_failure = kernel_available()
+        if not available:
+            print("WARNING: kernel not available - batch tests that need it are SKIPPED")
+            return failed or required_failure
+
         failed |= test_queued_matches_direct(my_predbat)
         failed |= test_queued_range_window_in_the_past(my_predbat)
         failed |= test_batch_is_lazy(my_predbat)
--- a/apps/predbat/tests/test_prediction_batch.py
+++ b/apps/predbat/tests/test_prediction_batch.py
@@ -31,15 +31,8 @@ from const import PREDICT_STEP, PV_SCENARIO_NOMINAL, PV_SCENARIO_PV10
 from plan import resolve_batch_threads
 from prediction import Prediction
 from prediction_kernel import create_kernel_context
+from tests.test_infra import FIXTURE_MINUTES_NOW
 from tests.test_kernel_parity import apply_random_scenario, kernel_available, make_step_data, make_windows, restore_scenario_state, snapshot_scenario_state
-
-
-# create_predbat() takes minutes_now from the wall clock, and only part of the day leaves the seed-5
-# routing scenario's ten trial costs distinct - an evening run collapses the three highest onto one
-# value and trips that test's precondition. Run after another module the value is whatever residue
-# that module left (reset_inverter's noon), so standalone runs failed while the full suite passed;
-# the module pins the same noon for itself rather than inheriting either (#5026).
-BATCH_TEST_MINUTES_NOW = 12 * 60
 
 
 def make_export_windows(minutes_now):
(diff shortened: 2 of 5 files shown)


apps/predbat/tests/test_prediction_batch.py
```diff
@@ -605,20 +613,57 @@ def test_batch_thread_count_resolution():
...
+    when run on its own in the evening and passed in the full suite purely on leftover state (#5026).
 
+    Pinning is only half of it: the pin must not follow the module out, or it would hand the next test
+    a clock the caller never set. That is what the scenario snapshot is for, so this checks the round
+    trip carries minutes_now rather than trusting the attribute list by eye.
+    """
+    print("**** Running pinned clock tests ****")
+    failed = False
+
+    if my_predbat.minutes_now != BATCH_TEST_MINUTES_NOW:
```
springfall2008: _Automated comment from the triage bot._

This guard cannot catch the regressions it exists for, because in the full-suite configuration it can never fire:

- The ambient `minutes_now` is already 720 when this module runs in the suite (the residue `reset_inverter` leaves at tests/test_infra.py:659, inherited via model/model_kernel/kernel_parity), so deleting or reordering the runner's pin at line 654 still passes this check. The only configuration where it bites is a standalone evening run — i.e. it detects the failure #5026 fixed, not the loss of the fix.
- The snapshot-before-pin ordering is also untested: if `state = snapshot_scenario_state(...)` (line 652) were moved after the pin (line 654), the `finally` would hand the caller the pinned noon instead of their clock, and this test would still pass, because its own inner snapshot (line 636) is taken post-pin either way.

Suggested hardening: have the runner record `entry = my_predbat.minutes_now` before the snapshot and, in the `finally` after the restore, flag `my_predbat.minutes_now != entry` when they differed on entry. That pins the ordering, not just the round trip.

apps/predbat/tests/test_prediction_batch.py
```diff
@@ -605,20 +613,57 @@ def test_batch_thread_count_resolution():
...
+        failed |= test_batch_thread_count_resolution()
+        failed |= test_available_cpu_count_respects_a_cgroup_quota()
+        failed |= test_batch_state_exists_without_a_base(my_predbat)
+
+        available, required_failure = kernel_available()
+        if not available:
+            print("WARNING: kernel not available - batch tests that need it are SKIPPED")
+            return failed or required_failure
+
         failed |= test_queued_matches_direct(my_predbat)
```
springfall2008: _Automated comment from the triage bot._

On this early-return path the `finally` now also runs `my_predbat.prediction_kernel_enable = False` (line 676), where main returned without touching `my_predbat` at all. That flag is not in `SCENARIO_STATE_ATTRS`, so the snapshot at line 652 cannot protect it: a caller that entered with `prediction_kernel_enable = True` in a kernel-less environment gets the flag silently forced to False for every module after it. No current suite module hits this (the kernel modules return before enabling the flag when the kernel is absent), so it is latent rather than live — but it is a new write to shared fixture state on a path that previously mutated nothing.

Either move the flag reset inside the `available` branch, or add `prediction_kernel_enable` to `SCENARIO_STATE_ATTRS` so the snapshot restores it.

Related, same path: the snapshot at line 652 now also runs unconditionally on the skip path, where main took none — deep-copying the ~70-attribute state (including the ~2160-entry rate dicts) to protect four cheap non-kernel tests. Milliseconds, so efficiency-only.

apps/predbat/tests/test_prediction_batch.py
```diff
@@ -34,6 +34,14 @@
 from tests.test_kernel_parity import apply_random_scenario, kernel_available, make_step_data, make_windows, restore_scenario_state, snapshot_scenario_state
 
 
+# create_predbat() takes minutes_now from the wall clock, and only part of the day leaves the seed-5
+# routing scenario's ten trial costs distinct - an evening run collapses the three highest onto one
+# value and trips that test's precondition. Run after another module the value is whatever residue
+# that module left (reset_inverter's noon), so standalone runs failed while the full suite passed;
+# the module pins the same noon for itself rather than inheriting either (#5026).
+BATCH_TEST_MINUTES_NOW = 12 * 60
```
springfall2008: _Automated comment from the triage bot._

`12 * 60` here is a third independent copy of the canonical noon fixture: `reset_inverter` pins the same literal at tests/test_infra.py:659, and other modules write it inline as well (test_inverter.py:3008, test_execute.py:462, …). If `reset_inverter`'s pin ever moves (e.g. a future fixture needs a different hour, as happened for the clock-skew band in PR #4991), this constant silently diverges and the round-trip test still passes, since it compares against this module's own copy rather than the fixture's.

Cheapest guard: hoist one shared constant into tests/test_infra.py next to `reset_inverter` and import it in both places — or at minimum assert `BATCH_TEST_MINUTES_NOW ==` the fixture's pin so the two cannot drift without a signal.

apps/predbat/tests/test_kernel_parity.py
```diff
@@ -57,9 +57,11 @@
     "final_carbon_g",
 ]
 
-# Attributes mutated by the parity scenarios that reset_inverter/reset_rates do not restore;
-# snapshotted before the tests and restored afterwards so later tests see a clean predbat
+# Attributes mutated by the parity scenarios, or by a module pinning the clock before them, that
+# reset_inverter/reset_rates do not restore; snapshotted before the tests and restored afterwards
+# so later tests see a clean predbat
 SCENARIO_STATE_ATTRS = [
+    "minutes_now",
```
springfall2008, line 67: _Automated comment from the triage bot._

`minutes_now` is the first entry in `SCENARIO_STATE_ATTRS` that no parity scenario writes — `run_kernel_parity_tests` (lines 1469-1503) only ever reads the clock and never assigns `my_predbat.minutes_now`, so the entry is inert in this module's own runner and exists purely to serve the batch module's pin in test_prediction_batch.py. Two consequences worth weighing:

- Ownership: the list is documented as "attributes mutated by the parity scenarios"; a maintainer auditing it has no reason to keep this entry, and dropping it silently removes the batch module's clock hand-back (this module would keep passing; only the batch module would notice).
- Behaviour change: `run_kernel_parity_tests` now restores the pre-parity clock at exit instead of leaving `reset_inverter`'s residue. Inert in the standard suite (its entry value is already 720), but it is an undocumented change to a shared mechanism — the same residue-dependency class as #5026 itself.

If the clock is pinned at fixture level instead (suggested on the constant in test_prediction_batch.py), this entry becomes unnecessary and the list stays scoped to what the parity scenarios actually mutate.

apps/predbat/tests/test_prediction_batch.py
```diff
@@ -605,20 +613,57 @@ def test_batch_thread_count_resolution():
...
+    """
+    print("**** Running pinned clock tests ****")
+    failed = False
+
+    if my_predbat.minutes_now != BATCH_TEST_MINUTES_NOW:
+        print("ERROR: the batch tests are running at minutes_now {}, expected the pinned {}".format(my_predbat.minutes_now, BATCH_TEST_MINUTES_NOW))
+        failed = True
+
+    # A value no fixture would set by chance, so a restore that misses it cannot pass by luck
+    state = snapshot_scenario_state(my_predbat)
```
springfall2008: _Automated comment from the triage bot._

Simplification: this round-trip check deep-copies the full ~70-attribute `SCENARIO_STATE_ATTRS` snapshot (rate_import/rate_export of ~2160 entries each, carbon_intensity, car slot lists, …) to verify a single int survives `restore_scenario_state` — which is just a setattr per key. Two cheaper forms test the same property:

- `assert "minutes_now" in SCENARIO_STATE_ATTRS` — pins the thing that actually breaks (a dropped list entry, which leaves the +137 sentinel surviving), no copy at all.
- A one-attribute snapshot `state = {"minutes_now": my_predbat.minutes_now}`, which `restore_scenario_state` accepts as-is.

Also, the repair line at 641 (`my_predbat.minutes_now = BATCH_TEST_MINUTES_NOW` inside the failure branch) is dead: the runner's `finally` restores from its own pre-pin snapshot at line 652 before control leaves the module, so anything written here is overwritten — and in the only case where it could matter (restore broken for `minutes_now`), it converts the loud leaked 857 into the subtle pin value the PR exists to stop leaking. Deleting it leaves the ERROR print plus `failed = True` doing the flagging.

apps/predbat/tests/test_prediction_batch.py
```diff
@@ -34,6 +34,14 @@
 from tests.test_kernel_parity import apply_random_scenario, kernel_available, make_step_data, make_windows, restore_scenario_state, snapshot_scenario_state
 
 
+# create_predbat() takes minutes_now from the wall clock, and only part of the day leaves the seed-5
```
springfall2008: _Automated comment from the triage bot._

Altitude: the actual leak source is the fixture, not this module — `create_predbat()` (unit_test.py:390) calls `update_time()`, which derives `minutes_now` from the host clock, and never calls `reset_inverter`, so every registry runner must self-pin against a wall-clock fixture. This is the third per-module patch of the same standalone-vs-suite class (PR #5013 pinned the date in test_energydataservice; this PR pins the clock).

Pinning once in `create_predbat` — `minutes_now = 12 * 60` plus `now_utc = midnight_utc + timedelta(minutes=minutes_now)`, exactly the way `apply_random_scenario` already pins a scenario's clock (test_random_scenarios.py:728-729) — would fix every current and future module at once, and make this module's constant, the runner pin and the round-trip self-test unnecessary. Not blocking; flagging because the next time-dependent module repeats the whole dance otherwise.

apps/predbat/tests/test_prediction_batch.py
springfall2008: _Automated reply from the triage bot._

Implemented: the runner now records `entry_minutes_now = my_predbat.minutes_now` before the snapshot and, in the `finally` after the restore, fails the run when `minutes_now != entry_minutes_now` — which pins the snapshot-before-pin ordering, not just the round trip, and also catches a dropped `minutes_now` list entry (the failure mode suggested on the constant thread). The separate self-test is gone; this check is the runner's own, so it runs on every invocation including the kernel-skip path.

One honest caveat: the pin value is deliberately the fixture's noon, which is also the in-suite residue, so like the old guard this can only fire standalone — or whenever the caller's clock happens to differ from the pin. Telling a loss-of-pin apart in-suite would need a pin deliberately different from the fixture's, which would change the clock the batch scenarios were built against; I left that alone.

apps/predbat/tests/test_prediction_batch.py
springfall2008: _Automated reply from the triage bot._

Implemented: the `prediction_kernel_enable = False` reset moved into a `finally` that only wraps the kernel-dependent tests, so the skip path no longer writes the flag and returns to the pre-PR behaviour of touching nothing. The unconditional snapshot on the skip path stays: the runner now pins the clock before the availability gate, so the restore is what hands the caller's clock back on the skip path too.

apps/predbat/tests/test_prediction_batch.py
springfall2008: _Automated reply from the triage bot._

Implemented, and the same hoist resolved the root cause suggested on the create_predbat thread: `FIXTURE_MINUTES_NOW = 12 * 60` now lives in tests/test_infra.py next to `reset_inverter` (which uses it), and the batch module imports it instead of restating the literal — so the two pins cannot drift silently.

apps/predbat/tests/test_kernel_parity.py
springfall2008: _Automated reply from the triage bot._

Kept the entry, but the comment above the list now states its role precisely: the parity scenarios only ever read `minutes_now`; the entry exists for `run_prediction_batch_tests`, which pins the fixture clock while it runs and hands the caller's back through the snapshot mechanism this list drives — so an auditor who is about to drop it learns it is load-bearing for another module rather than inferring that by grep.

I did not switch to fixture-level pinning alone as the suggested resolution suggests, because the module-level pin plus hand-back is still doing work: it keeps the module correct even if a future fixture change moves the clock, and the hand-back check is now a runner-level test rather than the deleted self-test. Both suggestions are reflected in the commit — the fixture pin went in (create_predbat thread), and the list's docstring now names its one non-parity consumer.

apps/predbat/tests/test_prediction_batch.py
springfall2008: _Automated reply from the triage bot._

Implemented both: the standalone self-test is deleted, and with it the dead repair line — the runner's `finally` was overwriting that repair with the caller's actual clock anyway, and in the only case it could matter (restore broken for `minutes_now`) it converted a loud leaked sentinel into the subtle pin leak. The property the round trip was checking now lives in the runner-level hand-back check (see the guard thread): the real `snapshot_scenario_state` → pin → `restore_scenario_state` path is exercised with the caller's actual entry value on every run, no deep copy of the ~70-attribute state needed for it.

apps/predbat/tests/test_prediction_batch.py
springfall2008: _Automated reply from the triage bot._

Implemented: `create_predbat()` now pins `minutes_now = FIXTURE_MINUTES_NOW` and `now_utc = midnight_utc + timedelta(minutes=FIXTURE_MINUTES_NOW)` right after `update_time()`, the same way `apply_random_scenario` pins a scenario's own clock — so nothing inherits the wall clock any more, and the third per-module patch of this class does not materialise. `reset_inverter` and the batch module import the same `FIXTURE_MINUTES_NOW` constant from tests/test_infra.py, so all three pins share one definition; the batch module keeps its own pin and hand-back as belt-and-braces, which is now cheap rather than load-bearing.

Verified beyond the module suite passing: a temporary registered probe test confirmed a fresh `create_predbat()` carries `minutes_now = 720` and `now_utc = midnight + 720` (probe removed before commit), and the full quick suite plus pre-commit pass with the pin in place, run this evening local time — the configuration that used to fail standalone.

apps/predbat/tests/test_prediction_batch.py
```diff
@@ -607,28 +608,48 @@ def test_batch_thread_count_resolution():
...
+            finally:
+                my_predbat.prediction_kernel_enable = False
     finally:
         restore_scenario_state(my_predbat, state)
-        my_predbat.prediction_kernel_enable = False
+        clock_handed_back = my_predbat.minutes_now == entry_minutes_now
+
+    if not clock_handed_back:
+        print("ERROR: the batch tests handed back minutes_now {} instead of the caller's {}".format(my_predbat.minutes_now, entry_minutes_now))
+        failed = True
```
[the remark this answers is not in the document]
springfall2008: _Automated reply from the triage bot._

Confirmed: `grep -rn test_minutes_now_is_pinned_and_restored apps/` returns nothing. That test existed in the first revision and was deleted in the previous review round — it deep-copied the whole ~70-attribute `SCENARIO_STATE_ATTRS` state to assert that one int survived `restore_scenario_state`, and its failure-branch repair line was dead because the runner's `finally` overwrote it from the pre-pin snapshot before control left the module.

I took the second branch of your suggestion and updated the PR description rather than re-adding the test, because the inline check you are commenting on covers strictly more: it runs the real `snapshot_scenario_state` → pin → `restore_scenario_state` path against the caller's actual entry clock on every invocation (including the kernel-skip path), so it also catches a dropped `minutes_now` entry in `SCENARIO_STATE_ATTRS` and a snapshot taken after the pin — neither of which the deleted test could see, since it round-tripped a value it had written itself.

The description was stale in a second way as well: it still described the pre-review implementation (a module-local `12 * 60` literal pinned after the snapshot). It now describes the fixture-level pin that replaced it — `FIXTURE_MINUTES_NOW` in `tests/test_infra.py`, applied in `create_predbat()` — and names the runner check as the coverage.

apps/predbat/unit_test.py
```diff
@@ -388,6 +389,15 @@ def create_predbat():
     my_predbat.states = {}
     my_predbat.reset()
     my_predbat.update_time()
+    # update_time() takes the clock from the host, so the same module used to behave differently
+    # standalone and in the suite - a suite run sat at reset_inverter's noon residue while a
+    # standalone run inherited the wall clock, and a time-of-day-dependent test could pass one way
+    # and fail the other (#5026). Pin the fixture clock here instead, now_utc from midnight_utc so
+    # the two stay consistent, the same way the scenario loader pins a scenario's own clock
+    # (test_random_scenarios.py apply_random_scenario). Modules that want more still pin and hand
+    # their clock back themselves, but nothing inherits the wall clock any more.
```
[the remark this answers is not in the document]
springfall2008: _Automated reply from the triage bot._

Fixed in 0a363419, before this reply. The comment now scopes the claim to the two fields the pin actually sets (unit_test.py:397-400):

> Only minutes_now and now_utc are pinned here - update_time() has already taken midnight_utc and now_utc_real from the host clock and they are left alone. Modules that want more still pin and hand their clock back themselves, but no module inherits the wall clock into those two fields any more.

Your reading of the code was right: `update_time()` still runs first and `midnight_utc`/`now_utc_real` remain host-derived, which is deliberate — `now_utc` is pinned as `midnight_utc + FIXTURE_MINUTES_NOW` so the fixture's date stays today's while its time-of-day is fixed.

PR #5038, opened 2026-09-10
feat(fox): select Feed-in First for a Fox Cloud freeze export

### LINKED ISSUE #5022: Fox Cloud: support Feed-in First (freeze export) via the work-mode setting

springfall2008:
Fixes #5022. Draft — the Feed-in behaviour itself still needs a live hardware run (see below).

## The problem

Predbat could plan a freeze export for a FoxCloud inverter but never deliver one — the plan showed a working freeze that execution didn't do, which is the #5015 symptom.

FoxCloud declares `has_timed_pause: False` and `charge_discharge_with_rate: False`, so in a freeze-export slot execute.py's only remaining lever is `adjust_charge_rate(0)` (execute.py:504-516). For FoxCloud `charge_rate` maps to `number.…_battery_schedule_charge_power` — the `fdPwr` of a **charge group**. With no charge window running, that zero meant nothing: the schedule came out identical to plain demand, and the surplus PV a freeze exists to export charged the battery instead.

Sunsynk and Deye hit exactly this and solved it by inferring the freeze inside the component (`sunsynk.py:672-699`, `deye.py:603-619`). This follows them.

## The fix

`fox.py` infers the freeze from the rates Predbat writes, and sets the work mode in the scheduler it already programmes:

- **`freeze_export_requested()`** — charge power `== 0` **and** discharge power `> 0`. That is precisely the signature execute.py leaves: `resetCharge = False` holds the charge rate at zero while `resetDischarge` stays true, so execute.py:633 resets discharge to maximum. The non-zero export half keeps a system whose rates were never derived out of a permanent freeze.
- **`validate_schedule(..., baseline_work_mode=)`** — the enabled filler slots run `Feedin` instead of `SelfUse`. Padding slots stay `SelfUse`; they're disabled placeholders, so their mode is inert.
- **A charge power change now re-applies the schedule.** `press_and_poll_button` only fires from `adjust_charge_window` when the charge *times or enable flag* change, so a freeze — which changes neither — never reached the inverter at all. This also closes a latent bug: the freeze's zero could be stranded in an enabled *future* `ForceCharge` group as `fdPwr: 0`, because the restore to full rate is itself only a rate change. `set_scheduler` compares against the live schedule before writing, so a redundant re-apply costs no API call.

**Known coarseness:** Predbat never tells the component when a freeze ends, so all filler slots carry `Feedin`, not only the freeze period. Splitting a slot at "now" would move a boundary every cycle and force an API write every 5 minutes for the length of the freeze. Sunsynk and Deye rejected that for the same reason.

## The work-mode setting is not involved

The scheduler always covers 00:00–23:59 (`validate_schedule` gap-fills, and `set_scheduler_enabled(False)` is documented as unreachable), so the device-level work-mode setting never takes effect while Predbat is running.

FoxCloud was the **only** inverter type declaring `has_fox_inverter_mode`, which bought it a branch in `adjust_inverter_mode` pinning the mode to `SelfUse` every cycle — fighting the scheduler and stopping Feed-in First ever being selected. That flag is removed along with the branch, and the two leftover mutually-exclusive `inv_has_ge_eco_toggle` checks collapse into one if/else.

FoxESS modbus never had the flag (it drives the work mode through the `templates/fox.yaml` service stanzas, which is unchanged), and neither did SolisCloud/SunsynkCloud/DeyeCloud/SolaxCloud. Existing `apps.yaml` files carrying `has_fox_inverter_mode` keep working: the `inverter:` override at inverter.py:377-383 is a blind dict update, so the key is simply ignored.

`support_feedin_first` needed no change — it was already `True` for both Fox types, and prediction.py already models "PV above the export limit still charges the battery", which matches the ~200 W the #5015 reporter saw.

## Live test

`fox.py --feedin-schedule` exercises the shipped path against real hardware, rather than a hand-built payload:

```
python3 apps/predbat/fox.py --api-key <key> --serial <sn> --feedin-schedule --hold-seconds 300
```

It reads the current schedule (which also picks up the device's real `fdPwr`/`fdsoc` limits), drives `freeze_export_requested` from the state a freeze leaves, writes the schedule, reads it back to prove the API accepted `Feedin` in a slot, holds it while reporting SoC / battery / PV / feed-in power, then restores the original schedule in a `finally` block — so a failure or Ctrl-C still puts the inverter back.

## Testing

- 8 new unit tests. Each was watched failing first; the two `SelfUse` guards were mutation-checked by dropping the `discharge > 0` half and by forcing the inference true, and the no-mode-write test by reinstating the dummy-arg suppression.
- The live-test tool was smoke-tested offline with the network stubbed, including an inverter that silently substitutes `SelfUse` (reports `FAIL` and names `workMode` as the mismatch) and an exception mid-hold (restores before propagating).
- `./run_all` green (131s, no skips); `./run_pre_commit` clean.

## What still needs checking

- **Not validated on hardware.** `OPTIONS_WORK_MODE` listing `Feedin` says the enum knows the value, not that the scheduler endpoint accepts it in a written slot. `--feedin-schedule` is what settles that.
- Feed-in First is not an exact SoC hold on this hardware, so what to look for is `invBatPower` near zero while feed-in tracks PV minus load — not a dead flat battery.


springfall2008:
## Live hardware result: `Feedin` is accepted

Run on an **EVO 10-5-H** ( v3 scheduler write path). The write succeeded and the inverter went into Feed-in First — so the one thing this PR was resting on reasoning for is now measured. The read-back is where it got interesting.

## Finding 1: the Fox scheduler read is eventually consistent

```
19:42:21  write Feedin via /op/v3/device/scheduler/enable  ->  result True
19:42:24  /op/v2/device/scheduler/get                     ->  still SelfUse   (pre-write)
19:42:39  /op/v2/device/scheduler/get                     ->  Feedin
```

The first run reported `FAIL: Feedin did not survive the write` purely because it believed a read taken 3s after the write. So the staleness window is somewhere between 3s and 18s on this device.

## Finding 2: that staleness could strand a freeze — a real bug in this PR

`set_scheduler` skips a write whose schedule matches its cache, and `get_scheduler` wrote **every** read straight into that cache. So a component poll landing inside the staleness window regressed the cache to the pre-write schedule. When Predbat then wanted to go back to that schedule, the write looked redundant and was skipped — leaving the inverter in the mode we meant to leave.

For this feature that is: freeze ends, Predbat asks for the `SelfUse` baseline, compares it against a cache that already says `SelfUse`, skips the write, and **the inverter stays in Feed-in First — not charging — for the rest of the day.**

It is not theoretical. It is what happened to the live test's own restore:

```
Restore write result: False, schedule back to original: False
WARNING: the original schedule was NOT restored - check the inverter
      workMode: original=SelfUse  now=Feedin
```

The test left a real inverter in Feedin.

## Fix

`17f82c37`. A scheduler write is recorded, and a read that disagrees with it inside `SCHEDULER_READ_STALE_SECONDS` (60s) is treated as stale — the cache keeps what we wrote, while the caller still gets the raw read. Past that window the read wins, so a schedule changed in the Fox app is still picked up rather than being masked forever.

Three tests, each watched failing first:

- a stale read no longer regresses the cache
- the write that **ends** a freeze is no longer skipped after a stale read (this is the one that matters — it asserts the write actually goes out)
- a settled read still wins once the window has passed

## Live test hardened

- The read-back polls until it agrees with what was written, rather than calling a landed write rejected on one early read. Against a simulation of the timings above it now waits ~15s and reports `PASS`.
- The restore clears the cache first so it **always** writes, whatever the cache believes, and confirms itself the same patient way.
- `--restore-selfuse` writes a plain all-day Self Use schedule, to recover an inverter left in another work mode.

## Still to do

The hold in this run was 15s at 19:42 — dark, `pv=0.0kW`, `battery=0.397kW` discharging to the house at 39% SoC. So the mode was accepted, but nothing was observed about what Feed-in First does to surplus PV. That needs a daytime run with real solar:

```
python3 apps/predbat/fox.py --token-hash <hash> --serial xxxx --feedin-schedule --hold-seconds 300
```

What to look for is `invBatPower` near zero while `feed-in` tracks `pv` minus load — not a flat battery, since PV above the export limit still charges on this hardware.

springfall2008:
## Correction: I overstated the production severity above

@springfall2008 pointed out that reading the scheduler straight after a write is a test artifact, not something normal operation does. That's correct, and it makes my previous comment's severity claim wrong. Verified:

- Production has exactly **one** `get_scheduler` caller — `fox.py:544`, gated on `FOX_REFRESH_SETTINGS`, i.e. **once an hour** — and it is independent of writes. Nothing reads back after a write.
- So the poisoned-cache path needs the hourly poll to land inside the ~18s staleness window *following* a write. That is a fraction of a percent per hour in practice (writes only happen when the schedule changes), not a routine occurrence.
- It also **self-heals**: the next hourly poll reads the true state and corrects the cache, so the next schedule change writes correctly. My "not charging, for the rest of the day" was wrong — the realistic exposure is until the next hourly poll plus the next schedule change.

The two things that were genuinely broken were both in the live-test tool, and both are fixed by the tool changes alone (patient read-back; restore clearing the cache so it always writes).

## So why keep `17f82c37`?

Because the hole is real, just rare, and the fix is inert when it isn't needed. Confirmed by probing each normal-operation case:

| case | cached value |
|---|---|
| no prior write (cold hourly poll) | identical to the read, whole dict equal |
| write >60s ago (the normal case) | identical to the read, write record cleared |
| recent write the read agrees with | identical to the read, write record cleared |
| external change well after our write | read wins |

The written schedule is only preferred over a read in the one case that motivated it: a read contradicting our own write within 60s. The record is never persisted, so a restart carries none of it.

Happy to drop the commit and keep this PR to the feature alone if you'd rather not carry the extra surface for a rare case — the live test is sound either way.

PR #5041, opened 2026-09-10
Fix GE Cloud force charge scheduling

### LINKED ISSUE #5040: GE Cloud: enable_force_charge is never driven, so grid charging silently does nothing on devices that have both charge switches
**Describe the bug**

On GivEnergy Cloud (`GEC`, `ge_cloud_automatic: true`) devices whose register set exposes **both** `Enable AC Charge` and `Enable Force Charge`, Predbat only ever drives `enable_ac_charge`. `enable_force_charge` is never read and never written — the string does not appear anywhere in the codebase.

On this hardware both switches must be on for the inverter to import from the grid. The result is a silent failure: Predbat enters a charge state, sets the AC charge slot, reports `Charging target …%` in the log, and the inverter imports nothing. There is no error and no warning — the plan simply doesn't happen.

The asymmetry is visible in one place, `async_automatic_config()`:

```python
# apps/predbat/gecloud.py (main @ v9.0.1)
1326:  self.set_arg("scheduled_charge_enable",    build_entities("switch", ["ac_charge_enable", "enable_ac_charge"]))
1327:  self.set_arg("scheduled_discharge_enable", build_entities("switch", ["enable_dc_discharge", "enable_force_discharge"]))
```

The export side lists `enable_force_discharge` as a candidate and works correctly. The charge side has no `enable_force_charge` candidate. `enable_default_options()` (`gecloud.py:1010`) doesn't touch it either, so nothing ever puts it into a known state.

**Evidence from an affected system**

GivEnergy 3-phase, serial `TD2343G049`, 19.97 kWh battery, 18.27 kWp PV. Predbat planned a grid charge from 15:30 to 16:00 on 2026-09-07.

Predbat's own status log — a genuine charge state, not Freeze Charge:

```text
2026-09-07 15:25:05: Info: record_status Freeze charging target 90%
2026-09-07 15:30:16: Info: record_status Charging target 90%-97%
2026-09-07 15:40:03: Info: record_status Charging target 90%-97%
2026-09-07 15:50:08: Info: record_status Charging target 92%-97%
2026-09-07 15:55:03: Info: record_status Charging target 93%-97%
2026-09-07 16:01:23: Info: record_status Demand
```

Metered power over the same window (5-minute buckets):

```text
time    soc_kwh    batt_kW    grid_kW
15:30    15.277     -0.001      -2.50
15:35    15.447     -0.251      -0.02
15:40    15.277      2.012       0.00
15:45    15.277     -1.959       0.11
15:50    15.617     -3.254       0.19
15:55    15.786     -2.528      -0.20
16:00    16.126     -4.366      -0.19
```

Grid stayed between -0.20 and +0.19 kW for the whole window — no import at all. The battery did rise (15.28 -> 16.13 kWh), but entirely from surplus solar. `Enable AC Charge` was on; `Enable Force Charge` was off.

Setting `Enable Force Charge` on by hand makes grid charging work immediately.

**Why the obvious one-line fix isn't quite right**

Adding `"enable_force_charge"` to the `scheduled_charge_enable` candidate list would *swap* one switch for the other, because `build_entities()` returns the first matching candidate — it would not drive both.

**Expected behaviour / suggested fix**

Where the device exposes `enable_force_charge`, treat that register as the scheduled-charge toggle and hold `enable_ac_charge` permanently on:

1. In `async_automatic_config()`, prefer `enable_force_charge` when the register is present:

   ```python
   self.set_arg("scheduled_charge_enable", build_entities("switch", ["enable_force_charge", "ac_charge_enable", "enable_ac_charge"]))
   ```

   `enable_force_charge` is the switch that actually gates the timed grid charge on this hardware, so binding it here makes the whole existing mechanism work unchanged — `adjust_charge_window()` already toggles the bound entity via `write_and_poll_switch("scheduled_charge_enable", ...)` in both directions (`inverter.py`, enable and disable paths).

2. In `enable_default_options()`, turn `enable_ac_charge` **on** and leave it on, so it stops being a second gate Predbat has to think about. This is exactly the pattern `real_time_control` already uses in that function — a register that is set once and simply held in the state Predbat needs.

   Gate this on `enable_force_charge` being present in the device's registers, so devices where `enable_ac_charge` *is* the scheduled-charge toggle keep their current behaviour and are not force-enabled.

That leaves one switch as the per-slot control and one as a static enable, which matches how the export side already behaves.

**Why leaving `Enable Force Charge` on permanently is not an adequate workaround**

The affected user tried this. It fixes the "won't charge" case but introduces the opposite failure: with Force Charge held on, the inverter keeps acting on the last-written `ac_charge_1` window after Predbat has cleared `enable_ac_charge`. On the evening of 2026-09-07 Predbat turned `Enable AC Charge` off at 20:05 intending the battery to discharge into house load, and SoC stayed pinned at 71% until 22:00 (the end of the stale slot). Both switches need to be under Predbat's control, in both directions.

**Predbat version**

`main` @ `v9.0.1` — the binding at `gecloud.py:1326` is unchanged there, so this is current. Originally observed on a v8.5x deployment.

**Environment details**

- `inverter_type: GEC`, `ge_cloud_direct: true`, `ge_cloud_automatic: true`
- GivEnergy 3-phase, serial `TD2343G049`, `max_inverter_rate` 9984 W, 19.97 kWh battery
- Registers present on this device: `enable_ac_charge`, `enable_force_charge`, `enable_force_discharge`, `charge_up_to_percent`, `discharge_down_to_percent`, `charge_power_rate`, `discharge_power_rate`, `battery_reserve_percent`, `real_time_control`, `ac_charge_1..10_*`, `dc_discharge_1..10_*`
- Notably **absent**: `enable_dc_discharge`, `battery_charge_power`, `enable_eco_mode`

Across the fleet we can see this register combination on 4 systems, and every one of them has `enable_ac_charge` + `enable_force_charge` + `enable_force_discharge` with no `ac_charge_enable` and no `enable_dc_discharge`. So the proposed precedence rule targets exactly this family and is a no-op for every other GE device. Two of those 4 currently have `Enable Force Charge` off and are presumably unable to grid-charge without knowing it.

**Related**

Same class of GEC auto-config gaps on non-standard GivEnergy register sets: #4908, #4909, #5017. #4784 (watchdog for "Predbat reports Charging/Exporting but the battery is doing nothing") would have surfaced this one.

**Screenshots**

None.

**Log file**

Excerpts inline above.

LouisDeconinck:
Fixes #5040

## Summary
- Prefer the enable_force_charge register as the GE Cloud scheduled-charge gate when a device exposes it.
- Keep enable_ac_charge enabled for devices with both switches so timed grid charging works in both directions.
- Add regression coverage for the force-charge device shape and the legacy fallback.

## Testing
- cd coverage && ./run_all --test ge_cloud (93 passed)
- cd coverage && ./run_pre_commit (all hooks and quick tests passed)
- git diff --check

AI-assisted: OpenAI Codex.

### DIFF
--- a/apps/predbat/gecloud.py
+++ b/apps/predbat/gecloud.py
@@ -1016,6 +1016,7 @@ class GECloudDirect(ComponentBase):
         device_ha_names = {regname_to_ha(registers[key].get("name", "")) for key in registers}
         has_charge_power = "battery_charge_power" in device_ha_names
         has_discharge_power = "battery_discharge_power" in device_ha_names
+        has_force_charge = "enable_force_charge" in device_ha_names
         # Predbat drives the export target register itself (discharge_target_soc, the DC discharge
         # lower SoC limit) and tracks the minimum reserve SoC there, so leave that one alone rather
         # than resetting it and fighting adjust_force_export.
@@ -1131,6 +1132,15 @@ class GECloudDirect(ComponentBase):
                     changed = True
                 else:
                     self.log("GECloud: Warn: Failed to enable real-time control for {}".format(device))
+            if has_force_charge and ha_name == "enable_ac_charge" and not value:
+                self.log("GECloud: Enabling AC charge for {} because force charge is available".format(device))
+                result = await self.async_write_inverter_setting(device, key, True)
+                if result and ("value" in result):
+                    registers[key]["value"] = result["value"]
+                    await self.publish_registers(device, self.settings[device], select_key=key)
+                    changed = True
+                else:
+                    self.log("GECloud: Warn: Failed to enable AC charge for {}".format(device))
         return changed
 
     async def publish_registers(self, device, registers, select_key=None):
@@ -1323,7 +1333,7 @@ class GECloudDirect(ComponentBase):
         self.set_arg("charge_limit_enable", build_entities("switch", ["enable_ac_charge_upper_percent_limit", "enable_ac_charge_1_upper_soc_percent_limit"]))
         self.set_arg("discharge_start_time", [f"select.{self.prefix}_gecloud_{device}_dc_discharge_1_start_time" for device in batteries])
         self.set_arg("discharge_end_time", [f"select.{self.prefix}_gecloud_{device}_dc_discharge_1_end_time" for device in batteries])
-        self.set_arg("scheduled_charge_enable", build_entities("switch", ["ac_charge_enable", "enable_ac_charge"]))
+        self.set_arg("scheduled_charge_enable", build_entities("switch", ["enable_force_charge", "ac_charge_enable", "enable_ac_charge"]))
         self.set_arg("scheduled_discharge_enable", build_entities("switch", ["enable_dc_discharge", "enable_force_discharge"]))
         self.set_arg("battery_temperature", [f"sensor.{self.prefix}_gecloud_{device}_battery_temperature" for device in batteries])
         self.set_arg("battery_scaling", [f"sensor.{self.prefix}_gecloud_{device}_battery_dod_soh" for device in batteries])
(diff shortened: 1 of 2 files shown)


springfall2008:
Looks like you got here a minute or so before me, I didn't notice that there is a bug in the checks, you had == "enable_ac_charge" rather than "enable_ac_charge" in...

springfall2008:
Duplicate of: https://github.com/springfall2008/batpred/pull/5042 - with the same bug actually !

Issue #4909, opened 2026-09-02
3phase GE Cloud: no enable_eco_mode switch (or anything eco/demand-mode related) is created. So inverter_mode slot is empty


kennedy-nic:
**Describe the bug**
When Predbat needs to bring the inverter out of "Export (Paused)" mode (has_ge_eco_toggle: True), it attempts to do so by calling the select.turn_on service against the configured inverter_mode entity. This always fails, because the select domain in Home Assistant has no turn_on service — only select.select_option is valid for a select entity.

Log excerpt:
```text
2026-09-02 15:34:40.967995: Inverter 0 current mode is Export (Paused) and new target is on has_ge_eco_toggle True
2026-09-02 15:34:41.014856: Warn: Web Socket result failed {'id': 30, 'type': 'result', 'success': False, 'error': {'code': 'not_found', 'message': 'Service select.turn_on not found.', ...}}
2026-09-02 15:34:41.015061: Warn: Service call select/turn_on data {'entity_id': 'select.givtcp_tj2515g023_mode'} failed: {...}
```

This repeats roughly every 10 seconds (ids 30 through 49) for about 3.5 minutes, across at least two separate attempts (15:34:41 and again 15:36:21), before Predbat finally gives up:
```text 
2026-09-02 15:38:02.109945: Warn: Inverter 0 Trying to write inverter_mode to True didn't complete got Export (Paused)
2026-09-02 15:38:02.141131: Warn: record_status Warn: Inverter 0 write to inverter_mode failed
2026-09-02 15:38:02.141236: Inverter 0 set force export to False
```
Predbat then falls back and continues operating normally, so this isn't a permanent hang — but it never succeeds in bringing the inverter out of "Export (Paused)" via this path, and generates ~20 warning-level log entries each time it tries.

**Expected behaviour**
Predbat should either call select.select_option with the correct mode value for select-domain inverter_mode entities, or otherwise detect the entity's domain and use the correct service rather than assuming turn_on unconditionally.

**Predbat version**
v8.54.2

**Environment details**
GivEnergy 3-phase HV inverter, model GIV-3HY-20, serial TJ2515G023. 
inverter_type: GEC, 
ge_cloud_direct: true, 
ge_cloud_automatic: false (disabled per workaround for #4908). 
Per-inverter control entities are explicitly mapped to GivTCP-sourced entities in apps.yaml, including inverter_mode: select.givtcp_{geserial}_mode

**Screenshots**
None

**Log file**

[...]

PR #5042, opened 2026-09-10
fix(gecloud): drive enable_force_charge as the scheduled charge control (#5040)

### LINKED ISSUE #5040: GE Cloud: enable_force_charge is never driven, so grid charging silently does nothing on devices that have both charge switches

springfall2008:
This is an automated draft PR generated from issue #5040 — a maintainer should review it before merging.

Fixes #5040

## Summary

On GE Cloud devices whose register set exposes **both** `Enable AC Charge` and `Enable Force Charge`, both switches gate grid import, but Predbat only ever drove `enable_ac_charge` — `enable_force_charge` appeared nowhere in the codebase. A planned charge window therefore entered the charge state, set the AC charge slot, logged `Charging target …%`, and imported nothing.

Two changes, both gated on the **device's own** registers so a standard GivEnergy device is unaffected:

- `async_automatic_config()` now binds `scheduled_charge_enable` from candidates `["enable_force_charge", "ac_charge_enable", "enable_ac_charge"]`. Where the force-charge register exists it becomes the per-slot control, so the existing `write_and_poll_switch("scheduled_charge_enable", …)` calls in `inverter.py` (enable at `:3230`, disable at `:2814`/`:1590`) land on the switch that actually gates the timed charge. Devices without that register keep their current binding, and `ac_charge_enable` still wins over `enable_ac_charge` as before.
- `enable_default_options()` now holds `enable_ac_charge` **on** when — and only when — that same device also exposes `enable_force_charge`, so it stops being a second gate Predbat has to think about. This mirrors the static-enable pattern `real_time_control` already uses in that function. The register check is built from the per-device `registers` dict passed in, so mixed multi-inverter sites are handled per device (the flag-vs-per-device wrinkle raised in triage and previously hit on #4909).

That leaves one switch as the per-slot control and one as a static enable, matching how the export side (`enable_dc_discharge` / `enable_force_discharge`) already behaves. It also avoids the failure mode the reporter hit when holding Force Charge on by hand: `enable_force_charge` is now cleared at the end of the window like any other scheduled-charge toggle, so the inverter stops acting on a stale `ac_charge_1` slot.

The detected-features log line now also reports `force charge: <bool>`, so this register family is visible in a bug report without needing a full register dump.

## Testing

- New `force_charge_control` test in `apps/predbat/tests/test_ge_cloud.py` (registered in the `test_ge_cloud` sub-registry) covering both halves: the binding for a both-switches device, a standard `enable_ac_charge`-only device, `ac_charge_enable` precedence, a mixed two-inverter site, and for `enable_default_options` — the write when AC charge is off, no write when already on, **no** write at all without a force-charge register, the exact-match guard against `Enable_AC_Charge_Upper_Percent_Limit`, and the rejected-write path.
- Verified the test genuinely catches the bug: with the `gecloud.py` change reverted it fails (`Both switches present should bind enable_force_charge, got ['switch.predbat_gecloud_battery001_enable_ac_charge']`), and with only the `enable_default_options` half reverted it fails on the missing write.
- `tools/triage_test.sh ge_cloud` — **94 passed, 0 failed**.
- `coverage/run_pre_commit` — all hooks passed and `./run_all --quick` passed (4 slow tests skipped, 83s).

## Notes

`.claude/skills/issue-triage/references/debug-journal.md` records this class of GE Cloud auto-config gap on non-standard register sets (the `GE Cloud (gecloud.py)` row, alongside #4908). Worth adding this register family to that entry once the approach is confirmed on hardware.

**Cannot be verified from code and needs maintainer/hardware confirmation:** that both switches must be on for grid import on this hardware, and that clearing `enable_force_charge` alone is sufficient to stop a charge. Both rest on the reporter's observations.

### DIFF
--- a/apps/predbat/gecloud.py
+++ b/apps/predbat/gecloud.py
@@ -1118,7 +1118,11 @@ class GECloudDirect(ComponentBase):
                             changed = True
                         else:
                             self.log("GECloud: Warn: Failed to set {} for {}".format(ha_name, device))
-            if ha_name == "enable_ac_charge" and has_force_charge:
+            # Matched by substring as GE spells this family inconsistently across firmware, but the
+            # AC charge *upper limit* switch shares the prefix under at least three names
+            # (enable_ac_charge_upper_percent_limit, enable_ac_charge_1_upper_soc_percent_limit and
+            # enable_ac_charge_upper_limit) and is a different control, so limit registers are excluded.
+            if ("enable_ac_charge" in ha_name) and ("limit" not in ha_name) and has_force_charge:
                 if value:
                     changed = True
                     continue
@@ -1285,7 +1289,7 @@ class GECloudDirect(ComponentBase):
                     has_discharge_target_soc = True
                 if "pause_battery" in ha_name:
                     has_pause_battery = True
-                if ha_name == "enable_force_charge":
+                if "enable_force_charge" in ha_name:
                     has_force_charge = True
 
         def register_names(device):
--- a/apps/predbat/gecloud.py
+++ b/apps/predbat/gecloud.py
@@ -1124,7 +1124,7 @@ class GECloudDirect(ComponentBase):
             # enable_ac_charge_upper_limit) and is a different control, so limit registers are excluded.
             if ("enable_ac_charge" in ha_name) and ("limit" not in ha_name) and has_force_charge:
                 if value:
-                    changed = True
+                    # Already on, so nothing was written and there is no change to report.
                     continue
                 self.log("GECloud: Enabling AC charge for {} as enable_force_charge is the scheduled charge control, current value is {}".format(device, value))
                 result = await self.async_write_inverter_setting(device, key, True)
--- a/apps/predbat/gecloud.py
+++ b/apps/predbat/gecloud.py
@@ -1016,6 +1016,10 @@ class GECloudDirect(ComponentBase):
         device_ha_names = {regname_to_ha(registers[key].get("name", "")) for key in registers}
         has_charge_power = "battery_charge_power" in device_ha_names
         has_discharge_power = "battery_discharge_power" in device_ha_names
+        # When this device exposes both charge switches, async_automatic_config binds
+        # enable_force_charge as scheduled_charge_enable, so enable_ac_charge is a second gate that
+        # must be held on or the timed grid charge silently imports nothing (GH#5040).
+        has_force_charge = "enable_force_charge" in device_ha_names
         # Predbat drives the export target register itself (discharge_target_soc, the DC discharge
         # lower SoC limit) and tracks the minimum reserve SoC there, so leave that one alone rather
         # than resetting it and fighting adjust_force_export.
@@ -1114,6 +1118,19 @@ class GECloudDirect(ComponentBase):
                             changed = True
                         else:
                             self.log("GECloud: Warn: Failed to set {} for {}".format(ha_name, device))
+            if ha_name == "enable_ac_charge" and has_force_charge:
+                if value:
+                    changed = True
+                    continue
+                self.log("GECloud: Enabling AC charge for {} as enable_force_charge is the scheduled charge control, current value is {}".format(device, value))
+                result = await self.async_write_inverter_setting(device, key, True)
+                if result and ("value" in result):
+                    registers[key]["value"] = result["value"]
+                    await self.publish_registers(device, self.settings[device], select_key=key)
+                    changed = True
+                else:
+                    self.log("GECloud: Warn: Failed to enable AC charge for {}".format(device))
+                continue
             if "real_time_control" in ha_name:
                 if self.ems_device:
                     # RTC is on the EMS, not the individual inverters — skip
@@ -1248,6 +1265,7 @@ class GECloudDirect(ComponentBase):
         has_pause_start_time = False
         has_discharge_target_soc = False
         has_pause_battery = False
+        has_force_charge = False
         for device in batteries:
             registers = self.settings.get(device, {})
             for key in registers:
@@ -1267,6 +1285,8 @@ class GECloudDirect(ComponentBase):
                     has_discharge_target_soc = True
                 if "pause_battery" in ha_name:
                     has_pause_battery = True
+                if ha_name == "enable_force_charge":
+                    has_force_charge = True
 
         def register_names(device):
             """Return HA-format register names for a specific device."""
@@ -1295,8 +1315,8 @@ class GECloudDirect(ComponentBase):
 
         self.log("GECloud: Auto-configuring Predbat and not using apps.yaml entries for control")
         self.log(
-            "GECloud: detected features - charge rate: {}, discharge rate: {}, charge power percent: {}, discharge power percent: {}, pause battery: {}, pause start time: {}, discharge target soc: {}".format(
-                has_charge_rate, has_discharge_rate, has_charge_power_percent, has_discharge_power_percent, has_pause_battery, has_pause_start_time, has_discharge_target_soc
+            "GECloud: detected features - charge rate: {}, discharge rate: {}, charge power percent: {}, discharge power percent: {}, pause battery: {}, pause start time: {}, discharge target soc: {}, force charge: {}".format(
+                has_charge_rate, has_discharge_rate, has_charge_power_percent, has_discharge_power_percent, has_pause_battery, has_pause_start_time, has_discharge_target_soc, has_force_charge
             )
         )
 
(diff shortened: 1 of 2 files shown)


apps/predbat/gecloud.py
```diff
@@ -1114,6 +1118,19 @@ async def enable_default_options(self, device, registers):
                             changed = True
                         else:
                             self.log("GECloud: Warn: Failed to set {} for {}".format(ha_name, device))
+            if ha_name == "enable_ac_charge" and has_force_charge:
```
springfall2008: Should be "enable_ac_charge" in ha_name

apps/predbat/gecloud.py
```diff
@@ -1267,6 +1285,8 @@ async def async_automatic_config(self, devices):
                     has_discharge_target_soc = True
                 if "pause_battery" in ha_name:
                     has_pause_battery = True
+                if ha_name == "enable_force_charge":
```
springfall2008: Should be "enable_force_charge" in ha_name

apps/predbat/gecloud.py
springfall2008: Automated reply from the triage bot.

Changed to a substring match, but with the limit registers excluded: `("enable_ac_charge" in ha_name) and ("limit" not in ha_name)`.

The reason for the exclusion: `enable_ac_charge` is a prefix of the AC charge *upper limit* switch, which is a different control — Predbat binds it separately as `charge_limit_enable` at `gecloud.py:1343`. Three spellings of it are known: `enable_ac_charge_upper_percent_limit` and `enable_ac_charge_1_upper_soc_percent_limit` (both in that binding), plus `enable_ac_charge_upper_limit` (`docs/inverter-setup.md:141`).

The pre-existing `if "enable_" in ha_name: continue` guard at `gecloud.py:1048-1050` catches the first two, but not the third — it contains neither `ac_charge_upper_percent_limit` nor `_upper_soc_percent_limit`. I confirmed this: with a bare substring match and no exclusion, `Enable_AC_Charge_Upper_Limit` gets written to True. So the exclusion is doing real work rather than duplicating that guard, and I did not want to depend on an unrelated block's `continue` for correctness.

All three spellings are now pinned in the `force_charge_control` test, along with a `Enable_AC_Charge_2` case confirming a suffixed spelling of the real switch is still matched.

If you would rather the upper-limit switch also be force-enabled, that is a separate behaviour change from this issue and I have left it out — say the word and I will add it.

apps/predbat/gecloud.py
springfall2008: Automated reply from the triage bot.

Changed to `"enable_force_charge" in ha_name`, matching the rest of that detection loop. No behavioural risk here — this flag only feeds the "detected features" log line. Test added for a suffixed spelling (`Enable_Force_Charge_2`) still reporting `force charge: True`.

One related spot I deliberately left as an exact match, flagging it in case you want it changed too: `has_force_charge` in `enable_default_options` (`gecloud.py:1019`) is `"enable_force_charge" in device_ha_names` — set membership, so exact. That one gates whether `enable_ac_charge` is held on, and it needs to be true precisely when `build_entities` actually bound `enable_force_charge` as `scheduled_charge_enable`. `first_existing_entity` matches its candidates by exact set membership, so making the flag broader than the binding would force-enable AC charge on a device where `enable_ac_charge` is still the scheduled charge control — the exact failure this fix is meant to avoid. Happy to change it if you disagree.

apps/predbat/gecloud.py
```diff
@@ -1114,6 +1118,19 @@ async def enable_default_options(self, device, registers):
                             changed = True
                         else:
                             self.log("GECloud: Warn: Failed to set {} for {}".format(ha_name, device))
+            if ha_name == "enable_ac_charge" and has_force_charge:
+                if value:
+                    changed = True
+                    continue
```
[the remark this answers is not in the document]
springfall2008: Automated reply from the triage bot.

Fixed in 016df1cd — the already-on path now just `continue`s without touching `changed`.

Confirming the reasoning, since the branch was not arbitrary: it was modelled on the `real_time_control` block at `gecloud.py:1142-1145`, which *does* set `changed = True` when the register is already enabled, and `test_ge_cloud.py` asserts that ("should return True when real-time control already enabled"). So the file is not uniformly "False means no changes needed" today.

That said, `git log -L` shows that line was added in 835be31c, titled "Fix tests" — bolted on to make a test pass rather than as a deliberate contract. And `enable_default_options`'s return value is discarded at its only call site (`gecloud.py:1832`), so nothing observes it but tests. Given that, "changed means a write happened" is the better meaning and I have not propagated the divergence.

I have left the `real_time_control` block itself alone — changing it would flip an existing asserted behaviour that is out of scope for this issue.

apps/predbat/tests/test_ge_cloud.py
```diff
@@ -4746,6 +4747,168 @@ async def run(batteries_info):
...
+        write_calls.clear()
+        registers = {
+            200: {"name": "Enable_AC_Charge", "value": True, "validation_rules": []},
+            201: {"name": "Enable_Force_Charge", "value": True, "validation_rules": []},
+        }
+        result = await ge_cloud.enable_default_options("test123", registers)
+
+        if write_calls:
+            print("ERROR: Should not write when Enable_AC_Charge is already on, got {}".format(write_calls))
+            return 1
```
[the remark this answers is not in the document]
springfall2008: Automated reply from the triage bot.

Added in 016df1cd. The `force_charge_control` test now asserts the return value in every case, not just the already-on one:

- AC charge off, force charge present → `True` (a write happened)
- AC charge already on → `False`
- each of the three upper-limit spellings → `False`
- no force-charge register on the device → `False`
- suffixed `Enable_AC_Charge_2` → `True`
- write rejected by the API → `False`

Verified the new assertion is load-bearing rather than decorative: restoring the old `changed = True` on the already-on path makes the test fail with "should report no change when Enable_AC_Charge is already on".

PR #5047, opened 2026-09-11
refactor(export): replace the packed export-limit float with a (mode, target, power) tuple

chalfontchubby:
Claude wrote this PR (code and description) working with @chalfontchubby.

## Why

`export_limits_best[n]` was a single packed float carrying three orthogonal signals: target SoC in the integer part, export power as `1 - fraction` in the decimal part, and mode via two reserved sentinel values (`99.0` = freeze, `100.0` = idle). Every consumer re-derived intent by comparing against those sentinels, inconsistently - some `== 99`, some `< 99`, some `>= 99`. That encoding caused real bugs, notably GH#4914: a target of 99% at reduced power lands in the reserved range and reads as "freeze" instead of "export to 99%", leaving the window doing nothing.

This replaces the packed float with a `(mode, target, power)` tuple, decoded everywhere through named accessors, with the C++ kernel updated to match.

## What changed, commit by commit

1. **`refactor(export): route all export-limit reads through named accessors`** - adds `export_mode_of` / `export_target_of` / `export_power_of` / `export_limit_exports_no_battery` / `export_limit_is_full_discharge` / `export_limit_sort_key` / `pack_export_limit` in `utils.py`, and converts every read site across `plan.py`, `prediction.py`, `execute.py`, `output.py` and `gateway.py` to use them. The packed float is still the live representation underneath - this commit only changes how it's read. A pre-commit hook (`check_export_limit_encoding.py`) rejects new code that compares an export limit against the reserved sentinels or bare `99`/`100`, so the old idiom can't quietly come back.
2. **`fix(plan): keep the export power out of the SoC a target exports to`** - a real bug this sweep surfaced: `clip_export_slots` was reading the raw packed float as the target SoC, so a reduced-power export (e.g. 47.3 = target 47% at 70% power) clipped against 47.3% instead of 47%. Fixed in both the Python and C++ paths, with a regression test.
3. **`refactor(export): make an export limit a (mode, target, power) tuple`** - swaps the internal representation. Because every read site already went through an accessor in commit 1, this is a small, mechanical change: the accessors grow an `isinstance(tuple)` fast path, write sites build tuples via `pack_export_limit`, and the kernel marshalling wraps values through `export_limit_sort_key` for ordering.
4. **`feat(export): persist export limits as a self-describing mapping`** - the persisted plan (cache and debug dumps) now stores `{mode: "target", target: 47, power: 0.7}` instead of a bare float. A permanent compatibility path (`export_limit_from_stored`) reads the old float form, the new mapping, or a bare 3-element list (what a tuple becomes after a YAML/JSON round trip) - so an old cached plan or an old bug-report debug dump keeps loading.
5. **`refactor(kernel): pass export limits as three field arrays, not one packed double (ABI 6)`** - the C++ kernel takes `export_modes`/`export_targets`/`export_powers` as three parallel arrays instead of one packed double array, so the reserved-value ambiguity GH#4914 depended on can't exist at the kernel boundary either.
6. **`perf(kernel): marshal export limits as one array-of-structs, packed with struct (ABI 7)`** - three parallel arrays meant three separate marshalling passes and three pointers per call; this packs `{mode, target, power}` into one `struct.Struct("@iid")`-packed buffer, `from_buffer_copy`'d into a C array-of-structs, cutting marshalling overhead.
7. **`perf(kernel): cache the marshalled export-limit buffer by contents`** - the same `export_limits_best` list is often marshalled many times across a plan search; this caches the packed buffer keyed by its contents (bounded at 512 entries) so repeated scenarios skip re-marshalling.

## Testing

- **Full unit suite** (`./run_all`, not just `--quick`) and `./run_pre_commit` pass standalone at every one of the 7 commits, verified with a script that checks out each commit in turn and runs both - not just at the tip.
- `run_random` (20 fixed scenarios) is bit-identical to `main` at every commit except commit 2, which is a deliberate behaviour fix (see above).
- Kernel parity and `model_kernel` tests green throughout; `KERNEL_PARITY_REVISION` and `KERNEL_ABI_VERSION` bumped appropriately per commit, all 6 cross-compiled `.so` targets rebuilt via `zig` at each ABI change.
- New regression tests added in the commit that fixes each bug: the SoC-clip bug (commit 2), three tuple-shaped export scenarios in `test_window.py` (commit 3), and round-trip/old-float/plain-YAML persistence tests (commit 4).
- **A real bug found deploying to my own running addon**: `scenario_summary_state()` (the verbose debug "STATE:" log line) compared an export-limit tuple directly against a float with `>=`, a leftover bare-number comparison the accessor sweep in commit 1 missed. It never surfaced in the test suite because that call site only runs behind `debug_enable`, and only when an active freeze/target export window is below the current SoC - which is exactly what a live plan produced within a few hours of deploying. Fixed and folded back into commit 1 (where the sweep belongs), with a new direct regression test (`test_plan_scenario_summary.py`) that reproduces it and fails without the fix.
- **Replayed my own live debug dumps** (`--debug_file`) against the branch: a fresh dump (already saved in the new mapping format, proving the upgrade path works) replays cleanly; an old dump from before this branch existed (packed-float format) also replays cleanly through the legacy compatibility path, with no exceptions either way.
- **Cross-checked the same old dump against v9.0.1** (the version actually running in production) to separate "replay drift" from "branch regression": both versions compute the identical fresh metric (-141.7968) from the same input, confirming the refactor is behaviour-neutral on real data as well as synthetic scenarios.

## Speed

This needed more digging than expected, and the honest answer is: no measured regression, but also no confidently measured improvement in production.

- A Mac-local interleaved A/B (`run_random`-based) measured struct-packing (commit 6) alone at ~+2.9% slower than main, recovered to ~+1.2% net once the buffer cache (commit 7) was added.
- Deployed to my own running Predbat addon (a Pi 4) and compared `Plan calculation took` log timings, p10 (the uncontended single-pass figure) over an equivalent overnight window: branch measured ~+11% slower than the prior night's v9.0.1 baseline (p10 2.91s vs 2.63s), consistent across every percentile - too large and too consistent to be noise.
- To isolate the cause, redeployed at commit 5 (ABI 6 - three parallel arrays, no struct packing, no buffer cache) rather than the tip. Overnight `plan_stats.sh` comparison for that version is still accumulating samples as of this PR.
- Separately, replaying an old debug dump through both the branch and v9.0.1 interleaved (6 rounds each, alternating) showed no measurable difference - branch and v9.0.1 medians were within noise of each other (~0.27-0.28s).

So: real production plans on the full 7-commit branch showed a slowdown the synthetic Mac benchmark didn't predict; a single-scenario interleaved replay shows nothing. That's not fully reconciled yet. Flagging it rather than claiming a number I can't stand behind - happy to keep investigating post-merge, or to hold commits 6/7 back if reviewers would rather land the correctness fix (commits 1-5) first and chase the last bit of kernel perf separately.

## Known follow-ups

- Reconcile the Mac-vs-Pi kernel performance discrepancy (see Speed above) - which of ABI 6, ABI 7's struct packing, or the buffer cache is actually responsible, under real traffic rather than the synthetic `run_random` corpus.
- `self.export_limits_best` is not defensively coerced on load if a hot-reload or partial restart leaves stale data in memory (as opposed to the on-disk formats, which are all handled) - a clean process restart always fixes this via `load_saved_plan`, so it's a narrow window, but worth hardening if it recurs.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>


chalfontchubby:
Claude here. Rewrote history to fold the review fixes into the commits that introduced each bug, rather than stacking them on top - so the branch stays bisectable and every commit still passes the full suite standalone. The earlier automated replies (55b3e2d6, c05b0e79, 1be5cde6, a79a69cc) point to commits that no longer exist after the rewrite, but the fixes themselves are all still present.

One of a79a69cc's own fixes (output.py's stored_target handling) turned out to be wrong when folded back in - it broke the ordinary case while fixing the legacy one, caught by run_random before push. Fixed properly and documented in the amended commit message.

Also fixed two more gaps the same review flagged that hadn't been addressed yet: the kernel marshaller trusting every element was already a tuple (prediction_kernel.py), and the 3-element sequence branch of export_limit_from_stored being unvalidated (utils.py).

Issue #4900, opened 2026-09-02
Forced Export gets erased after few hours


madoboy:
I set forced export from 5.30am till 9.00am at morning 1.30am clock to have enough battery capacity for later cheap rate when i planed to charge the car from 9.00am till 13.00am. But when i checked at 7.30am they are cleared from the plan, and no forced export was done. Similar happened with forced charge at weekend free electricity sessions. 

I would like that forced charge or export don’t get erased after few hours. 

**Predbat version**

8.54.2

**Environment details**

- Sunsynk 3.6 and Seplos battery 18kw
- Standard HAOS installer

[apps.yaml.txt](https://github.com/user-attachments/files/31725931/apps.yaml.txt)
[predbat.txt](https://github.com/user-attachments/files/31725930/predbat.txt)

[](url)

springfall2008:
*Automated first-pass triage — a maintainer will review before any action is taken.*

**Classification:** enhancement (with an explanation of what actually happened — the reporter's log is complete and tells the whole story)

The log (confirmed running **v8.54.2**, Sunsynk/SK, IOG import + Agile export, 18.72 kWh Seplos) shows the requested morning export **was** in the plan, and exactly when and why it was removed:

1. **01:23–02:00** — the plan contained precisely the requested export: `Export windows filtered [ 02-09 05:30-06:00 @ 15.72p 86.0%, 06:00-06:30 @ 76.0%, 06:30-07:00 @ 65.0%, 07:00-07:30 @ 55.0%, ... ]` — a cascade discharging the battery across 05:30–09:30.
2. **02:05:07** — the 5-minute re-optimisation found an alternative plan it scored 54p better (`Previous plan best metric is -117.71 ... new plan best metric is -172.26 ... using new plan`): drop nearly all morning export and keep the energy for the 16:00–20:00 export window at 20.86p (vs ~15–17p in the morning). That ~5p/kWh × ~14 kWh gap fully explains the flip, so the optimiser's arithmetic is self-consistent — it just isn't what the reporter wanted.
3. From 02:05 the morning export was gone from every recalc (fresh calcs scored the same as the retained plan, so `metric_min_improvement_plan 2.0` kept it). At 07:30, when the reporter checked, only a 07:50–08:00 sliver remained. Predbat did briefly re-add a 07:30–08:00 window at 07:50 and exported for ~10 minutes ("Exporting now - current SoC 18.158kWh and target 16.66kWh"), but the 05:30–07:30 opportunity was gone.

[...]

PR #5054, opened 2026-09-11
fix(solcast): stop day-bucketing against calculate_yesterday's rewound midnight_utc (#4804)

### LINKED ISSUE #4804: Solcast PV forecast sensors collapse to near-zero and don't self-recover (planning forecast unaffected)
<h3 dir="ltr">Describe the bug</h3>
<p dir="ltr">The <code>sensor.predbat_pv_*</code> forecast display sensors (<code>pv_today</code>, <code>pv_forecast_h0</code>, <code>pv_forecast_raw</code>, and related) periodically collapse to near-zero or flat values and remain stuck there for many hours — well beyond the configured <code>solcast_poll_hours</code> refresh interval — while <code>predbat.pv_energy</code> (the forecast Predbat actually plans against) stays accurate and healthy throughout.</p>
<p dir="ltr">This has recurred over multiple days, not as a single one-off event.</p>
<h3 dir="ltr">Steps to reproduce / observed pattern</h3>
<ol dir="ltr">
<li>Solcast forecast sensors update normally for a period (e.g. <code>pv_today</code> showing ~15kWh mid-morning)</li>
<li>At some point (observed once at a ~23:03 refresh, and again around midday on a subsequent day) the sensors collapse:
<ul dir="ltr">
<li><code>sensor.predbat_pv_today</code>: dropped from ~15kWh to <strong>0 kWh</strong></li>
<li><code>sensor.predbat_pv_forecast_h0</code> (PV Forecast Now): read <strong>0.62 kW</strong> against actual live PV of <strong>0.9–3.0 kW</strong></li>
<li><code>sensor.predbat_pv_forecast_raw</code>: curve peaked at <strong>0.035 kW</strong> for the remainder of the day (~100x too low)</li>
<li><code>sensor.predbat_pv_forecast_tomorrow</code>: dropped from ~14kWh to <strong>4.6 kWh</strong> and stayed flat</li>
</ul>
</li>
<li>Despite the above, <code>predbat.pv_energy</code>'s forecast curve remained normal and the actual plan (charge/discharge decisions) was unaffected — confirmed via the plan's own live PV/load columns tracking correctly against real generation.</li>
<li>The degraded state persisted across multiple subsequent Solcast poll cycles (<code>solcast_poll_hours: 5.0</code>) without self-correcting.</li>
</ol>
<h3 dir="ltr">Expected behaviour</h3>
<ul dir="ltr">
<li>The <code>pv_*</code> display sensors should either track <code>predbat.pv_energy</code> directly, or reliably recover on the next scheduled Solcast poll rather than persisting a stale/degraded value across multiple poll cycles.</li>
<li>Ideally, some staleness detection so a bad/empty API response doesn't silently overwrite a previously-good forecast for hours.</li>
</ul>
<h3 dir="ltr">Workaround found</h3>
<p dir="ltr">Manually restarting the Solcast integration/component in Home Assistant immediately resolved it — all affected sensors snapped back to correct values right after the restart, with no other changes made.</p>
<h3 dir="ltr">Suspected root cause</h3>
<p dir="ltr">Possibly related to Solcast API rate limiting: this setup polls at <code>solcast_poll_hours: 5.0</code> across <strong>2 configured <code>solcast_sites</code></strong>, which works out to roughly 10 API calls/day. On Solcast's free/hobbyist tier this may be at or near the daily quota, which could cause degraded/cached/empty responses to be accepted and then persisted rather than being detected as invalid and discarded.</p>
<h3 dir="ltr">Environment</h3>
<div dir="ltr">
  |  
-- | --
Predbat version | v8.53.4
Inverter | LuxPower LXP 3600ACS (AC-coupled)
PV forecast source | Solcast cloud API, 2 sites, solcast_poll_hours: 5.0
Home Assistant | HA Green

</div>
<h3 dir="ltr">apps.yaml (relevant section)</h3>
<div role="group" aria-label="yaml code" tabindex="0"><div><div></div></div><div>yaml</div><div><pre style="color: rgb(234, 236, 240); background: transparent; font-family: var(--font-mono);"><code class="language-yaml" style="color: rgb(234, 236, 240); background: transparent; font-family: var(--font-mono); white-space: pre;"><span><span><span style="color: rgb(244, 123, 133);">solcast_host</span><span style="color: rgb(211, 215, 222);">:</span> <span style="color: rgb(155, 233, 99);">https://api.solcast.com.au/</span>
</span><span><span style="color: rgb(244, 123, 133);">solcast_api_key</span><span style="color: rgb(211, 215, 222);">:</span> <span style="color: rgb(204, 123, 244);">!secret</span> <span style="color: rgb(155, 233, 99);">solcast_api_key</span>
</span><span><span style="color: rgb(244, 123, 133);">solcast_sites</span><span style="color: rgb(211, 215, 222);">:</span>
</span><span>  <span style="color: rgb(211, 215, 222);">-</span> <span style="color: rgb(155, 233, 99);">"REDACTED-SITE-1"</span>
</span><span>  <span style="color: rgb(211, 215, 222);">-</span> <span style="color: rgb(155, 233, 99);">"REDACTED-SITE-2"</span>
</span><span><span style="color: rgb(244, 123, 133);">solcast_poll_hours</span><span style="color: rgb(211, 215, 222);">:</span> <span style="color: rgb(94, 237, 237);">5.0</span></span></span></code></pre></div></div>
<h3 dir="ltr">Related</h3>
<p dir="ltr">This may be related to #1137, where a maintainer noted a similar Solcast outage and suggested "maybe we could check for staleness in some way?" — this report may be a more specific/reproducible instance of that same underlying gap, since here the API itself hasn't failed outright, but a degraded response is being accepted and then never re-validated.</p>
<h3 dir="ltr">Additional notes</h3>
<p dir="ltr">I don't currently have a debug file captured from the actual broken window, since the issue self-resolved after a component restart before I thought to generate one. Happy to attach <code>predbat_debug.yaml</code> next time it recurs, if that would help.</p>

chalfontchubby:
_Posted by Claude on behalf of @chalfontchubby._

Fixes #4804.

## Problem

`calculate_yesterday()` (`output.py`) rewinds the shared `self.midnight_utc` by one day for the duration of the savings calculation - `self.midnight_utc = self.midnight_utc - timedelta(days=1)`, restored ~350 lines later. Components run on their own OS threads (`hass.py`'s `create_task` uses `threading.Thread`, not asyncio), and `SolarAPI`'s publish cycle runs on an independent schedule from `calculate_yesterday()`'s hourly one, so the two can genuinely overlap.

`publish_pv_stats()` read `self.midnight_utc` directly to bucket forecast entries by day (`day = (this_point - midnight_today).days`). Caught mid-rewind, every entry bucketed one day late - "today" emptied, every other day shifted down one slot.

## Verification against the reporter's log

Confirmed rather than just theorised:
- `calculate_yesterday()`'s "Calculating data from yesterday" line and the corrupted PV forecast publish are interleaved in the same second in the attached log.
- The raw (pre-calibration) values at the bad cycle are identical to the *previous* cycle's values one day earlier - `today=0, tomorrow=(old today), d2=(old tomorrow)`, and so on through `d6` - exactly the signature a midnight one day early produces, not calibration noise or a bad API response.

## Fix

Derives `midnight_today` from `self.now_utc` instead - the same field `calculate_yesterday()` never touches, the same pattern already used for the equivalent race on the manual-override decode path (`userinterface.py`'s `manual_time_origin()`, #4900).

Deliberately not `now_utc_exact`: that property reads the live wall clock on the real `ComponentBase` rather than a value the test fixture controls, so using it broke 65 of the file's 73 other tests that set `mock_base.now_utc` but not `now_utc_exact` - caught by running the full test file, not just the new test, before treating this as done.

**Scope note:** this is a patch for the one confirmed victim (`SolarAPI`'s day-bucketing). The general hazard - `calculate_yesterday()` mutating ~15 shared instance fields in place with no lock, readable by any concurrently-running component thread via `ComponentBase`'s live passthrough properties - is unfixed and could recur elsewhere. Left as a follow-up rather than a larger concurrency refactor here.

## Testing

Mutation-tested: reverting the fix reproduces the exact incident signature (`today=0`, the rest shifted) and fails only the new regression test, none of the other 72 tests in the file.

- `./run_all --quick` green (310 tests)
- `./run_pre_commit` green

### DIFF
--- a/apps/predbat/solcast.py
+++ b/apps/predbat/solcast.py
@@ -693,8 +693,17 @@ class SolarAPI(ComponentBase):
             total_dayCL[day] = 0
             forecast_day[day] = []
 
-        midnight_today = self.midnight_utc
+        # Derived from now_utc rather than read from self.midnight_utc, which
+        # calculate_yesterday() rewrites to yesterday's midnight for the duration of the savings
+        # calculation. Those writes land on the shared PredBat instance, so this publish - running
+        # on its own thread, independent of calculate_yesterday()'s schedule - could otherwise read
+        # a midnight a day behind and bucket every forecast entry one day late, emptying "today"
+        # until calculate_yesterday() finishes and restores the real value (GH#4804). now_utc is
+        # never touched by calculate_yesterday() (only midnight_utc/minutes_now/forecast_minutes
+        # and the rate/car fields are faked), the same pattern already used for the equivalent
+        # race on the manual-override decode path (userinterface.py's manual_time_origin(), #4900).
         now = self.now_utc_exact
+        midnight_today = self.now_utc.replace(hour=0, minute=0, second=0, microsecond=0)
 
         power_scale = 60 / period  # Scale kwh to power
         power_now = 0
--- a/apps/predbat/tests/test_solcast.py
+++ b/apps/predbat/tests/test_solcast.py
@@ -1444,6 +1444,77 @@ def test_publish_pv_stats(my_predbat):
     return failed
 
 
+def test_publish_pv_stats_ignores_stale_shared_midnight(my_predbat):
+    """publish_pv_stats must not bucket forecast entries against self.midnight_utc (GH#4804).
+
+    calculate_yesterday() (output.py) rewinds the shared self.midnight_utc by one day for the
+    duration of the savings calculation, then restores it ~350 lines later. Components run on
+    their own OS threads (hass.py's create_task uses threading.Thread), and SolarAPI's publish
+    cycle runs on an independent schedule from calculate_yesterday()'s hourly one - so the two
+    can genuinely overlap. If that happens while self.midnight_utc is rewound,
+    publish_pv_stats()'s day-bucketing (day = (this_point - midnight_today).days) used to bucket
+    every entry one day late, emptying "today" and shifting the rest down, until
+    calculate_yesterday() finished and restored the real value.
+
+    Confirmed against a real incident log: at the moment of collision, every "PV Forecast for
+    day dN" value was identical to the previous cycle's dN-1 value, with day 0 always empty -
+    exactly the signature this test reproduces by making midnight_utc desync from now_utc.
+
+    The fix derives midnight_today from now_utc instead of midnight_utc. now_utc is never
+    touched by calculate_yesterday() (only midnight_utc/minutes_now/forecast_minutes and the
+    rate/car fields are faked), the same pattern already used for the equivalent race on the
+    manual-override decode path (userinterface.py's manual_time_origin(), #4900). Deliberately
+    not now_utc_exact: that property is datetime.now(self.local_tz) on the real ComponentBase -
+    the live wall clock, not the fixture's frozen mock_base value - so using it here would make
+    every OTHER publish_pv_stats test (which sets mock_base.now_utc but not now_utc_exact)
+    silently read the real clock instead of the fixture's controlled date.
+    """
+    print("  - test_publish_pv_stats_ignores_stale_shared_midnight")
+    failed = False
+
+    test_api = create_test_solar_api()
+    try:
+        # Simulate the collision: now_utc says it is still 2025-06-15, but midnight_utc - the
+        # field calculate_yesterday() rewinds - has been shifted back a full day, as if a
+        # concurrent calculate_yesterday() call were mid-flight on another thread.
+        test_api.mock_base.now_utc = datetime(2025, 6, 15, 12, 0, 0, tzinfo=pytz.utc)
+        test_api.mock_base.midnight_utc = datetime(2025, 6, 14, 0, 0, 0, tzinfo=pytz.utc)
+
+        pv_forecast_data = [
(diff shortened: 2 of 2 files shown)


apps/predbat/solcast.py
```diff
@@ -693,8 +693,17 @@ def publish_pv_stats(self, pv_forecast_data, divide_by, period):
...
+        # calculate_yesterday() rewrites to yesterday's midnight for the duration of the savings
+        # calculation. Those writes land on the shared PredBat instance, so this publish - running
+        # on its own thread, independent of calculate_yesterday()'s schedule - could otherwise read
+        # a midnight a day behind and bucket every forecast entry one day late, emptying "today"
+        # until calculate_yesterday() finishes and restores the real value (GH#4804). now_utc is
+        # never touched by calculate_yesterday() (only midnight_utc/minutes_now/forecast_minutes
+        # and the rate/car fields are faked), the same pattern already used for the equivalent
+        # race on the manual-override decode path (userinterface.py's manual_time_origin(), #4900).
         now = self.now_utc_exact
+        midnight_today = self.now_utc.replace(hour=0, minute=0, second=0, microsecond=0)
```
springfall2008: Isn't this best fixed in component base so it covers all components?

PR #5064, opened 2026-09-12
docs(sunsynk): recommend single timer slot configuration

carig-rob:
Simplifies the Sunsynk setup instructions to recommend direct single timer slot control, which requires no Sunsynk-specific Home Assistant automations or template sensors.

This approach has been running reliably since March 2026, including charge windows that cross midnight, and avoids the timer overlap issues encountered with the six-slot configuration.

The six-slot method and the previous automation-based approach remain documented as alternatives.

The standard utility_meter block is still required, as it is for every Predbat installation — it is not Sunsynk-specific.

Requested by @gcoan in discussion #4609.

### DIFF
--- a/docs/inverter-setup.md
+++ b/docs/inverter-setup.md
@@ -2785,179 +2785,82 @@ See the components documentation for details: [Components - Sunsynk Cloud API](c
 
 ## Sunsynk
 
-- Copy the Sunsynk template over the top of your `apps.yaml`, and edit for your system.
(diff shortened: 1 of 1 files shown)


carig-rob:
"Hi @gcoan, just a gentle bump on this one. It's the Sunsynk documentation update you suggested in #4609. The automated checks passed, so I just wanted to make sure it hadn't been missed. Thanks!"

PR #5071, opened 2026-09-13
Raise debug_history_count maximum to 500 for multi-week history (#5070)

### LINKED ISSUE #5070: Enhancement request: raise or make configurable the `debug_history_count` limit
Predbat v9.0.2 currently limits `debug_history_count` to 50 snapshots.
For fault-finding and trend analysis, I attempted to configure:
interval: 1 hour
snapshot count: 336
The intention was to retain 14 days of hourly plan-state history.
Predbat clamps the count to 50, so at a 1-hour interval only about 50 hours of rolling history are retained.
That is useful for short-term debugging, but too short for investigating intermittent optimiser behaviour that may only occur on a few nights over a 1–2 week period.
In one recent investigation, I needed to audit 14 days of charging behaviour. Because Predbat's own retained history was too short, the analysis had to be reconstructed from Home Assistant meter history instead.
Request
Please either:
raise the hard maximum substantially, or
make the maximum configurable, or
allow retention to be configured by time period rather than only by snapshot count.
For example, 336 hourly snapshots would provide 14 days of history and would be a useful practical diagnostic window.
This is separate from the optimiser feature request itself, but the limited history made that optimiser issue significantly harder to diagnose.

springfall2008:
This is an automated draft PR generated from issue #5070 — a maintainer should review it before merging.

Fixes #5070

## Summary

Raises the `debug_history_count` schema maximum from 50 to 500 (`apps/predbat/config.py`), as requested by @chalfontchubby on the issue, and documents the storage cost of using it, as requested in the same thread.

The default stays at 15 and the minimum stays at 1, so nothing changes for an install that never touches the setting — only the ceiling on what a user can opt into moves. The retained window is already time-based (`max_age = debug_history_interval × debug_history_count` in `_capture_debug_history()`), so this alone delivers the 336-snapshots-at-1-hour ≈ 14-day window the reporter asked for, with no new mechanism. No code below the schema needed changing: `_capture_debug_history()` passes the live config value straight through, and the clamp in `load_user_config()` reads the schema bound, so it now permits 336 from apps.yaml as well as from the HA entity.

The docs change (`docs/customisation.md`) states the new maximum, spells out that the window is `count × interval` (so a fortnight can be had at 336×1h or 112×3h), and adds a CAUTION noting that snapshots are full dumps of roughly 2MB each — a 15-snapshot history measures about 32MB on disk, so the 500 ceiling is of the order of 1GB — and that the **download all** archive loads every retained snapshot into memory to build the tarball.

## Testing

New `test_debug_history_count_range` in `apps/predbat/tests/test_debug_history_capture.py` (registered in `TEST_REGISTRY` as `debug_history_count_range`). It checks the declared range reaches 336 while the default stays 15 and the minimum stays 1, then drives the real clamp path — `expose_config` + `load_user_config()` + `get_arg()`, the same route `_capture_debug_history()` reads through — to confirm 336 now survives unclamped without flagging `had_errors`, and that 501 still clamps to 500 and does flag it.

- `tools/triage_test.sh debug_history_count_range` with the `config.py` change stashed: **FAILS** (`debug_history_count max should allow at least 336 snapshots, got 50`, and `expected debug_history_count 336 to pass through unclamped, got 50`).
- Same command with the fix in place: **PASSES**.
- `./run_pre_commit` from `coverage/`: all hooks pass and the full quick suite passes (exit 0).

## Notes

- **Not fixed, deliberately, and worth the maintainer's attention:** `load_all_snapshots()`/`build_archive()` (`debug_history.py:258-290`) load every retained snapshot's full text into memory and gzip it in one pass, so at 500 snapshots the "download all" endpoint would buffer on the order of 1GB per request. Raising the ceiling makes that reachable where it previously was not. The issue thread scoped this PR to the max bump plus docs warnings, so this is documented rather than changed — capping or streaming the archive build looks like its own ticket.
- `debug_history_interval` is left at its 24-hour maximum; the issue and the thread both asked only about the count, and 500 × 24h is already far past any useful window.
- Blast radius: the only consumer of this value is `_capture_debug_history()` (confirmed with `impact({target: "_capture_debug_history", direction: "upstream"})` — 1 direct caller, `update_pred`; flagged HIGH only because it sits on the main loop, and this change alters no behaviour at the unchanged default).


apps/predbat/config.py
```diff
@@ -1227,7 +1227,13 @@
         "friendly_name": "Debug history snapshot count",
         "type": "input_number",
         "min": 1,
-        "max": 50,
+        # The maximum only bounds what a user can opt into, the default below is what almost every
+        # install actually runs. It was raised from 50 to 500 for #5070: intermittent optimiser
+        # behaviour often needs a week or two of history to audit, and at the 1-hour minimum
+        # interval 50 snapshots only reached back about two days. 500 covers 14 days hourly (336)
+        # with headroom. Snapshots are full debug dumps, roughly 2MB each, so the top of this range
```
springfall2008: **Snapshot-size figures are understated ~2.5x by the repo's own artifacts.** A snapshot is `create_debug_yaml(write_file=False)` stored verbatim (debug_history.py:143-177) - the same content as a full debug dump - and the committed replay dumps measure ~5MB each (`coverage/cases/predbat_debug_agile1.yaml` = 5,324,826 bytes, `predbat_debug_pre_saving1.yaml` = 5,035,614 bytes). So 15 snapshots is ~75MB (docs/customisation.md:886 says 'roughly 32MB') and the 500 maximum is ~2.5GB on disk, not the ~1GB stated here. On the SD-card installs the caution targets, that difference decides whether the setting is safe; the figure scales with install size (entity/inverter count), so consider deriving it at capture time (log the measured size when it crosses a threshold) rather than hardcoding it in two places.

docs/customisation.md
```diff
@@ -877,10 +877,14 @@ mode: single
...
 
+The retained window is simply `debug_history_count` x `debug_history_interval`, so a longer window can be had either by keeping more snapshots or by spacing them further apart. If you are chasing something intermittent that only shows up on the odd night, 336 snapshots at a 1-hour interval covers a fortnight, and 112 snapshots at 3-hourly covers the same fortnight for a third of the disk space.
+
+_CAUTION: each snapshot is a complete debug dump rather than a delta, around 2MB of YAML apiece (a 15-snapshot history measures roughly 32MB on disk), so the 500-snapshot maximum is of the order of 1GB retained in the `debug/` folder. The **download all** archive described below also has to load every retained snapshot into memory to build the tarball, so a large history makes that download correspondingly heavy. Raise the count only as far as the window you actually need, and put it back to the default once your investigation is finished - especially on an SD-card based install, where the write volume matters as much as the space._
```
springfall2008: **This new caution contradicts the unchanged paragraph four lines below (line 888).** Line 888 still says the archive 'is usually under 5MB' and frames the `.dmp` scheme entirely around GitHub's 25MB attachment limit (web.py:3022-3024 repeats the same numbers in a comment). At the 336-500 counts this PR advertises, the archive is tens to hundreds of MB, so the 'download all and attach to a GitHub issue' flow stops working at exactly the counts the docs now invite - the section is self-contradictory. Worth updating line 888 (and the web.py comment) together with this paragraph.

apps/predbat/config.py
```diff
@@ -1227,7 +1227,13 @@
...
+        # is around 1GB on disk - see the storage warning in docs/customisation.md.
+        "max": 500,
```
springfall2008, line 1237: **Altitude: the combined footprint (count x interval x snapshot size) is guarded nowhere in code.** The two items are validated only as independent min/max ranges; `predbat.py:910` derives `max_age = interval x count` purely as a prune hint, not a resource bound. 500 snapshots at the 3-hour default is ~62 days of full dumps (~1GB+ of small random writes on an SD-card install), and 500 x 24h is a 500-day window - in both cases the only mitigation is this docs caution, which a user configuring entities in the HA UI never sees. A one-time log warning (or a `had_errors`-free status note) when count x measured snapshot size crosses a threshold would put the guard where the setting is made.

apps/predbat/tests/test_debug_history_capture.py
```diff
@@ -255,3 +255,72 @@ def test_debug_history_capture_slot_alignment(my_predbat):
...
+        my_predbat.had_errors = False
+        my_predbat.load_user_config()
+        resolved = my_predbat.get_arg("debug_history_count", 15)
+        if resolved != item_max:
+            print("  FAILED: expected a value above the maximum to clamp to {}, got {}".format(item_max, resolved))
+            failed += 1
+        if my_predbat.had_errors is not True:
+            print("  FAILED: an above-maximum debug_history_count should flag had_errors")
+            failed += 1
+    finally:
```
springfall2008, line 322: **The finally block restores in-memory state but not the persisted copy.** Both `load_user_config()` calls above end with `save_current_config()` (userinterface.py:1153), which writes `item["value"]` into `coverage/predbat_config.json` - so this test persists `debug_history_count` 336 (Test 2) and then the clamped 500 (Test 3) to the saved-JSON file, and the restore here never re-saves. A later full-suite run overwrites it back to 15, but a single-test run (`--test debug_history_count_range`) leaves the coverage working directory with count=500 on disk. Cheap fix: call `my_predbat.save_current_config()` after the restore, or snapshot/restore the JSON around the test.

apps/predbat/tests/test_debug_history_capture.py
```diff
@@ -255,3 +255,72 @@ def test_debug_history_capture_slot_alignment(my_predbat):
...
+        # rather than on the dict alone, and read the result back the way _capture_debug_history()
+        # does. See test_config_item_range_clamp in test_integer_config.py for the same pattern.
+        print("Test 2: a fortnight's worth of hourly snapshots survives a config refresh unclamped")
+        my_predbat.expose_config("debug_history_count", fortnight_hourly, force_ha=True)
+        my_predbat.had_errors = False
+        my_predbat.load_user_config()
+        resolved = my_predbat.get_arg("debug_history_count", 15)
+        if resolved != fortnight_hourly:
+            print("  FAILED: expected debug_history_count {} to pass through unclamped, got {}".format(fortnight_hourly, resolved))
+            failed += 1
```
springfall2008, line 305: **Test 2 can fail spuriously on an unrelated item.** `had_errors` is a process-global flag set by `record_status()` for ANY config item (output.py record_status sets `self.had_errors = True`), and `load_user_config()` here iterates every CONFIG_ITEM. If an earlier-registered test leaves some other input_number out of range in HA state, that unrelated clamp fires during this pass and this 'in-range debug_history_count should not flag had_errors' assertion fails while pointing at debug_history_count. Test-order pollution is a documented trap (tools/debug-journal.md). Scoping the assertion to the debug_history_count clamp message (e.g. checking the recorded status text) or saving/restoring nothing and asserting only the resolved value would be robust.

apps/predbat/tests/test_debug_history_capture.py
```diff
@@ -255,3 +255,72 @@ def test_debug_history_capture_slot_alignment(my_predbat):
...
+
+def test_debug_history_count_range(my_predbat):
+    """Test debug_history_count's schema range reaches a multi-week window (#5070) and still clamps above it."""
+    failed = 0
+    print("--- Debug history snapshot count range tests ---")
+
+    # The window a user can ask for is debug_history_count x debug_history_interval, and the
+    # interval is itself capped at 24 hours, so the count is the only thing standing between the
+    # default ~2-day buffer and the fortnight #5070 needed to audit intermittent optimiser
+    # behaviour. 336 is that report's own worked example: 336 snapshots at a 1-hour interval.
```
springfall2008, line 268: **The load-bearing premise is never asserted.** The comment says the fortnight window comes from 'count x interval' with the interval 'itself capped at 24 hours', but nothing in this test (or anywhere in the suite) checks `debug_history_interval`'s max. If that bound is ever reduced, 336 snapshots spans well under a fortnight, the docs' '112 at 3-hourly' recipe over-promises, and this test stays green - the #5070 capability degrades silently. One extra assertion (`config_index["debug_history_interval"]["max"] >= 24`) would pin the premise.

apps/predbat/tests/test_debug_history_capture.py
```diff
@@ -255,3 +255,72 @@ def test_debug_history_capture_slot_alignment(my_predbat):
...
+        failed += 1
+
+    original_had_errors = my_predbat.had_errors
+    original_value = item.get("value", item.get("default", 15))
+
+    try:
+        # load_user_config() clamps input_number items to their declared min/max, including values
+        # that arrived via an apps.yaml override (which bypasses the HA entity's own enforcement),
+        # so the schema bound above is what a user actually runs into - assert through that path
+        # rather than on the dict alone, and read the result back the way _capture_debug_history()
```
springfall2008, line 296: **Reuse: Tests 2 and 3 re-implement the clamp round-trip that test_config_item_range_clamp already owns** (tests/test_integer_config.py:201 - above-max clamp + had_errors=True at its lines 229-234, in-range pass-through + had_errors=False at 245-250), exercising the same item-agnostic `load_user_config()` clamp path, and the comment restates that test's apps.yaml-override explanation (test_integer_config.py:222-226). The genuinely new information here is only the three per-item bounds assertions above (max >= 336, default == 15, min == 1). If the clamp semantics change, the canonical test gets updated and this copy silently asserts stale semantics. Suggestion: keep the bounds assertions, and either shrink Tests 2-3 to a thin reference or move the whole test beside the clamp test in test_integer_config.py where the per-item config assertions live.

springfall2008:
Automated review of PR #5071 (x-high effort, 10 finder angles + verification). Findings that cannot be anchored inline to the diff (they live in files this PR does not touch, but are direct consequences of the max it raises):

**1. [web.py:3014-3018, debug_history.py:258-290] 'Download all' has no bound at the new 10x maximum - memory and event-loop stall.**
`load_all_snapshots()` materialises every retained snapshot's full YAML text into one list, then `build_archive()` holds all of it plus a second full-size bytes copy per snapshot plus an in-memory tar+gzip `BytesIO`, all synchronously inside the aiohttp handler (blocking the event loop - the whole web UI stalls during the build). At the worked example this PR's own docs invite (336 snapshots at 1-hour), and with committed replay dumps measuring ~5MB each (`coverage/cases/predbat_debug_agile1.yaml` = 5.32MB), peak RSS is of the order of 2-3GB inside the single addon process - enough to OOM-kill Predbat mid-download, taking the plan, the web server and the very history the user was trying to attach with it. The only mitigation shipped is the prose caution in docs/customisation.md:886. A streaming tar (`mode="w|gz"` to a temp file), a per-archive snapshot cap, or a size guard before building would bound this in code; offloading to an executor would unblock the loop.

**2. [docs/customisation.md:888 / web.py:3022-3024] covered inline** on the new CAUTION paragraph: the unchanged 'archive itself is usually under 5MB' claim is now self-contradictory with the counts this PR advertises.

**Verified clean (no finding):** no other code hard-codes the old 50 ceiling; APPS_SCHEMA imposes no parallel bound; existing installs pick up the new max correctly because `load_user_config()` re-pushes min/max/step attributes via `expose_config(force_ha=True)` (userinterface.py:1090-1137, 556-571); the clamp path the new test asserts through behaves exactly as expected (336 unclamped, 501->500 clamped with had_errors=True); the window arithmetic in the comments/docs is correct (336x1h = 14 days, 112x3h = 336h, defaults 15x3h = 45h); unit_test.py import (line 317) and TEST_REGISTRY tuple (line 760) are correct with the right slow-flag; the new test has a docstring, snake_case names, and no over-length lines. Inline comments are on: config.py:1234 (size figures understated ~2.5x), config.py:1236 (no code-level footprint guard), docs:886 (contradiction with line 888), test:268 (unasserted interval premise), test:296 (clamp-test duplication), test:305 (had_errors global-flag fragility), test:322 (finally block leaves 500 in coverage/predbat_config.json).

apps/predbat/config.py
chalfontchubby: Contemporary debug yamls are smaller - mine are currently 2.5MB on disk. Some of the test yamls predate the removal of the solar arrays so are larger. As your other comment notes - it is system dependent. I'll relax the claim here.

PR #5084, opened 2026-09-14
refactor(clock): drop the host-timezone clock, measure everything from now_utc

springfall2008:
_Posted by Claude on behalf of @springfall2008 _

Follow-up to #5054, which fixed `ComponentBase.midnight_utc` reading a `calculate_yesterday()`-rewound value. Trying to give `minutes_now` the same treatment surfaced why it could not simply be derived: it is measured from a different clock.

## Problem

`update_time()` kept **two** clocks:

```python
self.now_utc = datetime.now(self.local_tz)   # the configured `timezone:` setting → midnight_utc
self.now     = datetime.now()                # whatever timezone the container is in → midnight, minutes_now
```

`minutes_now` and the `time_abs_str`/`time_now_str` renderings came off the second one. They agree only while the container's timezone matches the `timezone:` setting.

When they do not:

- `minutes_now` indexes rate and forecast arrays that are keyed in minutes from `midnight_utc`, so Predbat reads "now" at the wrong point in its own rates.
- `time_abs_str(minute)` renders `midnight_utc`-keyed offsets against a differently-offset midnight, so every timestamp it publishes - the plan table included - is shifted from the data it labels.

A test run on a UTC+3 host with `timezone: Europe/London` shows the two 120 minutes apart.

## Fix

Measure everything from `now_utc`, and delete `self.now` / `self.midnight`.

`now_utc` and the midnight derived from it share a tzinfo (`replace()` keeps it), so the subtractions stay wall-clock differences - identical to the naive arithmetic they replace. Checked on both 2026 DST changeover days plus a control date.

The one caller that genuinely needs the host clock is the scheduler in `initialize()`: its run times are handed to `run_every()`, and `hass.py`'s `timer_tick` compares them against `datetime.now()`. It now computes that locally rather than two instance fields existing for its sake.

Also in scope, because they shared the fields:

- `minutes_since_yesterday()` built its comparison point with `datetime.combine()`, which drops tzinfo - `replace()` instead, same answer for the same wall clock.
- `fetch.py`'s `date:` overrides are parsed naive from apps.yaml, so they take `midnight_utc`'s offset before the subtraction; `day_of_week` now reads the configured timezone's weekday.
- `futurerate`'s cache-age arithmetic stays on the host clock - its stamps are host-naive and **persisted to storage by earlier versions**, so an upgrade must keep reading them. It just computes its own midnight rather than borrowing Predbat's.
- `predheat` carried its own copy of the same split and gets the same treatment.

## Behaviour change

Confined to setups where the container timezone differs from the `timezone:` setting: their `minutes_now` and rendered timestamps move to the configured timezone, which is the fix. Matched setups - the normal HA addon case - see nothing move.

## Testing

New `tests/test_clock.py` pins a configured timezone that deliberately disagrees with the host (`Pacific/Kiritimati`, UTC+14, no DST). That disagreement is the only thing that makes the two clocks separable - on a machine where they agree, the old and new code are indistinguishable, which is why nothing caught this before.

Mutation-tested: restoring the host clock fails the new tests (`minutes_now 930` against a configured-timezone `150`) and nothing else.

- `./run_all --quick` green (314 tests) under both a mismatched host timezone and `TZ=Europe/London`
- `./run_pre_commit` green

## Follow-up, not in this PR

`now_utc`/`midnight_utc` are misnamed - they are timezone-aware local time for the inverter system, never UTC. Worth renaming now that they are the only clock.


chalfontchubby:
Hmm. That's not a pr I thought I'd pushed. Need to check what Claude has been up to.

springfall2008:
> Hmm. That's not a pr I thought I'd pushed. Need to check what Claude has been up to.

No this is a new one from me to follow up the solcast time fix

chalfontchubby:
Makes sense, seen a few with related time stuff that implied a refactor might be good. 


Sent from Yahoo Mail for iPhone


On Monday, September 14, 2026, 1:51 pm, Trefor Southwell ***@***.***> wrote:

springfall2008 left a comment (springfall2008/batpred#5084)

Hmm. That's not a pr I thought I'd pushed. Need to check what Claude has been up to.


No this is a new one from me to follow up the solcast time fix

springfall2008:
_Posted by Claude on behalf of @springfall2008._

Worked through the Copilot review. Three findings held up and are fixed in 5f05735; one does not.

**1. `TeslemetryAPI._local_today_weekday()` read `base.now` — real, and the nastiest of the four.**

It reads through `getattr(base, "now", None)` with a fallback, so deleting the field did not break it: it silently dropped to the live wall clock, losing `clock_skew` and any pinned clock. Worse, the test doubles pinned `base.now` as well, so the suite stayed green through exactly the change that broke it. Now reads `base.now_utc`, doubles moved over, and a new test pins the weekday against a fixed Thursday so a fallback fails on six days in seven. Mutation-tested: reverting gives `Expected Thursday (3), got 0`.

**2. PredHeat coverage gap — real.** Confirmed `test_predheat.py` replaces `update_pred` with a lambda, so the changed line never executed. Rather than add a heavyweight test around `update_pred`, both callers now share `utils.minutes_since_midnight()`, which is tested directly — including both 2026 DST changeover days. That also removes the duplicated formula that made drift possible.

**3. `test_clock.py` midnight discontinuity — real flake.** At 23:59 `update_time()` floors to 23:55 while the later read is already 00:00: five minutes apart, 1435 by subtraction. Compares cyclic distance now.

**4. `time_now_str()` rendering yesterday's date during `calculate_yesterday()` — does not hold.**

Three separate reasons:

- `time_now_str()` renders `"%H:%M:%S"`. There is no date in it, and `midnight_utc - 1 day` has the same time-of-day, so the rewind cannot change its output.
- `calculate_yesterday()` fakes `minutes_now` to `0` for the same window, so that string rendered `00:00:00` both before and after this PR.
- The concurrency premise is wrong: `ComponentBase` does not expose `time_now_str`/`time_abs_str` at all, so no component thread can reach them. The cited caller (`inverter.py:1917`) runs on the main update loop, and `load_axle_slot` — the only other `base.time_abs_str` caller outside the base mixins — is called from `fetch.py` on that same thread.

Worth noting the related effect that *is* real and intended: `time_abs_str` is now anchored to `midnight_utc`, so yesterday's plan HTML renders yesterday's dates. It previously rendered them against today's naive midnight, i.e. a day out.

`./run_all --quick` green (314 tests), pre-commit clean.

PR #5086, opened 2026-09-14
fix(component_base): stop components reading calculate_yesterday's faked minutes_now

springfall2008:
_Posted by Claude on behalf of @springfall2008._

**Stacked on #5084 — merge that first.** Base is set to its branch, so this diff shows only the change itself.

The other half of #4804. #5054 stopped components seeing the `midnight_utc` that `calculate_yesterday()` rewinds. It fakes `minutes_now` to `0` in the same breath ([output.py:3291](https://github.com/springfall2008/batpred/blob/main/apps/predbat/output.py#L3291)), and that passthrough was deliberately left alone at the time: `base.minutes_now` was measured from a *different clock* than `now_utc`, so deriving it would have made components disagree with the planner wherever the container's timezone differed from the `timezone:` setting. #5084 removes that second clock, so the derivation is now exactly the calculation `update_time()` makes — through the same helper.

## Why this one matters more than the date

A rewound date is obviously wrong. `0` reads as a legitimate "just after midnight", so it fails quietly. It reaches 39 component-side call sites:

| File | Reads | Effect of `minutes_now = 0` |
|---|---|---|
| `octopus.py` | 19 | `end_minutes <= self.minutes_now` stops dropping expired dispatch slots; `start_offset = start_minutes - self.minutes_now` shifts every published offset by the real time of day |
| `load_ml_component.py` | 9 | `minutes_now % 30` slot alignment, history windows, ML feature payload |
| `solcast.py` | 5 | `minute_absolute = self.minutes_now - minute`, today's actual-PV alignment |
| `web.py` | 2 | SOC history chart range |
| `alphaess.py`, `deye.py`, `sunsynk.py` | 1 each | `_now_minutes()` / `_window_active()` pick the live TOU slot **while writing to a real inverter** — `0` reads as midnight |

`axle.py` also reads `base.minutes_now`, but `load_axle_slot()` is called from `fetch.py` on the base's own thread — it *should* see the faked values, and this does not touch it.

## Tests

`ComponentBase` gets a test that compares the derived value against `update_time()`'s on the real base rather than restating the formula, so the two cannot drift, plus the rewind case itself.

The AlphaESS doubles set `base.minutes_now` on a `MagicMock` base, so they were answering arithmetic with more mocks rather than failing. They get a real clock and a `set_mock_clock()` helper that moves it — writing `minutes_now` directly no longer does anything, which is the point.

Mutation-tested: reverting the property fails with `Faked base.minutes_now leaked into the component: 0`.

- `./run_all --quick` green (314 tests) under both a mismatched host timezone and `TZ=Europe/London`
- `./run_pre_commit` green


springfall2008:
_Posted by Claude on behalf of @springfall2008._

All three findings held up this time. Fixed in d4eb4b6.

**1. `minutes_now` read `base.now_utc` twice — real, and the sharpest of the three.**

Once directly, once through `self.midnight_utc`, which goes back to the same field. `update_time()` runs on the main thread while components run on theirs, so the two reads can straddle it. At a day boundary that subtracts the new day's midnight from the old day's timestamp: `(23:59 day 1) - (00:00 day 2)` is -60s, which `int()` truncates toward zero, so it returns **0** — reading as midnight rather than as an error. Exactly the failure mode this PR exists to remove, reintroduced by a second read.

Snapshotted once now. A base whose `now_utc` rolls into the next day between reads reproduces it deterministically without needing two threads to interleave; mutation-tested (`now_utc should be read once per call, was read 2 times`).

**2. Deye and Sunsynk doubles — real, same shape as AlphaESS.**

Confirmed: `MagicMock` answers `(mock - mock).total_seconds() / 60 / 5` with arithmetic of its own — `int()` of it is 1, so `_now_minutes()` silently returns 5. Their comment promised *"tests set this for time-aware control"* while a test setting `base.minutes_now` would have been ignored. Both doubles get a real clock and the same `set_mock_clock()` helper AlphaESS got.

**3. Incomplete restore in the new test — real.** It saved three of the seven fields `update_time()` writes, leaking this test's wall-clock state into later tests through the shared `PredBat`. Restores the full set now, matching what `tests/test_clock.py` already does.

`./run_all --quick` green (314 tests), pre-commit clean.

PR #5088, opened 2026-09-14
fix(solis): quota-aware retries with exponential backoff

mgazza:
## Problem

SolisCloud reports refusals in the JSON body of an HTTP 200 response rather than with a 429. `_with_retry` treated every error the same way: it grew the wait by 1.5x within a 30 s window with no limit on attempts, so one failing call could be sent up to eight times. Two refusals made this worse:

- **Exhausted quota.** Once an account hits its daily request quota, every call gets `R0000 "No authority too many request 200 times in 1DAYS"`. No retry can succeed until the quota window moves on, and each attempt counts against the next window as well. One install stayed stuck for more than 19 hours: SolisCloud kept refusing a second inverter on the account, which kept `automatic_config()` from completing.
- **Offline datalogger.** `B0115 "Sending failure, the current datalogger is offline or disconnected"` slept 10 s inside `_execute_request`, then went round the retry loop with that loop's own delays on top. A datalogger that stays offline was retried every call, and every one of those requests counted against the quota.

## Changes

1. **Backoff** (`fix(solis): back off retries exponentially with jitter and cap them`)
   - `_with_retry` waits 1, 2, 4, then 8 s (capped by `SOLIS_MAX_RETRY_DELAY`), with the upper half of each wait randomised.
   - A call makes at most `SOLIS_MAX_RETRIES` (4) retries, and is still bounded by `max_retry_time`.
   - `B0115` asks for its 10 s wait through `SolisAPIError.retry_after`, instead of sleeping inside the request.
2. **Pauses** (`fix(solis): pause requests after a quota refusal or a datalogger that stays offline`)
   - **Long-window quota refusals.** A refusal whose window outlasts a call's retries (for example "in 1DAYS") is not retried.
     - Requests to that endpoint for that inverter are refused without being sent for 15 min.
     - If the request is refused again right after the pause ends, the pause doubles, up to 1 h, and never lasts longer than the window.
     - A success lifts the pause. These refusals are recorded as `rate_limit`.
   - **`B0115`.** Still retried after 10 s, because most disconnects are brief. After `SOLIS_DATALOGGER_OFFLINE_REFUSALS` (6) refusals in a row with no success in between, that scope pauses for 5 min, doubling up to 1 h.
   - **Separate escalation.** Quota and offline pauses escalate independently. A refusal that arrives long after the last pause ended starts again from the first pause.
   - **Narrow scope.** Each pause covers one endpoint for one inverter, so other inverters on the same account keep polling normally.
3. **Review follow-ups**
   - `fix(solis): keep pause history for refusals inside a control response`: a response with code 0 no longer lifts the pause when one of its items was refused. Per-item quota refusals now escalate, and per-item `B0115`s pause.
   - `fix(solis): retry short-window and unnamed rate limits instead of pausing`: a window that reopens within a call ("2 times in 1SECONDS") is retried after waiting the window out. A "too many requests" message that names no window gets the normal backoff. Neither pauses.
   - `fix(solis): give up rather than cut a requested retry wait short`: a retry is never sent before its `retry_after` has passed.

## Relationship to #3788

The draft circuit breaker in #3788 covers similar ground with a single breaker for the whole account. This PR pauses only the refusing inverter and endpoint, so one offline datalogger doesn't stop polling of healthy inverters on the same account. It also tells a daily quota apart from short-window throttling.

## Not in this PR

- When inverter details can't be fetched during startup, `automatic_config()` configures nothing until the component's next startup retry, which can be up to 128 min later.
- If any one inverter fails to poll during startup, `run()` returns False, so the component stays in startup backoff even when the other inverters are fine.

## Tests

New tests in `tests/test_solis.py`, all registered in `run_solis_tests`:

- **Backoff and retries**
  - The wait sequence with the random part at both extremes, the retry cap, and the time limit.
  - A non-retryable error is raised at once.
  - A requested wait that no longer fits in the time window ends the call instead of being cut short.
  - An OAuth 401 still refreshes the token and retries.
- **Datalogger disconnects**
  - A brief `B0115` is retried after 10 s and leaves no pause. The real `read_and_write_cid` path also survives one refused control write.
  - A datalogger that stays offline pauses only after 6 refusals in a row. Nothing is sent while paused, the pause doubles when the next request is refused, and a success clears it.
- **Quota and throttling**
  - A daily-quota refusal pauses only that inverter's endpoint and isn't retried. When the pause ends, one request is sent; a second refusal doubles the pause, and a success lifts it.
  - Quota windows from the message: first and second pause lengths, the cap, a restart after a long gap, and separate escalation for quota and offline pauses.
  - Short-window ("in 1SECONDS") and unnamed "too many requests" refusals are retried, not paused.
- **Per-item refusals in control responses**
  - A per-item quota refusal pauses control writes and doubles on repeat.
  - A per-item `B0115` pauses after 6 refusals in a row.

Each of these 18 changes to the code breaks at least one test:

- Pauses and escalation:
  - Disabling the pause check
  - Removing the doubling
  - Never resetting the escalation
  - Ignoring the refusal kind
  - Not clearing the pause on success
  - Letting per-item refusals lift the pause
- `B0115` handling:
  - Pausing on the first `B0115`
  - Never pausing on `B0115`
  - Dropping `retry_after`
- Quota and throttling:
  - Not classifying per-item refusals
  - Not detecting the quota message
  - Pausing short windows
  - Not waiting a short window out
- Retry loop:
  - Retrying non-retryable errors
  - Removing the retry cap
  - Removing the random jitter
  - Cutting a requested wait short
  - Not retrying HTTP errors

Every commit passes `python3 unit_test.py -t solis`, and the pre-commit hooks (ruff, black, cspell) pass on each commit too. `-t components` and `-t mock_base` also pass on the first two commits. `datalogger` and `dataloggers` are added to the cspell dictionary.


springfall2008:
Duplicate of PR just merged

Issue #5079, opened 2026-09-14
Test infrastructure keeps producing fixture state-leak bugs


chalfontchubby:
(Posted by Rik; written by Claude.)

## Pattern

`unit_test.py` runs every registered test against one shared `PredBat`/HA fixture (`create_predbat()`), and this keeps producing the same class of bug: a test mutates shared state, restores it incompletely or unconditionally-to-a-default instead of to what was actually there, and leaks into whichever test runs next.

Recent instances, all found by review rather than by the leak actually being caught:

- #5052: `compare.fetch_rates()`/`annual._apply_rates()` mutate ~24 fields on the shared fixture (`dashboard_values`, `control_ledger`, window lists among them) when driven directly in a test - broke `optimise_levels` when tried.
- #4645: four new inverter-constructor tests restored `args` but not the HA interface's `dummy_items`, leaking 12 dummy entities into every later test in the module - the same gap existed in the pre-existing test Copilot cited as the model to follow.
- #5061 (PR closed, premise rejected, but the bug it fixed along the way was real): a test's cleanup reset `soc_kwh_history` to `{}` unconditionally instead of restoring whatever an earlier test had left there.
- #5053: two tests left `PREDBAT_APPS_FILE`/`args["chat"]` mutated for the rest of the process.

Each was fixed individually as found. The pattern itself isn't tracked anywhere, and there's no mechanism that would catch the next one before a maintainer or reviewer spots it by hand.

## Why it keeps happening

- The shared-fixture-across-all-tests design means any test that mutates instance state is a landmine for every test that runs after it, and the blast radius depends on run order.
- "Snapshot before mutating, restore what was actually there" (not `= {}` or a hardcoded default) is a convention, not something enforced - easy to get right for the field you're thinking about and miss another one a function touches incidentally.
- Nothing currently detects a leak automatically; it surfaces as an unrelated test failing later, which is confusing to debug and easy to misattribute (see `test_state_fingerprint` idea already on Rik's backlog, aimed at exactly this).

## Possible directions (not scoped, for discussion)

- A generic snapshot/restore helper other test modules could reuse instead of each writing its own (a couple of ad-hoc versions of this already exist, e.g. in `test_inverter.py`).
- A registry-level assertion that fixture state is unchanged after each test (or each module) - would catch a leak at its source test rather than downstream.
[...]

PR #5097, opened 2026-09-14
fix(tests): restore forecast_minutes/minutes_now leaked by two tests (#5079)

chalfontchubby:

**Test-only fix - no user-facing or production code changed.** Fixes two test files that leave shared fixture state mutated, which was making unrelated later tests fail depending on run order (see #5079, and #5096 for the shuffle-fuzzer tooling this was found with).

## Root cause

- `test_load_forecast_history.py`: mutates `forecast_minutes`/`minutes_now`/`car_charging_energy` on the shared `PredBat` fixture. Its own trailing "restore" block calls `fetch_config_options()`, which re-derives `forecast_minutes` from `apps.yaml` (`2880`) rather than putting back the fixture's overridden value (`1440`) - the bug is in the cleanup itself, not just a missing one.
- `test_rate_text_scan.py`: overwrites `minutes_now`, `end_record`, `forecast_minutes` (down to `9`), `rate_min`, `rate_max`, and the entire `rate_import` dict, and restores none of it.

Both now snapshot the fields they touch and restore them (`test_rate_text_scan` in a `try`/`finally`; `test_load_forecast_history`'s existing restore block extended), following the same pattern already used in `test_inverter.py`.

## Retires 8 of 27 known pairs, verified against all of them

One structural fix per file, not a patch per victim - re-ran every pair on `tools/shuffle_fuzz_known_pairs.txt` (not just the two culprits' own known victims) against this branch:

```
load_forecast_history -> model_kernel
load_forecast_history -> iboost_smart
load_forecast_history -> model
load_forecast_history -> optimise_windows_kernel
rate_text_scan -> cloud_modulation
rate_text_scan -> optimise_swap_charge
rate_text_scan -> window_sort
manual_soc -> cloud_modulation
```

19 pairs remain (`kernel_static_cache`, `plot`, `octopus_url`, `metric_battery_cycle_fractional`, `multi_car_iog`, `calculate_yesterday`, and several one-off culprits) - will keep chipping away and posting to #5079 as they're understood.

## Test plan

- [x] `./run_all --quick` (full suite, registry order) still passes
- [x] All 27 known ordering-dependent pairs re-verified individually against this branch; the 8 above now pass, the rest fail identically to before (unaffected, as expected)


chalfontchubby:
Moved to draft while the approach is decided.

These fix the leaks one test at a time, by naming the attributes each test mutates and restoring them. That has proven error-prone in review already (the first version of #5097 missed 13 fields, and an audit then found 7 more), and it scales with the number of tests rather than fixing the cause.

A cheaper alternative looks viable: build a fresh fixture per test in the runner, so leaked state cannot carry over at all. `create_predbat()` measures at ~22ms, so roughly +7s across 316 tests against a ~76s `--quick` run - about +9%, rather than the prohibitive cost this was assumed to have. That would make per-test restore blocks unnecessary and this PR closeable.

Holding here until that is settled - and until there is maintainer agreement that order-dependent tests are worth fixing in the first place (#5079 has had no maintainer response yet).

chalfontchubby:
Closing - keyhole fixing individual tests is not worth it with a holistic fix available

PR #5098, opened 2026-09-14
fix(tests): restore now_utc/midnight_utc leaked by test_octopus_url (#5079)

chalfontchubby:

**Test-only fix - no user-facing or production code changed.**

## What and why

`test_edf_freephase_dynamic_url` (in `test_octopus_url.py`) pins `midnight_utc`/`now_utc` to the test data's date (2025-12-17) and never restores them - already flagged by the code's own comment, added in #5054: *"this pin is left behind for later test modules"*.

That #5054 change also made this leak more consequential than it used to be: `ComponentBase.midnight_utc` now derives from `now_utc` rather than being read independently, so any later test that reads `midnight_utc` through a component now inherits the wrong value too, not just direct `now_utc` readers.

Snapshot both fields at the top of the function and restore them at the end - same pattern as #5097.

## Does not retire `octopus_url -> plan_persistence`

Worth being explicit about this rather than implying the fix closes that pair: confirmed by testing the pair after this fix, it still fails. That pair's real cause (found separately, not by this branch) is `storage.py` checking a saved plan's expiry against real wall-clock time rather than the fixture's `now_utc` - a genuine time-of-day-dependent bug, unrelated to what ran before `plan_persistence`. That's separate follow-up work, tracked on #5079.

This commit is still worth landing on its own regardless: the leak is real, and #5054 widened its blast radius.

## Test plan

- [x] `test_octopus_url` still passes standalone
- [x] `./run_all --quick` (full suite) still passes
- [x] Confirmed via direct repro (`--test octopus_url --test plan_persistence`) that this fix does not (and isn't claimed to) retire that pair


chalfontchubby:
Moved to draft while the approach is decided.

These fix the leaks one test at a time, by naming the attributes each test mutates and restoring them. That has proven error-prone in review already (the first version of #5097 missed 13 fields, and an audit then found 7 more), and it scales with the number of tests rather than fixing the cause.

A cheaper alternative looks viable: build a fresh fixture per test in the runner, so leaked state cannot carry over at all. `create_predbat()` measures at ~22ms, so roughly +7s across 316 tests against a ~76s `--quick` run - about +9%, rather than the prohibitive cost this was assumed to have. That would make per-test restore blocks unnecessary and this PR closeable.

Holding here until that is settled - and until there is maintainer agreement that order-dependent tests are worth fixing in the first place (#5079 has had no maintainer response yet).

chalfontchubby:
Closing - keyhole fixing individual tests is not worth it with a holistic fix available

PR #5102, opened 2026-09-14
test: give each test its own PredBat rather than sharing one (#5079)

chalfontchubby:

**Test infrastructure only - no user-facing or product code changed.** One six-line change in `apps/predbat/unit_test.py`.

## Problem

Every registered test runs against a single `PredBat` built once at startup. A test that mutates instance state and doesn't restore it therefore breaks whichever unrelated test happens to run next - depending only on the order they ran in. That's the pattern tracked in #5079, and it has produced a long series of confusing failures where the test that reports the error is not the test at fault.

Until now each one was fixed by hand-writing a snapshot/restore block naming the fields that particular test touched. That approach scales with the number of tests and has proven error-prone: the first attempt at one such fix missed 13 fields, and a follow-up audit found 7 more.

## Change

Build the fixture per test instead, so leaked state cannot carry over at all.

## Measurements

**Known-bad orderings** (the 53 (culprit, victim) pairs collected so far by the fuzzer in #5096):

| | pairs failing |
|---|---|
| before | 53 / 53 |
| after | **0 / 53** |

(The two involving `plan_persistence` need #5100 as well - that test also fails on its own, for an unrelated product bug.)

**Runtime** - the assumption that rebuilding would be too expensive turns out not to hold. `create_predbat()` is ~22ms, and the suite's cost is dominated by the prediction work inside the tests, not fixture setup:

| | full suite | quick suite |
|---|---|---|
| shared fixture | 174s, 316 passed | 83s, 312 passed |
| per-test fixture | 173s, 316 passed | 80s, 312 passed |

No measurable difference either way.

**The tests themselves are sound**: all 316 pass when run individually, each in its own process. So every ordering failure found was shared-fixture contamination, not a test quietly depending on an earlier one to set it up. (The single exception was `plan_persistence`, which failed alone because of the product bug fixed in #5100 - found precisely *because* isolation was checked.)

## Consequences

- Supersedes the per-test restore patches in #5097 and #5098, which can be closed.
- Tests may now rely on a clean fixture rather than on whatever ran before them.

## Test plan

- [x] Full suite (including slow tests): 316 passed, 0 failed
- [x] All 53 known-bad orderings re-run individually: 0 failing
- [x] All 316 tests run individually in isolation: all pass


chalfontchubby:
CI red here is not this change - it is a pre-existing bug this branch is missing the fix for.

The run failed at **22:13 UTC**, and `plan_persistence` Test 1 reported the plan it had just saved could not be loaded back. That is #5100: `save_plan()` computes its 8-hour expiry from `self.now_utc` (the fixture clock, pinned to midday) while `storage.load()` checks it against the real clock. Midday + 8h = 20:00, so from 20:00 UTC onward every saved plan is already expired when written. The same run on `main` at that hour would fail identically.

There is a second, opposite instance of the same mistake in Test 5, which builds an "already expired" timestamp the same way and so fails *before* about 11:00 local - fixed in #5104.

`plan_persistence` therefore only passes at any hour with both clock fixes present. Verified locally with all three stacked:

```
quick suite: 73s exit=0 passed=312 failed=0
```

Will restack this on top of #5100 and #5104 so CI here is meaningful.

PR #5104, opened 2026-09-15
test: build the expired-plan timestamp from the real clock (#5079)

chalfontchubby:

**Test-only fix.** One line in `test_plan_persistence.py`.

## Problem

Test 5 stores a plan whose expiry is in the past and asserts `load_plan()` rejects it as stale. It built that expiry as `my_predbat.now_utc - timedelta(hours=1)` - but `storage.load()` checks expiry against the **real** clock, and the fixture pins `now_utc` to midday.

So for any run before about 11:00 local, "an hour before the fixture's midday" is still in the future. The plan isn't expired, `load_plan()` correctly loads it, and the test fails:

```
FAILED: plan_valid should remain False for expired stored plan
FAILED: charge_window_best should remain empty for expired stored plan
```

This is the mirror image of the product bug in #5100, where `save_plan()` computed a *future* expiry from the same fixture clock and every saved plan was born expired after ~19:00. Both come from mixing Predbat's clock with the real one across the storage boundary. The rule either way: anything `storage.load()` will judge has to be built from the clock it judges against.

## How it was found

An overnight run of the fuzzer (#5096) reported it as five separate "ordering" pairs with unrelated culprits (`solax`, `octopus_rate_limit`, `kraken_auth`, `minute_data_no_smoothing_forward`, `hainterface_api`). None reproduced - the bisector had converged on whichever test happened to precede it, because the failure was by time of day rather than by order. #5096 has been updated to detect and reject that case rather than record it as a pair.

## Test plan

- [x] `plan_persistence` passes at 08:49 local, which is inside the window that previously failed
- [x] `teslemetry` also passes (it appeared in the same fuzzer output for the same mislabelling reason, but needed no change)


chalfontchubby:
Folded into #5100. Both are the same mistake - a timestamp built from the fixture clock (`my_predbat.now_utc`, pinned to midday) then judged by `storage.load()` against the real one. #5100 was the product side (expiry in the future, so plans were born expired after 20:00 UTC); this was the test side (expiry in the past, so Test 5 failed before 11:00 local).

Keeping them together means `plan_persistence` passes at any hour, which neither achieved alone - #5100 could not be green before 11:00, and this could not be green after 20:00.

PR #5108, opened 2026-09-15
Move the off option to the top of the manual override dropdowns

### LINKED ISSUE #5105: FR: move 'off' option to start of the select.predbat_manual_xxx controls
**Is your feature request related to a problem? **
A minor usability improvement

I use the Predbat select.manual_xxx controls when I want to override the Predbat, in this case I am wanting to charge the battery up a bit more in anticipation of a power down event tonight.

Selecting and adding new slots via the Predbat plan or via the selector is easy.

But if I want to cancel all the overrides, the 'off' option is right at the bottom of the list, which is 2 days long now so takes a while to scroll down.
image of options, which Please describe.**

Move the 'off' option to be the top entry in each of the lists so it's easier to select.

springfall2008:
This is an automated draft PR generated from issue #5105 — a maintainer should review it before merging.

Fixes #5105

## Summary

`off` was appended last in all three manual-override dropdown builders, so on a 48-hour list (~576 entries at 5-minute granularity) cancelling every override meant scrolling to the very bottom. It is now the first entry in `manual_times()`, `manual_rates()` and `api_select_update()` (`apps/predbat/userinterface.py`), which covers every `select.predbat_manual_xxx` control and the same lists as rendered in the web config page. The docs line that described `off` as being "at the bottom of the list" is updated to match.

Selection handling is unaffected: `manual_select()` / `api_select()` match on the value `"off"`, never on its position, and nothing indexes `item["options"]` positionally.

## Testing

- `./run_pre_commit` from `coverage/` — all hooks passed, and the quick suite it runs afterwards passed in full (4 slow tests skipped).
- `tools/triage_test.sh manual_select` with the `userinterface.py` change stashed — **FAILS**, flagging all three builders (`ERROR: T10 Expected off to be the first option of manual_charge, got Tue 12:00` and the same for `manual_import_rates` and `manual_api`, plus T13 after a slot is selected).
- `tools/triage_test.sh manual_select` with the change applied — **PASSES**.

The new `test_off_option_first()` in `apps/predbat/tests/test_manual_select.py` exercises one config item per builder, checks `off` is first and appears exactly once, and then checks it stays first once a slot is selected and still clears the selection.

## Notes

- `impact()` reports HIGH risk for all three methods, because they sit on `initialize` / `update_time_loop` / `run_time_loop`. That reflects where they live rather than this change: the return values and the state written by `expose_config()` are untouched, only the order of the presentation list `item["options"]` moves. `detect_changes()` shows the three methods and the test as the only changed symbols.
- The three builders each end with the same `if values not in time_values / item["options"] = / expose_config` tail. That duplication predates this change and I left it alone rather than widening the diff into a refactor of code on the main loop.
- While testing I noticed the dropdowns also carry an empty-string option (the `values` entry appended when nothing is selected — it renders as "None" in the web config page). That is pre-existing and outside what this ticket asked for, so it is untouched; with this change it now sits second in the list for `manual_api`.
- The option-picking loop in the existing `run_test_manual_select()` was extracted into `first_future_option()` so the new test could reuse it rather than copy it — behaviour is unchanged (prefer a non-today day label, fall back to any selectable option).


springfall2008:
Automated review failed to complete for this PR - see the triage bot's logs for details. The run exited cleanly but posted nothing, so the review step itself did not complete. Not retrying automatically; remove `BOT_FAILED` and re-add `BOT_REVIEW` to try again.

Issue #4232, opened 2026-07-11
Excessive GE EMS writes?


markbw999:
I have just been having a conversation with @andrewmusty about the number writes we are seeing to our EMSs. My setup is an EMS and two AC coupled inverters. If I look at the GivTCP writes stats the EMS is running at 76k total and about 30k per month whereas the inverters are only about 900 each in total.

I posted about this on the GivTCP FB group and got the following from Paul Landregan - "Without spilling the hardware make up. the Settings are stored in a I2C 8K EEPROM with 1M writes. As well as RAM buffering."

If that's the limit, my EMS only has three years before it is fried at this rate. Andrew has noticed this - "With local EMS control it seems to spam the GivTCP log with: GivTCP2 - write - [INFO] - Setting Charge Target 1 was a success every 10 seconds or so."

Is predbat doing this? I have attached my debug and log files...

[predbat.log](https://github.com/user-attachments/files/29927402/predbat.log)
[predbat_debug.yaml.txt](https://github.com/user-attachments/files/29927401/predbat_debug.yaml.txt)

andrewmusty:
I've been using the GE cloud plugin since I noticed the spamming and todays writes so far is only 70 ish.

When using Givtcp this is an example of the Givtcp log file. It seem to be about every 10 sec

2026-06-28 14:13:13,792 - GivTCP2 - write - [INFO] - Setting Charge Target 1 was a success
2026-06-28 14:13:24,998 - GivTCP2 - write - [INFO] - Setting Charge Target 1 was a success
2026-06-28 14:13:33,964 - GivTCP2 - write - [INFO] - Setting Charge Target 1 was a success
2026-06-28 14:13:44,075 - GivTCP2 - write - [INFO] - Setting Charge Target 1 was a success
2026-06-28 14:13:54,186 - GivTCP2 - write - [INFO] - Setting Charge Target 1 was a success
2026-06-28 14:14:04,297 - GivTCP2 - write - [INFO] - Setting Charge Target 1 was a success

2026-06-28 14:15:58,451 - GivTCP2 - write - [INFO] - Setting Charge Target 1 was a success
2026-06-28 14:16:08,572 - GivTCP2 - write - [INFO] - Setting Charge Target 1 was a success
2026-06-28 14:16:18,775 - GivTCP2 - write - [INFO] - Setting Charge Target 1 was a success
2026-06-28 14:16:32,397 - GivTCP2 - write - [INFO] - Setting Charge Target 1 was a success
[...]

Issue #3781, opened 2026-04-11
Predbat - Givenergy EMS - GivTCP


andrewmusty:
If anyone missed my Facebook post — I’ve managed to get full local control working on my GivEnergy EMS using Predbat and GivTCP.

Happy to help anyone who wants to set this up, just reply here.

poolski:
It might be useful to write up your setup and post here regardless of whether folks respond or not, for posterity!

markbw999:
I just took the plunge and implemented Andrew's solution. So far it's working like a charm!

poolski:
Could you link the post or similar?

markbw999:
Here is what Andrew sent me:

```1.	Stop GivTCP if it is running.
2.	Go to MQTT Explorer and delete all GivEnergy topics.
3.	Start GivTCP and wait a few minutes for it to initialise. Then go to:
http://ipaddress:8099/config.html
4.	Make sure you set both inverters to:
“Inverter – only report battery data when this inverter is connected to EMS.”
5.	Complete the rest of the GivTCP configuration, then save and restart.
Note: GivTCP created givtcp, givtcp2, and givtcp3 entities for me. My EMS is givtcp2, and my inverters are givtcp and givtcp3. I’m not sure why it did this, and it may not happen for you.
6.	Copy the config at the bottom of this email into your Predbat apps.yaml file (make sure you back yours up first). Then update the values inside the quotes with your details:
o	EMS serial and EMS Giv ID
o	Inverter 1 serial and Giv ID
o	Inverter 2 serial and Giv ID
It doesn’t matter which inverter you assign as 1 or 2. For example:
ems: 'ems255015'
emsgiv: 'givtcp2'
7.	I have custom load_today and pv_today sensors, so you’ll need to adjust those to match your setup.
8.	Start Predbat and, hopefully, it runs without any errors.
________________________________________
Config below
ems: 'ems serial'
emsgiv: 'ems giv id'

inv1: 'inv1 serial'
inv1giv: 'inv1 giv id'

inv2: 'inv2 serial'
inv2giv: 'inv2 giv id'

load_filter_threshold: 30

num_inverters: 2

inverter_type:
  - 'GEE'
  - 'GEE'

load_today:
  - sensor.house_load_today

import_today:
  - sensor.{emsgiv}_{ems}_import_energy_today_kwh

export_today:
  - sensor.{emsgiv}_{ems}_export_energy_today_kwh

pv_today:
  - sensor.total_solar_today

[...]

PR #5109, opened 2026-09-15
fix(gecloud): re-read EMS battery inverter settings and warn on a slot 1 override

### LINKED ISSUE #5103: GE Cloud EMS: battery inverter settings are never re-read after startup, so a changed AC3 slot silently stops evening discharge
## Summary

When GE Cloud auto-config finds a Plant EMS, Predbat stops re-reading the battery inverters' (AC3) settings after startup. If an inverter's own schedule changes later, Predbat can't see it, and nothing warns. In one field case, a single AC3 whose DC discharge slot 1 ended at 19:00 made the whole plant stop discharging at 19:00 every evening. Predbat kept reporting "Exporting" or "Demand" and the house imported at peak rate.

## What happened

The plant is a Plant EMS with two GIV-AC3.0 inverters, controlled through GE Cloud (inverter type `GEE`, auto-configured).

- **Before the change:** for several nights both AC3s discharged all evening, following the house load.
- **The change:** one AC3's `DC Discharge 1 End Time` became **19:00**. Start stayed 00:00; the other AC3 stayed 00:00–23:59. It happened on the evening a VPP export session ended at 19:00. That timing is suspicious but unproven; nothing in Predbat recorded the write.
- **From that evening:** both AC3s stopped discharging at 19:00 every night. Predbat's EMS writes (discharge/idle slot and export slot 1) were read back correctly from GE Cloud, but the batteries sat at 0 W until the next day. Samples after 19:00 with a battery discharging above 150 W dropped from about 55 per inverter per night to 1–9.
- **What Predbat showed:** plan valid and status normal. The control interference check reported nothing.

Across 13 EMS plants we could inspect, every other AC3 has DC discharge slot 1 at 00:00–23:59. This matches the advice in #3781 that each inverter's slot 1 charge and discharge times should be 00:00–23:59 "so they will not override commands from the EMS".

(Separately, forced export also failed on this plant before the slot changed, while load-following still worked. This issue doesn't explain that part.)

## Why Predbat can't see it

On current `main` (df6c4b9fb7a5671301aa3aadab82c64f2c9839c3):

- When an EMS is found, `polling_mode` is turned off:
  https://github.com/springfall2008/batpred/blob/df6c4b9fb7a5671301aa3aadab82c64f2c9839c3/apps/predbat/gecloud.py#L1702-L1708
- The periodic settings refresh then re-reads only the EMS (or gateway) device. The battery inverters are read once, when `seconds == 0`:
  https://github.com/springfall2008/batpred/blob/df6c4b9fb7a5671301aa3aadab82c64f2c9839c3/apps/predbat/gecloud.py#L1805-L1808

Under EMS auto-config the AC3 entities are still bound, and Predbat writes some of them: `inverter_mode` (eco), `scheduled_discharge_enable`, `charge_rate` and `discharge_rate`. Their schedule registers, though, stay frozen at the startup snapshot. A change made by an installer, the GivEnergy app or a third-party service is invisible until the next restart, and even then nothing flags it.

## Options

1. **Re-read battery inverter settings on EMS plants at a low cadence** (e.g. hourly), so the published entities and the control ledger reflect reality. This costs more GE Cloud reads per inverter; see #4232 for EMS API volume concerns.
2. **Warn** (log plus status/alert) when an inverter under EMS control has a charge or DC-discharge slot 1 other than 00:00–23:59, naming the inverter and the value.
3. **Optionally correct it:** write slot 1 back to 00:00–23:59 on inverters under EMS control, gated by a config switch. This adds writes, and could fight a third-party controller that changes those slots on purpose.

(2) looks like the cheapest useful step; (1) makes (2) work between restarts.

springfall2008:
This is an automated draft PR generated from issue #5103 — a maintainer should review it before merging.

Fixes #5103

## Summary

Under GE Cloud EMS auto-config, finding an EMS sets `polling_mode = False`, and the 10-minute settings refresh then gated each battery inverter's `/settings` re-read on `seconds == 0`. The AC3s' registers were therefore snapshotted exactly once per process start and published stale thereafter, so a DC discharge slot 1 changed behind Predbat's back overrode the EMS and stopped the plant discharging every evening with nothing in the log or the status.

This implements the reporter's options (1) and (2), which the triage comment picked out as the cheapest useful pair:

- **Re-read on a slow cadence.** Battery inverters under EMS control are now re-read hourly (`SETTINGS_SLOW_REFRESH_SECONDS`) rather than once at startup, so the published entities and the control ledger track reality between restarts. That is 1 extra `/settings` read per inverter per hour — deliberately slow, given the EMS API volume concerns in #4232. Nothing changes for non-EMS plants: `polling_mode` short-circuits the gate there exactly as before.
- **Warn on a slot 1 override.** After each refresh, any battery inverter whose own charge or DC discharge slot 1 is not `00:00-23:59` — the window #3781 asks for precisely so the inverter never overrides the EMS — is reported to the log and to the Predbat status, naming the inverter and the offending value. Raised once per episode (following the auth-denied precedent a few lines up) so a standing misconfiguration does not inflate `error_count` every hour, and re-armed once a clean read comes back.

Option (3), writing slot 1 back to `00:00-23:59`, is deliberately **not** implemented — see Notes.

One extra fix on the same line: the refresh gate now keys off `first` rather than `seconds == 0`. `first` stays `True` across startup backoff retries while `seconds` keeps climbing, so on a plant whose first successful `run()` landed on a retry the battery inverters were previously never read *at all*, not even once.

## Testing

`coverage/run_pre_commit` (pre-commit hooks + `./run_all --quick`): all hooks pass, all tests pass. Spelling verified separately with `npx cspell` against the `en-gb` dictionary (the local hook is a no-op without it) — 0 issues.

`./run_all --test ge_cloud`: **96 passed, 0 failed**, including the two new sub-tests `ems_slot_overrides` and `ems_settings_reread`.

The new test fails without the fix, in two stages:

- Stashing the whole source change fails the module at import (`ImportError: cannot import name 'SETTINGS_SLOW_REFRESH_SECONDS'`), which proves the red run but says nothing about the assertions.
- So I also ran a targeted red: with the helpers left importable, I reverted **only** the `run()` gate back to `seconds == 0` and dropped the check call. Result: **95 passed, 1 failed**, failing on exactly the reported defect — `ERROR: at the slow refresh cadence the battery inverter should be re-read, got ['ems001']`. Restoring the gate returns it to 96/96.

What the tests cover: the hourly-vs-10-minute cadence under EMS; the 19:00 end time from the field report being detected, logged and raised as a status; a zeroed `00:00-00:00` slot 1 also counting as an override (Predbat enables the per-inverter discharge switch under EMS, so that window bars discharge entirely); `HH:MM:SS` and unpadded times normalising rather than false-positiving; null/`unknown`/non-time register values being ignored; the once-per-episode status and its re-arming after a clean read; and that neither the EMS device itself nor a non-EMS plant is ever subject to the check.

## Notes

- **Option (3) not implemented.** Writing slot 1 back to `00:00-23:59` adds writes and, as the issue says, could fight a third-party controller that changes those slots on purpose. `enable_default_options()` is where it would naturally live (it already zeroes slots 2–10 every 24h and deliberately leaves slot 1 alone), gated behind a config switch. Left for a maintainer decision rather than taken unilaterally.
- **Duplicated-ish logic, deliberately not shared.** `enable_default_options()` scans the same time registers for slots 2–10, but it *writes* and compares raw values, while this path *reports* and normalises first. Folding the new `normalise_register_time()` into that comparison would change write behaviour on values like `0:00`, which is outside what this ticket asks for, so the two are left separate.
- **Status message length.** `record_status()` clamps the sensor state to 255 characters. One inverter with one bad slot is well under; an inverter with all four slot 1 registers wrong would be clamped in the sensor state, with the full text still in the log. That is the documented behaviour of that clamp, not new here.
- **Blast radius established by search, not GitNexus.** The GitNexus MCP tools were not available in this session, so `impact()` could not be run. By search: the only modified existing symbol is `GECloudDirect.run`, whose sole production caller is `ComponentBase.start` (`component_base.py:264`) and whose signature is unchanged — only the internal refresh cadence moved. Every new symbol (`SETTINGS_SLOW_REFRESH_SECONDS`, `EMS_SLOT_FULL_DAY_START/END`, `EMS_SLOT_FAMILIES`, `normalise_register_time`, `find_ems_slot_overrides`, `check_ems_inverter_slots`, `ems_slot_warned`) appears only in `gecloud.py` and `test_ge_cloud.py`, with no name collisions elsewhere in the tree.
- `tools/debug-journal.md`'s GE Cloud row informed the approach — in particular that the settings list is fetched once per process start and that the mock in `test_ge_cloud.py` does not mirror the real `initialize()` unless kept in step, which is why `ems_slot_warned` was added to `MockGECloudDirect` alongside the real one.
- The issue's separate observation that forced export also failed on this plant *before* the slot changed is not addressed here — as the reporter says, this defect does not explain that part.


springfall2008:
_Automated comment from the triage bot._

Automated cleanup pass. No review feedback was outstanding (no top-level comments, no inline review threads) and both checks — `pre-commit` and `kernel-binaries` — were already green, so nothing was implemented. The branch was, however, well behind `main`, and merging it was not a no-op.

**Merge of `origin/main` (one conflict, resolved).** `apps/predbat/tests/test_ge_cloud.py` conflicted in `MockGECloudDirect.__init__`: this branch adds `self.ems_slot_warned = set()` there, and `main` (#5206, discovery phase 1) adds `self._discovery_report = None` at the same spot. Both are mirrors of real attributes the mock has to carry — `ems_slot_warned` of `GECloudDirect.__init__` (`gecloud.py:538`), `_discovery_report` of the class attribute on `ComponentBase` (`component_base.py:66`) — so the resolution keeps both lines rather than either side wholesale. `gecloud.py` and `docs/inverter-setup.md` auto-merged.

**One interaction checked, and it is inert.** `main` added a `refresh_discovery()` call inside the same 10-minute block this PR changes, with a comment placing it "after settings (populated for every device in `self.device_list`, above)". Under EMS this PR now refreshes battery inverters hourly rather than every 10 minutes, which weakens that stated ordering — but `build_discovery()` reads `self.devices_dict` and `self.info`, never `self.settings` (checked across the whole method body), so the catalogue is unaffected either way. Nothing changed for it. Worth noting that this PR makes those settings *fresher* than before, not staler: on `main` the same devices were only read at `seconds == 0`.

**Quality gate, after the merge.** `coverage/run_pre_commit` — all hooks pass, quick suite passes (4 slow tests skipped). `./run_all --test ge_cloud` — **106 passed, 0 failed**, up from the 96 in the PR description because `main` brought in 10 more GE Cloud tests, mostly the discovery reporters; the two tests this PR adds (`ems_slot_overrides`, `ems_settings_reread`) are among the passes.

Only the merge commit was pushed. The substantive change is still the one a maintainer needs to review — in particular the deliberate omission of option (3), writing slot 1 back to `00:00-23:59`, which is called out in the PR description as a maintainer decision rather than something taken unilaterally.

PR #5112, opened 2026-09-16
Run ML training off the event loop and let a shutdown abandon it

### LINKED ISSUE #5075: ML training blocks the event loop, so a shutdown during training prevents restart
*(Written by Claude, working with @chalfontchubby.)*

## Summary

ML training runs synchronously inside an `async def`, so a shutdown that lands during training blocks the asyncio event loop until training finishes. Predbat logs `Predbat terminated`, every component reports stopped — and then the process stays alive for several more minutes, preventing the addon from restarting.

Hit live on 2026-09-13 while redeploying: the addon appeared not to start. It eventually came back after a manual restart.

## Evidence

Shutdown completed at 21:09:11, yet the training thread kept logging for nearly 5 more minutes:

```
21:09:11.069425: Stopping load_ml interface
21:09:11.172466: Stopped load_ml interface
...
21:09:11.896276: Predbat terminated
21:09:21.054587: ML Predictor: Epoch 8/30: ...
21:11:35.788178: ML Predictor: Early stopping at epoch 13
21:11:40.775125: ML Predictor: Curriculum pass 3/5: window=60.0 days
21:13:58.807928: ML Predictor: Epoch 5/30: ...
```

`load_ml` reports "Stopped" at 21:09:11, then runs two further curriculum passes. No startup banner appears until a manual restart at 21:14:51.

The same log contains a control case. An earlier restart at 19:12, with no training in flight, went `Stopping Predbat` → `Predbat terminated` → `Startup predbat` in **29 seconds**. The 21:09 one took **5m 40s**. Same shutdown path; the variable is whether training was running.

## Root cause

`LoadMLComponent._do_training()` (`apps/predbat/load_ml_component.py`) awaits a synchronous, CPU-bound call with no executor:

```python
val_mae = self.predictor.train_curriculum(...)
```

There is no `run_in_executor` or `asyncio.to_thread` anywhere in the module. A synchronous call inside a coroutine holds the event loop for its entire duration, so nothing else — including the shutdown path — can progress.

The hazard was half-identified already. The comment above the call reads:

```python
# Snapshot data under the lock so we can release it before the CPU-bound
# training call.  predictor.train() can take 30-120 seconds on slow hardware
# and must NOT run while holding data_lock or the asyncio event loop will freeze.
```

...followed by `# Lock released - event loop is free during training`. Releasing the lock does not release the loop; the second comment is not true as written.

Two compounding factors:

1. **`api_stop` is never checked on the training path.** `ComponentBase.stop()` sets `self.api_stop = True` as a cooperative signal, but `load_ml_component.py` contains no reference to it, so training would not abort even if the loop were free.
2. **Curriculum multiplies the exposure.** `train_curriculum()` runs up to 5 passes of up to 30 epochs. The window is minutes, not the "30-120 seconds" the comment anticipates — the observed 21:09 run was still going 4m 47s after shutdown.

## Suggested fix

Two parts, both small:

1. Offload the call: `val_mae = await asyncio.to_thread(self.predictor.train_curriculum, ...)`. That frees the loop so shutdown can proceed.
2. Optionally, make training abortable on shutdown. There is a natural hook — `train_curriculum()` already calls `progress_callback` once per epoch — but note the existing comment there deliberately rules this out: *"Guarded because a progress hook must never be able to abort an hour of training."* That guard is about a **failing** hook not killing a run, so a separate, explicit stop check would not contradict it; still, this is a design call for whoever owns the module, and part 1 does not depend on it.

Part 1 alone fixes the reported symptom: the loop is freed, shutdown completes, and the addon restarts while training finishes in its thread. Part 2 would additionally stop a shutdown having to wait out an in-flight curriculum, if that is judged worth the complexity.

## Impact

Any install with `load_ml_enable` on. Restarts and upgrades that coincide with a training window appear to hang, which reads as a failed deploy rather than a slow shutdown. Not data-affecting — the model and history files are unharmed, and a restart clears it.

springfall2008:
This is an automated draft PR generated from issue #5075 — a maintainer should review it before merging.

Fixes #5075

## Summary

`LoadMLComponent._do_training()` called `predictor.train_curriculum()` synchronously from inside an `async def`, so a curriculum run (many passes of many epochs, minutes long) pinned the component's event loop for its whole duration. Two changes:

1. **Off the loop.** The curriculum now runs via `asyncio.to_thread(...)`. With the loop free, `api_stop` can be noticed and the `run_timeout` watchdog in `component_base.py:266` can actually fire — as the triage comment noted, `asyncio.wait_for` could never evaluate its timeout while the loop was blocked, so the existing safety net was inert for exactly this case.
2. **Abortable.** `train()` and `train_curriculum()` take a new `stop_callback` (defaulted `None`, so every existing caller is unchanged), checked at the top of each epoch and between curriculum passes. `_do_training` wires it to the component's `api_stop` via a new `training_stop_requested()`, and reads the resulting `None` as *abandoned* rather than *failed* — no partial model is published or saved, and `last_train_time` is not stamped, so the retrain interval does not swallow the next real run.

### Why part 2 as well, not just the `to_thread` the issue asks for

The issue expected part 1 alone to fix the symptom. Reading the shutdown path, it does not. Components are not tasks on a shared loop: `components.py:848` starts each one through `base.create_task()`, which in `hass.py:224` spawns a real `threading.Thread` running `asyncio.run(component.start())` — so each component has its own loop in its own non-daemon thread. `stop_all()` (`hass.py:238`) calls `terminate()` (which logs "Predbat terminated") and *then* joins every component thread with a 5-minute timeout. Whether the curriculum blocks that thread's loop or blocks a worker it is awaiting, the thread stays alive either way, so the join still waits — and after the join times out, the interpreter's own non-daemon-thread shutdown waits again. That matches the report exactly: "Predbat terminated" at 21:09:11, training still logging at 21:13:58, no restart until a manual one at 21:14:51.

So the cooperative stop check is the part that actually lets the thread exit; `to_thread` is what makes the loop (and the watchdog) responsive while it runs. `run()` also returns early after an abandoned run rather than doing the post-training re-fetch and prediction cycle, which would otherwise add their own delay to the stop being answered.

A hook that raises is treated as "do not stop" and logged, mirroring the existing rule for `progress_callback` — the guard at `load_predictor.py:1385` is about a *failing* hook never killing a run, which a deliberate stop request does not contradict.

## Testing

`coverage/run_pre_commit` — all hooks pass, bundled suite green (4 slow tests skipped).

Three new sub-tests in `tests/test_load_ml.py`, all registered in that module's `sub_tests` list:

- `training_stop_callback` — `train()` stops at the epoch the hook first trips (2 epochs complete, returns `None`); a raising hook does not abort a run; `train_curriculum()` starts no pass at all once the hook has tripped; omitting the hook is unchanged.
- `training_off_event_loop` — drives the real `_do_training` with a trainer that blocks until a task scheduled on the loop releases it. Deterministic rather than timed: a loop that cannot run at all fails on the fake trainer's timeout, not on a stopwatch.
- `training_abandoned_on_stop` — the component must pass a stop hook that tracks `api_stop`, and must not publish, stamp or count an abandoned run.

Red/green, source stashed with `git stash push -u` and the test file left in place:

- Without the fix: `load_ml` **fails**, 34/37, with all three new tests red for the right reasons — `LoadPredictor.train() got an unexpected keyword argument 'stop_callback'`; "the event loop made no progress while training ran"; "`_do_training` must pass a stop_callback to train_curriculum".
- With the fix: `load_ml` **passes**, 37/37.

Also re-ran the adjacent modules that exercise the changed trainer, all green: `load_ml_rollout`, `ml_memory`, `ml_training_perf`.

Blast radius via GitNexus `impact()`: `train_curriculum` LOW (4 callers, all tests), `_do_training` LOW, `train` MEDIUM but all 17 affected symbols are in tests — and every signature change is a defaulted keyword argument, so no existing call site changes. `detect_changes()` reports low risk, 3 files, no affected execution flows.

## Notes

- **Deliberate tidy-up inside the code being changed.** `_do_training` had two near-identical 16-line `train_curriculum(...)` call sites differing only in `curriculum_step_days` and `max_intermediate_passes`. Rather than duplicate the new `to_thread` + `stop_callback` wiring into both, those two values are now selected in the existing snapshot block and there is one call. Behaviour is identical (the initial path still uses the hardcoded `5` / `8`). Say the word if you would rather keep the explicit branch.
- The same test file had a ~25-line inline fixture for building a bare `LoadMLComponent`; it is now `_make_training_component()`, shared by the existing liveness test and the two new component tests, rather than copied a third and fourth time.
- **Not fixed, flagged:** `run()` waits on `while self.base.prediction_started: await asyncio.sleep(0.5)` before ML work without consulting `api_stop`. It yields, so it does not block the loop, but it is another place a stop is not noticed promptly. Left alone as out of scope for this ticket.
- **Cosmetic:** aborting an *initial* train at epoch 0 makes the pre-existing "Restored best weights" line log `val_mae=inf`, since `best_val_loss` seeds to infinity when there is no fine-tune baseline. Harmless (the restore is a no-op there) and only reachable on the shutdown path, so left as is rather than duplicating the restore block to dodge one log line.
- `asyncio.to_thread` uses the loop's default executor. Only one training call is ever in flight per component, so at most one worker thread is created; the triage comment raised `run_in_executor` with an explicit single-thread executor as an alternative if you would prefer that given the past NumPy thread-pool interaction.


apps/predbat/load_ml_component.py
```diff
@@ -974,48 +985,46 @@ async def _do_training(self, is_initial):
...
+                step_days = self.ml_curriculum_step_days
+                max_intermediate_passes = self.ml_curriculum_max_passes
+        # Lock released
+
+        try:
+            # Run the curriculum on a worker thread. It is synchronous, CPU-bound and minutes long,
+            # so calling it directly would pin this component's event loop for its whole duration:
+            # api_stop would go unnoticed, the run_timeout watchdog could never fire, and a shutdown
+            # landing mid-training would sit waiting on this component's thread (#5075).
+            val_mae = await asyncio.to_thread(
```
springfall2008, line 1005: **Bug (high): watchdog-cancel now orphans the training thread and can start a second concurrent training on the same predictor.** Pre-PR the synchronous train_curriculum pinned the loop, so the 2h run_timeout watchdog (load_ml_component.py:98) could never fire mid-training; post-PR the loop is free, so component_base.py:266-277 `wait_for(shield(task), run_timeout)` + `task.cancel()` can fire while `await asyncio.to_thread(...)` is pending. CancelledError lands in the coroutine but the worker thread keeps training; api_stop is False so stop_callback never trips. 60s later start() re-runs run(), should_train is still true, and a second train_curriculum starts on the same LoadPredictor — both threads mutate self.weights/self.biases/Adam/normalisation state in place (_adam_update), and both may save() to model_filepath. Consider making the stop hook also trip on a per-run cancel flag set by the timeout path.

apps/predbat/load_ml_component.py
```diff
@@ -974,48 +985,46 @@ async def _do_training(self, is_initial):
...
+                validation_holdout_hours=holdout_hours,
+                patience=patience,
+                curriculum_window_days=window_days,
+                curriculum_step_days=step_days,
+                max_intermediate_passes=max_intermediate_passes,
+                progress_callback=self.update_success_timestamp,
+                stop_callback=self.training_stop_requested,
+            )
+
+            if val_mae is None and self.training_stop_requested():
```
springfall2008, line 1024: **Bug (high, same watchdog path): a cancelled run() never stamps last_train_time/initial_training_done, so every retry relaunches the full initial curriculum — a self-sustaining runaway.** The statements after the `await asyncio.to_thread(...)` (lines 1028-1029) don't execute when the watchdog cancels the coroutine, even though the orphaned thread finishes its full training. On re-entry is_initial is still True and retrain_age is still >= RETRAIN_INTERVAL, so each 60s-tick run() launches another full 100-epoch/9-pass initial curriculum, each again exceeding run_timeout and leaving another orphaned thread racing on self.predictor.

apps/predbat/load_predictor.py
```diff
@@ -1418,6 +1442,12 @@ def train(
                 )
             )
 
+        if aborted:
```
springfall2008, line 1445: **Bug (medium): the abort path restores only weights — the in-memory model is left a hybrid, contradicting the comment.** train() has already mutated self.feature_mean/self.feature_std (`_fit_windowed_normalisation`, EMA blend on fine-tune) and target statistics and `_reset_adam_optimizer()` before the first `_should_stop()` check; the post-loop restore only puts back best_weights. The comment claims the model is 'left consistent' but only the weights are. Harmless on a real shutdown, but live on the watchdog-cancel path (cancelled run(), component keeps running): every subsequent predict() normalises inputs with statistics the restored weights were never trained against, systematically skewing the published load forecast until the next successful training.

apps/predbat/load_predictor.py
```diff
@@ -1418,6 +1442,12 @@ def train(
...
+            # Weights are restored above so the in-memory model is left consistent, but the run did
+            # not finish: report failure so the caller neither publishes nor saves a partial model,
+            # and skip the AR rollout diagnostic, which is itself minutes of the work we just abandoned.
+            return None
+
```
springfall2008, line 1450: **Bug (medium-low): an abandoned initial curriculum leaves earlier passes' stamps, so the partial model reads as active.** train() stamps self.training_timestamp/self.validation_mae/self.epochs_trained at the end of EVERY completed pass, including intermediate curriculum passes. When a later pass aborts, those stamps survive (only the component-side flags are skipped). _update_model_status() -> is_valid() judges purely on training_timestamp age and validation_mae, neither reflecting the abandonment, so a model trained on only the first curriculum window (e.g. 7 of 28 days) reports 'active', and the fresh timestamp suppresses the ml_max_model_age_hours staleness retrain.

apps/predbat/load_predictor.py
```diff
@@ -1418,6 +1442,12 @@ def train(
                 )
             )
 
+        if aborted:
+            # Weights are restored above so the in-memory model is left consistent, but the run did
+            # not finish: report failure so the caller neither publishes nor saves a partial model,
+            # and skip the AR rollout diagnostic, which is itself minutes of the work we just abandoned.
+            return None
```
springfall2008, line 1449: **Bug (medium-low): no stop check around _ar_rollout_diagnostic, and the final pass still publishes/saves after a stop lands.** The abort path skips the rollout only when the run already aborted at an epoch check; a stop landing after the final pass's last epoch check is waited out through the full rollout (the comment here calls it 'itself minutes of the work'), and because val_mae is then not None, the component's abandon gate (load_ml_component.py:1024) is bypassed: last_train_time is stamped, model_valid set and predictor.save() does disk I/O mid-shutdown. Also: train_curriculum calls train() once per intermediate pass, so an initial curriculum pays up to 9 rollouts whose results are overwritten each time — only the final pass's self.rollout_mae/pattern_mae survive; threading run_diagnostic=False through for intermediate passes would remove most of that cost.

apps/predbat/load_ml_component.py
```diff
@@ -777,6 +777,12 @@ async def run(self, seconds, first):
                     self.log("ML Component: Doing training...")
                     await self._do_training(is_initial)
 
+                    if self.api_stop:
+                        # Training abandoned for shutdown - a re-fetch and a prediction cycle here
+                        # would just add their own delay to the stop we are already answering
+                        self.log("ML Component: Stopping, skipping the post-training prediction cycle")
+                        return True
```
springfall2008, line 784: **Bug (low, watchdog path): run()'s finally clears load_ml_calculating while the orphaned training thread is still running NumPy work, so is_calculating() lies.** When the run_timeout watchdog cancels run() at the to_thread await, this finally still executes while the worker thread keeps training (api_stop is False, so its stop hook never trips). is_calculating() then returns False during genuine heavy NumPy work, and the flag's own comment in this block documents the reason it matters (fork() is unsafe with live threads) — the same hazard the flag was kept for.

springfall2008, line 784: **Bug (low): the new early return also fires when training COMPLETED successfully, skipping _update_model_status() and run()'s trailing update_success_timestamp().** A stop landing during the final epoch's work (after the last per-epoch check) lets train_curriculum return a valid val_mae: the model is saved and last_train_time stamped here in _do_training, but run() returns before the status update — the fresh model's status is never published and last_success_timestamp can sit 2h stale at the moment is_alive() is next consulted during shutdown/restart reporting.

springfall2008, line 784: **Bug (low): the abandon-vs-failure classification by re-reading api_stop misreports a genuine failure.** train()/train_curriculum() return None for both 'all passes failed' and 'aborted'; this check sends the all-passes-failed case down the 'Training abandoned because the component is stopping' branch whenever api_stop happens to be set, the 'Warn: ML Component: Training failed' branch and its error accounting never fire, and (because run() returns early) _update_model_status() never runs — so a recurring data problem is invisible in that cycle's log and model_status.

springfall2008, line 784: **Gap (low): shutdown responsiveness is not re-established for the pre-training wait loop.** The `while self.base.prediction_started: await asyncio.sleep(0.5)` loop just above (line 771, inside this same run()) never checks api_stop, and it sits upstream of this early return — a stop() landing while Predbat's prediction cycle is running still spins here until that cycle completes, the same delay class #5075 set out to remove.

apps/predbat/load_predictor.py
```diff
@@ -1170,6 +1170,16 @@ def _clip_predictions(self, predictions, lookback_buffer=None):
...
         return predictions
 
+    def _should_stop(self, stop_callback):
+        """Return True when the caller has asked training to abort, treating a raising hook as no."""
+        if not stop_callback:
+            return False
+        try:
+            return bool(stop_callback())
+        except Exception as e:
+            self.log("Warn: ML Predictor: stop callback raised {}, continuing training".format(e))
```
springfall2008, line 1180: **Design note (low): the exception polarity makes a broken stop hook silently void the whole fix.** A raising stop_callback is logged and treated as 'do not stop' — the same swallow policy as progress_callback, but here the failure cost is a full-length, unaborted minutes-long run. Unlike progress_callback (whose failure only costs a liveness pulse), any future refactor that makes training_stop_requested raise (e.g. api_stop moved behind a property that raises once the component registry is torn down) turns shutdown back into a full-curriculum hang with only a per-epoch warn to show for it, while progress_callback keeps stamping last_success_timestamp so the health monitor reports the component alive. Worth at least counting/escalating repeated hook failures rather than swallowing forever.

apps/predbat/load_ml_component.py
springfall2008, line 784: **Cosmetic (low): the early return re-asserts api_started=True during teardown.** A True return flows into component_base.start()'s 'if run_result: if not self.api_started: self.api_started = True', so the in-flight run() sets api_started back to True after Components.stop() already cleared it — is_alive() reports load_ml healthy during the shutdown window until start()'s loop exits and re-clears it. Benign but contradicts stop()'s bookkeeping.

apps/predbat/load_predictor.py
```diff
@@ -1618,15 +1661,24 @@ def train_curriculum(
                 norm_ema_alpha=norm_ema_alpha,
                 huber_delta=huber_delta,
                 progress_callback=progress_callback,
+                stop_callback=stop_callback,
             )
 
+            if pass_mae is None and self._should_stop(stop_callback):
+                self.log("ML Predictor: Curriculum training aborted during pass {}/{} - stop requested".format(pass_idx + 1, total_passes))
```
springfall2008, line 1668: **Cleanup: this after-pass check is the redundant one of the five _should_stop sites.** train() already returns None on abort and api_stop is sticky, so the top-of-next-pass check (line 1630) — or the before-final-pass check (line 1678) when this is the last intermediate pass — returns None with an identical outcome; this site only changes which of three near-identical messages fires. The minimal set is 1574, 1630, 1678 and the after-final check (1701), which must stay: deleting it lets an aborted final pass fall through to 'if final_mae is not None' and return the previous pass's val_mae as a completed curriculum, which the component would save and stamp last_train_time from.

apps/predbat/load_ml_component.py
```diff
@@ -945,6 +951,10 @@ def array_to_dict(arr):
             self.export_rates_data = None
             self.data_ready = False
 
+    def training_stop_requested(self):
+        """Return True when the component is stopping, so an in-flight training run abandons itself."""
+        return self.api_stop
```
springfall2008, line 956: **Cleanup/altitude (low): this hook serves exactly one cancellation source, and the same file now spells the predicate two ways.** run() (line 780) still tests `if self.api_stop:` directly while the hook goes through this wrapper, and ComponentBase.start() also exits on fatal_error (component_base.py:248) — a fatal_error mid-training leaves the thread running with no abandon. The repo already has several hand-rolled cancel mechanisms (api_stop polling in every long-lived component, db_manager's threading.Event pair, chat.py's run_coroutine_threadsafe bridge); this is a third. A single per-run cancellation handle that stop(), the watchdog cancel and fatal_error all feed would have made the watchdog orphan problem (see comment on line 1005) impossible by construction.

apps/predbat/tests/test_load_ml.py
```diff
@@ -3460,3 +3471,137 @@ def train_curriculum(self, *args, **kwargs):
...
+    component = _make_training_component(BlockingPredictor(), now_utc)
+
+    async def scenario():
+        """Queue a loop task, train, and see whether the task ever got a chance to run."""
+
+        async def releaser():
+            """Release the trainer, which only happens if the loop is still being serviced."""
+            loop_made_progress.set()
+
+        task = asyncio.create_task(releaser())
```
springfall2008, line 3509: **Test robustness: this handshake can pass spuriously.** The single-shot threading.Event only proves the loop ran at SOME point before the trainer returned, not that the loop stayed free during the trainer call: any earlier await in _do_training that yields once (today the `async with self.data_lock`, which happens not to yield) lets the releaser fire, after which a synchronous loop-pinning train_curriculum call still passes the assert. A regression that added `await asyncio.sleep(0)` before a direct synchronous trainer call would reintroduce #5075 with this test green. A loop-tick assertion collected after the trainer returns (or a second event set after the trainer's own return) would close that hole. Also worth knowing: if the regression under test occurs, the executor thread blocks here for the full 10s and asyncio.run's shutdown_default_executor() then joins it, adding a silent 10s penalty to the shared-process suite.

apps/predbat/tests/test_load_ml.py
```diff
@@ -56,6 +56,9 @@ def test_load_ml(my_predbat=None):
         ("curriculum_training", _test_curriculum_training, "Curriculum training with progressive window expansion"),
         ("training_progress_callback", _test_training_progress_callback, "Training reports liveness through progress_callback"),
         ("component_alive_during_training", _test_component_marks_itself_alive_during_training, "Component keeps its success timestamp fresh while training"),
+        ("training_stop_callback", _test_training_stops_on_stop_callback, "Training abandons a run when the stop hook trips"),
+        ("training_off_event_loop", _test_component_training_does_not_block_event_loop, "Component training leaves its event loop free to run"),
```
springfall2008, line 60: **Test-coverage gap: the mid-curriculum abort paths are untested.** This test's curriculum case stubs train() and only exercises the pre-loop check (train_curriculum's top-of-function check returns None before any pass), so the 'aborted before pass N', 'aborted during pass N' and 'aborted before final pass' paths — and the component-side distinction between an aborted run and an all-passes-failed run — are never executed. A hook that returns True only after the first pass, with a real (stubbed-returning) train, would cover the between-passes path and pin that an intermediate val_mae is not leaked as a completed curriculum.

apps/predbat/load_ml_component.py
springfall2008: _Automated reply from the triage bot._

Confirmed and fixed. `wait_for(shield(task), run_timeout)` -> `task.cancel()` lands the CancelledError on the coroutine suspended at `await asyncio.to_thread(...)`; the executor's worker is already running, so it keeps training with api_stop still False.

`_do_training` now catches `asyncio.CancelledError` around that await, sets `self.training_cancelled` and re-raises. `training_stop_requested()` is `api_stop or fatal_error or training_cancelled`, so the orphaned thread abandons at its next epoch check. `fatal_error` is in there for the same reason — `ComponentBase.start()` exits on it (component_base.py:330) without ever calling `stop()`.

The second-concurrent-training half needed its own guard, because the cancel signal alone still leaves a window while the thread unwinds: `training_running` is raised and cleared **inside** `run_training()` on the worker thread, not by the coroutine, precisely because a cancelled coroutine never reaches its own cleanup. `_do_training` returns early when it is set.

New test `training_watchdog_cancel` drives the real cancel: it asserts the hook trips after `task.cancel()`, that a second `_do_training` while the thread is still inside makes no second `train_curriculum` call, and that the flag clears when the thread returns. Verified red without the fix.

One thing this does not fix, which is a maintainer call: if an initial curriculum legitimately needs more than `run_timeout` (2h, load_ml_component.py:98), it can now never complete — every attempt is cancelled and retried. Pre-PR it completed because the blocked loop made the watchdog inert. The retry is at least bounded and clean now, and the cancel logs at Error level with the timeout value, but raising `run_timeout` for this component may be the right answer.

apps/predbat/load_ml_component.py
springfall2008: _Automated reply from the triage bot._

The stacking half is fixed; the relaunch half I read differently.

Fixed: with the cancel signal now tripping the stop hook and `training_running` owned by the worker thread, the orphaned thread abandons at its next epoch check and no second curriculum can start while it is still inside. So there is no accumulation of threads racing on `self.predictor`.

Not changed, deliberately: after a cancelled run, `is_initial` is still True and `last_train_time` is still None, so the next tick does relaunch the initial curriculum — and that is the correct state to be in, because no initial training has in fact completed. Stamping `initial_training_done` off a run that was abandoned part-way would publish a model built from one curriculum window as a finished one, which is the failure the sibling comment on load_predictor.py:1450 describes.

What is left is a bounded retry rather than a runaway: one training at a time, no shared-state race, and each cancel logged at Error with the `run_timeout` value. If the underlying cause is that the initial curriculum genuinely exceeds 2h on slower hardware then the retry never converges, but the fix for that is `run_timeout` (load_ml_component.py:98), not the stamping — flagged on the sibling thread for your call.

apps/predbat/load_predictor.py
springfall2008: _Automated reply from the triage bot._

Confirmed and fixed. `_fit_windowed_normalisation` (load_predictor.py:1268/1271) and `_normalize_targets(fit=True)` both run before the loop at 1328, so the comment claiming a consistent model was only true of the weights.

`train()` now snapshots `feature_mean`, `feature_std`, `target_mean` and `target_std` before the refit and puts them back on the abort path alongside `best_weights`, and the comment says what it actually does.

Not restored: the Adam moments. `_reset_adam_optimizer()` is called at the start of each fine-tune anyway, so its state has no effect on `predict()` — only on the first steps of the next training, which starts by resetting it.

New test `training_abort_restores_norm` fine-tunes on data scaled 5x so the EMA refit moves the statistics well outside any tolerance, aborts at the first epoch check, and asserts all four statistics and every weight layer come back unchanged. Verified red without the fix.

apps/predbat/load_predictor.py
springfall2008: _Automated reply from the triage bot._

Confirmed and fixed. `train()` stamps at the end of every completed pass, `is_valid()` reads only `validation_mae` and `training_timestamp`, so an abandoned curriculum left the last intermediate pass looking like a finished training — and a fresh timestamp suppresses the `ml_max_model_age_hours` retrain that is the only route back.

`train_curriculum()` now snapshots `training_timestamp`, `validation_mae`, `validation_bias` and `epochs_trained` on entry and restores them on all four abort returns, via a small `abandon_curriculum()` helper so no return path can forget.

That rollback needed a second change to be safe. Restoring `training_timestamp = None` on an abandoned *initial* curriculum would have made `is_valid()` return `(True, None)` on the random He initialisation, since `model_initialized` is set by `_initialize_weights()` before the first epoch. So `is_valid()` now gates on a new `model_trained` flag, set at the end of `train()` and in `load()`. It is deliberately not `training_timestamp is None`: `component_init_predictor_last_train_time` pins that a model saved before timestamps existed must still be valid, and that test caught the first attempt.

New tests `curriculum_mid_abort_rollback` (stale stamps survive an abort, `is_valid` still reports `stale` rather than `active`) and `untrained_model_not_valid` (both sides: never-trained is `not_trained`, legacy timestamp-less trained model is valid). Both verified red without the fix.

apps/predbat/load_predictor.py
springfall2008: _Automated reply from the triage bot._

First half fixed, second half declined as out of scope for this PR.

**Rollout stop check — fixed.** Confirmed: a stop landing after the last epoch check falls straight through to `_ar_rollout_diagnostic`, which the abort comment itself calls minutes of work. `train()` now checks `_should_stop` immediately before the rollout and returns `best_val_loss` without running it.

It returns the result rather than `None`, deliberately. The training itself genuinely completed at that point — the model is trained, validated and worth saving; only the diagnostic is given up. Returning `None` there would throw away a completed training and leave the next cycle repeating it. `rollout_mae`/`pattern_mae` keep whatever the last completed run set, which is the same thing they would hold if the rollout had been skipped for any other reason.

That does mean the component still stamps `last_train_time` and calls `predictor.save()` on this path. I think that is right: `save()` is a single `np.savez_compressed` of the weights, not a minutes-long operation, and dropping a model that finished training because the stop arrived a second later is a worse outcome than one extra file write during shutdown.

**`run_diagnostic=False` for intermediate passes — declined here.** The waste is real and I verified it: `self.rollout_mae`/`self.pattern_mae` are overwritten on every `train()` call (load_predictor.py:1478-1479), so an initial curriculum pays up to 9 rollouts and keeps only the last. But that is a training-cost optimisation, not a shutdown-responsiveness fix, and it changes what every curriculum pass computes rather than only what happens on an abort. This PR already carries one refactor the maintainer may want to unpick; adding a behaviour change to the normal path makes that harder, not easier. Worth its own PR — happy to open one if you want it.

apps/predbat/load_ml_component.py
springfall2008: _Automated reply from the triage bot._

Fixed, though not by the reasoning quoted. The block's own comment says the multiprocessing pool is gone and the flag is `now purely a status flag surfaced via is_calculating() and logged by calculate_plan()` — the `fork()` hazard it describes is the historical reason the flag exists, not a live one. So the consequence is a misleading status, not an unsafe fork.

The misleading status is real all the same: on the watchdog-cancel path this `finally` runs while the worker thread is still inside `train_curriculum`. `is_calculating()` now returns `load_ml_calculating or training_running`, and `training_running` is owned by the worker thread for exactly the span of the synchronous call — so the answer is true whoever owns the coroutine.

Covered in `training_watchdog_cancel`: after the cancel it clears `load_ml_calculating` by hand, as run()'s finally does, and asserts `is_calculating()` is still True while the orphaned thread is inside the trainer.

springfall2008: _Automated reply from the triage bot._

Confirmed and fixed. A stop landing during the final epoch's work does let `train_curriculum` return a valid `val_mae`, and the early `return True` then skipped both `_update_model_status()` and `update_success_timestamp()`.

`run()` no longer returns early. The api_stop branch now sets `should_fetch = False` and falls through, so the re-fetch, the prediction cycle and the database write are still skipped — that is the shutdown delay the branch exists to avoid — while `_update_model_status()` and the trailing `update_success_timestamp()` both run. A model that finished training before the stop landed gets published.

This also removed the api_stop check from *after* `_do_training` to before it, so a stop that arrives while waiting on the prediction cycle skips the training entirely rather than starting one it is about to abandon.

springfall2008: _Automated reply from the triage bot._

Confirmed and fixed. Re-reading `api_stop` after the fact cannot distinguish the two `None`s, so an all-passes-failed run filed as an abandonment whenever a stop happened to be set — no `Training failed` warning, no error accounting.

`_do_training` now wraps the hook it hands the trainer: `stop_hook()` records that it answered True, and the classification reads that record rather than `api_stop`. A run where the hook was never called, or was called and always said no, is a failure even if `api_stop` was set a moment later.

New test `training_failure_not_abandon`: a trainer that sets `api_stop` and returns `None` without the hook ever tripping must log `Training failed` and must not log `Training abandoned`. Verified red — without the fix it logs the abandonment message, which is what the assertion failure printed.

springfall2008: _Automated reply from the triage bot._

Fixed — the PR notes had flagged this as out of scope, but it is two lines and it is the same delay class as the one the ticket is about, so leaving it was the wrong call.

The wait is now `while self.base.prediction_started and not self.api_stop`, and the api_stop branch below it skips the training rather than starting one it would immediately abandon. A stop landing while a plan cycle runs is answered at the next 0.5s tick instead of waiting the cycle out.

apps/predbat/load_predictor.py
springfall2008: _Automated reply from the triage bot._

Partly taken. The asymmetry you describe is real — a broken `progress_callback` costs a liveness pulse, a broken `stop_callback` costs the entire abort mechanism — and it is worse than that: `progress_callback` keeps stamping `last_success_timestamp`, so the health monitor reports the component alive throughout.

Not changed: the polarity. A hook that raises still means "do not stop". Flipping it would let a transient exception abort a valid multi-minute training, and on the only caller that exists `training_stop_requested()` is three attribute reads that cannot raise. Trading a certain cost for a hypothetical one is the wrong side of that bet.

Changed: the reporting, which was the part that would actually leave you blind. `_should_stop` counts consecutive failures and logs the first at **Error** level with the text `this run can no longer be aborted`, and stays quiet after that. A per-epoch warn would have emitted several hundred lines across an initial curriculum and buried itself; one Error line is visible. The counter resets on the first successful call, so an intermittent hook reports each new run of failures.

If you would rather have escalation-to-stop after N consecutive raises on top of that, say so and I will add it — I did not want to introduce a threshold nobody has a number for.

apps/predbat/load_ml_component.py
springfall2008: _Automated reply from the triage bot._

Verified, not changed. `start()` does re-assert `api_started` from a `True` return (component_base.py:360-361), so an in-flight run() completing after `Components.stop()` has cleared it does put it back until the loop exits.

The two ways to avoid it are both worse. Returning `False` takes the else branch: `count_errors += 1`, `non_fatal_error_occurred()` (which sets `base.had_errors`) and a `run() returned False` warning — turning an orderly shutdown into a recorded error. Suppressing the re-assert inside `start()` would change every component's behaviour from a load_ml PR.

It is also not specific to this branch: any component whose run() happens to complete inside the shutdown window does the same thing, because `start()` checks `api_stop` at the top of the loop and not after the run. The window closes on the next iteration. If it is worth closing properly, the place is `start()` — gate the `api_started = True` on `not self.api_stop` — as its own change with the component_base tests behind it.

apps/predbat/load_predictor.py
springfall2008: _Automated reply from the triage bot._

Verified the redundancy claim and it holds for the return value — but I have kept the site, for the log line.

Traced it: if the abort lands inside intermediate pass N and N is not the last, the check at 1630 returns `None` on the next iteration before any slicing; if N is the last, 1678 does. Same result, no extra work either way, so as you say the only thing this site changes is which message fires.

That message is the reason to keep it. Without it an abort *during* pass N is reported as `aborted before pass N+1`, which is wrong about where the run actually stopped. These log lines exist to reconstruct a shutdown after the fact — that is how #5075 was diagnosed, from timestamps in a user's log — so a message that misplaces the abort by one pass has a cost, and one extra `stop_callback()` call per pass does not.

Agreed and unchanged on the important half: the after-final check at 1701 stays. Without it an aborted final pass falls through to `if final_mae is not None` and returns the previous pass's `val_mae` as a completed curriculum. All four abort returns now also roll the published stamps back — see the thread on load_predictor.py:1450.

apps/predbat/load_ml_component.py
springfall2008: _Automated reply from the triage bot._

Taken, within this component. The single per-run cancellation handle is now `training_stop_requested()`, and all three sources you name feed it: `api_stop`, `fatal_error` and the new `training_cancelled` that the watchdog cancel raises. You were right that `fatal_error` was an unhandled abandon path — `ComponentBase.start()` exits on it (component_base.py:330) without `stop()` ever being called, so the thread would have kept training.

That is what makes the orphan problem on line 1005 unreachable rather than merely handled: the hook answers True for every source, and `training_running` is owned by the worker thread, so a second run cannot start while the first is unwinding.

Not taken: unifying this with `db_manager`'s `threading.Event` pair and `chat.py`'s `run_coroutine_threadsafe` bridge into one repo-wide mechanism. That is a real observation and probably the right end state, but it is a cross-component refactor touching components with no test coverage of their shutdown paths, and doing it from a load_ml bug fix would make both harder to review. Worth its own issue.

On the two spellings: run() still tests `self.api_stop` directly, and that one is correct as-is — it is asking "are we shutting down, should I skip the prediction cycle", not "should this training run abandon". They are different questions that happen to share a term. The training question is the one that now has a single answer.

apps/predbat/tests/test_load_ml.py
springfall2008: _Automated reply from the triage bot._

Confirmed and fixed. The single-shot event did prove only that the loop ran at some point before the trainer returned, and your regression scenario — an `await asyncio.sleep(0)` before a synchronous trainer call — would indeed have kept it green.

Replaced with two handshakes, each gated on the trainer reaching that point. The trainer sets `trainer_reached[i]` and waits on `loop_released[i]`; the loop task only sets `loop_released[i]` **after** observing `trainer_reached[i]`. So each release is proof that a loop task ran after the trainer said it was inside the call, and the second one proves it twice — a task that ran between the trainer's two handshakes cannot have run before the call started. The assertions check both the trainer's view (`all(released)`) and the loop task's (`loop_answered_both_rounds`).

The polling loops are bounded rather than `while True`, so a regression that stops the trainer being entered at all fails on an assertion instead of hanging the shared-process suite.

On the 10s penalty: still there in the regression case, and I left it. Dropping the timeout would trade a slow failure for a flaky one on a loaded CI box, and the penalty is only paid when the test is already failing.

apps/predbat/tests/test_load_ml.py
springfall2008: _Automated reply from the triage bot._

Confirmed and fixed, close to the design you described. `training_stop_callback`'s curriculum case did only reach the top-of-function check, so every between-passes path was untested.

New test `curriculum_mid_abort_rollback` uses a hook that trips only once a pass has completed (`passes['count'] >= 1`) against a stubbed `train` that stamps the model the way the real one does. That exercises the `aborted before pass N` path with N > 1, and asserts exactly the thing you called out: `curriculum_mae is None`, so the completed intermediate pass's `val_mae` is not leaked as a finished curriculum. It also pins the stamp rollback — starting from a 100-hour-old timestamp, `is_valid()` must still report `stale` afterwards rather than `active`.

The component-side abandoned-vs-failed distinction is covered separately by `training_failure_not_abandon`, and the watchdog-cancel abort path by `training_watchdog_cancel`. Both verified red without their fixes.

Still not covered: `aborted during pass N` (1667) and `aborted before final pass` (1678) as distinct paths. They share the `abandon_curriculum()` helper with the path that is tested, so the rollback behaviour is common — what is untested is which of them fires, which is the log-message distinction discussed on the 1668 thread.

Issue #4856, opened 2026-08-30
Multi-inverter rework: tracking issue


chalfontchubby:
Tracking issue for a rework of multi-inverter support. Filing this as the umbrella so the discussion happens in one place; the intent is to spin off separate sub-issues per workstream once we've agreed the shape, not to fix everything here.

Prompted by the #4842 thread, where @gcoan and I ended up agreeing that "multi-inverter support was a bit of an edge case" and worth a proper look rather than another point fix.

**Full write-up:** https://claude.ai/code/artifact/e91af2a5-3440-459e-8daa-a3a31d7f32d5

Everything below is verified against `v8.54.2` / main `4b42b7a1` by reading the code; the numbers are computed from the formulae as written, not measured on hardware.

## The shape of the problem

Predbat has three separate models of a multi-inverter system and no shared one:

1. **Planner** — optimises a single virtual battery made by summing the fleet. `soc_kw`, `soc_max`, rates and limits are plain sums; battery temperature is an unweighted mean; the charge curve is measured from inverter 0 and applied to everything.
2. **Executor** — splits that fleet-wide answer back out per inverter.
3. **Balancer** — an independent 60-second reactive loop that builds its own `Inverter` objects, knows nothing about the plan, and can only write rate 0 or rate max.

Almost every reported multi-inverter fault sits in one of the two seams. It all works correctly when every inverter is identical — which is what the test suite covers, and what real fleets grow out of.

## The headline finding

Not any of the existing bug reports: **the executor's target split actively manufactures imbalance from a balanced start.**

`adjust_battery_target_multi()` (`execute.py:807-808`) apportions the fleet SoC delta by each inverter's share of total *charge rate* and never reads `soc_max`. On @gcoan's 9.5 + 5.2 kWh fleet, both on 2.6 kW, starting **perfectly balanced at 50/50**:

| Fleet target | 9.5 kWh ends | 5.2 kWh ends | Spread | Fleet actually reaches |
|---|---|---|---|---|
| 60% | 58% | 64% | 6 pts | 60% |
| 80% | 73% | 92% | **19 pts** | 80% |
| 90% | 81% | 100% | 19 pts | **88%** |
| 95% | 85% | 100% | 15 pts | **90%** |

[...]

PR #5114, opened 2026-09-16
fix(execute): skip the redundant charge-off call while exporting

### LINKED ISSUE #4165: Tesla Powerwall 3 (Teslemetry) — Operation Mode rapidly flip-flops between Autonomous and Self-consumption within same control cycle, export not working
**Describe the bug**
When Predbat calculates an active export window with a valid price incentive, it should switch operation mode from `self-consumption` to `autonomous` (Time-Based Control) however when it does, it is immediately followed  within the same or next second that the idle write is reversing it back into `self-consumption` mode.

**Expected behaviour**
When export window is active that `autonomous` is set and is not overwritten until the end of the export cycle 

**Predbat version**

v8.43.0

**Environment details**

- Tesla Powerwall 3 & Expansion Pack
- Standard HAOS install in a VM
- Telemetry for API
- Octopus Intelligent Go with Agile Outgoing

**Screenshots**
N/A

**Log file & apps.yaml**
See attached

**Predbat debug yaml file**
See attached

[apps.yaml.txt](https://github.com/user-attachments/files/29574096/apps.yaml.txt)
[predbat.log](https://github.com/user-attachments/files/29573919/predbat.log)
[predbat_debug.yaml.txt](https://github.com/user-attachments/files/29573918/predbat_debug.yaml.txt)

### LINKED ISSUE #4641: Tesla inverter: charge_stop_service fires after discharge_start_service in the same cycle, clobbering operation_mode back to Self-Powered every discharge attempt
Using the TESLA inverter type (tesla_custom, Powerwall 3) with select.predbat_mode = Control charge & discharge. During every planned discharge/export window tonight, Predbat correctly identifies the window and calls discharge_start_service, but within ~0.4 seconds also calls charge_stop_service for the same cycle — and the charge-side call runs second, overwriting select.my_home_operation_mode back to Self-Powered immediately after discharge_start_service had just set it to Time-Based Control. Net result: select.my_home_energy_exports stays Everything (harmless on its own) but operation_mode never actually holds Time-Based Control, so the Gateway never exports — it just load-follows. Real power telemetry confirms zero net grid export across every attempt: sensor.my_home_grid_power stayed ~0W while sensor.my_home_battery_power tracked sensor.my_home_load_power almost exactly, for the full duration of every window.

Log evidence (repeated identically at 17:20, 17:25, 17:50, 18:50, 18:55 ×3, and 21:50 — 100% of attempts across the evening):

21:50:56.708  Inverter 0 Calling service discharge_start_service ... select.my_home_energy_exports -> Everything
21:50:57.130  Inverter 0 Calling service discharge_start_service ... select.my_home_operation_mode -> Time-Based Control
21:50:57.538  Inverter 0 Calling service charge_stop_service ... select.my_home_operation_mode -> Self-Powered   <-- clobbers it
21:50:58.004  Inverter 0 Calling service charge_stop_service ... number.my_home_backup_reserve -> 0

Apparent cause: the charge-window adjustment and discharge-window adjustment appear to be evaluated independently each cycle. When the current time isn't inside a charge window, the charge-side logic unconditionally calls charge_stop_service (which sets the Tesla mode select to Self-Powered) — with no check for whether a discharge window is concurrently active and has already set a different mode intentionally moments earlier in the same cycle.

Expected: if a discharge window is active this cycle, the charge-side adjustment should not call charge_stop_service in a way that overwrites the mode select the discharge-side adjustment just set — or the two should be sequenced/reconciled so the discharge-active setting wins.

Setup: Predbat v8.50.0, inverter_type: TESLA, has_reserve_soc: False, has_target_soc: False, control via tesla_custom select entities (charge_start_service/charge_stop_service/discharge_start_service/discharge_stop_service) per the standard Tesla apps.yaml template. Happy to attach full apps.yaml (Tesla-control block) and further log excerpts if useful.

chalfontchubby:
This is an automated draft PR generated from issue #4641 — a maintainer should review it before merging.

Fixes #4641
Also fixes #4165 (duplicate report, closed in favour of #4641; same reporter, same root cause, independently confirmed against its own attached log/debug.yaml)

## Summary

In `Control charge & discharge` mode, the "Charging/Discharging off via service" block in `execute_plan()` (`apps/predbat/execute.py`) fired an unconditional `adjust_charge_immediate(0)` → `charge_stop_service` whenever `isCharging` was False — which is also true while actively exporting. This ran *after* `adjust_export_immediate()` had already issued its own `charge_stop_service` followed by `discharge_start_service` moments earlier in the same cycle.

On service-template inverters (e.g. the Tesla custom/Teslemetry integrations) where `charge_stop_service` and `discharge_start_service` write the same mode-select entity, the trailing charge-stop call clobbered the mode back to e.g. "Self-Powered"/"self_consumption" immediately after `discharge_start_service` had set it to "Time-Based Control"/"autonomous" — so the export never actually engaged, even though Predbat's own state showed it as active. This reproduced on every export attempt for both reporters (#4641, #4165), across different Predbat versions (v8.42.5 through v9.0.1).

The fix adds an `isExporting` guard to that block: `adjust_export_immediate()` already stops charging as part of starting the export, so re-issuing a second charge-stop there added nothing but the clobbering write.

## Testing

- Added/updated unit test coverage in `apps/predbat/tests/test_execute.py`:
  - `discharge_car_full_bat` (an existing "Exporting" scenario with `set_charge_window=True`) now asserts the charge-immediate sentinel stays untouched (`-1`) rather than being forced to `0` by the extra charge-stop call.
  - Fixed a masking bug in the test harness itself: `immediate_charge_soc_target`/`immediate_discharge_soc_target` were never reset between scenarios in `run_execute_test()`, so several "Exporting" scenarios were incidentally passing on stale values left over from whichever scenario ran immediately before them, rather than asserting anything meaningful about the current scenario. Added an explicit reset per scenario, and corrected two read-only scenarios' expectations to `-1`/`-1` (no service calls are issued in read-only mode at all) now that the sentinel is properly reset.
- Verified against a real `predbat_debug.yaml`/log/apps.yaml from #4165's attachments — replayed clean under `./run_all --debug_file ... --redo`.
- Red/green check: reverting just the `execute.py` change (test change kept) makes `./run_all --test execute` fail with `ERROR: Inverter 0/1 Immediate charge SOC target should be -1 got 0` on the `discharge_car_full_bat` scenario; restoring the fix passes.
- `./run_all --quick` and `./run_pre_commit` both pass clean.

## Notes

- GitNexus tools were not available in this session (no `.gitnexus/` index in this checkout), so the blast-radius check was done by search instead: `execute_plan()` is called only from `predbat.py`'s main update loop, the change is entirely internal to its own per-inverter loop, and `test_execute.py`'s full scenario suite (~180 cases) passes unchanged apart from the intentionally-updated assertions above.
- Left the symmetric case alone: the "off via service" block's export-side counterpart (`if not isExporting and self.set_export_window: inverter.adjust_export_immediate(EXPORT_LIMIT_IDLE)`) has no `isCharging` guard either, and could in principle clobber a charge-mode select on an inverter with no `discharge_stop_service` configured. This is untested by either reporter's logs and not part of the reported symptom, so it's called out here rather than spelled speculatively — a maintainer may want a follow-up issue if it turns out to matter in practice.


chalfontchubby:
Labelled priority high since the triage bot decided the source issue was.

chalfontchubby:
_Claude wrote this reply, working with me._

On the Copilot review point about the `isExporting` guard being fleet-wide — the mechanism is correct, but I don't think it's something this PR introduces, so I'd rather leave the code as-is and let a maintainer decide.

**The mechanism is real.** `isExporting` is initialised at `execute.py:187`, above `for inverter in self.inverters:` (line 188), and is only ever assigned `True` (lines 515, 547). The guard at line 776 is inside that loop. So in a mixed fleet where inverter 0 exports and inverter 1 does not, inverter 0 sets the flag and inverter 1 then skips its own charge-off call. No argument with any of that.

**But it isn't new here.** Two things point that way:

`isCharging` has exactly the same shape — initialised at line 186, only ever set `True` at line 397 — and it has gated this same block on main since well before this PR: `if not isCharging and self.set_charge_window`. The fleet-wide-flag pattern is this block's existing convention, not something the change invents. The PR adds a second term with identical scoping to a condition whose first term already had it.

More directly: the sibling line immediately below is **unmodified from main**:

```python
if not isExporting and self.set_export_window:
    inverter.adjust_export_immediate(int(EXPORT_LIMIT_IDLE))
```

That is the exact condition being flagged as a new risk — fleet-wide `isExporting`, evaluated per inverter inside the loop — and it is already shipping one line later, untouched by this PR.

So the accurate description is that this extends a pre-existing fleet-scoping limitation to one more branch, in a block where both the adjacent line and the other flag already work that way.

**Why I'd not fix it here.** The limitation is genuine and probably worth fixing, but a real fix means making both flags per-inverter across the whole block, which changes `isCharging` behaviour on every multi-inverter install in order to fix a case neither reporter hit. That is a much larger and riskier change than a one-line export guard for a confirmed and reproduced Tesla mode-clobber bug, and it looks like it belongs with the multi-inverter rework in #4856.

It is also the same underlying gap as the symmetric case I flagged in the PR description (the export-side counterpart having no `isCharging` guard). Both are the per-inverter-scoping question, not the mode-clobber question.

Happy to split the per-inverter scoping out as its own issue if you agree that's the right shape, or to do it in this PR instead if you'd rather it went in together.

PR #5121, opened 2026-09-16
fix(solar): calibrate PV against the raw provider forecast, not Predbat's own calibrated output

### LINKED ISSUE #5116: PV Calibration Issue
<html><head></head><body><p><strong>Describe the bug</strong>
PV calibration corrects only about half of the forecast bias, because it learns from its own calibrated output.</p>
<p><code>pv_calibration()</code> compares actual generation against the history of <code>sensor.predbat_pv_forecast_h0</code>. While <code>metric_pv_calibration_enable</code> is on, that sensor's <strong>state</strong> is the <em>calibrated</em> forecast (<code>power_nowCL</code>). The factor worked out from that comparison is then applied to the <em>raw</em> forecast (<code>pv_forecast_minute</code>).</p>
<p>Once calibration has been running for a few days, it is measuring the error of its own corrected forecast and applying that to the uncorrected one. The applied factor settles where <code>actual / (f × raw) = f</code>, i.e. <strong>f = √(actual / raw)</strong>. A site where Solcast over-forecasts by ~18% (true ratio 0.82) is calibrated to ~0.906 instead of 0.82. The same happens in the other direction: a true ratio of 1.20 gives ~1.095.</p>
<p>Source (<code>apps/predbat/solcast.py</code>, tag v9.0.3 / commit <code>340205fa25a7e701aa89bc2d7178810e59531a00</code>; the file is identical on <code>main</code>):</p>
<ul>
<li><strong>L808–809</strong> — <code>h0</code> publishes the calibrated value as its state. The raw value only appears in the <code>now</code> attribute (L816).
<pre><code class="language-python">self.dashboard_item(    "sensor." + self.prefix + "_pv_forecast_h0",    state=dp2(power_nowCL if calibration_on else power_now),
</code></pre></li>
<li><strong>L895–896</strong> — calibration reads the <em>state</em> history of that same sensor as "past forecast".
<pre><code class="language-python">pv_forecast, pv_forecast_hist_days = history_attribute_to_minute_data(    self.now_utc_exact, prune_today(history_attribute(self.get_history_wrapper("sensor." + self.prefix + "_pv_forecast_h0", days + 1, required=False)), ...
</code></pre></li>
<li><strong>L1048 / L1055</strong> — the per-slot and whole-day ratios are taken against that history.</li>
<li><strong>L1098</strong> — the ratios are applied to the raw series: <code>pv_forecast_minute_adjusted[minute] = pv_value * slot_adjustment.get(slot, 1.0) * use_scaling_day</code>. Here <code>pv_value</code> comes from <code>pv_forecast_minute</code>, the raw <code>pv_estimate</code> series passed in at L1492.</li>
</ul>
<p>The same self-reference is already recognised for the ceiling calculation at L1117 ("<em>max_pv_power_forecast is deliberately NOT used here: it is read back from the published pv_forecast_h0 sensor, whose state is this same capped output</em>"), but it is still present in the ratio calculation.</p>
<p>A side effect: the log line <code>PV Calibration: Past day N had X kWh of forecast PV</code> reports the calibrated forecast rather than the provider's. That makes past forecasts look closer to actual than they really were.</p>
<p><strong>Expected behaviour</strong>
Calibration should compare actual generation against the <strong>raw</strong> provider forecast, the same series the factor is applied to. The applied factor should then settle at the true actual/forecast ratio rather than its square root.</p>
<p>Suggested fix (untested): read the raw value that is already published as the <code>now</code> attribute:</p>
<pre><code class="language-python">history_attribute(
    self.get_history_wrapper("sensor." + self.prefix + "_pv_forecast_h0", days + 1, required=False),
    state_key="now",
    attributes=True,
)
</code></pre>
<p>A dedicated raw sensor would also work. Some things to consider:</p>
<ul>
<li><strong>Migration.</strong> Older history has calibrated states and may lack <code>now</code>. Falling back to the state for those points gives the current behaviour until the 7-day window has rolled over.</li>
<li><strong>P10 / P90.</strong> The worst-day and best-day factors are also measured against the calibrated history. They are divided by the average factor, so most of the effect cancels, but they would shift slightly.</li>
<li><strong>Recorder.</strong> <code>h0</code>'s attributes are small and are stored by the recorder today.</li>
</ul>
<p><strong>Predbat version</strong></p>
<p>v9.0.3</p>
<p><strong>Environment details</strong></p>
<ul>
<li>Inverter and battery setup: AlphaESS hybrid inverter (DC-coupled, 5 kW AC limit), 13.68 kWh battery, controlled via a Modbus glue script using Predbat's GE-style control entities. 9.6 kWp PV on two strings / two Solcast sites.</li>
<li>Standard HAOS installer or Docker: HAOS (VM), Predbat as the HA add-on.</li>
<li>Anything else?:
<ul>
<li>PV forecast from the Solcast HA integration.</li>
<li><code>metric_pv_calibration_enable</code> switched on 2026-09-10, <code>pv_scaling</code> 1.0.</li>
<li>Tariff: Intelligent Octopus Go import, fixed export.</li>
<li>UK.</li>
</ul>
</li>
</ul>
<p><strong>Evidence from this system</strong></p>
<ol>
<li>
<p><strong>The <code>h0</code> state is the calibrated value</strong> (snapshot 2026-09-16 15:28):</p>

sensor.predbat_pv_forecast_h0 | value
-- | --
state | 3.9
attribute nowCL | 3.9
attribute now (raw) | 4.06

<p>Calibration clearly helps, but about 8% of the bias remains, which is what a √ correction predicts. Applying the full ratio (×0.816) would have put the total bias at about −1%. Five days is a small sample; the source analysis and the model below are the main evidence.</p>
</li>
<li>
<p><strong>Toy model</strong> — a day-level version of the same 7-day weighted ratio-of-sums, run with and without the feedback:</p>
<pre><code class="language-python">import random
def run(true_ratio, days=40, feedback=True, seed=1):
    random.seed(seed)
    raw, actual, hist, out = [], [], [], []
    for d in range(days):
        r = 30 * random.uniform(0.6, 1.4)
        raw.append(r); actual.append(true_ratio * r)
        num = den = 0.0
        for back in range(1, 8):
            i = d - back
            if i &lt; 0: continue
            w = max(1.0 - 0.1 * (back - 1), 0.3)
            num += actual[i] * w
            den += (hist[i] if feedback else raw[i]) * w
        f = num / den if d &gt;= 3 and den &gt; 0 else 1.0
        f = min(max(f, 0.1), 2.0)
        hist.append(f * r if feedback else r)   # what pv_forecast_h0 records
        out.append(f)
    return out
</code></pre>
<pre><code>true ratio 0.82 (sqrt 0.906)
  with feedback (as shipped): 1.000 1.000 1.000 0.820 0.856 0.908 0.904 0.904 0.906 0.905 0.906 ...
  learning from raw history : 1.000 1.000 1.000 0.820 0.820 0.820 0.820 0.820 0.820 0.820 0.820 ...
true ratio 0.70 (sqrt 0.837)   with feedback → 0.837   raw history → 0.700
true ratio 1.20 (sqrt 1.095)   with feedback → 1.095   raw history → 1.200
</code></pre>
<p>The factor reaches the true ratio as soon as calibration starts. It is then pulled back to the square root once its own output enters the 7-day window.</p>
</li>
</ol>
<p><strong>Related, and what I checked before filing</strong></p>
<ul>
<li><strong>Settings.</strong> The only calibration setting is <code>metric_pv_calibration_enable</code> (<code>config.py</code>). There's no option to choose which history calibration learns from, and nothing in <code>docs/</code> describes the partial correction as intended. <code>apps-yaml.md</code> says calibration "compares the last seven days of recorded forecasts against actual generation".</li>
<li><strong>#4390.</strong> Fixed the same <code>pv_forecast_h0</code> self-reference, but only for the cap (<code>max_pv_power_forecast</code>). The ratio calculation still reads the same history.</li>
<li><strong>#3986 / #4023.</strong> About curtailed export being mistaken for underperformance. That's a different input problem.</li>
<li><strong>Another consumer of the sensor.</strong> <code>web.py</code> L3345–3346 (PV chart) also treats the <code>h0</code> state as the raw forecast. It plots it as "Forecast History" next to the <code>nowCL</code> attribute as "Forecast History CL", so with calibration on those two series are identical. That suggests the state was meant to be raw.</li>
</ul>
<p>Apologies if I've missed something that already accounts for this.</p>
<p><strong>How this was found</strong>
I was building a Home Assistant chart comparing day-ahead Solcast and Predbat forecasts against actual generation. Predbat's calibrated figure was clearly better than Solcast's, but it was still consistently ~8% high, which is odd for a correction meant to remove systematic bias. Reading <code>pv_calibration()</code> showed where the forecast history comes from, and the <code>dashboard_item</code> call above it showed what that sensor publishes. A live state check confirmed state = <code>nowCL</code> ≠ <code>now</code>. The fixed-point algebra predicted f = √ratio, and recomputing the ratio against raw Solcast (0.824 → 0.908) matched the logged <code>0.9073x</code>.</p>
<p><strong>Screenshots</strong>
Not applicable. The sensor snapshot and day-ahead table above take their place.</p>
<p><strong>Log file</strong>
Excerpt from the calibration run (the full log is not attached because it contains Octopus account details):</p>
<pre><code>2026-09-16 15:28:40 SolarAPI: PV Calibration: Past day 7 had 42.22 kWh of forecast PV, and actual 37.8 kWh PV generation (weight 0.4)
2026-09-16 15:28:40 SolarAPI: PV Calibration: Past day 6 had 14.01 kWh of forecast PV, and actual 14.1 kWh PV generation (weight 0.5)
2026-09-16 15:28:40 SolarAPI: PV Calibration: Past day 5 had 33.4 kWh of forecast PV, and actual 28.0 kWh PV generation (weight 0.6)
2026-09-16 15:28:40 SolarAPI: PV Calibration: Past day 4 had 20.74 kWh of forecast PV, and actual 24.2 kWh PV generation (weight 0.7)
2026-09-16 15:28:40 SolarAPI: PV Calibration: Past day 3 had 31.5 kWh of forecast PV, and actual 29.0 kWh PV generation (weight 0.8)
2026-09-16 15:28:40 SolarAPI: PV Calibration: Past day 2 had 26.73 kWh of forecast PV, and actual 28.7 kWh PV generation (weight 0.9)
2026-09-16 15:28:40 SolarAPI: PV Calibration: Past day 1 had 39.81 kWh of forecast PV, and actual 28.9 kWh PV generation (weight 1.0)
2026-09-16 15:28:40 SolarAPI: PV Calibration: Worst day scaling factor 0.8, best day scaling factor 1.29 average day scaling factor 0.91 max historical power 8.4 max future predicted power 6.5
2026-09-16 15:28:40 SolarAPI: PV Calibration: PV production: 54.63 kWh, Total forecast: 60.21 kWh, adjustment 0.9073x max_hist 8.4kW max_forecast 6.5kW ...
</code></pre>
<p>Past days 1–6 (10–15 Sep) are after calibration was switched on, so their "forecast PV" is the calibrated value. For example, day 1 shows 39.81 while Solcast's raw figure for that day averaged ~42.0.</p>
<p><strong>Predbat debug yaml file</strong>
Not attached: this isn't a plan issue, and the source lines above reproduce it. Happy to provide one if it helps.</p></body></html>

springfall2008:
This is an automated draft PR generated from issue #5116, since reworked following maintainer review — a maintainer should review it before merging.

Fixes #5116

## Summary

`pv_calibration()` measured actual generation against the **state** history of `sensor.predbat_pv_forecast_h0`. While `metric_pv_calibration_enable` is on, that state is the *calibrated* forecast (`power_nowCL`), yet the factors derived from it are applied to the *raw* series (`pv_forecast_minute`). Calibration was measuring the error of its own output, so the applied factor settled at `f = sqrt(actual / forecast)` rather than the true ratio, correcting only about half of a systematic bias. That matches the reporter's algebra and log (`0.9073x` where the true ratio was ~0.82).

Calibration now reads a new sensor, **`sensor.predbat_pv_forecast_h0_uncalibrated`**, whose state is the provider's forecast for now.

### Why a new sensor rather than h0's `now` attribute

The first version of this PR read h0's `now` attribute, which holds the same raw value. Attribute history isn't reliable from Home Assistant, though. `get_history_window()` (`ha.py`) calls `/api/history/period` with HA's default `significant_changes_only=1`, which returns only the rows where an entity's **state** changed. A `now` that moves while the calibrated state rounds flat is recorded but never returned. In the worst case, a small calibration factor pins the state at `0.00` and starves calibration of the very history it needs to correct itself. `db_enable` defaults to off, so this is the default install's path. Predbat's own DB, by contrast, records attribute-only changes.

Putting the value in a state of its own means every change to it is recorded and returned. It also can't flip meaning when calibration is toggled, as h0's state does.

### The sensor

- Published by `publish_pv_stats()` next to h0, as `dp2(power_now)` — the same value as h0's `now` attribute.
- **Not** scaled by `pv_scaling`. Calibration scales it on read, as before, so a `pv_scaling` change takes effect across the whole window at once instead of decaying in over a week. Storing the scaled value would leave the factor wrong for a week after any change.
- Published whether or not calibration is on, so its history is there when calibration is switched on.
- `sensor.predbat_pv_forecast_raw` was not usable as a name: it already exists (the packed post-calibration minute series) and is in the FAQ's recorder `exclude` example.

### Cutover: merged on read, because HA history can't be back-dated

On the release that adds the sensor it has no history, which would disable calibration (`hist_days < 3`) for three days for every user. HA's REST API writes a state at *now* and the recorder has no state-import service, so a backfill can't be written. It is merged on read instead, in the new `pv_forecast_history()`:

1. Read the uncalibrated sensor, scaled by `pv_scaling`.
2. **Only if** that history doesn't reach the start of the window, fetch h0 and take the points strictly older than the sensor's first point. These come from `now`, falling back to the state per point — the earlier commits' `history_attribute(..., fallback_to_state=True)`. `now` was added in the same change that made the state calibrated (`c448c9ff`), so a point without it has a raw state.
3. Legacy points go first and sensor points after, so the result stays oldest-first for `prune_today()`.

This is permanent but lazy. It costs nothing once the sensor covers the window, and it also covers fresh installs and purged recorders. If neither sensor yields anything, a warning is logged.

### Also in this PR

- **PV chart** (`web.py`): "Forecast History" now plots the uncalibrated sensor. With calibration on it was drawing h0's calibrated state, the same line as "Forecast History CL".
- **Docs**: `output-data.md` lists the new sensor (and corrects h0's unit: kW, not Watts). `faq.md` adds it to the sensors that must not be excluded from the recorder. `customisation.md` describes the comparison, the cutover, and the `pv_scaling` hand-check.
- The ceiling comment on `max_pv_power_forecast` is updated to say where that value now comes from.

## Testing

Six new tests. Each was watched failing for the expected reason before the implementation existed.

| Test | Asserts |
|---|---|
| `test_pv_calibration_prefers_uncalibrated_sensor` | Sensor at 2.0 kW vs h0 `now` at 1.0 kW → `0.4`, not `0.8` |
| `test_pv_calibration_scales_uncalibrated_history_by_pv_scaling` | `pv_scaling` applied exactly once → `1.6` (twice `3.2`, never `0.8`, disabled `1.0`) |
| `test_pv_calibration_fills_uncalibrated_cutover_from_h0` | Sensor covers 2 days (below the 3-day minimum), h0 fills the rest → blended ~`0.52`, calibration enabled |
| `test_pv_forecast_history_fetches_h0_only_when_needed` | No h0 request when the sensor covers the window; merged result oldest-first, each point from one sensor |
| `test_publish_pv_stats_publishes_uncalibrated_forecast` | Raw value published with calibration on *and* off, unscaled at `pv_scaling` 0.5, equal to h0's `now` |
| `test_pv_chart_forecast_history_is_uncalibrated` | "Forecast History" plots the sensor, "Forecast History CL" plots `nowCL` |

The synthetic sensor history is deliberately offset 7 minutes from h0's timestamps. With shared timestamps, a merge that forgot to trim the overlap would pass by overwriting identical keys.

**Mutation sweep.** Seven deliberate regressions to `solcast.py` were each run against the suite, and all seven were caught:

- publishing the calibrated value
- publishing the pre-scaled value
- publishing only while calibration is on
- no overlap trim in the merge
- eager h0 fetch
- unscaled h0 fallback
- reverting to reading the h0 state

The last one reproduces the reported bug exactly: `total_adjustment 0.8852` against a true ratio of `0.8`, the square root.

- `./run_all --test solcast --test web_chart_pv_forecast --test web_chart_currency`: passed.
- `./run_pre_commit`: all hooks passed; the quick suite reports `All tests passed (4 slow tests skipped)`. Re-run after `main` was merged into the branch.

## Notes

- `impact` on `publish_pv_stats`, `pv_calibration` and `get_chart`: all **LOW**, one caller each, no signature changes. `detect_changes`: affected flows are `pv_calibration`'s own.
- The `significant_changes_only` behaviour is taken from HA's history API contract and not re-run against a live instance. Comparing row counts for `sensor.predbat_pv_forecast_h0` with and without `significant_changes_only=0` would confirm it and show how degraded the attribute series is. The design doesn't depend on the answer.
- **Follow-up, not folded in:** the PV *Accuracy* chart (`web.py`) reads `totalCL`/`remainingCL` attribute history from `sensor.predbat_pv_today` and has the same exposure. `remainingCL` changes every cycle while the state (`totalCL`) rarely does. It is display-only.


apps/predbat/solcast.py
```diff
@@ -891,10 +891,28 @@ def pv_calibration(self, pv_forecast_minute, pv_forecast_minute10, pv_forecast_m
...
+        # applied to - already carries it (see the minute_data calls in fetch_pv_forecast), and the ratio
+        # only means anything if both sides are on the same basis. A pv_scaling change part way through the
+        # window is applied to the whole of it, which settles as the window rolls over.
+        #
+        # Fall back to the state when no history point carries the attribute at all: "now" was added in the
+        # same change that made the state calibrated (c448c9ff), so a history without it was recorded by a
+        # version whose state was the raw forecast anyway.
+        forecast_history = self.get_history_wrapper("sensor." + self.prefix + "_pv_forecast_h0", days + 1, required=False)
+        forecast_history_raw = history_attribute(forecast_history, state_key="now", attributes=True, scale=self.pv_scaling)
+        if not forecast_history_raw:
```
springfall2008, line 916: _Automated comment from the triage bot._

**High - partial `now` coverage silently truncates the calibration window.** `history_attribute(..., attributes=True)` drops any point lacking the attribute (utils.py:828-829), but this fallback is all-or-nothing: it fires only when the attribute read returns an empty dict. A window mixing pre-attribute and post-attribute rows is truthy, so the fallback never runs and every pre-attribute day is dropped with no warning. Since `now` was added in the same change (c448c9ff) that made the state calibrated, any user upgrading inside the recorder's retention window (default purge_keep_days 10 vs the 8 days fetched here) has exactly this shape for their first days on the new version. Two silent outcomes: with fewer than 3 attribute-carrying days, `hist_days < 3` (line 919) disables calibration for up to a week behind the misleading 'Not enough historical data' log line; with 3 or more, calibration runs on the shortened window and the uncovered days land in `down_days` (missing `past_day_forecast[day]`), so factors are noisier until the window rolls over. The new test covers full `now` coverage and full absence but not the mixed case, so this path is untested. Suggestion: a per-point fallback, e.g. a `fallback_to_state` parameter filling missing points from `item[state_key]` inside history_attribute's existing loop (utils.py:805) - that keeps key order intact for prune_today's `last_time` grouping, which a post-hoc merge of the two dicts would not. The same parameter would serve the other attribute-vs-state read sites (web.py:231/234, 3345/3346, 3372/3373), and a log line counting dropped points would make the failure diagnosable. Minor related nit: the pre-`now` fixture in the new test still sets `nowCL` without `now`, but c448c9ff added both in one change, so that historical shape never actually existed.

apps/predbat/solcast.py
```diff
@@ -891,10 +891,28 @@ def pv_calibration(self, pv_forecast_minute, pv_forecast_minute10, pv_forecast_m
...
+            forecast_history_raw = history_attribute(forecast_history, scale=self.pv_scaling)
```
springfall2008: _Automated comment from the triage bot._

**Medium-high - the fallback can re-introduce the exact GH#5116 self-reference, with a double-scale on top.** The rationale above infers that a history lacking the `now` attribute was recorded by a pre-c448c9ff version whose state was raw. That inference breaks for lossy history sources: Predbat's own DB mirror returns rows with empty `attributes` when the stored attributes JSON fails to decode (db_engine.py:240-244), and on any post-c448c9ff history the h0 state is `power_nowCL` - already calibrated and already carrying `pv_scaling` (it derives from `pv_estimateCL` via `pv_forecast_minute`, which is built with `scale=self.pv_scaling`). If the whole retained window comes back without attributes, the attribute read returns an empty dict, this fallback fires, and calibration measures actuals against calibrated values multiplied by `pv_scaling` yet again - the sqrt fixed point this PR removes, now with extra `1/pv_scaling` error on top. Nothing logs either the decode failure or the fallback being taken, so a 'calibration stopped working after the update' report would be indistinguishable from a genuinely short history. Minimal mitigation: log when the fallback fires (and ideally when the attribute read is empty while the history is non-empty); the per-point fallback suggested above shrinks the trigger to the genuinely-legacy case.

apps/predbat/solcast.py
```diff
@@ -891,10 +891,28 @@ def pv_calibration(self, pv_forecast_minute, pv_forecast_minute10, pv_forecast_m
...
+        # The history is scaled by pv_scaling because pv_forecast_minute - the series these factors are
+        # applied to - already carries it (see the minute_data calls in fetch_pv_forecast), and the ratio
+        # only means anything if both sides are on the same basis. A pv_scaling change part way through the
+        # window is applied to the whole of it, which settles as the window rolls over.
+        #
+        # Fall back to the state when no history point carries the attribute at all: "now" was added in the
+        # same change that made the state calibrated (c448c9ff), so a history without it was recorded by a
+        # version whose state was the raw forecast anyway.
+        forecast_history = self.get_history_wrapper("sensor." + self.prefix + "_pv_forecast_h0", days + 1, required=False)
+        forecast_history_raw = history_attribute(forecast_history, state_key="now", attributes=True, scale=self.pv_scaling)
```
springfall2008: _Automated comment from the triage bot._

**Minor - the two paths filter `unavailable` differently.** The state path inside `history_attribute` skips `unavailable`/`unknown` rows (utils.py:836-838); the `attributes=True` path only checks that the key exists in `item['attributes']` (utils.py:827-830). A row whose state went unavailable while its recorded attributes still carried a stale `now` would be accepted by the primary read but rejected by the fallback. Low confidence this happens for this sensor in practice (HA clears attributes when a state goes unavailable, and the DB mirror stores what was published), but the invariant that unavailable rows are excluded is silently dropped on this branch - only the all-or-nothing fallback catches the case where every row is unavailable.

springfall2008: _Automated comment from the triage bot._

**Low-medium - the down-day thresholds now evaluate on raw x pv_scaling.** The `< 0.1` kWh floor and the `0.1 x forecast` ratio below (unchanged context, around line 941) are untouched, but their input now scales with `pv_scaling` through this read, where the old code measured the h0 state, which under calibration carried `pv_scaling` and the calibration factors (clamped 0.2-4.0). The direction of change is deterministic: with `pv_scaling < 1` (the docs recommend decreasing it), a marginal raw day in `[0.1, 0.1/pv_scaling)` kWh flips into `down_days`; enough flips and `valid_days < 3` silently disables calibration. The delta is bounded by the 0.2-4.0 clamp band, so this is a basis note rather than a concrete regression - but worth deciding whether the threshold is meant to compare raw or scaled forecasts.

docs/customisation.md
```diff
@@ -268,6 +268,8 @@ If `pv_metric10_weight` and `pv_metric90_weight` together exceed 1.0 they are sc
 
 **switch.predbat_metric_pv_calibration_enable** When turned On (the default), Predbat will use your historical solar generation data to calibrate your PV production estimates on a slot duration (default 30 minute) basis.<BR>
 This can be useful to adjust for your systems real performance.<BR>
```
springfall2008, line 270: _Automated comment from the triage bot._

**Doc nit - the comparison basis is raw x pv_scaling, not the raw provider figure.** The read passes `scale=self.pv_scaling` (solcast.py:912/914) to match the already-scaled `pv_forecast_minute` series the factors apply to, so a user with `pv_scaling = 0.8` who recomputes actual against the raw provider forecast by hand gets roughly 1.25 while `pv_scaling_total` settles at 1.0. One added sentence ('the history is scaled by pv_scaling so both sides are on the same basis') closes that support question. Related, pre-existing: the h0 sensor section in docs/output-data.md lists the `now10`/`now90` attributes but not `now`/`nowCL`, which this paragraph now relies on as the location of the raw forecast.

apps/predbat/solcast.py
chalfontchubby: Disclosure: posted by Claude (Anthropic's Claude Code), working on Rik's behalf.

Fixed in 91b2d8a7. `history_attribute()` now takes a `fallback_to_state` parameter that substitutes the point's own state, per-point, wherever that individual point lacks the attribute - rather than only falling back when the whole result comes back empty. Verified this was a real gap: added `test_pv_calibration_learns_from_mixed_upgrade_history` with a fixture giving pre/post-upgrade days distinct raw forecast levels (2.0kW vs 1.0kW) so the settled adjustment reveals which days were actually used; confirmed it fails against the old all-or-nothing logic and passes against the fix.

apps/predbat/solcast.py
chalfontchubby: Disclosure: posted by Claude (Anthropic's Claude Code), working on Rik's behalf.

Fixed as a side effect of the per-point fallback in 91b2d8a7: it's no longer all-or-nothing, so "every point missing the attribute" is now a genuinely exceptional case rather than the routine legacy-history one. If it does happen (e.g. the DB mirror decode failure you describe), the code now logs a warning instead of silently falling back and re-measuring against the calibrated state.

apps/predbat/solcast.py
chalfontchubby: Disclosure: posted by Claude (Anthropic's Claude Code), working on Rik's behalf.

Fixed in 91b2d8a7: the new fallback path checks `item["state"] in ("unavailable", "unknown")` before using it, matching the primary state-path's filter.

chalfontchubby: Disclosure: posted by Claude (Anthropic's Claude Code), working on Rik's behalf.

Left as-is - this is a deliberate, bounded basis choice (clamped 0.2-4.0) rather than a defect, and your comment already frames it that way ("worth deciding whether the threshold is meant to compare raw or scaled forecasts"). Not folded into this fix; flagging here so it isn't lost if someone wants to revisit the down-day threshold basis separately.

docs/customisation.md
chalfontchubby: Disclosure: posted by Claude (Anthropic's Claude Code), working on Rik's behalf.

Added a sentence in 91b2d8a7 covering the `pv_scaling` basis. Left the `now10`/`now90` doc gap as-is since it's pre-existing and out of scope for this fix.

docs/customisation.md
```diff
@@ -268,6 +268,9 @@ If `pv_metric10_weight` and `pv_metric90_weight` together exceed 1.0 they are sc
...
+The comparison is made against the raw forecast your solar provider gave, recorded in the `now` attribute of `sensor.predbat_pv_forecast_h0`, rather than against Predbat's own calibrated figure,
+so the scaling factor settles on the full measured ratio of actual to forecast generation.<BR>
+That history is scaled by `pv_scaling` before the comparison, so both sides are on the same basis - if you want to sanity-check the calibration factor by hand against the raw provider figure, multiply your result by `pv_scaling` first.<BR>
```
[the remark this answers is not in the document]
springfall2008: _Automated comment from the triage bot._

Correct, and fixed in abe83a67. Verified the direction from the code rather than the prose: `history_attribute` applies `scale` as a multiply (`state = float(state) * scale`, utils.py:853) and the calibration read passes `scale=self.pv_scaling` (solcast.py:915), so `total_adjustment = total_production / total_forecast` (solcast.py:1080) is `actual / (raw x pv_scaling)`. A hand-computed `actual / raw` therefore has to be *divided* by `pv_scaling` to land on Predbat's factor, not multiplied.

The sentence was also ambiguous about which quantity "your result" referred to, so rather than flipping the operator it now names both sides explicitly: compare actual generation against the raw provider forecast **multiplied by** `pv_scaling`, with the divide-your-ratio form given as the equivalent. Docs-only change; pre-commit and the `solcast` test module both pass.

Issue #2444, opened 2025-06-10
Growatt via Solax Modbus Addon: Unable to read inverter time string


MuadDib007:
**Describe the bug**
When using the rtc sensor from the Solax Modbus Integration (Connected to my Growatt MOD10KTL3-XH(BP)) as input for inverter_time in apps.yaml I receive the following error in the logs:

` Warn: Unable to read inverter time string 2025-06-10 15:44:42 using formats ['%Y-%m-%dT%H:%M:%S%z', '%Y-%m-%d %H:%M:%S%z', '%Y-%m-%dT%H:%M:%S']`

**Expected behaviour**
Expect Predbat to use this time string format without issue

**Predbat version**
v8.20.14

**Environment details**

- Growatt MOD10KTL3-XH(BP) with 30 kWh APX Batteries connected via Solax Modbus integration.
- HAOS running inside Vmware ESXi

**Screenshots**

![Image](https://github.com/user-attachments/assets/528ce62e-8033-48ca-8f7c-5ffeffc5f1d9)

Entity state of rtc:

![Image](https://github.com/user-attachments/assets/1d9c599a-65e3-401c-9c34-2b634f970bc6)

**Log file**
[predbat (2).log](https://github.com/user-attachments/files/20674062/predbat.2.log)

**Predbat debug yaml file**
[predbat_debug.txt](https://github.com/user-attachments/files/20674053/predbat_debug.txt)

springfall2008:
You can probably just comment out the inverter time for now, but its just a different time format

chalfontchubby:
From the issue post this looks like a space separator, no fractional seconds, a combination missing from the list of strings. The bot should be able to confirm that and add the required string.

springfall2008:
🤖 Automated first-pass triage — a maintainer will review before any action is taken.

**Classification:** bug (a small format-coverage gap in the inverter-time parser), `priority_low`. The existing `enhancement` framing fits too — it's a "support another time format" fix. The maintainer's workaround stands: remove `inverter_time` from apps.yaml (it only feeds clock-skew detection) or reformat the sensor via a template.

**What I checked** (against main at v9.0.3-35-g002f6b86, 2026-09-16):

- The attached log + debug yaml confirm the setup: `inverter_type: SA`, `inverter_time: [sensor.g10k_rtc]` (the Solax Modbus `rtc` sensor), `timezone: Europe/Amsterdam`.
[...]

PR #5126, opened 2026-09-17
fix(inverter): Persist inverters across planning cycles

### LINKED ISSUE #2328: Unneccessay updates to inverter causing eeprom wear
**Describe the bug**
Looking through the logs I see:

<html><body>

6140 | 2025-05-04 11:05:15.770579:  Inverter 0 Updated start and end charge window to 23:59:00 - 12:30:00 (old 23:59:00 - 12:30:00)
-- | --
6139 | 2025-05-04 11:05:15.770421:  Successfully pressed button button.solis8_update_discharge_times on Inverter 0
6138 | 2025-05-04 11:05:11.671597:  Successfully pressed button button.solis8_update_charge_times on Inverter 0
6137 | 2025-05-04 11:05:07.483607:  Inverter 0 write_and_poll_value: No write needed for charge_end_minute: 30 == 30 fuzzy 0
6136 | 2025-05-04 11:05:07.483543:  Inverter 0 write_and_poll_value: No write needed for charge_end_hour: 12 == 12 fuzzy 0
6135 | 2025-05-04 11:05:07.483472:  Inverter 0 write_and_poll_value: No write needed for charge_end_time: 12:30:00 == 12:30:00 fuzzy 0
6134 | 2025-05-04 11:05:07.483413:  Inverter 0 write_and_poll_value: No write needed for charge_start_minute: 59 == 59 fuzzy 0
6133 | 2025-05-04 11:05:07.483344:  Inverter 0 write_and_poll_value: No write needed for charge_start_hour: 23 == 23 fuzzy 0
6132 | 2025-05-04 11:05:07.483268:  Inverter 0 write_and_poll_value: No write needed for charge_start_time: 23:59:00 == 23:59:00 fuzzy 0
6131 | 2025-05-04 11:05:07.483128:  Adjust demand (idle) time computed is 12:30:00-23:59:00
6130 | 2025-05-04 11:05:07.482717:  Adjust idle time, charge 23:59:00-12:30:00 discharge 00:00:00-00:00:00

</body>
</html>

So ... it has written and polled some values ... and all of them got "No Write Needed" ... so at the end of the process, no button press was needed, yet, it still triggered a button press.

**Expected behaviour**
If no values are written, dont trigger a button press.

**Predbat version**

current, atest

**Environment details**

- Solis, but whatever
- 
**Screenshots**
If applicable, add screenshots to help explain your problem. The most useful ones can be your battery chart, the Predbat HTML plan and your current settings in HA.

**Log file**
Can you capture a log file from the time of the issue, debug mode is not normally required.

**Predbat debug yaml file**
Enable debug mode and let Predbat run for one cycle. Look in the 'debug' directory under the Predbat add-on and take the .yaml file for the time period in question.
Rename the predbat_debug_XX_XX_XX.yaml to predbat_debug_XX_XX_XX.yaml.txt and upload it to Github. This will allow your plan to be reproduced and also stores all your settings for review.

### LINKED ISSUE #4712: Inverter update button pressed on every idle cycle when no export is scheduled
## Summary

For inverters that need a button press to commit a schedule, Predbat presses the update button on **every** 5-minute cycle whenever *nothing* is being exported — i.e. for most of the day, not just during export windows.

Found while investigating #4709 (which covers the separate in-window case). Raised out of the discussion on #2328, whose reporter's original complaint was exactly *"no values were written, yet it still triggered a button press"*.

## Mechanism

`execute.py` calls `adjust_force_export(False)` with **no times at all** whenever no export is scheduled. There are two independent routes from there to the button press, and which one applies depends on the inverter type.

**Route 1 — `None` time comparison.** Inside `adjust_force_export`, `new_start` / `new_end` are `None`. The midnight-override branch that would replace them with `"00:00:00"` is skipped for inverters with `has_discharge_enable_time` set, so they stay `None` — while `old_start` / `old_end` still read back a real time string. `None != "00:00:00"` is never equal, so every idle cycle looks like a change.

**Route 2 — the `H M` unconditional write guards.** For `charge_time_format: "H M"`, the start/end write guards (`new_start != old_start or is_hm_format`) fire regardless, setting `changed_start_end = True`, which is the final term of the same `or`.

## Affected inverter types

Types with `time_button_press: True`:

| Type | Route | Notes |
|---|---|---|
| `GS_fb00` | 1 **and** 2 | Solis FB00, solax_modbus |
| `GS` | 2 only | Solis pre-FB00 |
| `FoxCloud`, `TESLA`, `EnphaseCloud`, `DeyeCloud`, `SunsynkCloud`, `AlphaESSCloud` | 1 only | `HH:MM:SS` format |

Measured on unmodified `main` (`b08b7716`), four idle `adjust_force_export(False)` calls, counting presses:

```
main    GS:      4 idle cycles -> 4 presses
main    GS_fb00: 4 idle cycles -> 4 presses
```

**Correction to an earlier revision of this issue:** it originally stated that plain `GS` was unaffected. That was wrong — it was measured with the #4711 fix already applied, not against `main`. `GS` is affected on `main`, via route 2.

## Interaction with #4711 / #4713

The two routes are fixed by different PRs, and neither alone is sufficient:

```
             GS      GS_fb00
main         4/4     4/4
+#4711       1       4/4     <- closes route 2
+#4713       1       1       <- closes route 1
```

- **#4711** drops `changed_start_end` from the condition and adds a commit-once check, closing route 2. `GS` settles after the first idle cycle.
- **#4713** closes route 1, still needed for `GS_fb00` and the cloud types — their `old_start` / `old_end` are never overwritten (nothing is written when `new_start` is `None`), so the `None` comparison stays true forever past #4711.

## Impact

Roughly 288 presses/day rather than a handful. On Solis this matters twice over: the presses are non-volatile register writes (the EEPROM-wear concern in #2328), and each press also zeroes the timed current registers (#4415 / #4709).

## Logs would help

**Confirmed by unit-level reproduction, but not yet against a live install**, and six of the eight affected types are cloud inverters that are hard to test here.

What would confirm it, from anyone running one of the above:

- `predbat.log` covering a full day, so it spans both an export window and a long stretch outside one
- Home Assistant history for the update button entity/entities over that same period
- masked `apps.yaml` — particularly `inverter_type` and which button entities are set

The specific thing to look for: **do the button presses continue at the same 5-minute cadence when no export is scheduled?**

Partial corroboration already exists in #2328's original log — every register reports `No write needed` and both buttons are pressed anyway — though the call site can't be pinned from those lines alone.

### LINKED ISSUE #5127: Solis: confirm whether GS_fb00 + charge_limit still micromanages charging
Split out of #2328 to survive that issue's close (fixed by #5126, unrelated mechanism).

**Unconfirmed report, needs fresh evidence.** @rszemeti reported hundreds of inverter writes from Predbat "micromanaging" SoC hold instead of letting the inverter's own timed-charge-to-target-SoC handle it. Investigation in #2328 established that GS_fb00 + a configured `charge_limit` entity *should* use the inverter's own target-SoC hold rather than software micromanagement - but the reporter confirmed they were on GS_fb00 with `charge_limit` pointed at a real entity (`number.solis8_timed_charge_soc`) and still saw the pattern:

> Let me check my config, but I am on GS_fb00 ... and yes, charge_limit is set and pointed at number.solis8_timed_charge_soc. I went through all the hoops suggested at the time, it didn't help.

No log against current head was ever supplied, so this was never actually confirmed against real code - see the full back-and-forth in #2328 for the detailed triage.

**To make progress on this** we need, from anyone on GS_fb00 with `charge_limit` configured and seeing frequent writes:
- `predbat.log` spanning several consecutive cycles
- masked `apps.yaml` (particularly `inverter_type` and `charge_limit`)
- a `predbat_debug.yaml` capture

Without that, this stays parked - not because it isn't real, but because we can't distinguish "still broken" from "config predates a since-fixed behaviour" without evidence against current main.

chalfontchubby:
*(Written by Claude, posted by @chalfontchubby.)*

## Problem

The update-button-commit mechanism on `time_button_press` inverters (Solis and others) was pressing the button every 5-minute cycle instead of once per genuine schedule change. A reporter on v9.0.3 supplied full logs (`predbat.log`/`apps.yaml`/`debug.yaml`) showing 377 presses in one night — every register logging "No write needed", the button pressed anyway, and `count register writes 0` on the same cycle, hiding the repeats from the write counter entirely.

That's the same signature as the original #2328 report:
```
Inverter 0 write_and_poll_value: No write needed for charge_start_time: 23:59:00 == 23:59:00 fuzzy 0
...
Successfully pressed button button.solis8_update_charge_times on Inverter 0
Successfully pressed button button.solis8_update_discharge_times on Inverter 0
```
On Solis this is a non-volatile write each time (EEPROM wear), and each press also zeroes the timed current registers (#4415/#4709).

## Root cause

`fetch_inverter_data()` defaults to `create=True` and runs one or more times per cycle — the plan run re-reads the inverter again before `execute_plan()` — so the `Inverter` objects were rebuilt before any commit-once guard could ever latch. #4711 added exactly such a guard on the export side, and it worked in the unit tests (which reuse one long-lived `Inverter`) but did nothing in production, because the object it lived on didn't survive to the next cycle.

The rebuild itself was never a deliberate design — before #1803 (Dec 2024) `fetch_inverter_data()` had no `create` parameter at all, it always rebuilt. That PR added the parameter to allow reuse but kept `create=True` as the default, and only one caller ever opted out, so the main loop kept rebuilding regardless.

## Fix

**Commit 1** — persist the `Inverter` objects across cycles instead of rebuilding them. `__init__` is split into a one-time identity half and a `refresh_config()` that runs every cycle, re-reading everything that's genuinely runtime-changeable (`battery_min_soc`, `battery_scaling`, the rate limits, `battery_calibration`) so persisting the object doesn't freeze a setting a user can change while Predbat runs. Also adds the charge-side commit-once guard that never existed (`adjust_charge_window()` had no equivalent to the export side's protection), with the same write-result accumulation (`schedule_write_ok`) that review required on the export side in #4711.

Measured with a real reporter config (GS, hour/minute entities only, a `charge_discharge_update_button`), driven through `fetch_inverter_data()` as production does:
```
main   presses/cycle = [2, 1, 1, 1, 1, 1]
fixed  presses/cycle = [2, 0, 0, 0, 0, 0]
```

**Commit 2** — persisting the objects exposed a mess in `__init__`: real one-shot state, real `None` sentinels, and placeholder values indistinguishable from genuine config (`reserve_percent = 4.0`, `battery_rate_max_raw = 2600.0`) were all set the same way in the same block. Harmless while the object was rebuilt every cycle; not harmless once it persists, since a stale placeholder can now live long enough to be read for real. Regrouped by actual lifetime into a new `reset_cycle_state()`, seeded to `None` rather than a real-looking value wherever nothing depends on a specific default.

That pass caught two real bugs the persistence work introduced along the way:
- `in_calibration` was write-only — only ever set `True`, never back `False` — previously masked by `__init__` reseeding it every cycle. On a persisted object a battery leaving calibration would stay stuck. Now a real two-way read.
- `battery_voltage` was seeded `52.0`, a guessed value used as a divisor in `set_current_from_power()`, reachable before `update_status()` has ever fetched a real reading. Seeded `None` instead; the divide site now checks explicitly and skips with a warning rather than computing from a made-up voltage.

## Scope

Deliberately **not** persisted across Home Assistant restarts — that would need HA storage, which is overkill for a cost of one re-commit on reboot.

The register-write counter fix (a button press wasn't counted, hiding the repeats from `count_register_writes`) is independent and already up as #5123.

## Testing

- Full quick suite green, pre-commit clean, on both commits individually and at the tip
- New regression tests: a commit survives object recreation-that-no-longer-happens, a stuck `scheduled_charge_enable` commits once not every cycle, a failed write/press retries on the next cycle rather than being recorded as done, a same-count `inverter_type` change at runtime rebuilds correctly, `in_calibration` clears when `battery_calibration` goes off, `set_current_from_power()` doesn't crash before `update_status()` has run
- `run_shuffle --hunt --quick` run against a static tree: 50/50 clean, no order-dependency pair found (#5079)

Fixes #4712. Fixes #2328 - matches that report's log signature exactly (every register "No write needed", button pressed anyway). Follow-ups were split off #2328 and #4712's later comments so they survive this close as their own tracked items, unrelated to this PR's fix: #5127 (unconfirmed GS_fb00 + `charge_limit` micromanagement report, needs fresh logs), #5128 (Feed-in priority modes 64/96/98 never reachable), #5129 (`charge_limit` should default on for GS_fb00), #5130 (a register read-back briefly bouncing through a wrong value after a write - a write-settling question, not this PR's object-recreation fix).


apps/predbat/inverter.py
```diff
@@ -306,24 +306,115 @@ def create_missing_arg(self, arg, default):
...
+        Split out of __init__ so the Inverter objects can persist across cycles. Everything here is
+        either live config a user can change at runtime (battery_min_soc, battery_scaling, the rate
+        limits, battery_calibration) or a per-cycle side effect (battery size tracking, clock skew
+        detection), so it has to keep running every cycle even though the object no longer does.
+        What stays in __init__ is the identity and the accumulated state that must survive - notably
+        the committed-schedule guard, which was being wiped by the rebuild (#4712).
+        """
         self.reserve_percent = self.base.get_arg("battery_min_soc", default=4.0, index=self.id, required_unit="%")
         self.reserve_percent_current = self.base.get_arg("battery_min_soc", default=4.0, index=self.id, required_unit="%")
         self.battery_scaling = self.base.get_arg("battery_scaling", default=1.0, index=self.id)
```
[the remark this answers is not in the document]
chalfontchubby: Fixed in 58c634bd — refresh_config() now resets inverter_limit/export_limit to the standing defaults (7500W/99999W) before the conditional re-read, so a key removed at runtime no longer leaves a stale override in place.

apps/predbat/tests/test_inverter.py
```diff
@@ -2874,6 +2880,314 @@ def test_force_export_unchanged_times_HM_format(test_name, ha, inv):
...
+    reports it back, that third term stays true forever, so the button was pressed every 5 minutes
+    for the whole window - the same failure #4711 fixed on the export side only. On Solis each press
+    is a non-volatile write (#2328) and zeroes the timed current registers (#4415/#4709).
+    """
+    failed = False
+    print("Test: {}".format(test_name))
+
+    # The suite shares one PredBat/HA fixture across every registry test, so anything mutated here
+    # has to be put back or a later test runs against altered config (batpred#5079, and the same
+    # review point raised on batpred#4711).
```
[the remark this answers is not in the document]
chalfontchubby: Fixed in 58c634bd — reworded to say run_inverter_tests() shares the fixture across its scenarios, not the whole suite.

chalfontchubby:
(Posted by Rik; working with Claude.)

Both Copilot findings on this PR are fixed in 58c634bd:

**refresh_config() left a stale inverter_limit/export_limit override in place.** Both were only overwritten `if "inverter_limit"/"export_limit" in self.base.args`, so removing either key from config at runtime (or a test popping it) left the previous cycle's value in place indefinitely instead of reverting to the standing default. Now reset to the default (7500W/99999W) at the start of every refresh, before the conditional re-read.

**A test comment described the wrong scope.** Said "the suite shares one PredBat/HA fixture across every registry test" when that's actually true of `run_inverter_tests()`'s own scenarios, not the whole suite. Reworded.

Both covered by the existing `inverter` test suite, green.

apps/predbat/inverter.py
```diff
@@ -3221,29 +3305,37 @@ def adjust_charge_window(self, charge_start_time, charge_end_time, minutes_now):
...
                     else:
-                        self.write_and_poll_option("charge_start_hour", start_hour_id, int(new_start[:2]))
+                        schedule_write_ok = self.write_and_poll_option("charge_start_hour", start_hour_id, int(new_start[:2])) and schedule_write_ok
                     start_minute_id = self.base.get_arg("charge_start_minute", indirect=False, index=self.id)
                     if start_minute_id and isinstance(start_minute_id, str) and start_minute_id.startswith("time."):
-                        self.write_and_poll_option("charge_start_minute", start_minute_id, new_start)
+                        schedule_write_ok = self.write_and_poll_option("charge_start_minute", start_minute_id, new_start) and schedule_write_ok
                     else:
-                        self.write_and_poll_option("charge_start_minute", start_minute_id, int(new_start[3:5]))
+                        schedule_write_ok = self.write_and_poll_option("charge_start_minute", start_minute_id, int(new_start[3:5])) and schedule_write_ok
```
springfall2008, line 3357: **Blocking:** this counts a write to an unconfigured entity as a failure, which latches the commit permanently pending and presses the button every cycle — the exact bug this PR fixes.

`adjust_force_export()` guards the same writes with `elif start_hour_id:` / `elif start_minute_id:` (lines 2786/2790); the charge side calls `write_and_poll_option()` unconditionally. With `charge_time_format "H M"` these rewrite every cycle regardless of change, so an unmapped (or literal-valued) `charge_start_minute` makes `check_write_entity()` return False forever -> `schedule_write_ok=False` -> `committed=False` -> `charge_schedule_commit_pending=True` -> the `elif pending` branch at 3412 forces `schedule_changed=True` -> `press_and_poll_button()` every 5 minutes, reintroducing the #2328 EEPROM wear and the #4415/#4709 current-register zeroing.

Reproduced on this branch, GS config with the two minute entities unmapped:

```
cycle 1: presses=1 pending=True committed=None
cycle 2: presses=1 pending=True committed=None
...
TOTAL presses over 5 identical cycles: 5
```

With all four mapped it behaves as intended — 1 press, then 0. The new test at `test_inverter.py:2947` maps every key and its comment acknowledges this ("an unmapped one fails and would keep the commit permanently pending"), so the case is known but not handled. Copying the export side's `elif` guards over fixes it.

apps/predbat/inverter.py
```diff
@@ -306,24 +306,115 @@ def create_missing_arg(self, arg, default):
...
         self.soc_max = None
         self.nominal_capacity = None
+        # Not a placeholder like the rest of this group: these are the real standing defaults
+        # (7500W / 99999W), only conditionally overridden if the user configures them - see the
+        # `if "inverter_limit" in self.base.args:` guard in refresh_config(). They also double as
+        # get_arg()'s own fallback when the key exists but the read is unusable, so they must
+        # always hold a real number, never None.
         self.inverter_limit = 7500.0 / MINUTE_WATT
         self.export_limit = 99999.0 / MINUTE_WATT
         self.inverter_time = None
```
springfall2008, line 374: **Blocking:** moving this seed out of the per-cycle path makes a dropped `inverter_time` reading keep the last good timestamp.

At line 582 `if ivtime:` skips the whole parse when the reading is absent or unavailable. Pre-PR `__init__` ran every cycle and reseeded this to None (main `inverter.py:326`), so the behaviour the comment at 576-579 promises held — "treated as 'no reading' — skew detection is skipped rather than misreporting it as inverter clock skew or triggering an auto-restart". Now that promise is broken: `self.inverter_time` keeps the stale stamp, `tdiff` grows 5 minutes per cycle, and at `INVERTER_CLOCK_SKEW_RESTART_MINUTES` `check_clock_skew()` calls `auto_restart()` — restarting the user's integration for a clock that is perfectly correct. A GivEnergy account losing its Premium entitlement, or any integration dropout, is enough.

Reproduced on this branch:

```
PROBE first refresh (good reading): inverter_time = 2026-09-19 12:00:00+01:00
PROBE after 'unavailable':          inverter_time = 2026-09-19 12:00:00+01:00
PROBE RESULT: STALE - skew detection runs against a dead reading
```

One line to fix: set `self.inverter_time = None` above the `if ivtime:` at 582 rather than relying on the construction-time seed. (Still applies after merging current main, which rewrote that block for #2444 but kept the None seed inside the `if`.)

apps/predbat/execute.py
```diff
@@ -951,6 +968,12 @@ def fetch_inverter_data(self, create=True):
                 self.inverters.append(inverter)
             else:
                 inverter = self.inverters[id]
+                try:
+                    inverter.refresh_config(quiet=True)
+                except Exception as e:
+                    self.log("Error: Failed to refresh inverter {}: {}, your configuration may be incorrect".format(id, e))
+                    self.inverters = []
+                    return False
             inverter.update_status(self.minutes_now, quiet=not create)
```
springfall2008: **Should fix before merge:** `create` now means "the very first cycle", so this silently switches every per-cycle inverter diagnostic off for the life of the process.

Pre-PR `fetch_inverter_data()` defaulted to `create=True` and `predbat.py:1083/1118/1139` used that default, so `quiet=False` on every call and `update_status()` logged each cycle. Now `create` is False from cycle 2 onwards, so all four `if not quiet:` blocks in `update_status()` (SoC/charge rate/discharge rate/battery power/voltage/grid/load/PV at 1599, "scheduled charge enable is" at 1704, charge windows and settings at 1719/1728, export windows at 1814/1853) plus `refresh_config()`'s own summary lines (604 "Inverter time ... difference N minutes", 660 "Inverter N with soc_max ...") emit once at startup and never again.

Every bug report that attaches `predbat.log` then has no inverter state history at all — which is the primary triage evidence for "the plan is wrong" and inverter-write reports, and is what this PR itself was diagnosed from. Nothing in the description suggests this was intended. Worth decoupling the two: keep the verbose cadence (or gate it on something deliberate) rather than tying it to object creation.

apps/predbat/inverter.py
```diff
@@ -3279,6 +3371,11 @@ def adjust_charge_window(self, charge_start_time, charge_end_time, minutes_now):
         if (old_charge_schedule_enable == "off" or have_disabled) and (new_start != new_end):
             # Enable scheduled charge if not turned on, unless the start and end are the same (disabled)
             if "scheduled_charge_enable" in self.base.args:
+                # Deliberately not folded into schedule_write_ok. On the hardware this issue is
```
springfall2008, line 3397: This comment is convincing — and if it's right, it means `adjust_force_export()` has the same bug, so only half of the reported 377 presses per night is fixed.

Line 2853 does the opposite of what this comment argues: with `force_export` True it folds `write_and_poll_switch("scheduled_discharge_enable", ...)` into `schedule_write_ok`. On the same hardware that returns False every cycle -> `committed` False -> `last_export_schedule_committed` never recorded -> line 2880 `if is_hm_format and export_schedule != self.last_export_schedule_committed:` forces `schedule_changed=True` -> `press_and_poll_button(side="discharge")` every cycle for the whole export window.

The #2328 log shows both `button.solis8_update_charge_times` **and** `button.solis8_update_discharge_times` pressed each cycle. Pre-existing from #4711 rather than introduced here, but this PR is the one claiming to close #2328, so it's worth either applying the same exclusion there or explaining why the export side differs.

springfall2008 (commented):
*(Review by Claude, posted by @springfall2008.)*

Reviewed the whole diff and verified the findings against the branch rather than just reading it. Two reproduce on the PR head; details are inline.

**Two things that check out first:** I ran an attribute-lifetime audit comparing main's `__init__` against the PR's `refresh_config()` + `update_status()` — apart from `inverter_time` (noted inline), nothing that used to be reseeded per cycle can now go stale, and both `count_register_writes` (zeroed by `execute.py:788`) and the window lists (rebuilt from `[]`) are safe to persist. I also test-merged into current main: clean, and #2444's multi-format clock parse survives the move into `refresh_config()`.

**Blocking, reproduced:**

1. A partially-mapped `apps.yaml` defeats the entire fix — the charge side folds write results into `schedule_write_ok` without the `elif start_hour_id:` guards the export side has, so one unmapped hour/minute entity latches `charge_schedule_commit_pending` forever and presses every cycle. Measured on this branch with a GS config missing `charge_start_minute`/`charge_end_minute`:

```
unmapped minute entities:  cycles 1-5, presses=1 each, pending=True, committed=None  -> 5 presses
all four mapped:           cycle 1 presses=1, cycles 2-5 presses=0                   -> 1 press
```

2. A dropped `inverter_time` reading now leaves a stale timestamp on the persisted object, so clock-skew detection runs against it and eventually calls `auto_restart()` for a clock that is correct:

```
PROBE first refresh (good reading): inverter_time = 2026-09-19 12:00:00+01:00
PROBE after 'unavailable':          inverter_time = 2026-09-19 12:00:00+01:00   <- stale
```

3. Separately, `quiet=not create` silently removes every per-cycle inverter diagnostic line from `predbat.log` for the life of the process — the evidence this PR itself was diagnosed from.

The core idea (persist the objects, split the per-cycle config re-read out of `__init__`) looks right to me, and the lifetime regrouping in commit 2 is a genuine improvement. The issues below are about the guard's edges, not the approach.

apps/predbat/inverter.py
```diff
@@ -3295,10 +3392,43 @@ def adjust_charge_window(self, charge_start_time, charge_end_time, minutes_now):
...
+        # indefinitely when the inverter does not read the written value back - an enable switch
+        # that never reports "on" pins the third term, and times that never read back pin the first
+        # two - which re-pressed the commit button on every 5 minute cycle for the life of the
+        # window. On Solis each press is a non-volatile write (#2328) and also zeroes the timed
+        # current registers (#4415/#4709), so the repeat is not merely wasteful.
+        #
+        # The observed old state is part of the key so that a window changed underneath us -
+        # externally, or re-presented in a different time format - is still re-committed rather than
+        # mistaken for a repeat of our own last commit.
+        charge_schedule = (new_start, new_end, old_start, old_end, old_charge_schedule_enable)
```
springfall2008: Including the observed `old_start`/`old_end` in the key guarantees it differs on the cycle right after a successful commit, so stuck-readback hardware gets two presses per window change instead of one.

Cycle N: window moves A->C, `old_start=A`, writes land, key `(C,end,A,old_end,"off")` recorded, button pressed. Cycle N+1: the registers now read back C, so the key is `(C,end,C,end,"off")` and no longer matches — while the stuck enable switch still pins the third term of `schedule_changed` at 3405, so the button is pressed a second time before the new key settles. Only from N+2 is it suppressed. On Solis that extra press is another non-volatile write and another zeroing of the timed current registers.

`adjust_force_export()` avoids this by keying on the target values alone: `export_schedule = (new_start, new_end, force_export)` (line 2874). The stated reason for the wider key — re-committing a window changed underneath us — is real, but it's already covered by the target comparison plus the pending flag in the cases that matter. The new test can't catch this: it seeds the dummy items to the same times it then requests, so `old == new` from the first call.

apps/predbat/inverter.py
```diff
@@ -3295,10 +3392,43 @@ def adjust_charge_window(self, charge_start_time, charge_end_time, minutes_now):
...
+
+        if charge_schedule == self.last_charge_schedule_committed:
+            # Already committed this exact schedule; suppress the repeat.
+            schedule_changed = False
+        elif self.charge_schedule_commit_pending:
```
springfall2008: Two ordering issues in this guard.

**The retry is an `elif` under the suppression check.** If a cycle's writes fail while the resulting key happens to equal a previously recorded key — a window that changes and changes back while the readbacks lag — suppression wins at 3409 and the failed commit is never retried, leaving the window programmed but uncommitted until the schedule changes again on its own. That is precisely the failure mode the pending flag was added to prevent. Reversing the two (`if pending: schedule_changed = True` first, `elif key == last_committed: ... = False` second) is strictly safer: a redundant press on a genuinely-committed schedule is the cheaper error here.

**`charge_schedule_commit_pending` is only ever cleared inside `if schedule_changed:`.** When suppression wins, the body at 3419 never runs, so a True flag survives untouched across arbitrarily many cycles rather than being resolved or cleared.

apps/predbat/inverter.py
```diff
@@ -3295,10 +3392,43 @@ def adjust_charge_window(self, charge_start_time, charge_end_time, minutes_now):
...
+            # schedule_changed above - retry instead of leaving the window programmed but
+            # uncommitted until it happens to change again on its own.
+            schedule_changed = True
+
+        if schedule_changed:
             # For Solis inverters and fox we also have to press the update_charge_discharge button to send the times to the inverter
+            button_committed = True
             if self.inv_time_button_press:
-                self.press_and_poll_button(side="charge")
+                button_committed = self.press_and_poll_button(side="charge")
```
springfall2008, line 3434: Third of four `press_and_poll_button()` call sites, and the guard is now on two of them.

`disable_charge_window()` (line 2960) still presses unconditionally whenever `old_charge_schedule_enable` reads back "on" — which on the hardware described at 3374, where the enable write is accepted but never reported back, is pinned True. Types where that path is reachable (button press + `has_charge_enable_time`): `GS_fb00`, `FoxCloud`, `TESLA`, `EnphaseCloud`, `DeyeCloud`, `SunsynkCloud`, `AlphaESSCloud` — several of them cloud APIs, so it's a request per cycle as well as a register write. `adjust_battery_target()` (2157) is the fourth.

This is the third bespoke `last_*_committed` / `*_pending` pair added as each call site gets reported. A single "has this schedule already been committed" check inside `press_and_poll_button()` itself would cover all four at once and stop the next report needing a fourth pair.

apps/predbat/execute.py
```diff
@@ -934,10 +934,27 @@ def fetch_inverter_data(self, create=True):
...
+        # Create the inverters only when we don't already have the right ones. The objects persist
+        # across cycles: rebuilding them every cycle wiped any state they accumulated, which is why
+        # the committed-schedule guard never suppressed a repeated commit (#4712). The config they
+        # read is not cached with them - refresh_config() re-reads it below, so a runtime change to
+        # battery_min_soc, battery_scaling or the rate limits still takes effect immediately.
+        #
+        # The type is checked as well as the count. A component's automatic_config() normally runs
+        # once at startup, before any inverter exists, but a deferred startup (e.g. AlphaESS
+        # retrying until telemetry arrives) can set inverter_type after these objects were built -
+        # and the capability flags derived from it are the half refresh_config() does not re-derive.
```
springfall2008: This justification is factually wrong — `refresh_config()` **does** re-read `inverter_type` and re-derive every capability flag.

`inverter.py:435` re-reads `self.inverter_type`, and 464-492 re-derive `inv_has_mqtt_api`, `inv_output_charge_control`, `inv_charge_time_format`, `inv_time_button_press`, `inv_support_charge_freeze`, `inv_has_idle_time` and the rest from `INVERTER_DEF`; 679-721 re-create the type-named dummy entities. So a deferred `automatic_config()` changing the type is already handled by the refresh path, and this branch throws away `last_charge_schedule_committed` / `last_export_schedule_committed` — forcing exactly the redundant commit (and button press) the PR exists to prevent — plus a `get_arg("inverter_type")` per inverter per cycle.

There may be a real reason to rebuild (stale dummy entities under the old type name is my guess), but as written the comment doesn't state it, and the test at `test_inverter.py:3073` now asserts the rebuild — so a wrong comment has become load-bearing. Either drop the branch or replace the reasoning.

apps/predbat/execute.py
```diff
@@ -951,6 +968,12 @@ def fetch_inverter_data(self, create=True):
                 self.inverters.append(inverter)
             else:
                 inverter = self.inverters[id]
+                try:
+                    inverter.refresh_config(quiet=True)
+                except Exception as e:
+                    self.log("Error: Failed to refresh inverter {}: {}, your configuration may be incorrect".format(id, e))
+                    self.inverters = []
```
springfall2008: Wiping the whole list on one inverter's failure discards every other inverter's accumulated guard state.

`refresh_config()` has transient ways to raise on a live system — `int(self.base.get_arg("set_reserve_min" + postfix))` at `inverter.py:610` raises `TypeError` if that entity reads back None, and `get_arg("inverter_limit_override", 0, ...)` feeding `if inverter_limit_override > 0` at 560 raises on a non-numeric read. On a two-inverter fleet, inverter 1 hiccupping wipes inverter 0's `last_charge_schedule_committed` too, so the next successful cycle re-commits both schedules and presses again on hardware where that's a non-volatile write.

Recovering only the failing inverter (or leaving the list intact and returning False) keeps the guard this PR is built around.

apps/predbat/inverter.py
```diff
@@ -306,24 +306,115 @@ def create_missing_arg(self, arg, default):
...
+        """
+        # Read every cycle by refresh_config().
         self.soc_max = None
         self.nominal_capacity = None
+        # Not a placeholder like the rest of this group: these are the real standing defaults
+        # (7500W / 99999W), only conditionally overridden if the user configures them - see the
+        # `if "inverter_limit" in self.base.args:` guard in refresh_config(). They also double as
+        # get_arg()'s own fallback when the key exists but the read is unusable, so they must
+        # always hold a real number, never None.
         self.inverter_limit = 7500.0 / MINUTE_WATT
```
springfall2008: Nit: these two are unreachable, and they duplicate the magic numbers.

`__init__` calls `reset_cycle_state()` and then `refresh_config()`, and `refresh_config()` unconditionally re-assigns both at 652-653 before anything can read them — so these assignments can never be observed, while carrying six lines of comment explaining why they must not be None. The 7500/99999 pair now appears twice each and has to be kept in sync by hand; a named constant in `const.py` would be better than either copy.

apps/predbat/inverter.py
```diff
@@ -306,24 +306,115 @@ def create_missing_arg(self, arg, default):
...
+        # even though the times now read back as already correct.
+        self.charge_schedule_commit_pending = False
+
+        self.reset_cycle_state()
+
+        # Everything below here is re-read every cycle, not just at construction - see
+        # refresh_config().
+        self.refresh_config(quiet=quiet)
+
+    def reset_cycle_state(self):
```
springfall2008: Nit: the name invites exactly the call the docstring forbids.

"Not called again afterwards" is doing a lot of work here — a future maintainer wiring `reset_cycle_state()` into the per-cycle refresh (which is what the name suggests it is for) would silently null `charge_enable_time`, `soc_kw`, `battery_voltage` and the window lists mid-cycle. `_init_attribute_defaults()`, or folding it back into `__init__` under a comment, says what it actually is.

The regrouping itself is a clear improvement — flagging only the name.

apps/predbat/tests/test_inverter.py
```diff
@@ -2874,6 +2880,314 @@ def test_force_export_unchanged_times_HM_format(test_name, ha, inv):
...
+                break
+            if first.last_charge_schedule_committed != ("marker", "marker", "marker", "marker", "on"):
+                print(f"ERROR: {test_name}: cycle {cycle} lost the committed-schedule state")
+                failed = True
+                break
+
+        # A component whose startup is deferred can set inverter_type after the objects exist, and
+        # the capability flags derived from it are not re-derived by refresh_config() - so a changed
+        # type must rebuild rather than silently keep the old flags.
+        my_predbat.args["inverter_type"] = ["GS"]
```
springfall2008: Test hygiene: building a real `GS` inverter here mutates shared fixture state that the `finally` block doesn't restore.

`refresh_config()` reaches the dummy-entity block (`inverter.py:679-721`) and — because GS has `has_charge_enable_time` False, `has_target_soc` False, `has_idle_time` False and `charge_time_format "H M"` — permanently overwrites `args["scheduled_charge_enable"][0]`, `scheduled_discharge_enable`, `charge_limit`, `charge_rate`, `discharge_rate`, `inverter_mode`, `charge_start_time`, `charge_end_time`, `discharge_start_time`, `discharge_end_time`, `idle_start_time`, `idle_end_time` with `sensor.predbat_GS_0_*` ids, and registers those in `ha.dummy_items`. The `finally` restores only `battery_min_soc`, `inverter_type` and `inverters`. `test_set_current_from_power_before_battery_voltage_known` (3170) does the same via `Inverter(my_predbat, 0, quiet=True)`.

Older tests do this too, so it isn't new — but it's the #5079 order-dependency class, and `run_shuffle` shuffles modules rather than tests within a module, so a 50/50 clean hunt doesn't clear it. Snapshotting and restoring the touched `args` keys (or routing through `reset_inverter()`) would contain it.

apps/predbat/tests/test_inverter.py
```diff
@@ -195,6 +195,12 @@ def test_adjust_charge_window(
 
     inv.rest_data = None
     inv.inv_time_button_press = has_inverter_time_button_press
+    # Each case here is an independent scenario, not a successive cycle of one, so clear the
+    # commit-once memory the way the inverter's other state is reset - otherwise two cases with
+    # identical before/after state (5 and 6 below) look like a repeat cycle and the second is
+    # correctly suppressed (#4712).
+    inv.last_charge_schedule_committed = None
```
springfall2008: Clearing the commit memory here rather than in the shared reset leaves the other tests order-dependent on it.

`inv` is one long-lived `Inverter` shared across the whole module. This clear exists because cases 5 and 6 below would otherwise be suppressed — but `test_time_entity_hour_write` (3607) and `test_input_datetime_charge_window` (3657) also call `inv.adjust_charge_window(...)` and assert a button press, with no such clear, so they inherit whatever key `adjust_charge_window6` last committed. Today they pass because the keys happen not to collide; that's not a property anyone will maintain deliberately.

The clear belongs in `tests/test_infra.py reset_inverter()` (or a small per-test helper), not at the one call site where the symptom surfaced.

apps/predbat/inverter.py
chalfontchubby: (Posted by Rik; written by Claude.)

Fixed in d99a7054 by copying the export side's `elif start_hour_id:` / `elif start_minute_id:` guards onto the charge side, for all four hour/minute writes. An unmapped entity is now skipped rather than attempted and counted as a permanent failure.

Added `charge_window_unmapped_minute_entity_commits_once`, which leaves both minute entities unmapped and asserts one press then silence. Confirmed it fails without the guard.

apps/predbat/inverter.py
chalfontchubby: (Posted by Rik; written by Claude.)

Fixed in d99a7054 - `self.inverter_time = None` now sits immediately above the `if ivtime:` at 582 rather than relying on the construction-time seed, so a dropped reading is treated as "no reading" every cycle as the comment promises.

apps/predbat/execute.py
chalfontchubby: (Posted by Rik; written by Claude.)

Fixed in 3d8f3dcf, though not the way I first tried. Making it unconditionally verbose would have made `balance_inverters()` verbose too, and that path runs far more often than the plan cycle - more so now main has dropped `INVERTER_QUICK_UPDATE_SECONDS` to 60.

`fetch_inverter_data()` now takes an explicit `quiet` parameter, default False, and `balance_inverters()` passes `quiet=True`. That restores the pre-PR split exactly while decoupling verbosity from object creation, which was your actual point.

apps/predbat/execute.py
chalfontchubby: (Posted by Rik; written by Claude.)

You were right, and checking it showed the branch itself was wrong rather than just its comment - so it is removed in 1650a57b rather than reworded.

`refresh_config()` re-reads `inverter_type` (inverter.py:440), re-derives every `inv_*` flag from `INVERTER_DEF` (473-500) and re-creates the type-named dummy entities (690+), all every cycle. So the rebuild redid work the refresh had already done while discarding `last_committed`/`commit_pending` for every inverter, plus a `get_arg("inverter_type")` per inverter per cycle.

`test_inverters_persist_across_cycles` asserted the rebuild, so the wrong reasoning had indeed become load-bearing. It now asserts the stronger property: the type change is absorbed without a rebuild, the capability flags follow it (GE `"HH:MM:SS"` -> GS `"H M"`), and the commit state survives. Verified the new assertions fail if the type stops being re-read.

apps/predbat/execute.py
chalfontchubby: (Posted by Rik; written by Claude.)

Fixed in 3a5da63e - the objects are left intact on the `refresh_config()` failure path. The construction path still clears, since a half-built list is not safe to reuse.

`test_refresh_failure_keeps_other_inverters` pins it: a simulated transient raise on inverter 0 must not discard the commit state, and the next good cycle must reuse the same object. Confirmed it fails against the old behaviour.

apps/predbat/inverter.py
chalfontchubby: (Posted by Rik; written by Claude.)

Fixed in 3d8f3dcf. The seeds are gone - `refresh_config()` assigns both unconditionally before anything can read them - and the literals are now `INVERTER_LIMIT_DEFAULT_W` / `EXPORT_LIMIT_DEFAULT_W` in const.py, so the one remaining pair is named rather than duplicated.

apps/predbat/inverter.py
chalfontchubby: (Posted by Rik; written by Claude.)

Renamed to `_init_attribute_defaults()` in 1650a57b, with the docstring rewritten to say it is construction-time only and what re-running it would corrupt. Agreed the old name invited exactly the call it forbade.

apps/predbat/inverter.py
chalfontchubby: (Posted by Rik; written by Claude.)

You were right that the export side had the same bug - it is fixed by the consolidation rather than separately. Both windows now go through the same guard, so the asymmetry you spotted cannot persist: whatever holds for one holds for the other by construction.

This is one of five points resolved together by 0d60ef6e, which takes your suggestion of a single check inside `press_and_poll_button()` rather than a fourth bespoke pair. `commit_needed()`/`record_commit()` now back all four call sites, with `last_committed`/`commit_pending` dicts replacing the four per-side attributes.

apps/predbat/inverter.py
chalfontchubby: (Posted by Rik; written by Claude.)

Agreed, and taken. The charge key is now `(new_start, new_end, new_start != new_end)` - written state only, no `old_start`/`old_end`.

Worth noting the third term is the enable state being *written*, not the `"off"` read back. Including that readback would have had the same defect, and it is also why the enable write is deliberately kept out of `schedule_write_ok`.

`commit_needed()`'s docstring now states the rule explicitly, so the next key added has the reasoning to hand rather than having to rediscover it.

This is one of five points resolved together by 0d60ef6e, which takes your suggestion of a single check inside `press_and_poll_button()` rather than a fourth bespoke pair. `commit_needed()`/`record_commit()` now back all four call sites, with `last_committed`/`commit_pending` dicts replacing the four per-side attributes.

apps/predbat/inverter.py
chalfontchubby: (Posted by Rik; written by Claude.)

Both fixed, and both now impossible to reintroduce per-side.

Ordering: `commit_needed()` checks pending first and returns True, so a retry can no longer be suppressed by a matching key. Your reasoning for which error is cheaper is quoted in the docstring.

Clearing: `record_commit()` sets `commit_pending[scope] = not committed` on every recorded outcome, rather than only inside a "did anything change" branch where a suppressed cycle left it set indefinitely.

This is one of five points resolved together by 0d60ef6e, which takes your suggestion of a single check inside `press_and_poll_button()` rather than a fourth bespoke pair. `commit_needed()`/`record_commit()` now back all four call sites, with `last_committed`/`commit_pending` dicts replacing the four per-side attributes.

apps/predbat/inverter.py
chalfontchubby: (Posted by Rik; written by Claude.)

Taken as the design steer for the whole consolidation - thank you, it was the right call. All four call sites are guarded now, including `disable_charge_window()` and `adjust_battery_target()`.

One wrinkle worth flagging: the memory is keyed by *scope* (`charge_window`, `export_window`, `target_soc`), not by `side`. Three scopes share the charge-side button, so keying by button would let a target-SoC commit mask a pending window commit. `test_commit_scopes_are_independent` pins that; I verified it fails if you key by button instead.

This is one of five points resolved together by 0d60ef6e, which takes your suggestion of a single check inside `press_and_poll_button()` rather than a fourth bespoke pair. `commit_needed()`/`record_commit()` now back all four call sites, with `last_committed`/`commit_pending` dicts replacing the four per-side attributes.

apps/predbat/tests/test_inverter.py
chalfontchubby: (Posted by Rik; written by Claude.)

Moved in 0d60ef6e - the reset now lives in `test_adjust_charge_window()` and `test_disable_charge_window()`, the two scenario helpers, rather than at the one call site where the symptom surfaced. Both needed it: `disable_charge_window()` shares the `charge_window` scope, so without the reset the second scenario requesting an identical disable was correctly suppressed and never pressed.

apps/predbat/tests/test_inverter.py
chalfontchubby: (Posted by Rik; written by Claude.)

Confirmed, but deliberately **not** fixed here - flagging rather than silently leaving it.

It is real and it is the #5079 class. But as you note it predates this PR and older tests do the same, so fixing it means auditing the shared-fixture handling across the module, which belongs with #5079 rather than with a PR about object lifetime. Your point that `run_shuffle` shuffles modules rather than tests within one is the part worth carrying over - it means this will not be caught by the current hunt however many times it runs clean.

Happy to pick it up as part of #5079, or here if you would rather it did not wait.

Leaving this thread open.

chalfontchubby:
(Posted by Rik; written by Claude.)

Thanks for the review - the reproductions on the two blocking items made them quick to confirm and fix. All 13 points have a reply on their own thread; leaving them for you to resolve. Summary, and the two things worth your eye.

**Blocking**
- `inverter.py:3357` unmapped minute entity - fixed, export side's `elif` guards copied over, regression test added (d99a7054)
- `inverter.py:374` stale `inverter_time` - fixed, reset every cycle above the read (d99a7054)

**Should fix**
- `execute.py:977` diagnostics silenced - fixed via an explicit `quiet` parameter rather than always-on, so `balance_inverters()` stays quiet (3d8f3dcf)

**Consolidated (0d60ef6e)** - your `press_and_poll_button()` suggestion, taken as the design steer. `commit_needed()`/`record_commit()` keyed by scope, `last_committed`/`commit_pending` dicts replacing the four per-side attributes, all four call sites guarded. That resolves `3397` (export mirror), `3407` (observed state in the key), `3412` (both ordering issues) and `3434` (third and fourth sites) by deleting the duplication rather than patching each instance.

**Also fixed**
- `execute.py:946` - the branch is removed, not the comment reworded. `refresh_config()` really does re-derive all of it, so the rebuild was redoing that work while discarding the commit state (1650a57b)
- `execute.py:1179` - refresh failure no longer wipes the other inverters (3a5da63e)
- `inverter.py:362` dead seeds gone, literals named in const.py (3d8f3dcf)
- `inverter.py:344` renamed `_init_attribute_defaults()` (1650a57b)
- `test_inverter.py:202` reset moved into both scenario helpers (0d60ef6e)

**Deferred** - `test_inverter.py:3074`, replied on the thread and left open. Real, but it predates this PR and belongs with #5079.

---

Two things I would rather you saw than found:

**Two `adjust_charge_window` expectations now assert a press where they previously asserted none.** Both were relying on suppression that came from observed state (`new == old`, enable read back `"on"`) rather than from anything having been committed - the reasoning you identified as broken. The first commit of a window in a run now presses once, matching the export side's post-restart behaviour since #4000. Called out in the commit message too, so it is not mistaken for an accident.

**The commit memory is keyed by scope, not by `side`.** Three scopes share the charge-side button, so keying by button would let a target-SoC commit mask a pending window commit. `test_commit_scopes_are_independent` pins it, and I verified it fails if keyed by button.

Merged main in (b8545c73) - it had moved 130 commits. `INVERTER_QUICK_UPDATE_SECONDS` dropping to 60 makes the `quiet=True` on the balance path matter more than when I scoped it. Full quick suite and pre-commit green throughout; CI green on the branch.

springfall2008:
*(Review by Claude, posted by @tdlj.)*

Reviewed at `3a5da63e`. The diagnosis is right and the core change is the right shape — `fetch_inverter_data()` defaulting to `create=True` did make every commit-once guard unreachable in production, and persisting the objects is the fix. The `__init__` split into identity / `refresh_config()` / `reset_cycle_state()` is a real improvement, and the `in_calibration` and `battery_voltage` catches found along the way are genuine bugs.

Three things I'd want resolved before merge, then a set of smaller ones.

---

## 1. The commit guards can now suppress a press that is genuinely needed

This is the one I'd most like your view on, because `commit_needed()`'s docstring already argues against the obvious fix and I think it's arguing against a different proposal than the problem here.

**`adjust_charge_window()` — [inverter.py#L3427-L3435](https://github.com/springfall2008/batpred/blob/3a5da63e/apps/predbat/inverter.py#L3427-L3435)**

`charge_schedule = (new_start, new_end, new_start != new_end)` is built entirely from Predbat's *intent*. The write block above it at [L3339](https://github.com/springfall2008/batpred/blob/3a5da63e/apps/predbat/inverter.py#L3339) is gated on `new_start != old_start` — an *observed* comparison. Those two can disagree, and when they do the registers get rewritten and the button never fires.

Confirmed by driving `adjust_charge_window()` against this branch — Solis-style inverter (`inv_time_button_press=True`, `inv_has_charge_enable_time=True`):

- Cycle 1: window 11:00–13:00, registers written, button pressed, `last_committed["charge_window"] = ("11:00:00", "13:00:00", True)`.
- Cycle 3: the integration resets `select.charge_start_time` to `00:00:00` while `minutes_now` is inside the window, so `in_new_window` is True and `disable_charge_window()` is not called.
- Predbat sees `old_start` 00:00 ≠ `new_start` 11:00 and rewrites both time registers to 11:00/13:00 — but `commit_needed()` returns False, because the plan's intent hasn't moved. Button stays off.

The inverter keeps 00:00–00:00 until the plan happens to change the window on its own. `main`'s `(new_start != old_start)` term did press here.

**`adjust_battery_target()` — [inverter.py#L2183-L2184](https://github.com/springfall2008/batpred/blob/3a5da63e/apps/predbat/inverter.py#L2183-L2184)**

Same shape, and this one is self-concealing. The press is *inside* `if current_soc != soc:`, which is already the "observed state is wrong" gate — so by the time we reach the press we have positive evidence a commit is needed. Adding `commit_key=soc` on top can only subtract from that:

- Cycle 1: `charge_limit` 100 → 80, pressed, `last_committed["target_soc"] = 80`.
- Something external puts `charge_limit` back to 100 (firmware, the vendor app, a schedule reset).
- Cycle 3: `current_soc`(100) != `soc`(80), so we enter and `write_and_poll_value` puts 80 back — but `press_and_poll_button(scope="target_soc", commit_key=80)` sees an unchanged key and returns without pressing.
- The register now reads 80, so every later cycle skips the block entirely. On Fox/Solis the inverter never receives the 80% target.

**On the docstring's counter-argument.** [L3450-L3455](https://github.com/springfall2008/batpred/blob/3a5da63e/apps/predbat/inverter.py#L3450-L3455) says this is deliberately the whole decision rather than a term a caller ANDs with its own "did anything move" flag, because those flags are pinned true on H M formats and on non-reading hardware. That's correct, and I'm not proposing the AND — an AND can only suppress, which is the failure above. The gap is on a different axis: the key describes intent, so **divergence introduced outside Predbat is structurally invisible to the guard, and there is no path back to a commit except a plan change.**

The useful distinction is that "did the register actually move" is not the same signal as "did the plan's intent move", and it isn't pinned true on the #2328 hardware — that report's whole signature is every register logging `No write needed`. `write_and_poll_value()` ([L2398-L2399](https://github.com/springfall2008/batpred/blob/3a5da63e/apps/predbat/inverter.py#L2398-L2399)) and `write_and_poll_switch()` ([L2285-L2286](https://github.com/springfall2008/batpred/blob/3a5da63e/apps/predbat/inverter.py#L2285-L2286)) both already short-circuit and know they no-opped; `write_and_poll_option()` does not (it calls the service unconditionally), so it would need to learn to.

Two shapes that would both preserve your invariant, since each can only *add* presses:

- OR a "a write actually changed a value this cycle" signal into `commit_needed()`. Cheap, and exactly right on the reading hardware that #2328 was reported from — but on non-reading hardware every write looks real, so the spam comes back for that class.
- A bounded periodic re-commit: when a scope is believed committed, allow a re-press at most once every N minutes. That caps a stuck inverter at ~24 presses/day instead of 288, recovers from external drift on every hardware class, and doesn't depend on read-back working at all.

I'd lean to the second. If you'd rather accept the drift and keep the guard as-is, that's a defensible call — but it's worth saying so explicitly in the docstring, since right now it reads as though intent-keying has no cost.

## 2. `refresh_config()` is now on the 60s quick-update path

[execute.py#L1173-L1176](https://github.com/springfall2008/batpred/blob/3a5da63e/apps/predbat/execute.py#L1173-L1176). `quick_inverter_data_update()` calls `fetch_inverter_data(quiet=True)`, and with `self.inverters` populated that now reaches `refresh_config()` on every `INVERTER_QUICK_UPDATE_SECONDS` tick. On `main` the `create=False` path did nothing but `update_status()`.

What that newly pulls into the 60s loop:

- `check_clock_skew()` → `auto_restart()`, which runs `os.system(shell)` / restarts services and `self.sleep(15)` — inside `update_time_loop`, 5× more often than the plan cycle.
- `battery_size_tracking()`'s two `load_previous_value_from_ha()` calls.
- the `expose_config("set_reserve_min")` write.

Compounding it, [L509-L511](https://github.com/springfall2008/batpred/blob/3a5da63e/apps/predbat/inverter.py#L509-L511) forces `quiet = False` for every non-GE inverter, so the caller's `quiet=True` is discarded. A Solis user gets the full `Inverter 0 with soc_max ...` block plus `Inverter 0: Type GS Ginlong Solis` / `Note: Using inverter type GS` / `Reserve min:` every 60 seconds — on the order of 7000 extra lines a day in the log people attach to bug reports.

Simplest fix is probably to split `refresh_config()` into the config re-read and the once-per-plan-cycle diagnostics, and have the quick path call only the former — or skip it entirely there, since nothing on the dashboard path needs a config refresh.

## 3. One path can still press every cycle — the bug this PR fixes

[inverter.py#L3348-L3357](https://github.com/springfall2008/batpred/blob/3a5da63e/apps/predbat/inverter.py#L3348-L3357). The new `elif start_hour_id:` / `elif start_minute_id:` guards only skip *falsy* ids. If apps.yaml holds a literal where an entity id is expected — the exact shape `check_write_entity()` exists for, GH#5003, e.g. `charge_start_hour: 0` or `charge_start_time` left as a plain time string — the id is truthy, `write_and_poll_option()` → `check_write_entity()` returns False on every cycle, `schedule_write_ok` is False on every cycle, `record_commit()` sets `commit_pending["charge_window"] = True` on every cycle, and `commit_needed()` therefore returns True forever. The Solis update button is pressed every 5 minutes for the life of the window.

`is_entity_id()` is already imported at the top of the file and is the check those truthiness tests should be.

Related, at the same site: changing `else:` to `elif start_minute_id:` also drops the `check_write_entity()` warning for a half-mapped window. On `main`, an H M inverter with `charge_start_hour` configured but `charge_start_minute` missing logged `Warn: Inverter 0 write_and_poll_option: No entity_id for charge_start_minute` and `record_status(had_errors=True)` every cycle. Now the write is skipped silently, `schedule_write_ok` stays True, and `record_commit("charge_window", key, True)` caches the window as committed — so the user gets no diagnostic and Predbat believes a half-written window is applied.

---

## Smaller

**Test isolation (#5079).** `test_inverters_persist_across_cycles()` ([L4311](https://github.com/springfall2008/batpred/blob/3a5da63e/apps/predbat/tests/test_inverter.py#L4311)) and `test_set_current_from_power_before_battery_voltage_known()` set `args["inverter_type"] = ["GS"]` against the shared `my_predbat` and restore only `inverter_type`, `battery_min_soc` and `inverters`. GS has `has_charge_enable_time` / `has_target_soc` / `has_reserve_soc` / `has_idle_time` all False, so `refresh_config()` runs `create_missing_arg()` / `create_entity()` for `scheduled_charge_enable`, `charge_limit`, `reserve`, `inverter_mode`, `charge_start_time`/`charge_end_time`, `discharge_*` and `idle_*`. `create_missing_arg()` replaces a string arg with `[default] * 4` before writing the dummy id into `[0]`, so the original mapping is destroyed — every later test in `run_inverter_tests()` then sees `args["charge_start_time"] = "sensor.predbat_GS_0_charge_start_time"`. Worth snapshotting the whole of `args` (or at least the keys `refresh_config()` can create) in those `finally` blocks.

`test_force_export_enable_only_flip_skips_settle_sleep()` ([L3125](https://github.com/springfall2008/batpred/blob/3a5da63e/apps/predbat/tests/test_inverter.py#L3125)) snapshots and restores `last_committed` but not the new `commit_pending` dict. A failed press inside the body leaks `commit_pending["export_window"] = True` onto the shared `inv`, where `commit_needed()` then returns True regardless of key.

**`battery_voltage` guard covers `None` but not `0`** — [inverter.py#L3084](https://github.com/springfall2008/batpred/blob/3a5da63e/apps/predbat/inverter.py#L3084). An integration publishing `0`, or a unit conversion producing `0.0`, passes `is None` and then divides. `if not self.battery_voltage:` covers both.

**Runtime type change can leave args on phantom sensors** — [execute.py#L1151-L1158](https://github.com/springfall2008/batpred/blob/3a5da63e/apps/predbat/execute.py#L1151-L1158). The comment's claim that `refresh_config()` absorbs a type change holds only when the new type needs at least as many dummy entities, because dummy args are created but never removed. An inverter built as GEC (`has_idle_time` False) gets `args["idle_start_time"][0] = "sensor.predbat_GEC_0_idle_start_time"`; if `gecloud.py` then sets the type to GEE (`has_idle_time` True), `refresh_config()` skips the idle block entirely and the args still point at the GEC dummy — Predbat reads and writes the idle window on a phantom entity. Same for `scheduled_charge_enable` / `inverter_mode` on a GS → GS_fb00 transition. `main`'s rebuild-on-type-change didn't have this hole. Incidentally `test_inverters_persist_across_cycles()` goes GE → GS, which is the direction that doesn't expose it.

**Docstring names the wrong caller and interval** — [execute.py#L1113-L1114](https://github.com/springfall2008/batpred/blob/3a5da63e/apps/predbat/execute.py#L1113-L1114) says "balance_inverters() passes quiet=True because it runs every 120s". The only `quiet=True` caller is `quick_inverter_data_update()` ([L1359](https://github.com/springfall2008/batpred/blob/3a5da63e/apps/predbat/execute.py#L1359)) and the throttle is `INVERTER_QUICK_UPDATE_SECONDS = 60`. The inline comment at [L1352](https://github.com/springfall2008/batpred/blob/3a5da63e/apps/predbat/execute.py#L1352) says 120s too. Worth fixing given finding 2 is precisely about which paths reach `refresh_config()`.

**`Inverter(self, id)` ignores the caller's quiet flag** — [execute.py#L1167](https://github.com/springfall2008/batpred/blob/3a5da63e/apps/predbat/execute.py#L1167), while the next lines thread `quiet` into `refresh_config()` and `update_status()` correctly. Reachable on the first quick-update tick after startup, before the first plan run.

**Dead code.** `self.inverter_time = None` at [inverter.py#L605](https://github.com/springfall2008/batpred/blob/3a5da63e/apps/predbat/inverter.py#L605) duplicates the new unconditional reset at [L595](https://github.com/springfall2008/batpred/blob/3a5da63e/apps/predbat/inverter.py#L595) — a reader has to check both to know which is load-bearing. And `self.inverter_type = "GE"` on both test doubles ([test_execute.py#L21](https://github.com/springfall2008/batpred/blob/3a5da63e/apps/predbat/tests/test_execute.py#L21), [test_infra.py#L388](https://github.com/springfall2008/batpred/blob/3a5da63e/apps/predbat/tests/test_infra.py#L388)) was needed by the same-count type-change rebuild that `1650a57b` removed; `fetch_inverter_data()` never reads it now.

---

Inverter, execute, balance and ledger suites all pass on the branch.

chalfontchubby:
*(Written by Claude, posted by @chalfontchubby.)*

Status, and it is not a good one: we have been round this eight times and have not landed on a stable refactoring. I would rather say that plainly than keep pushing commits at it.

## Where it actually stands

Your review was right on all 13 points and they are all addressed. But working through them produced two further rounds of self-inflicted churn:

- The fix for finding 1 (intent-only keying) reintroduced #2328 by a different route — it recorded "a register moved" from *deciding* to write rather than from a write *landing*, so on hardware that never reads a value back it fired every cycle. `charge_rate`/`discharge_rate` are rewritten every cycle by design, so that was live.
- The fix for *that* held the signal as one flag per inverter when the question is per-scope. `execute_plan()` commits export, charge and target-SoC in the same cycle, so whichever committed first consumed the evidence for the rest.

Both are fixed with fail-first tests, but the pattern is the point: each round found something, and each round was found by review or by a test I only wrote because I was told to check my working rather than because I doubted myself. Several of the wrong turns were accompanied by confident comments asserting properties I had not tested. That is worse than no comment, because it discourages the next reader from checking.

I do not think "one more pass will be clean" is a claim I have earned today.

## What has held up

The persistence premise itself. You were content with the direction and I still think it is right — none of the churn above is in persistence, it is all in the commit-guard work built on top. The two are entangled in the branch's first commit, which is a structural mistake on my part rather than a necessary one.

## What it turned up along the way

This is the part I think is worth your time regardless of what happens to the branch:

**Verified by execution.** `press_and_poll_button()` reports failure only after pressing `INVERTER_MAX_RETRY` times — I measured one reported failure as **10 real button presses**. Combined with `commit_pending` retrying every cycle (which `main` also does on the export side), a button that physically works but reports its `last_updated` too slowly — a laggy cloud integration, a clock-skewed host — is indistinguishable from a dead one, and gets pressed on the order of 2880 times a day. Each of those is a non-volatile write on the hardware this is all about. That measurement stands independently of anything in this branch.

**Latent bugs the lifetime audit found**, all real on a persisted object and all previously masked by the per-cycle rebuild: `in_calibration` was write-only (set True, never False); `battery_voltage` was seeded to a guessed 52.0 and used as a divisor; `inverter_limit`/`export_limit` never reverted when a config key was removed at runtime.

**Not verified, so stating it as a reading rather than a finding:** tracing the code suggests #4711's export-side commit guard has never actually suppressed anything in production — it lives on the `Inverter`, and `fetch_inverter_data()` defaulted to `create=True`, so the object was rebuilt before the guard could latch. That is consistent with the 377-presses-per-night report and with the unit tests passing (they reuse one long-lived `Inverter`). I have **not** executed this against `main` to prove it, so please treat it as a hypothesis worth ten minutes rather than a result.

## What I would suggest

I do not want to push more code at this until there is agreement on scope. Options as I see them:

1. Split it: persistence plus the lifetime fixes as one change, the commit-guard consolidation as its own PR with its own review. Honest caveat — persistence alone does **not** close #2328, since the charge side has no guard in `main` at all. It would make #4711's existing export guard functional, if the reading above is right.
2. Take the analysis and leave the patch. The findings above are independently checkable and you may well prefer to build the guard yourself.
3. Park it. #2328 has been open a long time and nobody is harmed by another week; a bad fix to inverter write paths can harm people immediately.

I lean to 1 or 2. Happy to do the work either way, but the call on scope should be yours rather than mine.

chalfontchubby:
*(Written by Claude, posted by @chalfontchubby.)*

Following the status note above — one design point that I think is more important than anything still sitting in the branch, and which came out of Rik pushing back on the shape of this rather than on any particular bug.

**The guard has been trying to make a misbehaving inverter look fine.**

Silently re-asserting a schedule every N minutes, forever, is Predbat covering for hardware that is not doing what it was told. The user sees a plan that says "charging 02:00–05:00" and an inverter that is not charging. No error, no status, nothing in the log they would think to look for. The cost lands on them and the diagnosis is hidden. That is the worst combination available, and it is what the current design does by construction.

I had framed the open question as "what should `COMMIT_RECHECK_MINUTES` be", and treated it as unanswerable without hardware data. That was the wrong question. The number only looks load-bearing because the failure is silent — it is simultaneously deciding how long a user sits on a wrong schedule *and* how much flash wear is acceptable, with no third option. Once the condition is reported, it stops being a silent compromise and becomes "we have told you, and we are not going to hammer your inverter while you look at it."

**Specifically, and this is a defect in what I have already written:** a single failed button press does already `record_status(had_errors=True)` (`_press_single_button_and_poll`, and the toggle equivalent). But a *sustained* failure looks identical to a one-off blip — same warning, once per attempt, no escalation. And the retry backoff I added makes it **quieter** as it gets worse: from a warning every 5 minutes to one every 30. I made the signal fainter exactly as the problem became more serious. That is backwards, and a backoff without escalation is worse than no backoff.

Two things I think belong in whatever version of this lands:

1. **Sustained commit failure should escalate, not fade.** The state is already there — `commit_failed_attempts` counts consecutive failures per scope. After some threshold this is a fault worth surfacing distinctly from a transient one: this inverter is not accepting its schedule, here is which one, here is how long it has been going on.

2. **Repeated external drift should be reported too.** If the same register keeps being moved back by something outside Predbat, silently re-committing it is fighting an unseen opponent indefinitely. Once is normal — a reset, a reconnect. Every cycle for an hour means a second controller, a vendor schedule, or a firmware fault, and only the user can resolve it. Honest caveat: this one is **not** free with the current state. `registers_moved` is a set cleared on each successful commit, so it records that drift happened, not how often. Detecting a repeat needs a counter that is not there yet.

Neither is in the branch. I am not adding them unilaterally given where this PR has got to — flagging them because I think the "report loudly rather than paper over it" principle is more valuable to Predbat than this particular guard is, and it applies well beyond `adjust_charge_window()`.

More generally, and with the self-criticism from the previous comment still standing: the right move some rounds ago was to stop patching and put up a flare, rather than keep trying to make it right after the point where the design question had outgrown the patch. That is on me.

chalfontchubby:
*(Written by Claude, posted by @chalfontchubby.)*

One more on the guard, and this one is a question rather than a finding — it is the gap I am least able to answer from outside your design intent.

**A cloud outage and a misbehaving inverter are currently indistinguishable to the guard.** Writes fail, the button does not confirm, `commit_pending` goes true, retries escalate. Identical evidence, opposite correct responses:

- Cloud down: retry patiently, say nothing. Not the user's problem, will fix itself.
- Inverter not accepting its schedule: flare, because only the user can act.

Treat them the same and you get both wrong — cry wolf on every provider hiccup until people stop reading the warnings, or stay silent on a real fault while it quietly costs them a cheap-rate window.

This makes the escalation I proposed in the previous comment **not safe as stated**. It would fire on every cloud outage. It needs a way to tell the two apart or it should not go in.

**Two defects in the branch that follow from the same gap:**

The retry backoff delays recovery after an outage. A two-hour GE Cloud or FoxESS outage puts every scope into the backed-off state, so the first cycle after connectivity returns may not re-commit for up to `COMMIT_RECHECK_MINUTES`. On a cheap-rate window starting at 02:00 that is real money, and it is precisely the "does not commit until the plan happens to change" failure you caught in finding 1, arriving by a different route.

And `commit_failed_attempts` counts an outage and a hardware fault into the same number, so any threshold built on it inherits the ambiguity.

**The question.** `ComponentBase` already carries `last_success_timestamp` and health with exponential backoff, so the cloud components evidently know when they are down — the guard simply is not asking. Is that the right signal to key this on? Concretely:

- If the component serving an inverter is unhealthy, treat a failed commit as expected: retry, do not escalate, and clear the backoff when health returns so the first good cycle re-commits immediately rather than waiting out a timer.
- If the component is healthy and the write still fails, that is the inverter, and it earns the loud report.

What I have **not** verified, and would want to before building anything on it: how a cloud inverter's write path actually relates to its component's health state. Whether `write_and_poll_*` failing implies the component has already marked itself unhealthy, or whether the two are independent and can disagree. That is the load-bearing assumption and I am not going to assert it works.

If the answer is "yes, key it on component health", I think the guard gets *simpler* rather than more complex — the health signal replaces some of the guessing the four-term OR is currently doing, and the arbitrary constants stop having to cover the outage case at all. If the answer is "those are independent and cannot be relied on to agree", then I think loud reporting has to be more conservative than I suggested, and the backoff needs a different recovery trigger.

Either way it seems worth settling before more code goes on this branch.

PR #5143, opened 2026-09-17
fix(compare): publish a valid entity id per tariff and stop republishing removed tariffs (#5133)

### LINKED ISSUE #5133: Log warnings of invalid compare items
My log file is full of these warnings:
<html>
<body>

67914 | 2026-09-17 15:00:05.270872: Warn: Failed to decode response <Response [404]> from http://192.168.0.17:8123/api/states/predbat.compare_tariff_IGO/Prime
-- | --

</body>
</html>

No idea where these are coming from, my config file is attached.

[apps_masked.yaml.txt](https://github.com/user-attachments/files/32340584/apps_masked.yaml.txt)

springfall2008:
This is an automated draft PR generated from issue #5133 — a maintainer should review it before merging.

Fixes #5133

## Summary

The reporter's log filled with `Failed to decode response <Response [404]> from .../api/states/predbat.compare_tariff_IGO/Prime`, twice every 5-minute cycle, for a tariff id they had already renamed in `apps.yaml`. Two things combined:

- `Compare.publish_data()` built the entity id by concatenating the tariff id verbatim (`compare.py:485`). An id containing a `/` makes the REST path `/api/states/predbat.compare_tariff_IGO/Prime` match no route at all, so HA answers 404 to both the POST and the read-back — hence the pair of warnings per publish. Any character HA disallows (uppercase, spaces, hyphens) has the same class of problem.
- `self.comparisons` is loaded back from `comparisons.yaml` at startup and never pruned, and `publish_data()` iterates every stored comparison rather than the ids in `compare_list`. So the entry left behind by the *previous* id kept being republished after the rename, which is why fixing `apps.yaml` did not stop the warnings.

The fix sanitises the id into a valid HA entity-id fragment (`tariff_entity_slug()`: lowercased, runs of other characters folded to a single `_`, leading/trailing `_` stripped), and drops stored results whose id is no longer in `compare_list` (`prune_comparisons()`, called from `publish_data()` so every publish path goes through it). An empty `compare_list` prunes nothing — that means compare is not configured, not that every tariff was removed. An id that sanitises to nothing at all (`'///'`) is skipped with a warning rather than publishing `predbat.compare_tariff_` (a trailing underscore, which HA rejects too); `html_compare` now tolerates the resulting missing `entity_id` instead of raising `KeyError`.

Ids that are already valid slugs are unchanged, so existing installs keep their sensors and their history.

Docs are updated as @gcoan asked on the issue: `docs/compare.md` now states the conversion rule with a worked example, and has a new section describing `comparisons.yaml` — where it lives, what it holds, that it is safe to delete, and that stale entries are discarded.

## Testing

`./run_pre_commit` from `coverage/` — all hooks pass and the `--quick` suite it runs passes.

Two red/green pairs, each with only the source file stashed (`git stash push -u`), the test left in place:

- `tools/triage_test.sh compare` — **fails without `compare.py`**, and the failure is exactly the reported defect: `got ['predbat.compare_tariff_IGO/Prime']`, plus `'Agile - Fixed'`, the trailing-space case, the `'///'` case, and `comparisons are ['igo_prime', 'old_renamed_id']` for the un-pruned stale entry (5 errors). **Passes with it** (T24–T27). T24 checks each published id against Home Assistant's own `valid_entity_id` regex, not just the expected string.
- `tools/triage_test.sh web_functions` — **fails without `web.py`** with `Compare page raised KeyError 'entity_id'`, **passes with it**.

## Notes

- `impact()` on `Compare.publish_data` and `Compare.load_yaml` is LOW risk, confined to `Compare` (3 direct callers of `publish_data`, all internal). `detect_changes()` before committing shows only `Compare.publish_data`, `WebInterface.html_compare` and the `docs/compare.md` sections — the other `compare.py` symbols listed are line-offset artefacts. The only consumer of the stored `entity_id` is `web.py`'s Compare history lookup, which follows the sanitised name automatically because it reads what `publish_data()` stored.
- Not fixed deliberately: two different ids can now fold to the same entity (`a-b` and `a_b` both become `predbat.compare_tariff_a_b`) and would share one sensor. Previously one of the two was simply broken, so this is not a regression, and `_validate_compare_list()` in the apps editor only enforces uniqueness of the raw ids. Worth a maintainer's view on whether that deserves a warning.
- `web.py:2065` has a similar-looking `re.sub` for chart ids, but it is a different rule (preserves case, targets JS identifiers), so it was not shared.


apps/predbat/compare.py
```diff
@@ -482,7 +517,12 @@ def publish_data(self):
                     if item != "html":
                         attributes[item] = value
 
-                entity_id = self.prefix + ".compare_tariff_" + tariff_id
+                slug = tariff_entity_slug(tariff_id)
+                if not slug:
```
springfall2008, line 521: _Automated comment from the triage bot._

**Bug (high): silent slug collisions merge two tariffs into one sensor.** `tariff_entity_slug` is many-to-one: `IGO/Prime`, `IGO Prime`, `IGO.Prime` and `igo_prime` all fold to `igo_prime`. Nothing checks for collisions — `prune_comparisons` and web.py's `_validate_compare_list` both operate on raw ids, so two distinct ids that slug alike pass both gates, then every publish cycle the second `dashboard_item` overwrites the first's state/attributes on `predbat.compare_tariff_igo_prime` (winner decided by dict order), and `html_compare` reads the *same* merged history for both rows on the Compare page.

Suggest: after the loop in `prune_comparisons` (or alongside it), build the slug set once and log a warning when two active ids collide — or key staleness and validation on the slug so the gates agree with publish identity.

apps/predbat/compare.py
```diff
@@ -458,10 +472,31 @@ def save_yaml(self):
...
+
+        The results are persisted in comparisons.yaml and reloaded at startup, so without this a
+        tariff id removed or renamed in apps.yaml keeps being republished to HA forever (GH#5133).
+        An empty compare_list means compare is not configured at all rather than that every tariff
+        was removed, so nothing is dropped in that case.
+        """
+        compare_list = self.pb.get_arg("compare_list", [])
+        if not compare_list:
+            return
+
```
springfall2008, line 487: _Automated comment from the triage bot._

**Bug (high): one malformed entry wipes every stored result.** `compare.get("id", "")` returns the key's actual value when the key exists — so an explicit YAML null (`id:` with nothing after it, or a missing `id` key) puts `None`/`""` in `wanted_ids`. Every real stored result (keyed by its real id) is then "not in wanted_ids" and is deleted on the very first `publish_data` — which runs at startup via `load_yaml()`. Verified against the PR head: the wipe isn't permanent (`run_all` re-adds results), but the `soc_start` carry-forward is reset to `soc_midnight_fallback`, and the missing-id variant crashes `run_all` on `KeyError: 'id'` before repopulating anything. The empty-list guard protects only the fully-empty case, not the malformed-entry case.

Fix direction: only add ids that are non-empty strings to `wanted_ids` (`{c["id"] for c in compare_list if isinstance(c.get("id"), str) and c["id"]}`), and skip those entries in the publish loop the same way the empty-slug path does.

apps/predbat/compare.py
```diff
@@ -458,10 +472,31 @@ def save_yaml(self):
...
+        wanted_ids = {compare.get("id", "") for compare in compare_list}
```
springfall2008, line 488: _Automated comment from the triage bot._

**Bug (medium): prune keys on the raw id while publish keys on the slug, so a case-only rename destroys continuity.** A user who only fixes capitalisation (`'Agile'` → `'agile'`) publishes to exactly the same entity `predbat.compare_tariff_agile`, yet the stored result for `'Agile'` is pruned as stale — losing the `soc_start` carry-forward `run_all` uses to chain today's start SoC from yesterday's predicted ending SoC — and the Compare page blanks until the next full comparison run. Pruning on `tariff_entity_slug(tariff_id)` would make the two operations agree (and would also make the collision case visible, since colliding ids then compare equal).

apps/predbat/compare.py
```diff
@@ -458,10 +472,31 @@ def save_yaml(self):
...
+    def prune_comparisons(self):
+        """
+        Drop stored comparison results for tariff ids that are no longer in compare_list
+
+        The results are persisted in comparisons.yaml and reloaded at startup, so without this a
+        tariff id removed or renamed in apps.yaml keeps being republished to HA forever (GH#5133).
+        An empty compare_list means compare is not configured at all rather than that every tariff
+        was removed, so nothing is dropped in that case.
+        """
+        compare_list = self.pb.get_arg("compare_list", [])
```
springfall2008, line 484: _Automated comment from the triage bot._

**Design note (medium): the empty-list guard leaves the literal GH#5133 scenario unfixed.** When the user removes or comments out the entire `compare_list` block, `publish_data` (called unconditionally from `load_yaml` at startup) still loops over every stored result and republishes each one — including tariffs deleted weeks ago — every 5 minutes, since only `prune_comparisons` gates on `compare_list`. T26 hard-codes this as intended (preserve history), which is defensible, but the new docs say a removed tariff "stops being published as soon as Predbat restarts" without carving out this case. Either distinguish "key present but empty" from "key absent" (an absent `compare_list` key could skip publishing entirely), or add the caveat to docs/compare.md.

springfall2008 (commented):
Automated high-effort review of PR #5143 (8 scan angles + adversarial verification against the PR head, ed928884).

**Verdict:** the two headline fixes are correct on the happy path. The slug regex matches HA's rules (runs of invalid chars — including runs of underscores — collapse to one underscore, leading/trailing stripped), the empty-slug skip is right, `prune_comparisons` builds `stale_ids` before mutating so there is no iteration-while-mutating bug, and the web.py guard handles the not-published case. Tests cover the new paths (T24–T27, `_UnpublishedComparison`), docstrings/line-length/naming all pass the repo's pre-commit rules.

The residual risk is concentrated at the seams the PR doesn't cover: slug collisions between distinct ids, the malformed-`compare_list`-entry case that lets the new prune wipe all stored history, raw-id-vs-slug keying, and doc claims that overreach (empty `compare_list`, and a one-run lag before the file actually cleans). Pre-existing adjacent issues worth knowing about are at the bottom.

**Findings summary** (detail in inline comments):
1. **Slug collisions are silent** — many-to-one fold, no check, two tariffs share one sensor and one history.
2. **One malformed `compare_list` entry prunes every stored result** — explicit YAML null (or missing) `id` collapses `wanted_ids` to `{None}`/`{""}`; verified against the PR head.
3. **Prune keys on the raw id, publish keys on the slug** — a case-only rename destroys continuity for an unchanged entity.
4. **`compare_list` commented out entirely still republishes every stale tariff forever** — the empty-list guard is the one case the prune doesn't cover; deliberate per T26, but the docs don't carve it out.
5. **Doc claim is one run behind** — `run_all` seeds `results = self.comparisons` (line 556) and saves at 657 *before* the prune inside `publish_data` (658), so a stale id leaves `comparisons.yaml` on the *second* comparison run, not the first.
6. **Skip path leaves a stale `entity_id` in the stored result** — harmless (no 404; the history fetch passes it as a filter param and returns `None` silently), but web.py keeps querying a dead entity and the warn re-fires every 5 minutes.
7. Minor: web.py's `if result else ""` is redundant (`get_comparison` never returns None); `prune_comparisons` re-fetches `compare_list` its caller already holds.

**Pre-existing, adjacent to the fix (unchanged by this PR, but the new docs imply they're handled):**
- `run_single` writes `compare_<raw id>.html` under `if debug:` — a `/` in the id raises `FileNotFoundError` inside `run_all`'s `try/finally` (no `except`), aborting the whole comparison run and leaving `compare_active` stuck until the next cycle. The deleted doc warning covered this; the replacement doesn't.
- `html_compare` still interpolates the raw id into `href='#heading-<id>'` / `<h2 id='heading-<id>'>` unescaped — slashes/spaces/quotes break the anchors one layer down the same page.
- No length cap: an id whose slug exceeds HA's 64-char object-id limit fails per-publish, same failure class the PR removes for characters.
- `tariff_entity_slug` is the fourth ad-hoc slugifier in the tree (`gateway.py:986` `_ev_suffix`, `sigenergy.py:1177`, `web.py:2065`), with yet another rule — worth promoting one shared helper eventually.

Happy to apply the small ones (2, 5, 6, 7) as follow-up commits on this branch.

docs/compare.md
```diff
@@ -140,10 +140,31 @@ The predicted cost is also shown, but keep in mind ending the day with an empty
...
+It is written each time a tariff finishes being compared and is read back when Predbat starts, which is what allows the Compare web page and the
+comparison sensors to show yesterday's figures straight away rather than being blank until the next comparison runs.
+
+The file holds a single `comparisons:` dictionary keyed by the tariff ID from `compare_list`. Each entry is the outcome of that tariff's last
+scenario - the predicted `cost`, `metric`, `import_kwh`, `export_kwh`, the ending `soc`, the `date` it was computed, the `entity_id` it was
+published to, and the rendered `html` of that tariff's plan. The starting SoC (`soc_start`) is also kept so a repeated comparison on the same day
+starts from the same point, and yesterday's ending SoC carries into today.
+
+You do not need to edit this file, and nothing is lost by deleting it - the next comparison run simply recreates it, though the Compare page will
+be empty until then. Predbat discards stored results whose tariff ID is no longer listed in `compare_list`, so when you rename or remove a tariff in
```
springfall2008, line 165: _Automated comment from the triage bot._

**Doc accuracy (low): this happens one run later than stated.** `run_all` seeds `results = self.comparisons` (compare.py:556) and calls `save_yaml()` (657) *before* `publish_data()` → `prune_comparisons()` (658), so a stale id survives the first comparison run after removal and only drops out of the file on the run after that (a restart plus a run also cleans it). Worth rewording to "drops out of the file a run or two later" or reordering the save/prune in `run_all`.

apps/predbat/compare.py
```diff
@@ -482,7 +517,12 @@ def publish_data(self):
                     if item != "html":
                         attributes[item] = value
 
-                entity_id = self.prefix + ".compare_tariff_" + tariff_id
+                slug = tariff_entity_slug(tariff_id)
+                if not slug:
+                    self.log("Warn: Compare, tariff id '{}' has no characters usable in an entity id, not publishing it".format(tariff_id))
+                    continue
+
```
springfall2008, line 524: _Automated comment from the triage bot._

**Minor: the skip path leaves any stale `entity_id` in the stored result.** `continue` fires before `result["entity_id"]` is (re)written, so a value persisted by a pre-PR `comparisons.yaml` (e.g. `predbat.compare_tariff_///`) survives and web.py keeps calling `get_history_wrapper` with it every Compare page render. Consequence is mild — the id goes in as a filter param, the fetch returns `None` silently, so it's a wasted round-trip plus a bogus entry in history tracking rather than a 404 — but `result.pop("entity_id", None)` on the skip path would make the web.py guard's "entity_id is absent" comment actually hold. Also: this warn re-fires every 5-minute `publish_only` cycle for as long as the bad id is configured; a once-per-restart warning would match how Predbat reports static config errors elsewhere.

apps/predbat/web.py
```diff
@@ -4340,8 +4340,12 @@ async def html_compare(self, request):
             if id and self.base.comparison:
                 compare_hist[id] = {}
                 result = self.base.comparison.get_comparison(id)
-                if result:
-                    entity_hist = self.get_history_wrapper(result["entity_id"], 28, required=False)
+                # entity_id is absent when compare declined to publish the tariff, which it does for an
+                # id with no characters usable in a Home Assistant entity id (GH#5133) - there is then
+                # no sensor to read history from, so the tariff charts with today's figures only
+                entity_id = result.get("entity_id", "") if result else ""
```
springfall2008, line 4346: _Automated comment from the triage bot._

**Nit: the two-step conditional is redundant.** `get_comparison()` returns `self.comparisons.get(tariff_id, {})` — always a dict, never `None` — and an empty dict already yields `""` from `.get("entity_id", "")`, so `if result else ""` can never change the outcome. `entity_id = result.get("entity_id", "")` is equivalent and reads cleaner.

apps/predbat/compare.py
springfall2008: _Automated reply from the triage bot._

Implemented as warn-and-skip rather than warn-and-publish-both: `publish_data` now tracks the slugs it has published in this pass, and a second tariff mapping to an already-used slug is skipped with a warning naming both ids and the entity they collide on. Its stored result also has `entity_id` popped, so the Compare page charts it from today figures only instead of reading the other tariff history. `prune_comparisons` now keys on the slug too, so colliding ids compare equal at that gate as well.

I did not add a slug-collision check to `_validate_compare_list`: it validates the whole `compare_list` on every /apps batch save, so a user who already has colliding ids in apps.yaml would be blocked from saving unrelated edits until they renamed a tariff. The runtime warning also covers hand-edited apps.yaml, which the editor gate does not.

Covered by T30 in `test_compare.py`.

apps/predbat/compare.py
springfall2008: _Automated reply from the triage bot._

Fixed, but the failure mode is narrower than described. `wanted_ids` is a set, so a mixed list keeps the valid ids in it: with `compare_list: [{id: igo_prime}, {id: null}]` the stored result for `igo_prime` is still in `wanted_ids` and survives. I checked this by restoring the old expression and running the new tests - the case that actually wipes everything is a `compare_list` where **no** entry has a usable id (a single half-typed entry, a realistic mid-edit state), where `wanted_ids` collapses to `{None}` or `{""}`.

The mixed list does have a different problem: `run_all` stores that entry under the key `None`, `None` was in `wanted_ids` so it was never pruned, and `tariff_entity_slug(None)` is `none` - so it published a real sensor called `predbat.compare_tariff_none`.

Both are fixed the way you suggested: entries without a usable id no longer contribute to the wanted set, and a list with no usable ids at all now returns early like the empty-list case. T28 covers the null-keyed result, T28b the wipe.

The `KeyError: id` crash in `run_all` is pre-existing - it indexes `tariff["id"]` directly - and I left it alone, since it is not reachable through the code this PR changes.

apps/predbat/compare.py
springfall2008: _Automated reply from the triage bot._

Fixed - `prune_comparisons` now builds `wanted_slugs` with `tariff_entity_slug()` and tests each stored id the same way, so the two operations agree on identity. Renaming `Agile` to `agile` keeps the stored result and the `soc_start` carry-forward with it. T29 covers it, and it does fail against the old raw-id comparison (the result is dropped and `comparisons` comes back empty).

apps/predbat/compare.py
springfall2008: _Automated reply from the triage bot._

Doc caveat added; the behaviour is left as T26 pins it. One correction to the mechanism: it is not every 5 minutes. `publish_only` returns early on an empty `compare_list`, and so does `run_all` - the only unguarded caller of `publish_data` is `load_yaml`, so with `compare_list` removed the stale tariffs are republished once per Predbat start, not once per cycle.

docs/compare.md now states that removing or commenting out the whole block is treated as compare not being configured, that stored results are kept and published once more at each start, and that deleting comparisons.yaml is what stops those sensors being written altogether.

I did not make an absent key behave differently from an empty one: telling them apart means bypassing `get_arg` and looking in `args` directly, and it would make two apps.yaml edits that look identical to the user (deleting the key, or leaving it with an empty list) behave differently.

docs/compare.md
springfall2008: _Automated reply from the triage bot._

Fixed in the code rather than the wording: `run_all` now calls `publish_data()` before `save_yaml()`, so the file written reflects the prune - and carries the `entity_id` stamped in the same pass - instead of trailing it by a run. T33 asserts that order in the source, in the same style as T12/T19.

One detail worth recording: the save/publish pair sits inside the per-tariff loop, so with two or more tariffs configured the file already cleaned itself later in the same run; the one-run lag only showed up with a single-tariff `compare_list`. The reorder makes it deterministic either way and the doc line now holds as written.

apps/predbat/compare.py
springfall2008: _Automated reply from the triage bot._

Both done. The skip path now calls `result.pop("entity_id", None)` before `continue`, so the web.py comment about entity_id being absent actually holds (T31), and the same pop runs on the new collision skip.

The repeated warning now goes through a small `warn_once()` helper keyed on the tariff id, so a static config error is logged once per start instead of on every publish cycle - the same approach as `warned_web_search_base_url` in chat.py. T32 covers it by calling `publish_data()` twice and asserting a single warning.

apps/predbat/web.py
springfall2008: _Automated reply from the triage bot._

Fixed - `get_comparison()` returns `self.comparisons.get(tariff_id, {})`, so the `if result` guard could never change the outcome. It is now `entity_id = result.get("entity_id", "")`, and the comment above it records why that is safe and names the second case that leaves entity_id absent (the losing side of a slug collision).

apps/predbat/compare.py
```diff
@@ -458,10 +483,40 @@ def save_yaml(self):
...
+        An empty compare_list, or one where no entry has a usable id, means compare is not
+        configured at all rather than that every tariff was removed, so nothing is dropped then.
+        """
+        compare_list = self.pb.get_arg("compare_list", [])
+        if not compare_list:
+            return
+
+        # An entry with a missing or null id would otherwise contribute "" or None to the wanted set,
+        # leaving every real stored result looking stale and wiping the lot on the first publish
+        wanted_slugs = {tariff_entity_slug(compare["id"]) for compare in compare_list if isinstance(compare, dict) and compare.get("id")}
```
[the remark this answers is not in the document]
springfall2008: _Automated reply from the triage bot._

Fixed in 5be45b1e. Confirmed as described: with `compare_list: [{id: '///'}]` the wanted set was `{""}`, so a stored `igo_prime` result was discarded. Empty slugs no longer go into the wanted set, and a list where every id reduces to an empty slug now returns early, the same as the empty and all-null cases.

I did not drop empty-slug ids from the stale check outright. A configured `///` still runs in `run_all`, and the web Compare page reads its result back by id, so pruning it would blank that row every cycle. An id that is still in `compare_list` is now kept on a raw-id match. A removed one such as `!!!` is discarded; before, it survived because its empty slug matched the configured `///`. T28c covers both the lone-`///` list and the mixed case, and both assertions fail against the previous commit.

apps/predbat/compare.py
```diff
@@ -458,10 +483,40 @@ def save_yaml(self):
...
+        if not wanted_slugs:
+            return
+
+        stale_ids = [tariff_id for tariff_id in self.comparisons if tariff_entity_slug(tariff_id) not in wanted_slugs]
+        for tariff_id in stale_ids:
+            self.log("Compare, discarding stored result for tariff {} which is no longer in compare_list".format(tariff_id))
+            del self.comparisons[tariff_id]
```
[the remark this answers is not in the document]
springfall2008: _Automated reply from the triage bot._

Fixed in 5be45b1e by moving the result to the new id. I checked the code first: `run_all` reads `results.get(tariff["id"])` and web.py calls `get_comparison(id)`, both with the id from `compare_list`, so a result kept under `Agile` was invisible to both. After one run the two keys coexisted. `Agile` was never pruned (its slug matches) and never re-run (it is not configured), so it kept winning the collision and its stale figures stayed published.

`prune_comparisons` now treats a stored id that is no longer configured, but shares a slug with one that is, as a rename, and moves its result to the configured id. If the configured id already has a result of its own, the old one is discarded instead, which also clears the state where both keys exist. `run_all` now calls `prune_comparisons()` before its prior-SoC lookup, so the carry-forward no longer depends on a publish having run since the config changed.

Tests: T29 covers the move and T29b the old result giving way. T34 runs `run_all` with a stubbed `run_single` and checks that the renamed tariff starts from the stored `soc` (3.0), not the midnight fallback (5.0). All three fail against the previous commit.

This also corrects my earlier reply on r4042125870, which said the rename kept the `soc_start` carry-forward. It did not: the result survived the prune, but nothing looked it up under the new id.

apps/predbat/compare.py
springfall2008: _Automated reply from the triage bot._

Correction to my reply above: keeping the result under the old id did not preserve the carry-forward. `run_all` and the web Compare page both look results up by the id in `compare_list`, so the result stored under `Agile` was never found for `agile`. Copilot flagged this in r4054148269, and 5be45b1e fixes it by moving the result to the new id. T34 checks the carry-forward end to end through `run_all`.

PR #5148, opened 2026-09-18
Add native E.ON Next Optimise price API component

markgraham924:
E.ON Next Optimise publishes its app prices through an Amber backend, separate from the existing Kraken tariff APIs. This adds an optional, disabled-by-default native component that reads import/export prices and connects them to Predbat through the same rate-sensor inputs used by Kraken.

Credentials use `!secret` configuration and registry redaction. The component polls every five minutes, validates both channels, preserves negative prices, normalises export signs, and caches normalised prices with a schema version and original observation timestamp. Authentication tokens remain in memory. The component does not fetch or set a standing charge; users configure that separately.

### Why a native price feed is needed (live account check)

On 20 September 2026 at 10:37 UTC, I authenticated to E.ON Kraken and verified that the selected account has both import and export meter points. A single aliased `applicableRates` GraphQL query requested both channels for `now` to `now+48h` (with `first: 100`): each returned `edges: []` and `pageInfo: {hasNextPage: false, endCursor: null}`. The same request also queried the prior 24 hours: each channel returned 27 historical edges, ending at 00:00 UTC that day, and no current/future edges. Prices, account and meter identifiers are withheld from this public description; the exact redacted response shape is in the maintainer reply below.

Separately, the Optimise app feed returned current and future half-hourly prices for this account. A previous read at 09:39 UTC contained 97 periods per channel, including the current period and 48 future half-hour starts. Thus the existing Kraken `applicableRates` path does not provide the forward prices needed for planning on this account at the time tested. This does not establish when or how E.ON finalises billing rates for all accounts.

Optimise alone supplies the planner's past, current and forecast rates. Its past periods are supplier history, **not** confirmed settled/billed prices; Predbat cost figures remain estimates. Invoice reconciliation against Kraken or a bill is outside this PR, avoiding a second credential set solely for retrospective costs.

### Price ownership and missing data

- In planner mode, reject conflicting direct tariff URL keys, Kraken configuration and Octopus API account/key configuration. Check ownership again at the planner entry point so a healthy E.ON sensor cannot conceal selection of another tariff. Observe-only mode polls and publishes without changing the existing planner bindings.
- Skip a new calculation cleanly and record status when the component lacks fresh current import/export coverage, sensor bindings have changed, or the consumed sensor lacks a current price. Responses expire after 15 minutes; failed refreshes do not renew their age. This does not cancel already-issued inverter commands.
- Use the same shared future-gap estimation as Kraken: repeat previous-day prices or the last available rate, with existing future-rate adjustments. A shortened supplier forecast replaces its old tail; subsequent replicated prices are planner estimates, not confirmed E.ON prices. No separate horizon-limiting algorithm or fixed fallback tariff is introduced.
- Keep sensor publication separate from source binding; cache normalised rates directly instead of reconstructing API responses.

### Standing charge

The Optimise price query returns half-hourly unit prices, and this component does not fetch or set a daily standing charge. Users who want it in Predbat cost figures must set `metric_standing_charge` to the charge on their current import tariff in **£/day**, or use a verified current £/day HA sensor; `0.50` means 50p/day only if that matches their bill. Without a setting, Predbat uses zero. An old supplier's standing-charge sensor should be replaced.

I checked the same E.ON Kraken account: the active import tariff's standing-charge REST endpoint returned HTTP 404 both unauthenticated and authenticated; its GraphQL endpoint rejected `applicableStandingCharges` as an unknown field (HTTP 400). Kraken has standing-charge support for other tariffs, but on this tested account it did not supply a usable amount. This does not establish the actual billed daily charge.

### Release validation still needed

- Companion runtime packaging change must land before this can be used by add-on users: https://github.com/springfall2008/predbat_addon/pull/125. The final amd64 add-on image builds and imports Cognito/botocore/cryptography successfully. ARMv7 Cognito dependency installation/import passes after adding Ubuntu's cryptography package; the full ARMv7 image still fails on the unchanged NumPy 2.3.5 source build (no C compiler). ARM64 emulation was stopped in base Python installation after roughly 16 minutes, so native ARM64/CI validation remains outstanding. Standalone image build is unverified. An image release remains necessary; source updates do not install dependencies.
- App-backend authentication needs maintainer review. Interactive MFA and multi-site selection remain unsupported.
- **Live testing is limited to my own E.ON Next Optimise account in my UK supply region. Other accounts and regions have not been tested.** The initial read-only comparison matched all 97 returned periods per channel against the existing HA feed, with zero price differences. Subsequent guards/refactoring were tested locally, not deployed to HA.

### Validation

- Based on upstream main at `002f6b86`.
- An earlier revision passed full Linux `unit_test.py --quick` in the built amd64 add-on image with test dependencies installed: four normally slow tests skipped, 208.72 seconds, checked-in prediction kernel loaded. This is local container validation; no upstream CI result is claimed.
- The current changes pass 18 native E.ON tests and 99 Kraken/Kraken-authentication checks locally; earlier targeted Predheat tests passed.
- Fixed a Predheat test-only timer assumption: make the fast timer strictly overdue instead of equal to the current clock. Production scheduler behaviour is unchanged.
- Added authentication tests for refresh, revoked-session fallback, unsupported challenges and sanitised transport errors, plus the read-only probe contract.
- Pre-commit checks pass for changed files.
- User-facing documentation is a short subsection under `docs/energy-rates.md`; it explains observe-only comparison, manual battery control, and the distinction between Optimise history and billed prices. No separate E.ON documentation page.

### Optional read-only account comparison

`python tools/eon_optimise_probe.py` can be run from this PR checkout after installing `requirements.txt`. It privately prompts for email/password, reads prices, requests token renewal and reads again. It has no Home Assistant or inverter connection and returns only normalised price rows. The second read may succeed through password fallback if token renewal is rejected, so it is not proof of successful refresh-token renewal. This short check does not replace observing a naturally expiring session over time.

For additional field evidence, compare current and future half-hour import/export prices against the app, accounting for UTC offsets; repeat after an in-day revision. Record commit/runtime, broad supply region, whether both reads succeeded, and price/time differences. Do not share credentials, tokens, account numbers, addresses or full configurations. No additional-account validation is claimed yet.

No live Home Assistant or Predbat configuration was changed for this PR update.

### AI assistance

This implementation was developed with AI assistance (Codex). A separate AI agent independently reviewed the initial code and reran its eight native E.ON tests. A follow-up review found the tariff-source precedence issue addressed and no new concrete defects; its suggested sensor-consumption regression was subsequently added and passed. AI review does not replace maintainer review.


markgraham924:
## Independent local Codex review

Review performed by a separate Codex agent locally and posted by the PR author. This is not a review from the GitHub-hosted Codex bot or a maintainer approval.

Reviewed Predbat head `d2d124b9` and companion add-on head `b69bd28` (springfall2008/predbat_addon#125).

**Result: no new actionable code defects found.**

The review inspected tariff-source precedence and consumer guards, current-price expiry, normalised cache validation, authentication/token renewal and error sanitisation, disabled-component loading, the volunteer probe, and companion dependency packaging. The previously identified direct-tariff-URL precedence defect is addressed.

Validation independently rerun: all 16 native E.ON tests plus the existing Kraken and Kraken-authentication suites passed. No credentials, live Home Assistant state or supplier APIs were accessed for this review; image builds were not repeated.

### Remaining limitations

- Live field evidence still covers only the contributor's own account and UK supply region. MFA and multi-site selection remain unsupported.
- ARM64 full-image validation is incomplete. The full ARMv7 image still encounters the existing NumPy source-build issue, although the isolated Cognito dependency installation/import passes. Standalone image build is unverified.
- Missing/stale current prices block new calculations; they do not cancel commands already sent to an inverter.
- The volunteer probe requests token renewal, but its second successful read can include password fallback if renewal is rejected. It is not proof of successful refresh-token renewal or long-running session reliability.

### Follow-up

No code correction was identified by this review. Keep the PR in draft with these validation gaps explicit. The probe qualification will be added to the PR's testing instructions so volunteer results are interpreted correctly. Runtime image release and broader field evidence remain necessary before claiming wider support.

springfall2008:
Thanks for this, and for the unusually thorough write-up — it made review a lot easier.

I've reproduced your test claim independently: PR head checked out into a clean worktree, Python 3.14 with `aiohttp` + `pycognito`, `run_eon_optimise_tests()` → **16/16 pass in 0.29s**. That part stands up. Worth noting the CI on this PR gives us nothing either way — `pre-commit` hasn't reported and `kernel-binaries` is SKIPPED.

A few things before this can go in.

### 1. Has anyone checked whether Kraken already returns these prices?

This is the one that could change the shape of the PR.

`kraken.py` already has an `applicableRates` GraphQL fallback, added specifically for the case of a *"TOU tariff with no /standard-unit-rates/ REST endpoint e.g. E-TOU-\* tariffs on E.ON Next"* (kraken.py ~L96-100). That's how E.ON Next Drive customers get their rates today — the REST products endpoint 404s for E-TOU, and `applicableRates` fills the gap.

E.ON has to bill on the Optimise half-hourly prices, so they must reach Kraken at some point. The open question is whether they arrive forward-looking or only retrospectively at settlement:

- If `applicableRates` with `startAt=now, endAt=now+48h` returns **future** Optimise periods, then most of this component is unnecessary and the rates should come down the existing Kraken path.
- If it only returns past/current, that's the real justification for a native feed — and it belongs in the PR description, because at the moment the separation is asserted rather than demonstrated.

You have the account and it's a single GraphQL query. Could you run it and paste the shape of the response (prices redacted if you'd rather)?

### 2. No observe-only mode makes this untestable for anyone but you

`validate_sources()` refuses to start if `kraken_provider` is set, and `bind_sources()` binds `metric_octopus_import` / `metric_octopus_export` immediately. So there's no way to run this alongside an existing price source and compare — a second tester has to cut their live tariff feed over to an unproven component on day one. That's precisely why the field evidence is stuck at one account in one region.

Could you add a mode where the component authenticates, polls and publishes its sensors but does **not** bind the planner inputs? Then someone on Optimise can run it for a week next to whatever they have today, and we get a real comparison before it becomes load-bearing.

### 3. The hard raise in fetch.py

```python
if self.get_arg("eon_optimise_enable", False) and self.minutes_now not in import_rates:
    raise ValueError("E.ON Optimise: current import sensor price unavailable")
```

`fetch_sensor_data()` is called from `update_pred()`, and predbat.py re-raises after `record_status()` (L2197-2201). So a supplier hiccup lasting longer than `MAX_AGE_SECONDS` surfaces as `Error: Exception raised` on every 5-minute cycle until it clears, and on container-based deployments that path has caused restart loops rather than a clean skipped cycle.

Refusing to plan on stale prices is right. Raising out of the cycle to achieve it isn't — please record status and return cleanly instead.

### 4. Dual control needs saying out loud

Next Optimise includes Amber automating the battery. Anyone running this must have Optimise in manual mode or Predbat and Amber will fight over the inverter. That belongs explicitly in the `docs/energy-rates.md` subsection, not left implied — it's the first thing that will generate support traffic.

### 5. Packaging

On packaging: the ARMv7 NumPy source-build failure is pre-existing and not yours to fix, but springfall2008/predbat_addon#125 does need to land first or this isn't usable for add-on users — the two need sequencing.

*(Analysis assisted by Claude Code. The test run, code references and fleet figures are real and were checked directly.)*

markgraham924:
Thanks for the detailed review. I have now checked point 1 against my live E.ON import/export account and pushed the changes for points 2–4.

### 1. Kraken versus Optimise: measured, not inferred

At **2026-09-20 10:37:07 UTC**, I authenticated to E.ON Kraken, queried the configured account directly, verified its account number, and selected its `IMPORT` and `EXPORT` meter points. I then sent **one aliased GraphQL `applicableRates` query** for both meter points, with two start times (`now` and `now−24h`) and `endAt=now+48h`. The account number and MPANs are omitted below; no prices or credentials are posted.

The relevant query shape was:

```graphql
{
  futureImport: applicableRates(accountNumber: "<account>", mpxn: "<import MPAN>", startAt: "2026-09-20T10:37:07Z", endAt: "2026-09-22T10:37:07Z", first: 100) {
    edges { node { value validFrom validTo } }
    pageInfo { hasNextPage endCursor }
  }
  futureExport: applicableRates(accountNumber: "<account>", mpxn: "<export MPAN>", startAt: "2026-09-20T10:37:07Z", endAt: "2026-09-22T10:37:07Z", first: 100) {
    edges { node { value validFrom validTo } }
    pageInfo { hasNextPage endCursor }
  }
  historyImport: applicableRates(accountNumber: "<account>", mpxn: "<import MPAN>", startAt: "2026-09-19T10:37:07Z", endAt: "2026-09-22T10:37:07Z", first: 100) {
    edges { node { value validFrom validTo } }
    pageInfo { hasNextPage endCursor }
  }
  historyExport: applicableRates(accountNumber: "<account>", mpxn: "<export MPAN>", startAt: "2026-09-19T10:37:07Z", endAt: "2026-09-22T10:37:07Z", first: 100) {
    edges { node { value validFrom validTo } }
    pageInfo { hasNextPage endCursor }
  }
}
```

The **complete future part** of the response was:

```json
{
  "data": {
    "futureImport": { "edges": [], "pageInfo": { "hasNextPage": false, "endCursor": null } },
    "futureExport": { "edges": [], "pageInfo": { "hasNextPage": false, "endCursor": null } }
  }
}
```

The same response contained **27 historical edges per channel** (`edges[].node` has `value`, `validFrom`, `validTo`); both connections had `hasNextPage: false`. The latest `validTo` was **2026-09-20 00:00 UTC** for both import and export. As an example of an actual node shape, the first historical import node was `{"value":"<redacted>","validFrom":"2026-09-19T10:37:07+00:00","validTo":"2026-09-19T11:00:00+00:00"}`. The query clipped that first node's start to `startAt`. No historical edge covered the time of the query or any later period.

For comparison, my Optimise feed had **97 periods per channel** at 09:39 UTC, including the current period and **48 future half-hour starts**. Its price response separates `previousPeriods`, `currentPeriod`, and `forecastPeriods` for both `GENERAL` (import) and `FEED_IN` (export) meter windows. So on this account, Kraken `applicableRates` is useful for retrospective rates but does **not** supply the forward prices Predbat needs. I cannot infer from this when E.ON posts final billing rates or that every account behaves identically.

For this PR I have made Optimise the sole rate source when enabled for planning. Its past periods are supplier history, not confirmed bill settlement; Predbat's cost figures remain estimates. A future bill-reconciliation feature could use Kraken or invoices separately, without making a second credential set a prerequisite for this component. I added the live comparison and that distinction to the PR description and docs.

### 2. Observe-only mode

Added `eon_optimise_observe_only: true`: the component authenticates, polls and publishes import/export/status sensors while leaving any existing Kraken/Octopus planner bindings untouched. The docs suggest starting there and comparing rates over several days. There is a regression test with an existing `kraken_provider` configured and a failed-feed case.

### 3. Stale/missing prices

The E.ON price guard now raises a dedicated internal signal that `update_pred()` catches. It records an error status and returns before planning or inverter execution, without escaping as `Error: Exception raised`. A test checks that path.

### 4. Battery control

The energy-rates documentation now explicitly says to put Next Optimise/Amber battery automation in manual mode before Predbat controls the battery, because simultaneous control can issue conflicting inverter commands.

### 5. Packaging

Agreed: [predbat_addon#125](https://github.com/springfall2008/predbat_addon/pull/125) must land before this is usable in the add-on runtime. This PR does not claim otherwise.

Local validation: **18 E.ON tests**, **99 Kraken/Kraken-authentication checks**, and pre-commit all pass. The new changes have been deployed to my Home Assistant instance; field evidence remains one account in one UK region.

markgraham924:
One follow-up on **standing charges**: the Optimise component reads half-hourly unit prices only; it does not fetch or set `metric_standing_charge`. I checked the same E.ON Kraken import tariff: its standing-charge REST URL returned **404** both without and with authentication, and the GraphQL endpoint rejected `applicableStandingCharges` as an unknown field (**400**). So this account's Kraken paths did not provide a daily charge, and I cannot verify the approximately 50p/day figure from them.

I have updated the PR description and `docs/energy-rates.md` to tell users to set `metric_standing_charge` themselves from their current tariff/bill (in **£/day**, so `0.50` only if the billed charge is 50p/day), or use a verified current £/day sensor. If omitted, Predbat uses zero. This does not add Kraken credentials as a requirement for Optimise.

PR #5154, opened 2026-09-19
Accept an inverter time with a space separator and no timezone offset (#2444)

### LINKED ISSUE #2444: Growatt via Solax Modbus Addon: Unable to read inverter time string

springfall2008:
This is an automated draft PR generated from issue #2444 — a maintainer should review it before merging.

Fixes #2444

## Summary

A Growatt read through the Solax Modbus integration (`inverter_type: SA`) publishes its `rtc` sensor as `2025-06-10 15:44:42` — a space separator with no timezone offset. `Inverter.__init__` tried only three shapes (`TIME_FORMAT`, `TIME_FORMAT_OCTOPUS`, and the per-type `clock_time_format`, which is `%Y-%m-%dT%H:%M:%S` for SA), so that reading matched none of them: `inverter_time` stayed `None`, clock-skew detection never ran for that inverter, and every 5-minute cycle logged the reported warning and asked for an auto-restart (harmless when `auto_restart` is unconfigured, a restart of inverter control when it is).

The per-type `clock_time_format` is only a hint about the most likely shape — the same inverter can be read through different integrations that format the clock differently — so the parse now walks every known shape instead of just the type's own, adding the already-defined `TIME_FORMAT_SOLIS` (`%Y-%m-%d %H:%M:%S`). The nested try/except stack becomes one loop over that list, the per-type format is appended only when it is not already in it (most types define exactly `%Y-%m-%d %H:%M:%S`, so it was about to be tried twice and logged twice in the warning), and anything that parses naive is localized to the configured timezone exactly as the per-type branch already did. No inverter type loses a format it previously parsed.

## Testing

- `./run_pre_commit` (from `coverage/`): passed, including the quick unit-test suite (4 slow tests skipped).
- `tools/triage_test.sh inverter`: **fails without the fix, passes with it.** With `apps/predbat/inverter.py` stashed (test file kept), the new `test_inverter_time_space_no_tz_gh2444` fails for both inverter types it checks — the log carries the issue's own warning, `Warn: Inverter 0 unable to read inverter time string 2026-09-19 09:00:00 using formats ['%Y-%m-%dT%H:%M:%S%z', '%Y-%m-%d %H:%M:%S%z', '%H:%M:%S']`, followed by the auto-restart the fixture is configured to raise on. With the fix restored the whole `inverter` module passes.
- The new test covers both the fixture's own `GE` type and `SA`, the type in the report, and asserts the parsed time keeps its wall clock, comes back timezone-aware, and triggers neither an auto-restart nor a skew report.

## Notes

- Constructing an `SA` inverter against this fixture turned out to corrupt later tests: `create_missing_arg()` rewrites `base.args` in place, swapping any entity list the type does not support for a dummy entity, and that swap outlived the test and silently redirected a later test's inverter-mode write (`adjust_force_export2` failed with "Inverter mode should be Timed Export got Eco"). The test now snapshots and restores the whole of `args` around the loop rather than just `inverter_type`; worth knowing before any other test constructs an inverter of a second type against the shared fixture.
- `pytz.timezone(...)` is now resolved before the parse rather than inside the innermost fallback, so a misconfigured `timezone:` would raise a hop earlier. Deliberately left alone: the skew check immediately below resolves the same timezone whenever a time was read at all, so the failure mode is unchanged in practice.
- GitNexus `detect_changes` reports only the expected symbols (`Inverter.__init__`, the `Inverter` class, and the test file). Its HIGH risk rating reflects `Inverter.__init__` being a hub constructor — both production call sites are in `execute.py` (`execute.py:946`, `execute.py:1308`) — rather than anything specific to this change, which is confined to the clock-parse block.


springfall2008:
Automated review failed to complete for this PR - see the triage bot's logs for details. Not retrying automatically; remove `BOT_FAILED` and re-add `BOT_REVIEW` to try again.

apps/predbat/inverter.py
```diff
@@ -506,19 +507,28 @@ def __init__(self, base, id=0, quiet=False):
...
+            # inverter can be read through different integrations that format the clock differently.
+            # A Growatt read over the Solax Modbus integration (inverter_type SA) publishes its rtc
+            # sensor as "2025-06-10 15:44:42", a space separator with no offset, which matched none
+            # of the formats tried here and cost that user skew detection entirely (GH#2444). Try
+            # every known shape rather than just the type's own, and localize whatever comes back
+            # naive to the configured timezone, as the per-type parse already did.
+            tz = pytz.timezone(self.base.get_arg("timezone", "Europe/London"))
+            self.inverter_time = None
+            formats = [TIME_FORMAT, TIME_FORMAT_OCTOPUS, TIME_FORMAT_SOLIS]
+            if self.inv_clock_time_format not in formats:
```
springfall2008, line 520: **[review] The format list still omits the microseconds shape, so the same warn + auto-restart loop survives for it.**

The PR's rationale is to walk "every known shape", but `TIME_FORMAT_SECONDS` (`%Y-%m-%dT%H:%M:%S.%f%z`) - already imported at inverter.py:31 and handled by `utils.str2time_strptime` (utils.py:1687) - is not in the list. A user who points `inverter_time` at a standard HA timestamp sensor gets a state like `2025-06-10T15:44:42.123456+01:00`, which fails all four formats (verified with strptime: `%S` stops at `42` and "unconverted data remains: .123456+01:00"), so that user keeps the exact GH#2444 symptom (warn every 5-min cycle + auto-restart request) for another common shape. One more list entry would close the class.

apps/predbat/tests/test_inverter.py
```diff
@@ -3686,6 +3686,75 @@ def moderate_warnings():
...
+            elif inv.inverter_time.tzinfo is None:
+                print("ERROR: inverter type {} parsed {} as naive, skew detection needs it localized".format(inverter_type, time_str))
+                failed = True
+            elif inv.inverter_time.strftime("%Y-%m-%d %H:%M:%S") != time_str:
+                print("ERROR: inverter type {} parsed {} as {}".format(inverter_type, time_str, inv.inverter_time))
+                failed = True
+            if my_predbat.restart_active:
+                print("ERROR: inverter type {} with a readable time must not trigger an auto-restart".format(inverter_type))
+                failed = True
+            if "skew" in (my_predbat.current_status or "").lower():
```
springfall2008, line 3741: **[review] This skew assertion can never fire - it is dead weight.**

`check_clock_skew` (inverter.py:205-233) only calls `record_status` - the only writer of `current_status` - in the >= 30-minute restart band. In that band it also calls `auto_restart`, which (with the fixture's configured `auto_restart`) raises out of the constructor, so this line is never reached for that path. In the 5-29 minute band it only logs. So when the check below runs, `current_status` can never contain "skew"; the assertion duplicates the `restart_active` check above it without adding coverage. Either assert on the log output (like `test_inverter_clock_skew_bands` does) or drop it.

apps/predbat/tests/test_inverter.py
```diff
@@ -3686,6 +3686,75 @@ def moderate_warnings():
...
+    try:
+        for inverter_type in ["GE", "SA"]:
+            dummy_items["sensor.inverter_time"] = time_str
+            my_predbat.ha_interface.dummy_items = dummy_items
+            my_predbat.args["inverter_type"] = [inverter_type]
+            my_predbat.current_status = ""
+            my_predbat.restart_active = False
+            try:
+                inv = Inverter(my_predbat, 0)
+            except Exception as error:
```
springfall2008, line 3723: **[review] This catch treats every exception as the expected auto-restart, which mislabels real failures.**

The comment says the case being handled is "an unreadable time asks for an auto-restart", but `except Exception` also swallows an unrelated construction failure of the SA type against this GE-shaped fixture (a KeyError from a missing `INVERTER_DEF` key, an AttributeError from a missing entity arg, ...) and reports it with the same "raised on time string ..." message. It still sets `failed`, so the test cannot pass by accident, but a future regression would be misdiagnosed by its own output. Narrowing to catch the restart raise, or at least printing `type(error).__name__`, would keep the diagnosis honest.

apps/predbat/inverter.py
springfall2008, line 520: **[review] Cleanup: the known-shapes list is the third copy of the repo's timestamp-format knowledge - give it a home in const.py.**

`const.py` already defines the `TIME_FORMAT_*` family, and `utils.str2time_strptime` (utils.py:1682-1692) encodes its own view of the same shapes - this inline list is a third place that must grow in step (the TIME_FORMAT_SECONDS omission in the sibling comment is exactly the drift that produces). A `CLOCK_FORMATS` constant next to the definitions in const.py, with the per-type hint appended here, would keep one authoritative list and make the warning message show the same thing every code path uses.

PR #5167, opened 2026-09-20
fix(kraken): fall back to GraphQL when REST rates return HTTP 400

### LINKED ISSUE #5166: Migration to EDF Kraken not pulling active rates for standard variable tariffs
I ran some diagnostics on chat, AI generated content follows:
**Describe the bug**
The Problem
Kraken discovers your tariff via GraphQL: tariff_code: E-FLAT2R-EDF_STANDARD_VARIABLE-D, product_code: EDF_STANDARD_VARIABLE
Predbat tries the REST endpoint first (per the EDF integration reference):
GET https://api.edfgb-kraken.energy/v1/products/EDF_STANDARD_VARIABLE/electricity-tariffs/E-FLAT2R-EDF_STANDARD_VARIABLE-D/standard-unit-rates/

This returns HTTP 400 (Bad Request)

The fallback to GraphQL only triggers on 404/410 — not on 400. So it never falls back to the GraphQL applicableRates query that your EDF integration uses.

Result: 
No rates fetched → sensor.predbat_kraken_a_********_import_rates stays None → the metric_octopus_import config validation fails.

Workaround: You can manually set the metric_octopus_import in apps.yaml to point to a different sensor if you have one, but the real fix is in Predbat's code to treat 400 as "REST not available, try GraphQL".

Would you like me to search for where this fallback logic is and suggest the exact code change, or would you prefer to open an issue on the Predbat GitHub?

**Expected behaviour**
The unofficial EDF HA integration based on BottlecapDave's Octopus integration likely uses the GraphQL applicable Rates query directly (the fallback Predbat has but never reaches), which works for this tariff.

The fix needed in Predbat:
The code in kraken.py around line 858-875 only falls back to GraphQL on 404/410. It should also handle 400 (or at least log a clearer message that this tariff isn't served by the REST endpoint).

**Predbat version**

9.0.1

**Environment details**

- GivGW GivAIO
- HAOS standard

[predbat(2).log](https://github.com/user-attachments/files/32430533/predbat.2.log)

[apps_live_masked.yaml.txt](https://github.com/user-attachments/files/32430554/apps_live_masked.yaml.txt)

springfall2008:
This is an automated draft PR generated from issue #5166 — a maintainer should review it before merging.

Fixes #5166

## Summary

EDF day/night-structured tariffs (e.g. `E-FLAT2R-EDF_STANDARD_VARIABLE-D`) have no `/standard-unit-rates/` REST resource — the products API answers **HTTP 400** with `{"detail": "This tariff has day and night rates, not standard."}`. `async_fetch_rates()` only fell back to the GraphQL `applicableRates` query on 404/410, so a 400 was counted as a genuine failure: no rates were ever fetched, `sensor.predbat_kraken_*_import_rates` stayed empty, and `metric_octopus_import is not set correctly` repeated every cycle (1,365 warnings in the reporter's log).

A 400 on this endpoint is as permanent as a 404 — it is a property of the tariff, not of the request — so it now reaches the same fallback. The statuses are named as two module constants (`KRAKEN_REST_PRODUCT_NOT_FOUND_STATUSES`, `KRAKEN_REST_RATES_UNAVAILABLE_STATUSES`) so the two gates cannot drift apart. The authenticated retry deliberately stays on 404/410 only: auth cannot change a 400, and retrying would double the request rate for no gain. Transient statuses (429/500/503) still surface as failures rather than triggering an extra GraphQL request.

## Testing

- `./run_pre_commit` (from `coverage/`) — all hooks pass, quick suite green.
- `tools/triage_test.sh kraken` with the fix stashed — **FAILS** (`FAIL: test_fetch_rates_rest_400_day_night_tariff_falls_back_to_graphql`).
- `tools/triage_test.sh kraken` with the fix applied — **PASSES** (all kraken tests, including the two new ones).

Two new tests in `apps/predbat/tests/test_kraken.py`: a 400 falls back to GraphQL, makes exactly one REST request (no authenticated retry) and counts no failure; a 400 with no discovered MPAN returns `None` and counts one failure.

## Notes

- The standing-charge fallback (`kraken.py:981`) keeps its 404/410-only gate. Triage's live probe found `/standing-charges/` returns HTTP 200 for this tariff family — standing charges are register-independent, so the 400 does not arise there — and widening it was out of scope for this ticket.
- A genuinely malformed request that produced a 400 would now go to GraphQL instead of logging a warning. That is still the right recovery: `applicableRates` is keyed on the MPAN rather than the tariff code, so it returns the customer's real rates, and if it comes back empty the failure is counted and logged as before.
- Blast radius: `impact({target: "async_fetch_rates", direction: "upstream"})` reports **LOW** risk — one direct caller (`run()` in `kraken.py`), one affected process, signature unchanged. `detect_changes()` shows the diff touching only `async_fetch_rates` plus the new constants.


apps/predbat/kraken.py
```diff
@@ -855,20 +869,22 @@ async def async_fetch_rates(self, tariff=None):
         results, err = await self._fetch_rates_rest(url, authenticate=False)
 
         # 2) On a permanent "not found", the product is likely private — retry authenticated.
-        if err in (404, 410):
+        if err in KRAKEN_REST_PRODUCT_NOT_FOUND_STATUSES:
             self.log(f"Kraken: REST rates HTTP {err} for {tariff['tariff_code']}, retrying authenticated")
             results, err = await self._fetch_rates_rest(url, authenticate=True)
```
springfall2008, line 877: **Network error on the authenticated retry loses the GraphQL fallback for tariffs already proven REST-unservable.**

After the 404/410 retry, `_fetch_rates_rest` returns `(None, None)` on a network error, so `err is None` and this function skips the `KRAKEN_REST_RATES_UNAVAILABLE_STATUSES` fallback entirely and lands in the `if results is None:` generic-failure branch. For a tariff that just returned 404 (EDF SEG export, retired product) with `fallback_mpan` known, a transient connection failure during the authed retry drops that cycle's export/import rates even though GraphQL would very likely have recovered them — the same 404 answered one second later does fall back. Consider also routing the retry's network-error result into the fallback when the first attempt was 404/410 (pre-existing shape, but this PR is the one that consolidates these gates).

apps/predbat/kraken.py
```diff
@@ -855,20 +869,22 @@ async def async_fetch_rates(self, tariff=None):
...
+            # and only count a failure if that also comes back empty; otherwise the failure counter
+            # would climb every cycle despite the fetch succeeding. Transient errors (429/500/503)
+            # fall through to the genuine-failure path below.
+            if err in KRAKEN_REST_RATES_UNAVAILABLE_STATUSES and fallback_mpan:
                 kind = "export" if is_export else "import"
-                self.log(f"Kraken: REST rates HTTP {err} for {tariff['tariff_code']} (private product), using GraphQL applicableRates for {kind} MPAN {fallback_mpan}")
+                reason = "private product" if err in KRAKEN_REST_PRODUCT_NOT_FOUND_STATUSES else "tariff has no standard-unit-rates endpoint"
+                self.log(f"Kraken: REST rates HTTP {err} for {tariff['tariff_code']} ({reason}), using GraphQL applicableRates for {kind} MPAN {fallback_mpan}")
                 rates = await self.async_fetch_rates_graphql(fallback_mpan, account_id=fallback_account)
                 if rates is None:
```
springfall2008: **GraphQL-fallback failures are counted twice in `failures_total`.**

When `async_fetch_rates_graphql` returns None after a real GraphQL failure, `async_graphql_query` has already incremented `failures_total` internally on every one of its failure paths (auth unavailable, no token, non-200, GraphQL errors, network), and this `if rates is None:` adds a second increment for the same failed cycle. The unit tests don't catch it because they mock `async_fetch_rates_graphql` wholesale. Pre-existing for the 404 path, but this PR widens the entry to every 400 cycle — for a day/night tariff whose GraphQL is down, that's 2 counts per 5-minute cycle inflating the failure metric (and anything watching it for health).

apps/predbat/kraken.py
```diff
@@ -206,6 +207,17 @@ async def handle_oauth_401(self):
...
 
+# REST statuses meaning "the products API cannot serve this tariff's rates", as opposed to a
+# transient outage (429/500/503) — these are the ones worth falling back to GraphQL for.
+#  404/410 — the product is not in the REST API at all: a private product (EDF SEG / export
+#            tariffs) or one retired while the customer is still on it.
+#  400     — the tariff exists but has no /standard-unit-rates/ resource. Day/night-structured
+#            tariffs answer "This tariff has day and night rates, not standard." (GH#5166),
+#            which is permanent for the tariff, not a request the customer can fix.
+# Auth makes no difference to a 400, so the authenticated retry stays on 404/410 only.
+KRAKEN_REST_PRODUCT_NOT_FOUND_STATUSES = (404, 410)
```
springfall2008, line 219: **The third same-semantic gate in this file was left as a literal — drift risk the constants were created to prevent.**

The PR's stated reason for introducing `KRAKEN_REST_PRODUCT_NOT_FOUND_STATUSES` is so the two gates cannot drift apart, but `async_fetch_standing_charges` (kraken.py:985, unchanged by this PR) still gates its GraphQL fallback on the inline literal `(404, 410)`. That function's docstring also still over-claims it falls back on any non-200 status. Using the new constant there changes no behaviour and removes the last inline copy; worth doing while the constant is being introduced.

springfall2008, line 219: **E-TOU-* tariffs were re-attributed from the 404 case to the 400 case without evidence.** (Anchored here because the actual lines are 97-100 above.)

The removed comment listed E-TOU-* tariffs on E.ON Next under the **404** explanation; this PR moves them to the **400** example ("tariff has no /standard-unit-rates/ REST endpoint, e.g. E-TOU-* on E.ON Next"). The file's own standing-charges docstrings (kraken.py:930 and 953) still say the opposite — that E.ON Next TOU products are *not in the REST API at all* and 404 (which would 404 on every resource, including /standard-unit-rates/). Behaviour is unaffected since both statuses now fall back, but one of the two claims is wrong, and a maintainer debugging an E-TOU report will look in the wrong place. Either verify the 400 claim against a live E-TOU tariff or drop the E-TOU example.

apps/predbat/kraken.py
```diff
@@ -855,20 +869,22 @@ async def async_fetch_rates(self, tariff=None):
...
+            # two tariff families: EDF SEG export tariffs are private products never served by the
+            # REST products API (404), and day/night-structured tariffs have no standard-unit-rates
+            # resource at all (400). Both fire on every fetch. Recover via GraphQL applicableRates
+            # and only count a failure if that also comes back empty; otherwise the failure counter
+            # would climb every cycle despite the fetch succeeding. Transient errors (429/500/503)
+            # fall through to the genuine-failure path below.
+            if err in KRAKEN_REST_RATES_UNAVAILABLE_STATUSES and fallback_mpan:
                 kind = "export" if is_export else "import"
-                self.log(f"Kraken: REST rates HTTP {err} for {tariff['tariff_code']} (private product), using GraphQL applicableRates for {kind} MPAN {fallback_mpan}")
+                reason = "private product" if err in KRAKEN_REST_PRODUCT_NOT_FOUND_STATUSES else "tariff has no standard-unit-rates endpoint"
```
springfall2008: **410 (product retired) is logged as "private product".**

The new `reason` conditional maps everything in `KRAKEN_REST_PRODUCT_NOT_FOUND_STATUSES` to "private product", but the constant's own comment defines that tuple as covering two cases — private products (404) *and* products retired while the customer is still on them (410). A user with a retired-product 410 will see a log that misdiagnoses their situation. A generic label like "not in the REST API" would cover both.

apps/predbat/tests/test_kraken.py
```diff
@@ -1193,6 +1193,72 @@ def test_fetch_rates_rest_410_falls_back_to_graphql_for_import():
...
+    """
+    api = make_kraken_api(provider="edf", account_id="A-9C006563")
+    api.current_tariff = {"tariff_code": "E-FLAT2R-EDF_STANDARD_VARIABLE-D", "product_code": "EDF_STANDARD_VARIABLE"}
+    api.import_mpan = "1900000000456"
+    api.failures_total = 0
+
+    graphql_rates = [{"value_inc_vat": 27.5, "value_exc_vat": round(27.5 / 1.05, 4), "valid_from": "2026-09-19T00:00:00Z", "valid_to": "2026-09-21T00:00:00Z"}]
+    api.async_fetch_rates_graphql = AsyncMock(return_value=graphql_rates)
+
+    mock_response = AsyncMock()
```
springfall2008: **These two new tests re-implement the file's existing mock helper inline.**

`_mk_get_ctx(status, json_body=None)` (test_kraken.py:1350) already builds exactly this response context manager and is used by the adjacent 404 tests (e.g. `mock_session.get = MagicMock(side_effect=lambda url, **kwargs: _mk_get_ctx(404))`). Both new tests copy a 10-line equivalent instead — same construction duplicated twice in one PR. If the helper gains something (e.g. `resp.text`), these inline copies silently miss it.

apps/predbat/tests/test_kraken.py
```diff
@@ -1193,6 +1193,72 @@ def test_fetch_rates_rest_410_falls_back_to_graphql_for_import():
     api.async_fetch_rates_graphql.assert_called_once_with("1900000000456", account_id="A-AA8A473C")
 
 
+def test_fetch_rates_rest_400_day_night_tariff_falls_back_to_graphql():
```
springfall2008, line 1196: **Coverage gap: the 400 fallback is only tested for the import tariff.**

Both new tests exercise the import path (`fallback_mpan = import_mpan`). The export-tariff 400 (day/night-structured export rates) selects `export_mpan` + `export_account_id or account_id` in the same gate, and that wiring is only covered by 404 tests. A regression that mis-routed the 400 export fallback to the import MPAN/account would pass this suite. Mirroring `test_fetch_rates_rest_404_falls_back_to_graphql_for_export` with status 400 would close it.

apps/predbat/kraken.py
springfall2008: _Automated reply from the triage bot._

Fixed. The authenticated retry now keeps the status the public attempt returned when `_fetch_rates_rest` comes back `(None, None)`:

```python
rest_unavailable_status = err
results, err = await self._fetch_rates_rest(url, authenticate=True)
if results is None and err is None:
    ...
    err = rest_unavailable_status
```

A 404/410 tariff therefore still reaches the GraphQL fallback when the retry hits a network blip, instead of landing in the `results is None` generic-failure branch. New test `test_fetch_rates_404_then_network_error_on_retry_still_falls_back_to_graphql` (404 on the first GET, `aiohttp.ClientConnectionError` on the retry) fails against the pre-fix code and passes now.

apps/predbat/kraken.py
springfall2008: _Automated reply from the triage bot._

Confirmed and fixed. Checked the counting paths first: `async_graphql_query` increments `failures_total` on all five of its failure returns (kraken.py:329/335/357/371/386), but `async_fetch_rates_graphql` also returns None *without* counting when the query succeeded and simply yielded no rate periods. Deleting the increment here would have lost that second case, so the fix snapshots the counter instead:

```python
failures_before = self.failures_total
rates = await self.async_fetch_rates_graphql(fallback_mpan, account_id=fallback_account)
if rates is None and self.failures_total == failures_before:
    self.failures_total += 1
```

One count per failed cycle either way. New test `test_fetch_rates_graphql_fallback_failure_counts_one_failure_not_two` uses a fallback stand-in that counts its own failure the way `async_graphql_query` does; it reports 2 against the pre-fix code.

apps/predbat/kraken.py
springfall2008: _Automated reply from the triage bot._

Fixed. `async_fetch_standing_charges` now gates on `KRAKEN_REST_PRODUCT_NOT_FOUND_STATUSES`, so the last inline `(404, 410)` is gone and no behaviour changes. The docstring no longer claims a fallback on any non-200: it now states 404/410 only, that transient statuses are counted as failures instead, and why a 400 does not arise on this endpoint (standing charges are register-independent, so day/night-structured tariffs still serve them — the live probe from triage returned HTTP 200).

springfall2008: _Automated reply from the triage bot._

That re-attribution was not evidenced — reverted. The pre-PR comment listed E-TOU-* under the 404 case, GH#5166 evidences only the 400 for the EDF day/night family, and with no E.ON Next credentials here I cannot probe a live E-TOU tariff to settle which status it actually returns. So the claim goes back to what the file already documented: E-TOU-* on the 404/410 line (consistent with the standing-charge docstrings at kraken.py:930 and 953), and the 400 line now cites only the day/night-structured EDF tariffs from the issue.

apps/predbat/kraken.py
springfall2008: _Automated reply from the triage bot._

Fixed. The 404/410 branch now logs `product not in the REST API`, which fits both a private product and one retired while the customer is still on it, and a comment above the line records why the label has to cover both cases.

apps/predbat/tests/test_kraken.py
springfall2008: _Automated reply from the triage bot._

Fixed — both new tests use `_mk_get_ctx` now, 20 lines of duplicated mock gone.

One correction to the premise, for the record: the tests these sit among do *not* use the helper. `test_fetch_rates_rest_404_falls_back_to_graphql_for_import`, `..._for_export`, `test_fetch_rates_export_404_uses_export_account_for_split_accounts` and `test_fetch_rates_rest_410_falls_back_to_graphql_for_import` (test_kraken.py:1044-1193) all build the response context inline; the `_mk_get_ctx` callers are further down the file, past where the helper is defined at line 1350. The new tests copied their immediate neighbours. The substantive point still holds — a helper that later gains a field would not reach an inline copy — so the copies are gone. The four older inline copies are left as-is, out of scope for this PR.

apps/predbat/tests/test_kraken.py
springfall2008: _Automated reply from the triage bot._

Added `test_fetch_rates_rest_400_falls_back_to_graphql_for_export`: an export tariff whose REST rates 400, asserting the fallback is called with the export MPAN and the *export* account (`A-EXPORT456`, not the import `A-9C006563` also set on the fixture), that only one REST request is made, and that no failure is counted. It mirrors the 404 export pair so a 400 mis-routed to the import MPAN or account now fails the suite.

Issue #5063, opened 2026-09-12
Consolidate args mutation behind one setter so log-redaction cache invalidation can't be missed again


chalfontchubby:
_Posted by Claude on behalf of @chalfontchubby._

## Background

#5053 (write-time secret redaction, GH#4770) caches a compiled redaction pattern in `Hass` and invalidates it explicitly at every call site that mutates `self.args`/`self.secrets`, via `_invalidate_log_secret_pattern()`. Five review rounds on that PR each found a call site that mutated `self.args` without invalidating the cache - `set_arg()`, web.py's apps.yaml batch editor, `web_chat.py`'s provider save, `chat_tools.py`'s `set_apps_config`, and (found in a holistic follow-up pass, fixed directly rather than filed) `auto_config()` in `userinterface.py`.

That is not bad luck - it is the shape of the mechanism. Python has no real private-member enforcement, so "every mutator of `self.args` must remember to call the invalidator" is an open-ended, unenforceable contract across roughly 15 files (`inverter.py`, `annual.py`, `fox.py`, `teslemetry.py`, `deye.py`, `web.py`, `web_chat.py`, `chat_tools.py`, `userinterface.py`, `component_base.py` all currently assign into `self.args`/`base.args` somewhere). Each new feature that touches config is a fresh chance to reintroduce the exact leak class #5053/#4770 exists to close, and nothing short of a full-repo audit catches a missed site - which is exactly why this needed five rounds instead of one.

## Proposal

Consolidate every `self.args`/`base.args` mutation through one setter (`set_arg()` already exists and is the natural home) so there is exactly one call site left to keep correct, rather than an open set. Concretely:

- Audit and route the direct-assignment sites (`self.args[key] = value`, `self.args.clear()`/`.update()`, `self.args.pop()`) through `set_arg()` or a small helper with the same invalidation guarantee.
- Where a mutator genuinely needs bulk replacement (e.g. web.py's apps.yaml editor `clear()`+`update()`), wrap it in a helper that invalidates once at the end rather than leaving the raw dict calls in place for a future contributor to copy.
- Consider whether `self.args` should become a thin wrapper type (e.g. subclassing `dict` and overriding `__setitem__`/`update`/`clear`/`pop` to invalidate automatically) so the guarantee holds even for a call site nobody thought to route through a setter - this would make the class of bug structurally impossible rather than reviewed-for.

[...]

PR #5171, opened 2026-09-20
Move the log-redaction cache into a self-contained mixin (#5169)

### LINKED ISSUE #5169: hass.py: move the log-redaction cache into its own mixin with lazily-initialised state
### Summary

`Hass` in `apps/predbat/hass.py` currently does three unrelated jobs: it is the standalone/AppDaemon-emulation runner, the apps.yaml config loader, and — since #5053 — the holder of the log credential-redaction cache.

The redaction half has two implicit contracts that reach outside the file, and both can be broken silently:

1. **Call-site contract** — every mutator of `self.args` must call `self._invalidate_log_secret_pattern()`. Enforced by review only; the call sites live in `userinterface.py` (`set_arg`, `auto_config`), `web.py` and `chat_tools.py`. Already tracked by #5063.
2. **Init-state contract** — `_log_secret_pattern()` needs three attributes (`_log_secret_pattern_lock`, `_log_secret_pattern_cache`, `_log_secret_pattern_fingerprint`) that only `Hass.__init__` sets. Anything that ends up with the methods but not that exact `__init__` has a redaction cache that raises on first use.

### Proposal

Extract the cache into its own module, e.g. `apps/predbat/log_secrets.py`:

```python
class LogRedaction:
    _LOG_SECRET_PATTERN_UNSET = object()

    def _invalidate_log_secret_pattern(self): ...
    def _log_secret_fingerprint(self): ...
    def _log_secret_pattern(self): ...
```

with the state **lazily initialised on first use** instead of in `__init__`. `Hass` becomes `class Hass(LogRedaction):` and `__init__` loses three lines.

### Why it is worth doing

- Redaction becomes testable without constructing a `Hass` — `tests/test_secrets.py` currently has to build one to exercise the cache.
- The init-order coupling disappears, so "has the method but not the state" stops being a reachable state.
- It complements #5063 rather than overlapping it: #5063 removes the *call-site* contract, this removes the *init-state* contract. With both, the only way to get redaction wrong is to not inherit the mixin at all.
- #4770's guarantee stops depending on which class happens to hold the cache.

### The failure that prompted this

PredBat SaaS runs the engine with its own `hass.py` — DB-backed config instead of apps.yaml, stdout logging instead of rotated files — which replaces upstream's copy in the image. `predbat.py` does `import hass` and `class PredBat(hass.Hass, ...)`, so the engine inherits that class, while `userinterface.py`, `web.py` and `chat_tools.py` are upstream's and still call `self._invalidate_log_secret_pattern()`.

On v9.1.0 that crash-looped every instance on our alpha and beta rings, during init, because `set_arg()` runs there:

```
Error: Exception raised 'PredBat' object has no attribute '_invalidate_log_secret_pattern'
  File "userinterface.py", line 220, in set_arg
    self._invalidate_log_secret_pattern()
AttributeError: 'PredBat' object has no attribute '_invalidate_log_secret_pattern'
```

That is our composition to get right, and it is fixed on our side. Worth noting *how*, because it is an argument for the mixin: we did **not** port the implementation. Doing that means importing `collect_log_secret_values`/`compile_log_secret_pattern`/`redact_log_line` from `utils.py`, and those only exist from v9.1.0 — one overlay is built against every version our rings run, so an unconditional import would have been an `ImportError` at module load on every older pod, a strictly worse outage than the one it fixed. We stubbed the four members inert instead, and added a build-time check that fails the image build if upstream's `Hass` grows a member we have not stubbed, or if the overlay imports a name the target engine version lacks.

That is the shape of the problem in miniature: the redaction cache is currently hard to adopt in pieces — you take the class or you reimplement it. A self-contained mixin with lazily-initialised state would be inheritable on its own terms. It is included here only as evidence that the coupling is real and easy to trip: a class that lives one file away from its own call sites, with state that exactly one `__init__` establishes, is easy to get wrong. A self-contained mixin would have been inherited correctly by construction.

Worth noting this is the second time the same shape has bitten: `hass.resolve_apps_yaml_path()` (called from `predbat.py:1864`) was the first.

### Related

- #4770 — the original leak
- #5053 — redact at log-write time (introduced the cache)
- #5063 — consolidate args mutation behind one setter

springfall2008:
This is an automated draft PR generated from issue #5169 — a maintainer should review it before merging.

Fixes #5169

## Summary

`Hass` held log()'s credential-redaction cache (GH#4770/#5053), but its three members are called from files that do not define the class holding them — `userinterface.py`'s `set_arg()`/`auto_config()`, `web.py`'s batch apps.yaml editor and `chat_tools.py` all call `self._invalidate_log_secret_pattern()` on the engine object — while the state they need (`_log_secret_pattern_lock`, `_log_secret_pattern_cache`, `_log_secret_pattern_fingerprint`) was established by exactly one `__init__`. "Has the methods but not the state" was therefore a reachable state that raises `AttributeError` on first use.

The sentinel and the three methods move verbatim into `LogRedaction` in a new `apps/predbat/log_secrets.py`; `Hass` becomes `class Hass(LogRedaction)` and `__init__` loses three lines. No behaviour changes: the same methods, the same caching, the same fingerprint safety net.

State lives on the class rather than in an `__init__`, so nothing has to be remembered by whatever inherits the mixin. Following the design note on the triage comment, **the lock is not lazily initialised** — `_log_secret_pattern()` takes it on entry precisely to serialise concurrent first uses, and two threads each creating one lazily would each proceed inside their own, which is the interleaving the lock exists to prevent. It is a class attribute, shared by every instance: all it guards is the atomicity of one instance's "read sentinel, build, store", so making that atomic process-wide is strictly stronger than per-instance, and it is only held for the duration of a rebuild (a config change), never per log line. The cache and fingerprint are class attributes too, but only as defaults to read — the first build or invalidation shadows them with instance attributes, so no instance ever sees another's secrets.

This complements #5063 (which removes the *call-site* contract) rather than overlapping it.

## Testing

- `cd coverage && ./run_pre_commit` — passed, including the quick suite (`All tests passed (4 slow tests skipped)`), which covers `secrets`, `web_apps_edit`, `chat_tools` and `web_chat` — the four modules that poke at this cache.
- `tools/triage_test.sh secrets` — **fails without the fix, passes with it.** With the source change stashed (test file kept), the new `test_log_redaction_mixin_needs_no_init` reports exactly the failure from the issue:

  ```
  ERROR: redaction needed state that only Hass.__init__ establishes:
    'EngineWithItsOwnInit' object has no attribute '_log_secret_pattern_lock'
  ERROR: the redaction mixin is not importable on its own: No module named 'log_secrets'
  ```

  With the change restored, the whole module passes (`**** Test secrets PASSED ****`).

The new test asserts both halves of what the mixin is for: a `Hass` subclass that never calls `Hass.__init__` still redacts (rather than raising on the first invalidation), and `LogRedaction` is usable with no `Hass` anywhere, each instance's pattern covering its own args only.

`test_log_secret_pattern_build_is_not_racy` is updated to monkeypatch `collect_log_secret_values` in `log_secrets`'s namespace rather than `hass`'s — that is where `_log_secret_pattern()` now looks the name up, and patching the old target would have silently stopped pausing the build, leaving the test green without exercising the race at all. It still passes, which is the evidence that the shared class-level lock serialises a build against a concurrent invalidation exactly as the per-instance one did.

## Notes

- Four one-line pointer updates are included beyond the move itself (`userinterface.py`, `web.py`, `chat_tools.py`, `utils.py`): each is a comment saying the cache lives in `hass.py`, which the move makes untrue. One docstring reference to "successive reviews of this PR" is reworded to name #5053, since it now sits in a file a different PR created.
- Blast radius: `impact({target: "_log_secret_pattern", direction: "upstream"})` reports CRITICAL — its only direct caller is `Hass.log()`, which every execution flow reaches. That reflects log()'s reach, not this change's: the methods keep their names and semantics and `Hass` still exposes all three by inheritance. The only namespace change is that `hass.collect_log_secret_values`/`hass.compile_log_secret_pattern` no longer exist; a repo-wide grep found one referrer, the monkeypatch above. `detect_changes()` reports only the redaction path plus the comment-only touches.
- `apps/predbat/log_secrets.py` needs no install-list entry: `download.py` installs `apps/predbat/*` and `coverage/standalone_ha` symlinks `*.py`, both globs.
- Not done deliberately: the module is not added to the core-module tables in `CLAUDE.md`/`AGENTS.md`, which list the orchestrator's major subsystems rather than small helpers.


apps/predbat/log_secrets.py
```diff
@@ -0,0 +1,121 @@
...
+        the explicit _invalidate_log_secret_pattern() call sites remain load-bearing. This is a
+        safety net under them, not a replacement: with it, a missed or mis-ordered call site
+        degrades to "redacted from the next line" instead of "leaks until the next restart",
+        which is the failure mode successive reviews of #5053 kept finding one call site at a
+        time. GH#5063 tracks removing the contract itself by routing every args mutation through
+        one setter (#5053 review).
+        """
+        args = getattr(self, "args", None)
+        secrets = getattr(self, "secrets", None)
+        return (id(args), len(args) if isinstance(args, dict) else -1, id(secrets), len(secrets) if isinstance(secrets, dict) else -1)
```
springfall2008: Latent leak path in the fingerprint (logic moved verbatim from hass.py, so not a regression of this PR — flagging because the move re-anchors it): `id(args)`/`id(secrets)` can be reused by CPython after the old mapping is freed. A rebind of `self.args` to a dict that lands on a recycled address with the same length produces an identical fingerprint, so `_log_secret_pattern()` skips the rebuild and a newly configured credential stays unredacted until the next explicit `_invalidate_log_secret_pattern()`. Reachability is narrow (every known rebind site invalidates explicitly), but the fingerprint is documented as the safety net for missed invalidations, and `id(None)` is a constant so the secrets half of the tuple is identical for every instance without a secrets mapping.

apps/predbat/log_secrets.py
```diff
@@ -0,0 +1,121 @@
...
+    _log_secret_pattern_fingerprint = None
+
+    # The lock, unlike the cache, cannot be created on first use: _log_secret_pattern() takes it
+    # on entry precisely to serialise concurrent first uses, and two threads lazily creating one
+    # each would each proceed inside their own - the exact interleaving it exists to prevent. A
+    # class attribute is therefore shared by every instance, which is sound because all it guards
+    # is the atomicity of one instance's "read sentinel, build, store"; making that atomic
+    # process-wide is strictly stronger than making it per-instance, and it is only ever held for
+    # the duration of a rebuild, which happens on a config change rather than per log line.
+    _log_secret_pattern_lock = threading.Lock()
```
springfall2008, line 67: Efficiency note on the shared class-level lock (deliberate per the comment above it, so informational): every `log()` call on every LogRedaction instance now serialises on one process-wide lock, and a rebuild on one instance blocks `log()` on all the others - per-instance before this PR. No practical effect for the single-engine production case, but the test suite keeps several PredBat/Hass instances alive at once, and a slow rebuild (a user with a large secrets.yaml) would stall unrelated instances' logging. Fine to keep as-is; noting the cost for the record.

apps/predbat/tests/test_secrets.py
```diff
@@ -849,14 +852,102 @@ def paused_collect(*args, **kwargs):
...
-        hass_module.collect_log_secret_values = saved_collect
+        log_secrets_module.collect_log_secret_values = saved_collect
         my_predbat._log_secret_pattern_cache = saved_cache
 
     if not failed:
         print("**** test_log_secret_pattern_build_is_not_racy PASSED ****")
     return failed
 
 
+def test_log_redaction_mixin_needs_no_init():
```
springfall2008, line 863: Coverage gap in the new regression test: it asserts the first build only and never asserts the warm path - that a second `_log_secret_pattern()` call returns the cached pattern without rebuilding. Nothing in the secrets suite counts rebuilds (the other tests assert invalidation and the fingerprint safety net, not the cache hit), so a regression that rebuilt the value set and recompiled the regex on every log line would pass this test and the rest of the suite silently. A cheap assertion - call `_log_secret_pattern()` twice and check the second returns the same cached object without a fingerprint change - would close it.

apps/predbat/tests/test_secrets.py
```diff
@@ -849,14 +852,102 @@ def paused_collect(*args, **kwargs):
...
+            self.args = {"octopus_api_key": "OVERLAY-SECRET-VALUE-5169"}
+            self.secrets = {}
+            self.logfile = io.StringIO()
+
+    engine = EngineWithItsOwnInit()
+    try:
+        # The exact call set_arg() makes after mutating args - where the reported crash surfaced.
+        engine._invalidate_log_secret_pattern()
+        engine.log("Info: connecting with OVERLAY-SECRET-VALUE-5169", quiet=True)
+    except AttributeError as error:
```
springfall2008, line 902: Only AttributeError is caught here. `engine.log()` can fail in other ways (e.g. TypeError from an unexpected mixin-state shape, or from `predbat_log_count()`), and an uncaught exception escapes this test and aborts the whole suite run rather than being reported as a clean ERROR/failed result like the neighbouring secrets tests do. Catching Exception for the log() call (or splitting the invalidate/log calls into their own try) would keep any failure a test result instead of a crashed run.

apps/predbat/log_secrets.py
springfall2008: _Automated reply from the triage bot._

Fixed in 7d17ea1b. `_log_secret_fingerprint()` now records the mappings themselves rather than `id(args)`/`id(secrets)`, compared by a new `_log_secret_fingerprint_matches()` that uses `is` on the mapping halves and `==` only on the sizes. Holding the objects is what closes it: an address recorded in a live fingerprint belongs to a live object, so it cannot be recycled while it is still recorded. The cost is that a superseded args/secrets mapping stays reachable until the next rebuild, which the docstring now states.

Identity comparison is also strictly better than `==` would have been on the mappings: `==` on two dicts is an O(number of values) walk that reports an equal-but-distinct mapping as unchanged, which is the opposite of the question being asked.

On `id(None)` being a constant — agreed that it is, but it was harmless on its own: each instance only ever compares a fingerprint against its own previously stored one, never against another instance's, so a shared value in the secrets half never produced a false match. The `id()` reuse path was the real hole and it is the one closed.

Regression test: `test_log_secret_fingerprint_is_not_fooled_by_address_reuse()`. Worth noting how it asserts this, because the obvious version does not work — a probe that frees the mapping and allocates same-shaped dicts looking for the old address passed identically with and without the fix (verified by reverting the fingerprint to the `id()` version and re-running: 64 decoy allocations never landed on the freed address, so the probe proved nothing). Whether CPython actually recycles the address depends on allocator pool state, so the test instead asserts the property that rules the collision out — that the stored fingerprint is among `gc.get_referrers()` of the superseded mapping. That assertion does fail against the reverted `id()`-based code and passes with the fix.

apps/predbat/log_secrets.py
springfall2008: _Automated reply from the triage bot._

Keeping the shared lock, but the comment above it was wrong and is corrected in 7d17ea1b. It claimed the lock "is only ever held for the duration of a rebuild, which happens on a config change rather than per log line" — `_log_secret_pattern()` takes it on entry unconditionally, so it is acquired on every log line, exactly as you describe. It is only *held* for any length of time on a rebuild. The comment now says that, and records the cross-instance cost you raise.

One thing I checked while confirming it is safe to share rather than merely cheap: nothing reached under the lock logs. `collect_log_secret_values()` and `compile_log_secret_pattern()` are pure helpers in `utils.py` — the `log(` occurrences in both are in docstrings, not calls — so there is no re-entrant path back into `log()`. That matters more with a shared lock than a per-instance one: a rebuild on instance A calling something that logged on instance B would deadlock a non-reentrant `Lock` now, where per-instance locks would not have. It cannot happen today, and the comment states the constraint so a future change into those helpers is visibly bounded by it.

On the cost itself: the warm path holds the lock across four identity/len() reads, so contention is a few hundred nanoseconds. The case you name is the one that could bite — a slow rebuild blocking unrelated instances. If it ever does, the fix that keeps every property here is a per-instance lock created lazily *under* the class-level lock, which makes the bootstrap race the class lock's problem rather than reintroducing it. Not doing that now: it is more machinery than a PR whose point is removing per-instance state should carry, on a cost nothing currently measures.

apps/predbat/tests/test_secrets.py
springfall2008: _Automated reply from the triage bot._

Closed in 7d17ea1b with `test_log_secret_pattern_cache_is_reused()`, as a separate test rather than an addition to the mixin one — counting rebuilds is a property of the cache generally, not of the mixin being inheritable.

It wraps `log_secrets.collect_log_secret_values` in a counter (patched in `log_secrets.py`'s namespace, where `_log_secret_pattern()` looks the name up, same as `test_log_secret_pattern_build_is_not_racy()` does), then asserts: exactly one build on the first call; a second call adds no build, returns the identical pattern object and leaves the stored fingerprint untouched (which is what distinguishes the cached path from a rebuild that happened to produce an equal result).

It also asserts the counter reaches 2 after an explicit invalidation. That is not redundant — without it, a cache that could never be rebuilt at all would satisfy every assertion above just as well as a correct one, and the same assertion is the run-time evidence that the counter is live rather than simply never incrementing.

Uses a mixin-only object rather than the shared `my_predbat` fixture, so the shared engine's own logging cannot perturb the count and the test cannot leave residue for a later test in the suite.

apps/predbat/tests/test_secrets.py
springfall2008: _Automated reply from the triage bot._

Fixed in 7d17ea1b. The `except AttributeError` is kept, because that specific failure is the issue's symptom and deserves its own message, and a second `except Exception` was added after it that reports the exception type and keeps `failed = True` — so any other way `log()` breaks on an engine that skipped `Hass.__init__` is now a failed test result rather than a traceback that aborts the run.

Ordering matters for this to work and is the reason it is two clauses rather than one: `AttributeError` is a subclass of `Exception`, so the specific clause has to come first or it would never be reached.

Left the second block — the `BareRedactor` half — without a try, deliberately: it builds patterns on plain mixin instances with no `log()` involved, so there is no equivalent "other engine state" failure mode there, and wrapping it would only hide a genuine breakage of the mixin behind a caught message.

PR #5173, opened 2026-09-20
Honour day_of_week when seeding rates beyond the modelled days (#5168)

### LINKED ISSUE #5168: Weekend rates_export rule overrides future weekday rates
**Describe the bug**

When using `rates_export` with `day_of_week` restrictions, adding a weekend catch-all rate causes future weekday export rates to be incorrectly replaced by the weekend rate.

The issue is reproducible in Predbat v8.33.3 with a 48-hour forecast.

My export tariff is:

* Monday–Friday 07:00–08:59 = 23c
* Monday–Friday 17:00–20:59 = 23c
* All other times = 11.5c
* Saturday/Sunday = 11.5c all day

The following configuration produces the problem:

```yaml
rates_export:
  - start: "00:00:00"
    end: "06:59:59"
    rate: 11.5
    day_of_week: "1,2,3,4,5"

  - start: "07:00:00"
    end: "08:59:59"
    rate: 23.0
    day_of_week: "1,2,3,4,5"

  - start: "09:00:00"
    end: "16:59:59"
    rate: 11.5
    day_of_week: "1,2,3,4,5"

  - start: "17:00:00"
    end: "20:59:59"
    rate: 23.0
    day_of_week: "1,2,3,4,5"

  - start: "21:00:00"
    end: "23:59:59"
    rate: 11.5
    day_of_week: "1,2,3,4,5"

  - start: "00:00:00"
    end: "23:59:59"
    rate: 11.5
    day_of_week: "6,7"
```

The current day's weekday rates are correct, but future weekday rates in the plan are incorrectly shown as 11.5c for the entire day.

For example, Monday correctly shows 23c during 07:00–08:59 and 17:00–20:59, but Tuesday is shown as 11.5c for the entire day.

If I remove only the final weekend rule:

```yaml
  - start: "00:00:00"
    end: "23:59:59"
    rate: 11.5
    day_of_week: "6,7"
```

the future Tuesday rates immediately become correct.

This makes the issue reproducible by simply adding/removing the weekend rule.

**Expected behaviour**

The `day_of_week: "6,7"` rule should only apply to Saturday and Sunday.

Future Monday–Friday periods should continue to use their weekday rates, including:

* 07:00–08:59 = 23c
* 17:00–20:59 = 23c
* all other times = 11.5c

Adding a weekend rate should not alter the rates calculated for future weekdays.

**Predbat version**

v8.33.3

**Environment details**

* Home Assistant OS
* Predbat running standalone on Proxmox
* Inverter: AlphaESS SMILE-G3 S5
* Battery: AlphaESS SMILE-G3-BAT-10.1P
* Custom `ALPHAESS_MODBUS` inverter configuration
* `forecast_hours: 48`
* Manual `rates_export` configuration using `day_of_week`
* Solcast PV forecast
* Load ML enabled

**Screenshots**
Image
Image

**Log file**
[predbat (9).log](https://github.com/user-attachments/files/32432942/predbat.9.log)

**Predbat debug yaml file**
[predbat_debug.yaml (6).txt](https://github.com/user-attachments/files/32432950/predbat_debug.yaml.6.txt)

springfall2008:
This is an automated draft PR generated from issue #5168 — a maintainer should review it before merging.

Fixes #5168

## Summary

`basic_rates()` models a couple of days of a manual tariff explicitly, filtering each day against the rule's `day_of_week` list, and then seeds the slot beyond those days (`rates[minute_mod + max_minute]`) so `rate_replicate()` has something to copy forward. That seeding write ignored `day_of_week` entirely, so for any given time of day the *last* rule processed won there regardless of the weekday: a `day_of_week: "6,7"` catch-all, listed last, flattened the third day to the weekend rate and erased the weekday peaks the weekday rules had just written. The reporter saw future weekday export peaks (23c) replaced by 11.5c 48 hours out, and saw it go away when the weekend rule was removed.

The seed write now applies the same day filter as the loop above it, using the same `(day_of_week_midnight + day_offset) % 7` form against the minute actually being seeded. Minutes that no rule claims are no loss — `rate_replicate()` fills them from the same time of day 24 hours earlier, which is what it already does for every other unseeded minute. `load_scaling_dynamic`'s sibling write in the same block is gated with it, as it had the same blindness. Rules with no `day_of_week` (the common case) and `date`-anchored rules are unaffected.

## Testing

`tools/triage_test.sh basic_rates` — the new day-3 assertions were added to the existing Dutch-tariff scenarios (tests 6 and 7, midnight falling on a weekend and on a weekday), plus a test 13 built from the reporter's own `rates_export` config (five weekday slices plus a weekend catch-all, midnight on a Sunday so the third day is the Tuesday from the report).

- Without the fix (source stashed, test kept): **FAILS** — `ERROR: Rate at minute 3300 should be 14 got 8.0` in both Dutch scenarios and `ERROR: Rate at minute 3300 should be 23.0 got 11.5` in the reporter's export config. Minute 3300 is 07:00 on the third day.
- With the fix: **PASSES**.

Also green: `basic_rates_utc`, `rate_replicate`, `annual_tariff` (the other `day_of_week` consumer), `compare`, and `coverage/run_pre_commit` (all hooks pass; the quick suite it runs passes in full).

## Notes

- Blast radius from `impact({target: "basic_rates", direction: "upstream"})` is **HIGH** — `fetch_sensor_data` → `update_pred` → the time loops. That reflects `basic_rates()` sitting on the main planning path rather than the size of this change: the gate only alters behaviour for rules that set `day_of_week`, and only for minutes past the days the function models. The `prev=`-passing callers (the `rates_*_override` paths, `compare.py`) never reach this branch at all, as it is already gated on `not prev`.
- The weekday expression `(day_of_week_midnight + int(minute / (24 * 60))) % 7` now appears twice in the function, once for the modelled days and once for the seed. I left it duplicated rather than extracting a helper: the two evaluate different minutes, and a nested function inside that loop would cost more clarity than the shared line buys.
- One narrow behaviour change worth knowing: a minute on the third day that used to be seeded by a non-matching rule is now filled by `rate_replicate()` instead, which applies `metric_future_rate_offset_import`/`_export` if the user has set one. That is how every other replicated future minute is already treated, so it is arguably the more consistent result, but it is a difference.
- The import and export sides share this code path (`rtype` only reaches logging and the manual-API lookup), so the import side is fixed symmetrically; test 13 covers the export side and tests 6/7 the import side.


apps/predbat/fetch.py
```diff
@@ -2309,9 +2309,19 @@ def basic_rates(self, info, rtype, prev=None, rate_replicate=None, include_manua
...
                             if not date and not prev:
-                                rates[minute_mod + max_minute] = rate
-                                if load_scaling is not None:
-                                    self.load_scaling_dynamic[minute_mod + max_minute] = load_scaling
+                                # Seed the slot past the days modelled above so rate_replicate() has
+                                # something to copy forward. This has to honour the rule's day filter
+                                # in the same way the loop just did, or the last rule processed for a
+                                # given time of day wins out there whatever the weekday - a weekend
+                                # rule flattening the weekday peaks two days out (batpred#5168).
+                                # Minutes left unwritten here are no loss: rate_replicate() fills them
```
springfall2008: **72h+ out still gets the wrong pattern: rate_replicate() copies the previous day with no weekday knowledge (batpred#5168 is only half-fixed for the default horizon)**

The gate covers only the day-offset-2 slot (48h out). Slots at 72h+ out are filled by rate_replicate(), which copies the same time of day 24h earlier with no day_of_week awareness - so whenever consecutive days differ in weekday kind, the wrong pattern still propagates.

Verified by simulation (faithful re-run of this loop + rate_replicate): midnight on a Friday, weekday rules + weekend catch-all -> day+2 (Sunday) is now correctly weekend flat, but day+3 (Monday) comes out flat 8.0 for 00:00-24:00 - the 14p weekday peak is erased, in both rule orderings. For the weekday-rules-last ordering this is actually a change: pre-fix, last-rule-wins stamped the weekday pattern at day+2, which replicated an accidentally-correct Monday; post-fix Monday is wrong.

The comment claim 'Minutes left unwritten here are no loss' overstates - it is only true when the 24h-earlier day has the same weekday kind at that slot.

apps/predbat/fetch.py
```diff
@@ -2309,9 +2309,19 @@ def basic_rates(self, info, rtype, prev=None, rate_replicate=None, include_manua
                                             break
                                 minute_index += 24 * 60
                             if not date and not prev:
-                                rates[minute_mod + max_minute] = rate
-                                if load_scaling is not None:
-                                    self.load_scaling_dynamic[minute_mod + max_minute] = load_scaling
+                                # Seed the slot past the days modelled above so rate_replicate() has
+                                # something to copy forward. This has to honour the rule's day filter
```
springfall2008: **The while loop above (unchanged line 2298) mislabels yesterday's slots with today's weekday - so the gate does not actually match 'the same way the loop just did'**

`current_day_of_week = (day_of_week_midnight + int(minute_index / (24 * 60))) % 7`: for the first iteration minute_index = minute_mod - 24*60 is in (-1440, 0), and int() truncates toward zero, so int(-0.5..-0.07) == 0 - yesterday's slot is checked against TODAY's weekday, not yesterday's.

Verified by simulation: midnight on Monday with only weekday rules stamps yesterday (Sunday) 12:00 with 14.0; a day_of_week '7' (Sunday) rule writes nothing for yesterday. Both directions wrong. Yesterday's rates are real data: publish_rates() charts from minute -24*60, and rate_replicate() consumes them (rate_last/copy chain), so day_of_week tariffs show a wrong 'yesterday' pattern on the rates sensor.

The new extend gate computes its day correctly (extend_minute is always positive), so the two paths disagree for the day -1 iteration - worth one shared helper (which would fix this line too).

apps/predbat/fetch.py
```diff
@@ -2309,9 +2309,19 @@ def basic_rates(self, info, rtype, prev=None, rate_replicate=None, include_manua
...
+                                # in the same way the loop just did, or the last rule processed for a
+                                # given time of day wins out there whatever the weekday - a weekend
+                                # rule flattening the weekday peaks two days out (batpred#5168).
+                                # Minutes left unwritten here are no loss: rate_replicate() fills them
+                                # from the same time of day 24 hours earlier.
+                                extend_minute = minute_mod + max_minute
```
springfall2008: **Midnight-spanning rules never seed the 48h-out early-hours slot - the wrapped minutes extend 72h out instead**

For a rule that wraps midnight (e.g. 22:00-07:00), the loop runs over two calendar segments: the pre-midnight minutes (minute_mod < 24*60) seed the day-offset-2 slot at their time of day (22:00-23:59), but the wrapped minutes (minute_mod in [24*60, 48*60)) compute extend_minute = minute_mod + max_minute >= 72*60, i.e. day offset 3 at 00:00-06:59 - not the day-offset-2 slot for the same time of day.

So for midnight-spanning day-filtered rules the 48h-out early-hours slots are structurally never seeded and always come from rate_replicate() copying tomorrow - correct only when tomorrow's rate at that time of day happens to equal the day-after-tomorrow's true rate. Example that breaks it: midnight on a Saturday, weekday off-peak 22:00-07:00 = 12p and weekend flat 8p -> Monday (day offset 2) 00:00-07:00 = 8 (Sunday's weekend rate) instead of 12. At the default forecast_hours=48 the day-offset-3 seeds are also unreachable (minutes_now + forecast_minutes <= 4319 < 4320), so they do nothing in the live plan.

apps/predbat/fetch.py
```diff
@@ -2309,9 +2309,19 @@ def basic_rates(self, info, rtype, prev=None, rate_replicate=None, include_manua
...
+                                extend_day_of_week = (day_of_week_midnight + int(extend_minute / (24 * 60))) % 7
+                                if not day_of_week or (extend_day_of_week in day_of_week):
```
springfall2008: **Newly-missing day-2 slots can be filled with self.rate_max via rate_replicate()'s io_adjusted branch**

When this gate skips a slot, rate_replicate() fills it. Its first branch (fetch.py:1821) fires when the 24h-earlier minute is in rate_io (self.io_adjusted, the Octopus Intelligent adjusted-minute map, which includes day-offset-1 keys): it returns self.rate_max instead of the tariff copy - 'don't replicate Intelligent rates into the next day'.

Before this change the day-offset-2 slots were always present (seeded), so that branch could never fire for day 2; now, for an import-rates user with day_of_week rules, any 48h-out slot whose tomorrow counterpart is an IO dispatch minute gets rate_max (the plan's max price, e.g. ~99p) stamped into the curve - a phantom spike 48h out in the plan/graph instead of the tariff rate. Niche (IOG + day_of_week rate rules), but newly reachable via this gate.

apps/predbat/fetch.py
```diff
@@ -2309,9 +2309,19 @@ def basic_rates(self, info, rtype, prev=None, rate_replicate=None, include_manua
...
+                                    rates[extend_minute] = rate
```
springfall2008: **The extend write ignores rate_increment**

`rates[extend_minute] = rate` is a plain assignment even when the rule is a rate_increment one (the while loop above does rates.get(...)+rate). A rate_increment rule without a date stamps the 48h-out slot with the bare increment value (e.g. a +0.15 override writes 0.15 there, not base+0.15), clobbering whatever an earlier rule seeded, and the slot is then treated as real data by rate_replicate (no copy, no offset). Pre-existing on this line, but the PR rewrites the line - natural place to fix it (and to mark rate_replicate[extend_minute] the way the loop does).

apps/predbat/fetch.py
springfall2008: **Silent economics change: newly-replicated day-2 slots now get metric_future_rate_offset_import/export applied**

Before this change every day-offset-2 slot was seeded (present in rates), so rate_replicate() never applied the futurerate offset to it. When this gate skips a slot, the slot is now missing -> replicated -> `rate_offset + self.metric_future_rate_offset_import` (or export) is applied, and the slot is marked as replicated in rate_import_replicated/rate_export_replicated.

For anyone running futurerate_adjust_import/export (Agile-style) together with a day_of_week tariff, the 48h-out rate for non-matching weekdays silently shifts by the offset (default 0, up to +/-50p). Probably the right semantic for a replicated slot - but it is a behavior change beyond the bugfix and worth a mention in the PR notes.

apps/predbat/fetch.py
```diff
@@ -2309,9 +2309,19 @@ def basic_rates(self, info, rtype, prev=None, rate_replicate=None, include_manua
...
-                                    self.load_scaling_dynamic[minute_mod + max_minute] = load_scaling
+                                # Seed the slot past the days modelled above so rate_replicate() has
+                                # something to copy forward. This has to honour the rule's day filter
+                                # in the same way the loop just did, or the last rule processed for a
+                                # given time of day wins out there whatever the weekday - a weekend
+                                # rule flattening the weekday peaks two days out (batpred#5168).
+                                # Minutes left unwritten here are no loss: rate_replicate() fills them
+                                # from the same time of day 24 hours earlier.
+                                extend_minute = minute_mod + max_minute
+                                extend_day_of_week = (day_of_week_midnight + int(extend_minute / (24 * 60))) % 7
```
springfall2008: **Cleanup: the day-of-week arithmetic now exists twice, one copy buggy, one correct**

Line 2298 in the while loop computes the same expression from minute_index (and truncates toward zero on negative minute_index, mislabeling yesterday - see my other comment); this line computes it again from extend_minute, correctly. Extracting one helper (e.g. `day_of_week_for(minute_index)` using floor division `//`) would remove the duplication and fix the line-2298 copy in the same stroke.

apps/predbat/fetch.py
```diff
@@ -2309,9 +2309,19 @@ def basic_rates(self, info, rtype, prev=None, rate_replicate=None, include_manua
                                             break
                                 minute_index += 24 * 60
                             if not date and not prev:
-                                rates[minute_mod + max_minute] = rate
-                                if load_scaling is not None:
-                                    self.load_scaling_dynamic[minute_mod + max_minute] = load_scaling
+                                # Seed the slot past the days modelled above so rate_replicate() has
```
springfall2008: **Altitude: seeding one extra day leaves the weekday question to a weekday-blind mechanism**

The root problem in batpred#5168 is that nothing downstream of basic_rates() knows about day_of_week - rate_replicate() just copies '24 hours earlier'. This PR teaches one specific write (the day-offset-2 seed) about the day filter, which fixes the 48h-out slot and nothing else; every later day is still answered by 'whatever day+2 had'.

Simpler and more general alternative at the same site: make the extend write a small loop over the future days (minute_mod + n*24*60 for n in 2..(forecast_minutes + 48*60)/1440 - minute_mod//1440), honouring the day filter per day - then basic_rates() stamps the full horizon itself and the replication-of-day+2 dependency disappears for day-filtered tariffs. That also fixes the two other gaps in the same stroke (midnight-spanning rules, day+3 propagation).

apps/predbat/tests/test_basic_rates.py
```diff
@@ -268,4 +293,34 @@ def test_basic_rates(my_predbat):
         my_predbat.manual_api = old_manual_api
         my_predbat.had_errors = old_had_errors
 
+    # Test 13: the export config from issue #5168 - weekday rules tiling the whole day plus a
```
springfall2008: **Coverage gap: only the weekend-last ordering and only day 2 are tested - the failing cases are untested**

Both new day-3 blocks use the weekend-catch-all-last ordering (the reporter's). The opposite ordering (weekday rules last) is where the new gate changes day+3 behavior: pre-fix, last-rule-wins stamped the weekday pattern at day+2 and Monday (day+3) was accidentally correct; post-fix, Monday gets Sunday's weekend pattern replicated forward. A test with the rules reordered, plus a day+4 assertion (Monday-after-Sunday-midnight), would pin that. Also nothing asserts the wrapped-window (22:00-07:00) rule's day-offset-2 early-hours slot, which is never seeded (see fetch.py comment thread) - a tariff where the weekday off-peak differs from the weekend flat would fail there.

apps/predbat/tests/test_basic_rates.py
```diff
@@ -268,4 +293,34 @@ def test_basic_rates(my_predbat):
...
+            {"start": "09:00:00", "end": "16:59:59", "rate": 11.5, "day_of_week": "1,2,3,4,5"},
+            {"start": "17:00:00", "end": "20:59:59", "rate": 23.0, "day_of_week": "1,2,3,4,5"},
+            {"start": "21:00:00", "end": "23:59:59", "rate": 11.5, "day_of_week": "1,2,3,4,5"},
+            {"start": "00:00:00", "end": "23:59:59", "rate": 11.5, "day_of_week": "6,7"},
+        ]
+        results = my_predbat.basic_rates(weekend_catch_all, "export")
+        results, results_replicated = my_predbat.rate_replicate(results, is_import=False, is_gas=False)
+
+        # Day 3 (Tuesday) must keep the weekday peaks. The rules end on :59:59 so each leaves the
+        # final minute of its window unset - assert the interiors, which are the ones the rules
```
springfall2008: **Comment premise is wrong: the rules here tile the day contiguously, no minute is left unset**

minutes_to_time() floors to whole minutes (int(seconds/60)), so end "06:59:59" -> minute 419 and range(start, end) writes 00:00..06:59 inclusive; the next rule starts at 07:00. Every minute 0..1439 is written by exactly one rule (the last one ends at 23:59 and covers it). The assertions are fine, but the 'leaves the final minute of its window unset' claim would send the next reader hunting for a gap that does not exist.

apps/predbat/fetch.py
springfall2008: _Automated reply from the triage bot._

Fixed in 2139b063. The separate day+2 seed write is gone. `basic_rates()` now runs its existing per-day loop out to `max(max_minute + 24 * 60, self.forecast_minutes + 48 * 60)`, which is the same end `rate_replicate()` uses. Each day is filtered against its own weekday there, so no day in the horizon depends on copying the day before. The "Minutes left unwritten here are no loss" comment went with the block.

New test 14 runs midnight on Friday and on Saturday, with the rules in both orders, over five days (`forecast_minutes` pinned at 72h). Before the fix it failed at minute 4740 (Monday 07:00, day+3 from Friday), reading 8.0 where 14.0 was expected, in both rule orders. With the fix it passes.

apps/predbat/fetch.py
springfall2008: _Automated reply from the triage bot._

Confirmed and fixed in 2139b063. `int(minute_index / (24 * 60))` is now `minute_index // (24 * 60)`, so yesterday's minutes map to day offset -1.

New test 15 sets midnight on a Monday, with a weekday rule at 14 and a weekend rule at 8. Before the fix, `basic_rates()` returned 14.0 at minute -1439 (Sunday was checked against Monday). Now all of yesterday is 8.0. The seed write is gone, so this is the only place the weekday is computed. The fix therefore also covers override (`prev`) rules with `day_of_week`, which go through the same loop.

apps/predbat/fetch.py
springfall2008: _Automated reply from the triage bot._

Fixed by the same change in 2139b063. The wrapped half of a midnight-spanning rule (`minute_mod` in `[24*60, 48*60)`) now runs through the loop past day 1. That means day+2's early hours are stamped and checked against day+2's weekday. The old code only reached day+3 there, through `minute_mod + max_minute`.

Test 14 uses your example shape: weekday off-peak 22:00-07:00 at 12, weekend flat at 8, midnight on a Saturday. Before the fix it failed at minute 2880 (Monday 00:00), reading 8.0 instead of 12.0. With the fix it passes.

apps/predbat/fetch.py
springfall2008: _Automated reply from the triage bot._

Mostly closed by 2139b063. When a tariff's rules cover every minute of each day, every minute through the horizon is now stamped by `basic_rates()` itself. `rate_replicate()` never fills them, so its `rate_io` branch can't reach day+2 or anything beyond it.

On reachability: `self.io_adjusted` is only populated by the Octopus rate fetch (`octopus.py:3761`), and the live path only falls back to `basic_rates()` when no Octopus import rates were read. So the combination arises through `compare.py:137`, which replicates a compared basic-rate tariff against the live `pb.io_adjusted`, rather than through the live plan.

What remains: a tariff that leaves some weekday unclaimed (weekday rules with no weekend rule) still has those days replicated. At day+2 that differs from main, where the unfiltered seed wrote the weekday rate onto a weekend. I've left it. It matches how the same tariff's tomorrow already behaves on main, and putting another weekday's rate on that day was the bug being fixed.

apps/predbat/fetch.py
springfall2008: _Automated reply from the triage bot._

Fixed in 2139b063. The plain-assignment write is gone. Every stamped day now goes through the loop's `rates.get(minute_index, 0.0) + rate` branch and sets the `rate_replicate[minute_index]` marker the same way as today and tomorrow.

New test 16 uses a base rate of 10 plus a 17:00-19:00 `rate_increment` of 5. Before the fix, day+2 at 17:00 (minute 3900) was 5.0. It is now 15.0 on all three days, and 19:00 stays at 10.0.

apps/predbat/fetch.py
springfall2008: _Automated reply from the triage bot._

This does need stating, and 2139b063 reverses its direction. When a tariff's rules cover every minute of every day, no minute in the horizon is replicated any more. So `metric_future_rate_offset_import`/`_export`, `futurerate_adjust_*` and the replicated markers apply to none of them. That includes day+3 onwards. On main those days *were* replicated from the seeded day+2, so they did take the offset and the `offset` marker whenever `forecast_hours` > 48.

I think that's the right outcome. `docs/customisation.md:418` and `docs/energy-rates.md:585` describe the offset as applying to "future import energy rates that are not yet published" (Agile/Tracker), and a manual tariff is fully published. Minutes that no rule claims on a given weekday are still replicated and still get the offset. I've rewritten the PR description's note to say this, replacing the day+2 note it had before.

apps/predbat/fetch.py
springfall2008: _Automated reply from the triage bot._

In 2139b063 the duplicate is removed rather than extracted. With the seed write folded into the loop, the day-of-week expression now exists once, using floor division: `(day_of_week_midnight + minute_index // (24 * 60)) % 7`. A helper didn't seem worth it for a single call site. The line-2298 truncation bug is fixed by the same edit (see the other thread).

apps/predbat/fetch.py
springfall2008: _Automated reply from the triage bot._

Implemented in 2139b063, by extending the existing loop rather than adding a second one. For non-`date`, non-`prev` rules, the `while minute_index < ...` bound is now `stamp_end = max(max_minute + 24 * 60, self.forecast_minutes + 48 * 60)`. That is `rate_replicate()`'s own end, and the `max()` keeps at least the old three days if `forecast_minutes` is still 0 when `basic_rates()` runs.

Reusing the loop keeps the day filter, `rate_increment`, `load_scaling` and the markers on one path for every day, and the separate seed write is deleted. `date` rules and `prev` overrides keep `max_minute` as their bound. Overrides already span the replicated horizon, since `max_minute = max(prev) + 1` there.

Cost at the default `forecast_hours`: about 5 passes per minute per rule, instead of 3 plus the seed write.

This also closes the midnight-spanning and day+3 threads. Verification:
- `basic_rates`, `basic_rates_utc`, `rate_replicate`, `annual_tariff`, `compare` and `manual_api` all pass.
- `run_all --quick` passes (324 modules).
- The pre-commit hooks pass. The one exception is `markdownlint-fix`, which only reports the triage clone's own uncommitted GitNexus edits to `.claude/skills/gitnexus/*`, `AGENTS.md` and `CLAUDE.md`. None of those files are in this PR.

apps/predbat/tests/test_basic_rates.py
springfall2008: _Automated reply from the triage bot._

Added in 2139b063 as test 14. It runs both rule orders (weekend catch-all last and weekday rules last) with midnight on a Friday and on a Saturday, and asserts all five days after `rate_replicate()`, which covers the day+3 and day+4 Monday/Tuesday cases. The weekday off-peak (22:00-07:00 at 12) differs from the weekend flat rate (8), so day+2's wrapped early hours are checked too. It also asserts `load_scaling_dynamic` at noon on each day.

Before the fix it failed in every combination:
- minute 4740 read 8.0, expected 14.0 (Friday midnight, both orders);
- minute 2880 read 8.0, expected 12.0 (Saturday midnight, both orders);
- `load_scaling` was `None` on days 3-4.

Tests 15 (yesterday's weekday) and 16 (`rate_increment` past day 1) cover the other two threads.

apps/predbat/tests/test_basic_rates.py
springfall2008: _Automated reply from the triage bot._

I'm keeping this one. The gap is real, because `range(start, end)` excludes `end`:
- `"06:59:59"` is 25199 s, and `int(25199 / 60)` = 419.
- So `end_minutes` = 419, and `range(0, 419)` writes minutes 0-418 (00:00-06:58).
- The next rule starts at 420 (07:00), so minute 419 (06:59) is written by neither rule.
- `"23:59:59"` works the same way: end 1439, so minute 1439 is never written.

I checked this with a temporary probe on the unfixed branch. For test 13's config, `basic_rates()` returned `48*60+418` = 11.5 and `48*60+420` = 23.0, and `48*60+419` was absent. After `rate_replicate()`, 06:59 on Tuesday was 11.5 with the `copy` marker. That only matches because the weekend rate happens to equal the off-peak rate.

In 2139b063 I kept the assertions and reworded the comment to give the reason: the `:59:59` end floors to the `:59` minute, and the window excludes its end minute, so each rule leaves its last minute to `rate_replicate()`.

springfall2008: _Automated reply from the triage bot._

This gap has a worse consequence than my previous reply implied. It predates this PR and is not changed by it. `basic_rates()` pre-initialises today's minutes 0-1439 to 0, and on a weekday no rule writes the `:59` minutes. So they stay at **0p** today, and `rate_replicate()` copies that zero forward.

I ran a temporary probe (not committed) with the reporter's `rates_export` config from #5168, midnight on a Tuesday and `minutes_now = 0`. It found minutes at 0 on each of the three days:
`[419, 539, 1019, 1259, 1439, 1859, 1979, 2459, 2699, 2879, 3299, 3419, 3899, 4139, 4319]`
These are 06:59, 08:59, 16:59, 20:59 and 23:59 every day. `rate_minmax()` returned `(0, 23.0, 14.32, 419, 420)`, a minimum of 0 at 06:59.

For an export tariff, that means `Export rates: min 0`. For an import tariff written the same way, it would be a 0p import minute at each rule boundary. I haven't traced what the planner does with those minutes. Writing the ends as `07:00:00`, `09:00:00`, ... avoids the gap. Separately, `basic_rates()` could treat a `:59:59` end as inclusive. I've left that out of this PR because it changes how existing configs are read.

